PR #4, opened 2020-02-25
Add some bringup tests for SM 5.1 bindless

HansKristian-Work:
Test Buffer SRVs with NonUniformResourceIndex.


tests/d3d12.c
```diff
@@ -33843,6 +33843,535 @@ static void test_constant_buffer_sm51(void)
...
+    ok(SUCCEEDED(hr), "Failed to create root signature, hr %#x.\n", hr);
+
+    static const DWORD cs_code[] = {
+#if 0
+        StructuredBuffer<uint> Buffers[] : register(t4, space1);
+        StructuredBuffer<uint> AliasBuffers[64] : register(t4, space2);
+        StructuredBuffer<uint> StandaloneBuffer1 : register(t100, space2);
+        StructuredBuffer<uint> StandaloneBuffer2 : register(t200, space2);
+
+        RWByteAddressBuffer OBuffer : register(u0);
```
doitsujin: It might be useful to throw in a texture array for completeness. Doesn't have to be anything complex, but maybe something like
```
Texture2D<float4> Textures[2] : register(t0, space3);
// ...
result += uint(Textures[1].Load(int2(0,0)).x);
```
where the first texture is black and the second one is white or something.

HansKristian-Work: Added tests for this.

PR #5, opened 2020-02-25
Add various DXIL tests

HansKristian-Work:
Adds a DXIL alternative to most of the interesting shaders.

The overall strategy is to change the test entry points to take a flag which selects DXIL or DXBC. DXIL is more bloated than DXBC and is byte-aligned, which bloats the code size a bit, but not much we can do with the current test architecture.

The diff is rather large, but very repetitive.


HansKristian-Work:
Yeah, it's all naked `void func(void)` and no templates, so it's hard to do it better without rewriting.
The tests all pass on Windows and dxil-spirv when I get to PR-ing that implementation.

PR #8, opened 2020-02-26
Add test for SM 5.1 immutable sampler

HansKristian-Work:
This already worked, the FIXME was benign unlike the other cases where it would return NULL.


tests/d3d12.c
```diff
@@ -37730,7 +37752,7 @@ static void test_register_space(bool use_dxil)
     for (i = 0; i < 4; i++)
     {
         /* Start value of 9 is for the StructuredBuffer we write to. */
-        float reference = 2 * 3 * 4 * 5 * 6 * 7 * 8 * 10 * 11;
+        float reference = 2 * 3 * 4 * 5 * 6 * 7 * 8 * (10 * 10) * 11;
```
doitsujin: 10 comes from the additional texture lookup?

HansKristian-Work: Yes. Kinda messy to do things like this though, but it's basically a hardcoded value either way.

PR #10, opened 2020-02-26
vkd3d-shader: Rework SPIR-V capability enablement.

doitsujin:
Makes adding new capabilities significantly easier, and we're going to add quite a lot of them with bindless (as well as future features).

### DIFF
--- a/libs/vkd3d-shader/spirv.c
+++ b/libs/vkd3d-shader/spirv.c
@@ -285,9 +285,10 @@ static bool vkd3d_spirv_stream_append(struct vkd3d_spirv_stream *dst_stream,
 
 struct vkd3d_spirv_builder
 {
-    uint64_t capability_mask;
-    uint64_t capability_draw_parameters : 1;
-    uint64_t capability_demote_to_helper_invocation : 1;
+    SpvCapability *capabilities;
+    size_t capabilities_size;
+    size_t capability_count;
+
     uint32_t ext_instr_set_glsl_450;
     uint32_t invocation_count;
     SpvExecutionModel execution_model;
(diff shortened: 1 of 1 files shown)


libs/vkd3d-shader/spirv.c
```diff
@@ -1699,16 +1708,45 @@ static void vkd3d_spirv_builder_free(struct vkd3d_spirv_builder *builder)
...
+    VKD3D_SPV_KHR_SHADER_DRAW_PARAMETERS        = 0x00000001,
+    VKD3D_SPV_EXT_DEMOTE_TO_HELPER_INVOCATION   = 0x00000002,
+};
+
+struct vkd3d_spirv_extension_info
+{
+    enum vkd3d_spirv_extension extension;
+    const char* name;
+}
+vkd3d_spirv_extensions[] =
```
HansKristian-Work: Shouldn't this be static const?

libs/vkd3d-shader/spirv.c
```diff
@@ -1699,16 +1708,45 @@ static void vkd3d_spirv_builder_free(struct vkd3d_spirv_builder *builder)
...
+    {VKD3D_SPV_KHR_SHADER_DRAW_PARAMETERS,      "SPV_KHR_shader_draw_parameters"},
+    {VKD3D_SPV_EXT_DEMOTE_TO_HELPER_INVOCATION, "SPV_EXT_demote_to_helper_invocation"},
+};
+
+struct vkd3d_spirv_capability_extension_mapping
+{
+    SpvCapability capability;
+    enum vkd3d_spirv_extension extension;
+}
+vkd3d_spirv_capability_extensions[] =
```
HansKristian-Work: Ditto.

libs/vkd3d-shader/spirv.c
```diff
@@ -1699,16 +1708,45 @@ static void vkd3d_spirv_builder_free(struct vkd3d_spirv_builder *builder)
 
     rb_destroy(&builder->declarations, vkd3d_spirv_declaration_free, NULL);
 
+    vkd3d_free(builder->capabilities);
     vkd3d_free(builder->iface);
 }
 
+enum vkd3d_spirv_extension
+{
+    VKD3D_SPV_KHR_SHADER_DRAW_PARAMETERS        = 0x00000001,
```
HansKristian-Work: Isn't it supposed to be an array and not a bitmask now?

libs/vkd3d-shader/spirv.c
```diff
@@ -1720,22 +1758,28 @@ static bool vkd3d_spirv_compile_module(struct vkd3d_spirv_builder *builder,
...
-        vkd3d_spirv_build_op_extension(&stream, "SPV_KHR_shader_draw_parameters");
-    if (builder->capability_demote_to_helper_invocation)
-        vkd3d_spirv_build_op_extension(&stream, "SPV_EXT_demote_to_helper_invocation");
+    for (i = 0; i < ARRAY_SIZE(vkd3d_spirv_capability_extensions); ++i)
+    {
+        if (vkd3d_spirv_has_capability(builder, vkd3d_spirv_capability_extensions[i].capability))
+        {
+            uint32_t extension = vkd3d_spirv_capability_extensions[i].extension;
+
+            if (extension_mask & extension)
```
HansKristian-Work: Oh, I see, it's using the bitmask to not enable the same extension over and over.

libs/vkd3d-shader/spirv.c
doitsujin: Indeed, fixed.

doitsujin:
An alternative implementation wrt extensions would be to just do something like
```
enum vkd3d_spirv_extension {
  VKD3D_SPV_EXT_FROG,
  VKD3D_SPV_KHR_TOAD,
};
static const char* vkd3d_spirv_extension_names[] = {
  "SPV_EXT_frog",
  "SPV_KHR_toad",
};
```

This would simplify the lookup a bit, but requires that the extension enum values and extension names be declared in the exact same order.

HansKristian-Work:
Current implementation should be fine as-is unless we need to deal with > 64 extensions.

PR #12, opened 2020-02-27
Dump SPIR-V along DXBC/DXIL with VKD3D_SHADER_DUMP_PATH

HansKristian-Work:


libs/vkd3d-shader/vkd3d_shader_private.h
```diff
@@ -837,6 +837,9 @@ void vkd3d_dxbc_compiler_destroy(struct vkd3d_dxbc_compiler *compiler) DECLSPEC_
 
 void vkd3d_compute_dxbc_checksum(const void *dxbc, size_t size, uint32_t checksum[4]) DECLSPEC_HIDDEN;
 
+void vkd3d_shader_dump_spirv_shader(enum vkd3d_shader_type type, const struct vkd3d_shader_code *shader);
+void vkd3d_shader_dump_shader(enum vkd3d_shader_type type, const struct vkd3d_shader_code *shader);
```
doitsujin: This implies that we're also going to be calling these methods from outside `shader_main.c`; where exactly would that be? I assume this is DXIL-related, but I wonder if there's a way to potentially unnecessary avoid duplication.

HansKristian-Work: The dxil implementation will live in dxil.c and that function calls vkd3d_shader_dump_spirv_shader. I suppose dump_shader (raw DXBC container blob) doesn't really need to be placed in the private header, but I felt it made sense to keep them together like this.

PR #14, opened 2020-03-01
Temporary UAV counter rework

doitsujin:
In preparation for the bindless implementation, this makes the following changes to UAV counter handling:
- UAV counters are packed into the root signature's descriptor set layout for now. This allows us to get rid of per-pipeline descriptor set layouts and pipeline layouts, and gretaly simplifies descriptor updates involving UAV counters.
- UAV counter bindings are passed to vkd3d-shader as regular resource bindings in order to avoid some duplication.

As part of the bindless implementation, I plan to rewrite `d3d12_command_list_update_descriptors` as well as `d3d12_root_signature_init` completely from scratch. While those parts are currently messy, they hopefully won't stay that way for long.

It's probably worth testing WoW with thsese patches, which I haven't done yet. SotTR and Metro still work as expected and they don't introduce new test failures.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -2573,11 +2573,12 @@ static void d3d12_command_list_prepare_descriptors(struct d3d12_command_list *li
 static bool vk_write_descriptor_set_from_d3d12_desc(VkWriteDescriptorSet *vk_descriptor_write,
         VkDescriptorImageInfo *vk_image_info, const struct d3d12_desc *descriptor,
         uint32_t descriptor_range_magic, VkDescriptorSet vk_descriptor_set,
-        uint32_t vk_binding, unsigned int index)
+        uint32_t vk_binding, unsigned int index, bool uav_counter)
 {
     const struct vkd3d_view *view = descriptor->u.view;
+    bool is_uav = descriptor->magic == VKD3D_DESCRIPTOR_MAGIC_UAV;
 
-    if (descriptor->magic != descriptor_range_magic)
+    if (descriptor->magic != descriptor_range_magic || (uav_counter && !is_uav))
         return false;
 
     vk_descriptor_write->sType = VK_STRUCTURE_TYPE_WRITE_DESCRIPTOR_SET;
@@ -2601,7 +2602,7 @@ static bool vk_write_descriptor_set_from_d3d12_desc(VkWriteDescriptorSet *vk_des
         case VKD3D_DESCRIPTOR_MAGIC_UAV:
             /* We use separate bindings for buffer and texture SRVs/UAVs.
              * See d3d12_root_signature_init(). */
-            vk_descriptor_write->dstBinding = vk_binding + 2 * index;
+            vk_descriptor_write->dstBinding = vk_binding + (is_uav ? 3 : 2) * index;
             if (descriptor->vk_descriptor_type != VK_DESCRIPTOR_TYPE_UNIFORM_TEXEL_BUFFER
                     && descriptor->vk_descriptor_type != VK_DESCRIPTOR_TYPE_STORAGE_TEXEL_BUFFER)
                 ++vk_descriptor_write->dstBinding;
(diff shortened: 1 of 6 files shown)


HansKristian-Work:
Thanks, I'll look at it on Monday.

libs/vkd3d/state.c
```diff
@@ -812,8 +830,8 @@ static HRESULT d3d12_root_signature_init(struct d3d12_root_signature *root_signa
     /* XXX: Vulkan buffer and image descriptors have different types. In order
      * to preserve compatibility between Vulkan resource bindings for the same
      * root signature, we create descriptor set layouts with two bindings for
-     * each SRV and UAV. */
-    info.descriptor_count += info.srv_count + info.uav_count;
+     * each SRV and UAV. For UAVs we additionally have one UAV counter. */
+    info.descriptor_count += info.srv_count + 2 * info.uav_count;
```
HansKristian-Work: Does this also affect root UAVs (which do not have counters?)

libs/vkd3d/command.c
```diff
@@ -2607,15 +2602,26 @@ static bool vk_write_descriptor_set_from_d3d12_desc(VkWriteDescriptorSet *vk_des
...
                     || descriptor->vk_descriptor_type == VK_DESCRIPTOR_TYPE_STORAGE_TEXEL_BUFFER)
             {
-                vk_descriptor_write->pTexelBufferView = &view->u.vk_buffer_view;
+                if (uav_counter)
+                {
+                    vk_descriptor_write->dstBinding += 2;
+                    vk_descriptor_write->pTexelBufferView = &view->vk_counter_view;
+
+                    if (!view->vk_counter_view)
+                        return false;
```
HansKristian-Work: Is it okay to not write a descriptor here? Without PARTIALLY_BOUND_BIT we probably have to bind something here. Although this might be somewhat moot since we'll need to do PARTIALLY_BOUND and deferred updates for RS 1.0 soon anyways.

libs/vkd3d/command.c
```diff
@@ -2675,41 +2681,25 @@ static void d3d12_command_list_update_descriptor_table(struct d3d12_command_list
...
-                    bindings->descriptor_set, range->binding, j))
+                    bindings->descriptor_set, range->binding, j, false))
                 continue;
 
             ++descriptor_count;
             ++current_descriptor_write;
             ++current_image_info;
 
-            if (descriptor_count == ARRAY_SIZE(descriptor_writes))
+            if (vk_write_descriptor_set_from_d3d12_desc(current_descriptor_write,
```
HansKristian-Work: This could overflow I think.

libs/vkd3d/command.c
```diff
@@ -2675,41 +2681,25 @@ static void d3d12_command_list_update_descriptor_table(struct d3d12_command_list
...
+                    current_image_info, descriptor, range->descriptor_magic,
+                    bindings->descriptor_set, range->binding, j, true))
+            {
+                ++descriptor_count;
+                ++current_descriptor_write;
+                ++current_image_info;
+            }
+
+            if (descriptor_count >= ARRAY_SIZE(descriptor_writes) - 1)
```
HansKristian-Work: I suppose this takes care of the overflow, would be good with a comment to signal intent.

libs/vkd3d/state.c
doitsujin: `srv_count` and `uav_count` only count descriptor table entries, not root descriptors. Those are being counted separately.

libs/vkd3d/command.c
doitsujin: It should be fine as long as it isn't statically used by the pipeline, which would only be the case if the app does not provide a counter in a situation where it is required to do so.

HansKristian-Work: Right, I think I'm confusing things, from spec:
"• Descriptors that are statically used must have been populated before the descriptor set is
consumed.
• Entries that are not used by a pipeline can have undefined descriptors."

PARTIALLY_BOUND would be for bindless only in that case.

libs/vkd3d/command.c
doitsujin: Sure, will do.

PR #15, opened 2020-03-02
vkd3d: Implement RS 1.0 volatile descriptors.

HansKristian-Work:


libs/vkd3d/command.c
```diff
@@ -1341,7 +1341,11 @@ static VkDescriptorPool d3d12_command_allocator_allocate_descriptor_pool(
     {
         pool_desc.sType = VK_STRUCTURE_TYPE_DESCRIPTOR_POOL_CREATE_INFO;
         pool_desc.pNext = NULL;
-        pool_desc.flags = 0;
+        /* For a correct implementation of RS 1.0 we need to update packed descriptor sets late rather than on draw.
+         * If device does not support descriptor indexing, we must update on draw and pray applications don't rely on RS 1.0
+         * guarantees. */
+        pool_desc.flags = allocator->device->vk_info.EXT_descriptor_indexing ?
+                          VK_DESCRIPTOR_POOL_CREATE_UPDATE_AFTER_BIND_BIT_EXT : 0;
```
doitsujin: We should probably check for *all* the `UpdateAfterBind` features separately, and store whether we support volatile descriptors somewhere in the `d3d12_device`. Older Nvidia generations (up to Pascal) do not support `descriptorBindingUniformBufferUpdateAfterBind`.

libs/vkd3d/command.c
```diff
@@ -2656,22 +2662,103 @@ static void vk_write_descriptor_set_for_descriptor_info(VkDescriptorSet vk_descr
...
+}
+
 static void d3d12_command_list_update_packed_descriptors(struct d3d12_command_list *list,
         VkPipelineBindPoint bind_point)
 {
     struct vkd3d_descriptor_updates *updates = &list->packed_descriptors[bind_point];
     struct vkd3d_pipeline_bindings *bindings = &list->pipeline_bindings[bind_point];
     const struct d3d12_root_signature *root_signature = bindings->root_signature;
     const struct vkd3d_vk_device_procs *vk_procs = &list->device->vk_procs;
+    bool deferred_update = list->device->vk_info.EXT_descriptor_indexing;
```
doitsujin: Would need to change this check as well.

libs/vkd3d/command.c
```diff
@@ -2693,38 +2780,27 @@ static void d3d12_command_list_update_packed_descriptors(struct d3d12_command_li
...
+            d3d12_command_list_defer_update_descriptor_table(list,
+                    descriptor_set, updates,
+                    root_signature, base_descriptor,
+                    root_parameter_index);
+        }
+        else
+        {
+            /* Fallback, we update the descriptor set here.
+             * Will work in most cases, but it's not a correct implementation of RS 1.0.
+             * Will work if application uses RS 1.1 and declares STATIC descriptor flag. */
```
doitsujin: I guess we should respect that flag, might be worth adding a TODO for that somewhere. Currently we only have access to RS1.0 data structures though, so that will need some more work.

libs/vkd3d/state.c
```diff
@@ -288,20 +288,49 @@ static HRESULT vkd3d_create_descriptor_set_layout(struct d3d12_device *device,
...
+    unsigned int i;
     VkResult vr;
 
     set_desc.sType = VK_STRUCTURE_TYPE_DESCRIPTOR_SET_LAYOUT_CREATE_INFO;
     set_desc.pNext = NULL;
     set_desc.flags = flags;
     set_desc.bindingCount = binding_count;
     set_desc.pBindings = bindings;
+
+    if (device->vk_info.EXT_descriptor_indexing && (flags & VK_DESCRIPTOR_SET_LAYOUT_CREATE_PUSH_DESCRIPTOR_BIT_KHR) == 0)
```
doitsujin: Here's another check to fix up actually.

HansKristian-Work:
Should have fixed all issues.

PR #18, opened 2020-03-05
vkd3d: Create NULL buffer and image views.

doitsujin:
Currently, we leave descriptors undefined that are either left uninitialized by the application or have the wrong resource type bound. However, this causes issues when using descriptor update templates since we need to provide a valid view handle for all descriptors that are to be updated.

Providing these may then also improve stability with misbehaving applications that access an undefined descriptor in a shader.

Note that this is different from the explicit NULL descriptors that the application can create using `CreateShaderResourceView` and friends.

### DIFF
--- a/libs/vkd3d/resource.c
+++ b/libs/vkd3d/resource.c
@@ -17,6 +17,8 @@
  * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
  */
 
+#include <float.h>
+
 #include "vkd3d_private.h"
 
 #define VKD3D_NULL_BUFFER_SIZE 16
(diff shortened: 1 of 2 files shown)


libs/vkd3d/resource.c
```diff
@@ -4144,20 +4193,44 @@ HRESULT vkd3d_init_null_resources(struct vkd3d_null_resources *null_resources,
     if (!use_sparse_resources && FAILED(hr = vkd3d_allocate_image_memory(device, null_resources->vk_2d_storage_image,
             &heap_properties, D3D12_HEAP_FLAG_NONE, &null_resources->vk_2d_storage_image_memory, NULL, NULL)))
         goto fail;
+    if (!vkd3d_create_vk_image_view(device, null_resources->vk_2d_storage_image,
```
HansKristian-Work: Are these resources being used anywhere? I seem them being created and destroyed.

doitsujin: Not yet, but I'm building the root signature rework around this currently, in order to use descriptor update templates as mentioned in the comment.

Of course we can treat this as an RFC for now and merge it later.

HansKristian-Work: Alright, just wanted to make sure. No reason not to merge it.

PR #20, opened 2020-03-07
Binding model: Bindless part 2

doitsujin:
Implements initial support for bindless SRVs and Samplers, as well as dynamic and non-uniform descriptor indexing for Shader Model 5.1/6.0.

This consists of several parts, all of which depend on each other:
1. Shader compiler work to support these features, including some required additions to the interface between vkd3d and the shader compiler.
2. A complete rewrite of the root signature and resource binding code, which was needed since assumptions about the exact descriptor set layouts were duplicated in various places and the whole thing was barely maintainable, let alone extensible.
3. API-side additions to support bindless SRVs/Samplers.

The implementation passes both `test_bindless_srv_sm51` and `test_bindless_samplers_sm51` on AMDVLK as well as RADV with ACO (non-uniform indexing is broken on LLVM), and was further tested in Shadow of the Tomb Raider, Metro Exodus, and Resident Evil 2.

## Overview
Each descriptor heap holds one or more descriptor sets, one for each supported descriptor type. These sets are updated by functions such as `CreateShaderResourceView` and `CopyDescriptors`, and are bound to a command list via `SetDescriptorHeaps`.

The set and binding numbers for these sets are be the same for all pipeline layouts, so there is no need to rebind the sets when the root signature changes.

Descriptor table offsets are passed to the pipeline via push constants, one 32-bit integer per table, and are updated after each call to `Set*RootDescriptorTable`.

```glsl
/* Declarations */
layout(push_constant, std430) uniform push_cb {
    uint descriptor_tables[2];
} pc;
/* Descriptor heaps */
layout(set = 0, binding = 0) uniform sampler samplers[];
layout(set = 1, binding = 0) uniform usamplerBuffer srvBuf[];
layout(set = 2, binding = 0) uniform utexture2D srv2D[];
/* Root descriptor */
layout(set = 3, binding = 0, r32ui) uniform writeonly uimageBuffer u0;
/* Sample lookup */
texelFetch(srv2D[nonuniformEXT(pc.descriptor_tables[0u] + uint(floatBitsToInt(r0.x)))], ivec2(0), 0)
```

Root descriptors were also reworked to always use a separate set, even if `VK_KHR_push_descriptor` is not supported. This simplifies the design and may improve performance, since root descriptors likely change more frequently than descriptor tables. Static samplers also use a separate set now.

The packed descriptor set, root descriptor set and static sampler set are optional and will not be created if not needed by the root signature.

## Current issues
- It is not entirely clear which IDs we need to decorate with `NonUniformEXT` (is there any documentation for this?). RADV seems to be happy with pretty much anything, whereas AMDVLK requires it on both the index and the resource ID (of `OpTypeImage`/`OpTypeSampler`).
Decorating the pointer returned by `OpAccessChain` does *not* work, which complicates the design and causes some inevitable code duplication.
- We currently only enable bindless if the device reports `RESOURCE_BINDING_TIER_2` since we have no easy access to the device features and properties. A possible solution is to store a `vkd3d_physical_device_info` struct in `d3d12_device`.
- There is currently no clear path towards supporting bindless UAV counters. The original idea was to add a single SSBO descriptor to one of the bindless set layouts which would then contain the raw VAs, however this complicates the design significantly since we need to keep track of which binding to update etc. Another option which does not increase the overall set count would be to add it to the root descriptor set if necessary.
- We need a way to find the descriptor heap from a `D3D12_CPU_DESCRIPTOR_HANDLE`. The current (and somewhat ugly) solution is to initialize each descriptor with a pointer to the heap.
- There are still a few things I'd like to clean up, e.g. the way the descriptor table offset is calculated.

## Future work
- Inline uniform blocks should be used on implementations that only support 128 bytes of push constant space.
- Static samplers should be cached so that we don't exceed the limit on the number of samplers that may exist at a time. D3D12 allows up to [2032](https://docs.microsoft.com/en-us/windows/win32/direct3d12/hardware-support) unique static samplers.
- We may want to allocate static sampler descriptor sets from a global pool using `VK_DESCRIPTOR_POOL_CREATE_FREE_DESCRIPTOR_SET_BIT`. Currently we allocate a new set each time the root signature changes.
- Support volatile descriptor ranges properly for the packed descriptor set.
- Support dynamic descriptor indexing within the packed descriptor set.
- Use `VK_EXT_buffer_address` for root descriptors to replace the root descriptor set.
- Support bindless UAVs and UAV counters.
- Support bindless CBVs on hardware that can support it.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -1860,7 +1860,7 @@ static void d3d12_command_list_invalidate_current_render_pass(struct d3d12_comma
 }
 
 static void d3d12_command_list_invalidate_root_parameters(struct d3d12_command_list *list,
-        VkPipelineBindPoint bind_point)
+        VkPipelineBindPoint bind_point, bool invalidate_descriptor_heaps)
 {
     struct vkd3d_pipeline_bindings *bindings = &list->pipeline_bindings[bind_point];
 
@@ -1873,6 +1873,12 @@ static void d3d12_command_list_invalidate_root_parameters(struct d3d12_command_l
     bindings->root_constant_dirty_mask = bindings->root_signature->root_constant_mask;
 
     bindings->static_sampler_set_dirty = bindings->static_sampler_set != VK_NULL_HANDLE;
+
+    if (invalidate_descriptor_heaps)
+    {
+        struct d3d12_device *device = bindings->root_signature->device;
+        bindings->descriptor_heap_dirty_mask = (1ull << device->bindless_state.set_count) - 1;
+    }
 }
 
 static bool vk_barrier_parameters_from_d3d12_resource_state(unsigned int state, unsigned int stencil_state,
@@ -4099,7 +4105,7 @@ static void d3d12_command_list_set_root_signature(struct d3d12_command_list *lis
                 list->allocator, root_signature->vk_sampler_descriptor_layout);
     }
 
-    d3d12_command_list_invalidate_root_parameters(list, bind_point);
+    d3d12_command_list_invalidate_root_parameters(list, bind_point, false);
 }
 
 static void STDMETHODCALLTYPE d3d12_command_list_SetComputeRootSignature(ID3D12GraphicsCommandList2 *iface,
@@ -4885,7 +4891,7 @@ static void d3d12_command_list_clear_uav(struct d3d12_command_list *list,
     d3d12_command_list_end_current_render_pass(list);
 
     d3d12_command_list_invalidate_current_pipeline(list);
-    d3d12_command_list_invalidate_root_parameters(list, VK_PIPELINE_BIND_POINT_COMPUTE);
+    d3d12_command_list_invalidate_root_parameters(list, VK_PIPELINE_BIND_POINT_COMPUTE, true);
 
     if (!d3d12_command_allocator_add_view(list->allocator, view))
         WARN("Failed to add view.\n");
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -2342,6 +2342,7 @@ static void d3d12_command_list_reset_state(struct d3d12_command_list *list,
     list->current_render_pass = VK_NULL_HANDLE;
 
     memset(list->pipeline_bindings, 0, sizeof(list->pipeline_bindings));
+    memset(list->descriptor_heaps, 0, sizeof(list->descriptor_heaps));
 
     list->state = NULL;
 
@@ -2762,6 +2763,26 @@ static bool vk_write_descriptor_set_from_root_descriptor(VkWriteDescriptorSet *v
     return true;
 }
 
+static void d3d12_command_list_update_descriptor_heaps(struct d3d12_command_list *list,
+        VkPipelineBindPoint bind_point)
+{
+    struct vkd3d_pipeline_bindings *bindings = &list->pipeline_bindings[bind_point];
+    const struct d3d12_root_signature *root_signature = bindings->root_signature;
+    const struct vkd3d_vk_device_procs *vk_procs = &list->device->vk_procs;
+
+    while (bindings->descriptor_heap_dirty_mask)
+    {
+        unsigned int heap_index = vkd3d_bitmask_iter64(&bindings->descriptor_heap_dirty_mask);
+
+        if (list->descriptor_heaps[heap_index])
+        {
+            VK_CALL(vkCmdBindDescriptorSets(list->vk_command_buffer, bind_point,
+                root_signature->vk_pipeline_layout, heap_index, 1,
+                &list->descriptor_heaps[heap_index], 0, NULL));
+        }
+    }
+}
+
 static void d3d12_command_list_update_static_samplers(struct d3d12_command_list *list,
         VkPipelineBindPoint bind_point)
 {
@@ -2863,6 +2884,9 @@ static void d3d12_command_list_update_descriptors(struct d3d12_command_list *lis
 
     if (!rs)
         return;
+
+    if (bindings->descriptor_heap_dirty_mask)
+        d3d12_command_list_update_descriptor_heaps(list, bind_point);
 
     if (bindings->static_sampler_set_dirty)
         d3d12_command_list_update_static_samplers(list, bind_point);
@@ -4012,12 +4036,33 @@ static void STDMETHODCALLTYPE d3d12_command_list_ExecuteBundle(ID3D12GraphicsCom
 static void STDMETHODCALLTYPE d3d12_command_list_SetDescriptorHeaps(ID3D12GraphicsCommandList2 *iface,
         UINT heap_count, ID3D12DescriptorHeap *const *heaps)
 {
+    struct d3d12_command_list *list = impl_from_ID3D12GraphicsCommandList2(iface);
+    struct vkd3d_bindless_state *bindless_state = &list->device->bindless_state;
+    unsigned int i, j, set_index;
+    uint64_t dirty_mask = 0;
+
     TRACE("iface %p, heap_count %u, heaps %p.\n", iface, heap_count, heaps);
 
-    /* Our current implementation does not need this method.
-     *
-     * It could be used to validate descriptor tables but we do not have an
-     * equivalent of the D3D12 Debug Layer. */
+    for (i = 0; i < heap_count; i++)
+    {
+        struct d3d12_descriptor_heap *heap = unsafe_impl_from_ID3D12DescriptorHeap(heaps[i]);
+
+        if (!heap)
+            continue;
+
+        for (j = 0; j < bindless_state->set_count; j++)
+        {
+            if (bindless_state->set_info[j].heap_type != heap->desc.Type)
+                continue;
+
+            set_index = d3d12_descriptor_heap_set_index_from_binding(&bindless_state->set_info[j]);
+            list->descriptor_heaps[j] = heap->vk_descriptor_sets[set_index];
+            dirty_mask |= 1ull << j;
+        }
+    }
+
+    for (i = 0; i < ARRAY_SIZE(list->pipeline_bindings); i++)
+        list->pipeline_bindings[i].descriptor_heap_dirty_mask = dirty_mask;
 }
 
 static void d3d12_command_list_set_root_signature(struct d3d12_command_list *list,
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -2661,6 +2661,7 @@ static void d3d12_command_list_update_packed_descriptors(struct d3d12_command_li
     VkDescriptorSet descriptor_set = VK_NULL_HANDLE;
     union vkd3d_descriptor_info *vk_descriptor;
     const struct d3d12_desc *base_descriptor;
+    uint32_t table_offsets[D3D12_MAX_ROOT_COST];
     unsigned int root_parameter_index, i, j;
     unsigned int write_count = 0;
 
@@ -2687,6 +2688,8 @@ static void d3d12_command_list_update_packed_descriptors(struct d3d12_command_li
 
         table = root_signature_get_descriptor_table(root_signature, root_parameter_index);
         vk_descriptor = &updates->descriptors[table->first_packed_descriptor];
+
+        table_offsets[table->table_index] = d3d12_desc_heap_offset(base_descriptor);
 
         for (i = 0; i < table->binding_count; i++)
         {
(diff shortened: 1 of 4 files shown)


include/private/vkd3d_common.h
```diff
@@ -81,6 +82,36 @@ static inline bool vkd3d_bitmask_is_contiguous(unsigned int mask)
...
+    r -= (mask & 0x5555555555555555ull) ?  1 : 0;
+    return mask ? r : 64;
+#endif
+}
+
+/* find least significant bit, then remove that bit from mask */
+static inline unsigned int vkd3d_bitmask_iter64(uint64_t* mask)
+{
+    uint64_t cur_mask = *mask;
+    *mask = cur_mask & (cur_mask - 1);
```
HansKristian-Work: Is this correct? I would expect something like
```
*mask = ~(1ull << tzcnt(*mask));
```

include/private/vkd3d_common.h
```diff
@@ -81,6 +82,36 @@ static inline bool vkd3d_bitmask_is_contiguous(unsigned int mask)
...
 }
 
+/* Returns 64 for mask == 0 */
+static inline unsigned int vkd3d_bitmask_tzcnt64(uint64_t mask)
+{
+#ifdef _MSC_VER
+    unsigned long result;
+    _BitScanForward64(&result, mask) ? result : 64;
+    return result;
+#elif defined (HAVE_BUILTIN_CTZLL)
```
HansKristian-Work: Is this activated in autoconf?

doitsujin: Apparently I forgot to commit the change.

include/private/vkd3d_common.h
doitsujin: The idea is to clear the lowest set bit and leave the upper bits intact, which is what this does.

HansKristian-Work: That's what I figured, but I have a hard time seeing how that mask does that, it is some classic "hackers delight" trick? I've never seen that way of doing a bitloop.

HansKristian-Work: A comment explaining what the trick is doing is probably good enough.

HansKristian-Work:
vkd3d: Rework descriptor table binding refers to a descriptor update template. I don't see that being used. Thought we dropped it?

libs/vkd3d/command.c
```diff
@@ -2538,176 +2544,186 @@ static bool d3d12_command_list_update_graphics_pipeline(struct d3d12_command_lis
...
-    struct d3d12_desc *descriptor;
+    const struct d3d12_root_descriptor_table *table;
+    VkDescriptorSet descriptor_set = VK_NULL_HANDLE;
+    union vkd3d_descriptor_info *vk_descriptor;
+    const struct d3d12_desc *base_descriptor;
+    unsigned int root_parameter_index, i, j;
+    unsigned int write_count = 0;
 
-    descriptor_table = root_signature_get_descriptor_table(root_signature, index);
+    if (!vkd3d_descriptor_updates_resize_arrays(updates, root_signature->packed_descriptor_count))
```
HansKristian-Work: I suppose for RS 1.0 we can just append new updates here and flush on submit instead of updating immediately?

doitsujin: Haven't thought about that, but yeah, that might actually work.

doitsujin:
> vkd3d: Rework descriptor table binding refers to a descriptor update template. I don't see that being used. Thought we dropped it?

I fixed the commit message.

libs/vkd3d/command.c
doitsujin: Actually, it's not quite as simple and would need a (rather minor) rework of how we populate the array; it still shows that this was originally written with descriptor update templates in mind. But yes, I think this is the right way forward.

libs/vkd3d/vkd3d_private.h
```diff
@@ -618,8 +623,32 @@ struct d3d12_descriptor_heap
...
 struct d3d12_descriptor_heap *unsafe_impl_from_ID3D12DescriptorHeap(ID3D12DescriptorHeap *iface) DECLSPEC_HIDDEN;
 
+static inline unsigned int d3d12_descriptor_heap_sampler_set_index()
+{
+    return 0;
+}
+
+static inline unsigned int d3d12_descriptor_heap_cbv_set_index()
+{
+    return 0;
```
HansKristian-Work: Is this because we haven't really bothered hooking up CBV bindless yet?

libs/vkd3d/vkd3d_private.h
```diff
@@ -618,8 +623,32 @@ struct d3d12_descriptor_heap
...
+}
+
+static inline unsigned int d3d12_descriptor_heap_srv_set_index(bool is_buffer)
+{
+    return 1 + (is_buffer ? 0 : 1);
+}
+
+static inline unsigned int d3d12_descriptor_heap_uav_set_index(bool is_buffer)
+{
+    return 3 + (is_buffer ? 0 : 1);
```
HansKristian-Work: I suppose there is no argument for counters since we don't bother with bindless UAV counters yet?

libs/vkd3d/command.c
doitsujin: Nevermind, doesn't work. We need to pull in the correct *descriptors* at submission time, so nothing really changes from your previous attempt at implementing it.

libs/vkd3d/vkd3d_private.h
doitsujin: These are intended to be somewhat hard-coded (but therefore, very fast) functions to give us access to a descriptor set stored with a descriptor heap. The exact numbers don't really matter. This wouldn't change with bindless CBV.

libs/vkd3d/vkd3d_private.h
```diff
@@ -647,6 +647,12 @@ static inline unsigned int d3d12_descriptor_heap_uav_set_index(bool is_buffer)
     return 3 + (is_buffer ? 0 : 1);
 }
 
+static inline uint32_t d3d12_desc_heap_offset(const struct d3d12_desc *dst)
+{
+    size_t heap_start = (size_t)dst->heap->descriptors;
+    return ((size_t)dst - heap_start) / sizeof(struct d3d12_desc);
```
HansKristian-Work: Is this sizeof aligned to POT? Integer divide isn't that cool. Alternatively, we can store the offset directly.

libs/vkd3d/vkd3d_private.h
doitsujin: Kind of, in practice (assuming they get a separate set, or are incorporated into an existing set) we'd probably define a new function for UAV counters instead.

libs/vkd3d/vkd3d_private.h
doitsujin: > Is this sizeof aligned to POT?

No, but storing the offset (alongside the pointer to the heap, which we already have anyway) works too.

libs/vkd3d/command.c
```diff
@@ -2725,6 +2728,16 @@ static void d3d12_command_list_update_packed_descriptors(struct d3d12_command_li
                 root_signature->packed_descriptor_set,
                 1, &descriptor_set, 0, NULL));
     }
+
+    /* Set descriptor offsets */
+    if (root_signature->descriptor_table_count)
```
HansKristian-Work: This is called only inside update_packed_descriptors. This seems a bit sketchy if we only changed a root descriptor which uses descriptor tables. I think we technically don't need to update packed descriptors in that case, only push constants.

libs/vkd3d/device.c
```diff
@@ -1350,12 +1353,20 @@ static HRESULT vkd3d_init_device_caps(struct d3d12_device *device,
...
+     * number of samplers to support per set is 500k */
+    if (descriptor_indexing->runtimeDescriptorArray
+            && descriptor_indexing->shaderSampledImageArrayNonUniformIndexing
+            && descriptor_indexing->shaderUniformTexelBufferArrayNonUniformIndexing
+            && descriptor_indexing->descriptorBindingSampledImageUpdateAfterBind
+            && descriptor_indexing->descriptorBindingUpdateUnusedWhilePending
+            && descriptor_indexing->descriptorBindingPartiallyBound
+            && descriptor_indexing->descriptorBindingVariableDescriptorCount
+            && descriptor_indexing_properties->maxPerStageDescriptorUpdateAfterBindSampledImages >= 1000000)
         device->feature_options.ResourceBindingTier = D3D12_RESOURCE_BINDING_TIER_2;
```
HansKristian-Work: We'll probably need to report TIER_3 here anyways because WoW is broken and refuses to use D3D12 unless that is reported. Is there an "app-opt" thing in place we can use for this?

libs/vkd3d/state.c
```diff
@@ -343,42 +370,68 @@ struct d3d12_root_signature_info
...
-    if (range->NumDescriptors == 0xffffffff)
+    if (d3d12_root_signature_descriptor_range_is_bindless(device, range->RangeType))
     {
-        FIXME("Unhandled unbound descriptor range.\n");
-        return E_NOTIMPL;
+        switch (range->RangeType)
+        {
+            case D3D12_DESCRIPTOR_RANGE_TYPE_SRV:
+                /* separate image + buffer descriptors */
+                info->binding_count += 2;
```
HansKristian-Work: Probably best to set descriptor_count here and increment outside the branch. The two branches here to the same thing in different ways.

libs/vkd3d/device.c
doitsujin: > Is there an "app-opt" thing in place we can use for this?

No. I guess the best thing to do for now is to keep the old code for this and instead check for bindless features in `vkd3d_bindless_state_get_bindless_flags`.

HansKristian-Work: That works, please add comments explaining the situation in that case.

libs/vkd3d/state.c
doitsujin: The difference is that bindings and packed descriptors are decoupled here. Can of course introduce a separate `binding_count`, but this will lead to some duplication in the second branch.

HansKristian-Work: Hm, right, nvm.

libs/vkd3d/command.c
doitsujin: True, on the other hand we currently have no way to check whether a table has packed descriptors or not.

HansKristian-Work: If any SetDescriptorTable call will invalidate the packed set, I suppose this is fine, but I'd like a TODO comment explaining what optimization we can do later so we don't forget we need to move this out.

libs/vkd3d/state.c
```diff
@@ -712,9 +807,11 @@ static HRESULT d3d12_root_signature_init_static_samplers(struct d3d12_root_signa
 static HRESULT d3d12_root_signature_init(struct d3d12_root_signature *root_signature,
         struct d3d12_device *device, const D3D12_ROOT_SIGNATURE_DESC *desc)
 {
+    const struct vkd3d_bindless_state *bindless_state = &device->bindless_state;
+    VkDescriptorSetLayout set_layouts[3 + VKD3D_MAX_BINDLESS_DESCRIPTOR_SETS];
```
HansKristian-Work: If VKD3D_MAX_BINDLESS_DESCRIPTOR_SETS is 6, wouldn't this overflow the limit of 8?

doitsujin: This is really just here so that we have a large enough array. I can hard-code it to 8 if you want.

We'll never have >8 sets in practice,. since if everything is bindless, there will be no packed descriptor set.

HansKristian-Work: Defining VKD3D_MAX_DESCRIPTORS to 8 is probably fine.

libs/vkd3d/state.c
```diff
@@ -2764,8 +2764,22 @@ static HRESULT vkd3d_bindless_state_add_binding(struct vkd3d_bindless_state *bin
...
+    const struct vkd3d_vulkan_info *vk_info = &device->vk_info;
+    uint32_t flags = 0;
+
+    if (!vk_info->EXT_descriptor_indexing ||
+            !device_info->descriptor_indexing_features.runtimeDescriptorArray ||
+            !device_info->descriptor_indexing_features.descriptorBindingPartiallyBound ||
+            !device_info->descriptor_indexing_features.descriptorBindingVariableDescriptorCount)
+        return 0;
+
+    if (device_info->descriptor_indexing_properties.maxPerStageDescriptorUpdateAfterBindSampledImages >= 1000000 &&
```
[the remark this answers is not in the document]
doitsujin: Yeah, fixed.

libs/vkd3d/state.c
doitsujin: Done.

libs/vkd3d/device.c
doitsujin: Done.

libs/vkd3d/command.c
doitsujin: Added a comment.

I guess for an implementation, we could rework this further and not track each individual table, instead just adding a `packed_descriptors_dirty` and `descriptor_tables_dirty` flag similar to the static samplers one, and store whether a table has packed descriptors with the root signature.

HansKristian-Work: ye, I think that'd be a good thing to do since it's unlikely UAV/CBV ranges will overlap with SRVs for bindless scenarios. But it doesn't block this PR.

PR #21, opened 2020-03-09
Binding model: Shaders

doitsujin:
As requested.

### DIFF
--- a/libs/vkd3d-shader/spirv.c
+++ b/libs/vkd3d-shader/spirv.c
@@ -7365,6 +7365,68 @@ static const struct vkd3d_symbol *vkd3d_dxbc_compiler_find_resource(struct vkd3d
     return RB_ENTRY_VALUE(entry, struct vkd3d_symbol, entry);
 }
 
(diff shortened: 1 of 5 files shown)


libs/vkd3d-shader/spirv.c
```diff
@@ -7220,6 +7395,81 @@ static const struct vkd3d_symbol *vkd3d_dxbc_compiler_find_resource(struct vkd3d
...
+        ptr_type_id = vkd3d_spirv_get_op_type_pointer(builder,
+                symbol->info.resource.storage_class,
+                symbol->info.resource.type_id);
+
+        ptr_id = vkd3d_spirv_build_op_access_chain(builder,
+                ptr_type_id, ptr_id, &index_id, 1);
+    }
+
+    /* AMD drivers rely on the index being marked as nonuniform */
+    if (reg->modifier == VKD3DSPRM_NONUNIFORM)
```
HansKristian-Work: This is somewhat sketchy since index_id can be uninitialized. Shouldn't this only happen inside the branch?

doitsujin: Fixed.

doitsujin:
I added the WoW workaround for now, it also fixes SotTR.

We should probably test how this behaves on Windows though.

PR #25, opened 2020-03-10
Binding model: Root signature refactor part 4

doitsujin:
As requested.

### DIFF
--- a/libs/vkd3d/state.c
+++ b/libs/vkd3d/state.c
@@ -289,21 +289,6 @@ static enum vkd3d_shader_descriptor_type vkd3d_descriptor_type_from_d3d12_root_p
     }
 }
 
-static bool vk_binding_from_d3d12_descriptor_range(struct VkDescriptorSetLayoutBinding *binding_desc,
-        const D3D12_DESCRIPTOR_RANGE *descriptor_range, D3D12_SHADER_VISIBILITY shader_visibility,
-        bool is_buffer, uint32_t vk_binding)
-{
-    binding_desc->binding = vk_binding;
-    binding_desc->descriptorType
-            = vk_descriptor_type_from_d3d12_range_type(descriptor_range->RangeType, is_buffer);
-    binding_desc->descriptorCount = 1;
-
-    binding_desc->stageFlags = stage_flags_from_visibility(shader_visibility);
-    binding_desc->pImmutableSamplers = NULL;
-
-    return true;
-}
-
 static HRESULT vkd3d_create_descriptor_set_layout(struct d3d12_device *device,
         VkDescriptorSetLayoutCreateFlags flags, unsigned int binding_count,
         const VkDescriptorSetLayoutBinding *bindings, VkDescriptorSetLayout *set_layout)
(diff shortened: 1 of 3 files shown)


libs/vkd3d/state.c
```diff
@@ -580,9 +516,15 @@ static HRESULT d3d12_root_signature_init_root_descriptor_tables(struct d3d12_roo
 
         table = &root_signature->parameters[i].u.descriptor_table;
         range_count = p->u.DescriptorTable.NumDescriptorRanges;
+        range_descriptor_offset = 0;
 
         root_signature->parameters[i].parameter_type = p->ParameterType;
         table->range_count = range_count;
+        table->table_index = t++;
+        table->binding_count = 0;
+        table->first_binding = &root_signature->bindings[context->binding_index];
```
HansKristian-Work: What is the purpose of first_binding and first_packed_descriptor exactly? I assume this is to hold a range of all CBV/UAV descriptors which don't participate in bindless later?

libs/vkd3d/state.c
```diff
@@ -593,58 +535,84 @@ static HRESULT d3d12_root_signature_init_root_descriptor_tables(struct d3d12_roo
...
+            binding.shader_visibility = vkd3d_shader_visibility_from_d3d12(p->ShaderVisibility);
+            binding.binding.set = context->vk_set;
+            binding.binding.binding = context->vk_binding;
 
             /* Unroll descriptor range. */
             for (k = 0; k < range->NumDescriptors; ++k)
             {
-                uint32_t vk_current_binding = vk_binding + k;
-
+                VkDescriptorSetLayoutBinding vk_binding;
```
HansKristian-Work: Why unroll when we have descriptor_count now propagated to shader compiler?

libs/vkd3d/state.c
doitsujin: Yes, this is used to iterate over bindings to populate the packed descriptor set.

libs/vkd3d/state.c
doitsujin: We do not support indexing on packed descriptors in the shader compiler.

libs/vkd3d/state.c
HansKristian-Work: And context->binding_index is currently referring to what? A certain number of descriptors of some other type?

libs/vkd3d/state.c
HansKristian-Work: Ok, so all SRVs currently enter this path. I assume the idea is that further commits will only enter this loop for UAV/CBV?

doitsujin: Yes, bindless doesn't use this code path.

libs/vkd3d/state.c
```diff
@@ -580,9 +516,15 @@ static HRESULT d3d12_root_signature_init_root_descriptor_tables(struct d3d12_roo
 
         table = &root_signature->parameters[i].u.descriptor_table;
         range_count = p->u.DescriptorTable.NumDescriptorRanges;
+        range_descriptor_offset = 0;
 
         root_signature->parameters[i].parameter_type = p->ParameterType;
         table->range_count = range_count;
+        table->table_index = t++;
```
HansKristian-Work: Is this used so that when you call SetDescriptorTable() you can transform parameter_index to push constant offset?

libs/vkd3d/state.c
```diff
@@ -593,58 +535,84 @@ static HRESULT d3d12_root_signature_init_root_descriptor_tables(struct d3d12_roo
...
+                vk_binding.binding = binding.binding.binding = context->vk_binding++;
+                vk_binding.descriptorType = vk_descriptor_type_from_d3d12_range_type(range->RangeType, true);
+                vk_binding.descriptorCount = 1;
+                vk_binding.stageFlags = stage_flags_from_visibility(p->ShaderVisibility);
+                vk_binding.pImmutableSamplers = NULL;
+
+                binding.register_index = range->BaseShaderRegister + k;
+                binding.register_count = 1;
+                binding.descriptor_offset = range_descriptor_offset + k;
+                binding.flags = range->RangeType == D3D12_DESCRIPTOR_RANGE_TYPE_SAMPLER
```
HansKristian-Work: SAMPLER?

libs/vkd3d/state.c
doitsujin: It's basically just the current index into the `root_signature->bindings` array. The size is pre-computed, which makes this a bit ugly, but we need to know where to write to.

libs/vkd3d/state.c
doitsujin: Yes.

libs/vkd3d/state.c
HansKristian-Work: Oh, right, it's updated inside the loop.

libs/vkd3d/state.c
doitsujin: `SAMPLER` requires the `IMAGE` flag set in the shader compiler (this was true before the bindless changes as well) and is the only descriptor type that doesn't have a `BUFFER` variant.

PR #28, opened 2020-03-10
vkd3d: Track dirty bindings more accurately.

doitsujin:
Decouples descriptor table offset updates from updating the packed descriptor set and only updates the latter if necessary.


doitsujin:
Reported to cause a crash in Metro Exodus.

PR #35, opened 2020-03-17
vkd3d-shader: Rework constant buffer loads and root constant layout

doitsujin:
Unrolls the root constant buffer to something like the following:
```glsl
layout(push_constant)
uniform push_cb_t {
  float cb0_0;
  float cb0_1;
  float cb1_0;
  float cb1_1;
  float cb1_2;
  float cb1_3;
  uint table0;
  uint table1;
} push_cb;
```
This fixes issues where we would have misaligned and overlapping vec4 variables in the push constant buffer.

### DIFF
--- a/libs/vkd3d-shader/spirv.c
+++ b/libs/vkd3d-shader/spirv.c
@@ -3197,12 +3197,16 @@ static uint32_t vkd3d_dxbc_compiler_emit_load_constant_buffer(struct vkd3d_dxbc_
     assert(!reg->idx[0].rel_addr);
 
     last_index = 0;
-    if (register_info->cbv_binding && (register_info->cbv_binding->flags & VKD3D_SHADER_BINDING_FLAG_BINDLESS))
-        indexes[last_index++] = vkd3d_dxbc_compiler_get_resource_index(compiler, reg, register_info->cbv_binding);
 
-    indexes[last_index++] = vkd3d_dxbc_compiler_get_constant_uint(compiler, register_info->member_idx);
-    indexes[last_index++] = vkd3d_dxbc_compiler_emit_register_addressing(compiler,
-            &reg->idx[shader_is_sm_5_1(compiler) ? 2 : 1]);
+    if (register_info->cbv_binding)
+    {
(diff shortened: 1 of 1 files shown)


tests/d3d12.c
```diff
@@ -39195,99 +39195,99 @@ START_TEST(d3d12)
...
-    run_test(test_null_uav);
-    run_test(test_null_vbv);
-    run_test(test_get_copyable_footprints);
-    run_test(test_depth_clip);
-    run_test(test_depth_stencil_sampling);
-    run_test(test_depth_load);
-    run_test(test_depth_read_only_view);
-    run_test(test_stencil_load);
-    run_test(test_typed_buffer_uav);
+    // run_test(test_create_device);
```
HansKristian-Work: Hm?

libs/vkd3d-shader/spirv.c
```diff
@@ -3199,6 +3185,73 @@ static uint32_t vkd3d_dxbc_compiler_emit_load_scalar(struct vkd3d_dxbc_compiler
...
+            continue;
+
+        if (register_info->cbv_binding)
+        {
+            indexes[last_index] = vkd3d_dxbc_compiler_get_constant_uint(compiler, vkd3d_swizzle_get_component(swizzle, i));
+        }
+        else
+        {
+            /* Root constants are unrolled to individual floats */
+            indexes[last_index] = vkd3d_dxbc_compiler_get_constant_uint(compiler,
```
HansKristian-Work: Since we're unrolling members, we probably need to clamp the member index here for robustness?

tests/d3d12.c
doitsujin: Accident.

libs/vkd3d-shader/spirv.c
doitsujin: Hm, should we maybe return `OpUndef` if the index is out of bounds?

HansKristian-Work: Either that, or 0. Not sure how OOB for root constants are specified. But the most common case here is when a CBV only contains 8 bytes and app only accesses 8 bytes, but DXBC still deals in terms of 16 byte loads.

PR #37, opened 2020-03-18
vkd3d: Implement bindless CBV using SSBOs if necessary

doitsujin:
Older Nvidia generations up to Pascal don't support bindless uniform buffers.

Note that we don't expose `RESOURCE_BINDING_TIER_3` just yet on those GPUs since the code to determine d3d12 caps should probably be reworked as well. What should we check for?

Maybe something like the following:
- No bindless -> Tier 1
- Bindless SRV+Samplers -> Tier 2
- Bindless everything -> Tier 3


HansKristian-Work:
As for TIER_3 I think we'll need bindless UAV as well, but roughly, that's what we should expose I think.

PR #40, opened 2020-03-23
Do not attempt to enable primitive restart for non-strip topology

HansKristian-Work:


libs/vkd3d/state.c
```diff
@@ -2703,7 +2720,8 @@ VkPipeline d3d12_pipeline_state_get_or_create_pipeline(struct d3d12_pipeline_sta
     ia_desc.pNext = NULL;
     ia_desc.flags = 0;
     ia_desc.topology = vk_topology_from_d3d12_topology(topology);
-    ia_desc.primitiveRestartEnable = !!graphics->index_buffer_strip_cut_value;
+    ia_desc.primitiveRestartEnable = !!graphics->index_buffer_strip_cut_value &&
```
doitsujin: Could get rid of the `!!` since `&&` just checks for nonzero anyway.

PR #47, opened 2020-03-25
vkd3d: Clamp RTV/DSV array sizes to their real array lengths.

HansKristian-Work:
Avoids overflowing vkCreateFramebuffer later with UINT_MAX layers.

### DIFF
--- a/libs/vkd3d/resource.c
+++ b/libs/vkd3d/resource.c
@@ -3100,9 +3100,12 @@ static void vkd3d_create_texture_uav(struct d3d12_desc *descriptor,
             case D3D12_UAV_DIMENSION_TEXTURE3D:
                 vkd3d_desc.view_type = VK_IMAGE_VIEW_TYPE_3D;
                 vkd3d_desc.miplevel_idx = desc->u.Texture3D.MipSlice;
-                if (desc->u.Texture3D.FirstWSlice || desc->u.Texture3D.WSize != resource->desc.DepthOrArraySize)
+                if (desc->u.Texture3D.FirstWSlice ||
+                    ((desc->u.Texture3D.WSize != resource->desc.DepthOrArraySize) && (desc->u.Texture3D.WSize != UINT_MAX)))
+                {
                     FIXME("Unhandled depth view %u-%u.\n",
-                            desc->u.Texture3D.FirstWSlice, desc->u.Texture3D.WSize);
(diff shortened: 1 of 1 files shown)


libs/vkd3d/resource.c
```diff
@@ -3344,22 +3348,26 @@ void d3d12_rtv_desc_create_rtv(struct d3d12_rtv_desc *rtv_desc, struct d3d12_dev
                 vkd3d_desc.miplevel_idx = desc->u.Texture2DArray.MipSlice;
                 vkd3d_desc.layer_idx = desc->u.Texture2DArray.FirstArraySlice;
                 vkd3d_desc.layer_count = desc->u.Texture2DArray.ArraySize;
+                vkd3d_desc.layer_count = min(vkd3d_desc.layer_count, resource->desc.DepthOrArraySize - vkd3d_desc.layer_idx);
```
doitsujin: I think this would be more readable if there this came after the switch block since it applies to all arrayed types and shouldn't have any effect on non-arrayed types, or as `min(desc->u.Texture2DArray.ArraySize, resource->desc.DepthOrArraySize - vkd3d_desc.layer_idx)` etc.

libs/vkd3d/resource.c
```diff
@@ -3435,20 +3443,24 @@ void d3d12_dsv_desc_create_dsv(struct d3d12_dsv_desc *dsv_desc, struct d3d12_dev
...
             case D3D12_DSV_DIMENSION_TEXTURE2D:
                 vkd3d_desc.miplevel_idx = desc->u.Texture2D.MipSlice;
+                vkd3d_desc.layer_count = 1;
                 break;
             case D3D12_DSV_DIMENSION_TEXTURE2DARRAY:
                 vkd3d_desc.view_type = VK_IMAGE_VIEW_TYPE_2D_ARRAY;
                 vkd3d_desc.miplevel_idx = desc->u.Texture2DArray.MipSlice;
                 vkd3d_desc.layer_idx = desc->u.Texture2DArray.FirstArraySlice;
                 vkd3d_desc.layer_count = desc->u.Texture2DArray.ArraySize;
+                vkd3d_desc.layer_count = min(vkd3d_desc.layer_count, resource->desc.DepthOrArraySize - vkd3d_desc.layer_idx);
```
doitsujin: Same here and below.

libs/vkd3d/resource.c

libs/vkd3d/resource.c

PR #48, opened 2020-03-25
vkd3d: Disable index_buffer_strip_cut spam in normal runtime.

HansKristian-Work:
Demote this warning to TRACE only.

Control spams this alot.


doitsujin:
Do we know how Windows behaves? If we match that then I'd vote for nuking this warning entirely, at least for the case that we can handle correctly (DX:MD also spams this *a lot*).

HansKristian-Work:
Strip restart index is supposed to behave like the API wants (32-bit indices with 0xffff restart should work, which we cannot support), but of course there were lots of driver bugs here. I don't think it's all that meaningful though.

doitsujin:
What about the 0xFFFFFFFF case with uint16? Does that effectively disable primitive restart or just cut on 0xFFFF?

Which case does Control complain about?

HansKristian-Work:
I'll make an issue to make some focused tests for this case.

PR #49, opened 2020-03-25
vkd3d: Keep a weak reference to resources in views.

HansKristian-Work:
This way we avoid binding descriptors which resources have since been
freed. This can happen in CopyDescriptors and packed descriptor set
update, which will call vkUpdateDescriptors with a potentially dead
VkImage/VkBuffer backing the VkImageView/VkBufferView in d3d12_desc.

This mostly only affects validation layers now. I haven't seen a GPU crash yet because of this, so not sure if we should merge it, or keep it behind debug only runtime, so making this a draft PR for now.

### DIFF
--- a/libs/vkd3d/resource.c
+++ b/libs/vkd3d/resource.c
@@ -1106,6 +1106,10 @@ static void d3d12_resource_destroy(struct d3d12_resource *resource, struct d3d12
 {
     const struct vkd3d_vk_device_procs *vk_procs = &device->vk_procs;
 
+    resource->weak->alive = false;
+    if (!InterlockedDecrement(&resource->weak->refcount))
+        vkd3d_free(resource->weak);
+
     if (resource->flags & VKD3D_RESOURCE_EXTERNAL)
         return;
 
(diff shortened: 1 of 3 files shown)


libs/vkd3d/resource.c
```diff
@@ -2212,6 +2224,13 @@ static void d3d12_desc_update_bindless_descriptor(struct d3d12_desc *dst)
     is_buffer = dst->magic == VKD3D_DESCRIPTOR_MAGIC_CBV ||
             dst->u.view->type == VKD3D_VIEW_TYPE_BUFFER;
 
+    if ((dst->magic == VKD3D_DESCRIPTOR_MAGIC_SRV || dst->magic == VKD3D_DESCRIPTOR_MAGIC_UAV) &&
+        dst->u.view->weak && !dst->u.view->weak->alive)
+    {
+        WARN("Attempting to write descriptor %p which references a freed resource.\n", (const void*)dst);
```
doitsujin: How common is this? Might be a lot of log spam since this is perfectly legal in D3D12, maybe ignore?

libs/vkd3d/resource.c
```diff
@@ -2835,6 +2854,9 @@ static void vkd3d_create_buffer_srv(struct d3d12_desc *descriptor,
     descriptor->magic = VKD3D_DESCRIPTOR_MAGIC_SRV;
     descriptor->vk_descriptor_type = VK_DESCRIPTOR_TYPE_UNIFORM_TEXEL_BUFFER;
     descriptor->u.view = view;
+
+    descriptor->u.view->weak = resource->weak;
+    InterlockedIncrement(&descriptor->u.view->refcount);
```
doitsujin: I'd prefer this to happen inside `vkd3d_create_buffer_view_for_resource` and `vkd3d_create_texture_view`, respectively (the latter might need an - optional - resource parameter, or better yet, introduce `vkd3d_create_texture_view_for_resource` or something like that).

PR #52, opened 2020-03-25
vkd3d: Bindless UAV

doitsujin:
Passes the bindless UAV test *without* counters on both RADV and AMDVLK.

We currently completely ignore UAV counters for unbounded UAV descriptor ranges in order to avoid out of memory errors. This is not a regression as unbounded UAV descriptor ranges were completely unsupported before.

### DIFF
--- a/libs/vkd3d-shader/spirv.c
+++ b/libs/vkd3d-shader/spirv.c
@@ -5635,9 +5635,6 @@ static void vkd3d_dxbc_compiler_emit_resource_declaration(struct vkd3d_dxbc_comp
 
     if (binding && (binding->flags & VKD3D_SHADER_BINDING_FLAG_BINDLESS))
     {
-        if (is_uav && (uav_flags & VKD3D_SHADER_UAV_FLAG_ATOMIC_COUNTER))
-            FIXME("Bindless UAV counters currently not supported.\n");
-
         global_binding = vkd3d_dxbc_compiler_get_global_binding(compiler,
                 is_uav ? VKD3D_DATA_UAV : VKD3D_DATA_RESOURCE, resource_type,
                 sampled_type, storage_class, &binding->binding);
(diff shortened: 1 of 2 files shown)


HansKristian-Work:
From discussion on Discord, needs this patch:
```
diff --git a/libs/vkd3d-shader/dxbc.c b/libs/vkd3d-shader/dxbc.c
index 802a91d..d1c6fa2 100644
--- a/libs/vkd3d-shader/dxbc.c
+++ b/libs/vkd3d-shader/dxbc.c
@@ -1519,6 +1519,7 @@ static bool shader_sm4_read_param(struct vkd3d_sm4_data *priv, const DWORD **ptr
 
             case 0x20001:
                 param->modifier = VKD3DSPRM_NONUNIFORM;
+                *modifier = VKD3DSPDM_NONE;
                 break;
 
             default:
```

HansKristian-Work:
Passes bindless UAV test on NVIDIA as well. Control seems to render without glitches now, going to review / add DXIL support tomorrow. :tada:

libs/vkd3d-shader/spirv.c
```diff
@@ -5091,6 +5091,33 @@ static const struct vkd3d_shader_global_binding *vkd3d_dxbc_compiler_get_global_
             vkd3d_spirv_enable_capability(builder, SpvCapabilitySampledImageArrayNonUniformIndexingEXT);
         }
     }
+    else if (data_type == VKD3D_DATA_UAV)
```
HansKristian-Work: We're missing decorations here such as: Writeonly/Readonly and explicit format for non-typed UAVs. I think we need a unique variable for each variant of these combinations based on what the shader module wants to declare.

HansKristian-Work:
Tested Control, Wow, Metro and SOTTR, no obvious issues.

HansKristian-Work:
DXIL test passes as-is, just pushed a trivial bugfix to dxil-spirv.

PR #57, opened 2020-03-27
Replace descriptor mutex with spinlocks

### LINKED ISSUE #56: Optimize bindless descriptor updates and copies
All the locking currently causes major performance issues in Deus Ex: Mankind Divided, probably others.

HansKristian-Work:
Fix #56.


doitsujin:
Should probably close #56 .

PR #58, opened 2020-03-27
vkd3d: Bindless UAV counters

doitsujin:
Uses `VK_KHR_buffer_device_address` to pass UAV counter addresses to shaders via an SSBO. This SSBO is part of the root descriptor set in order to avoid complexity around descriptor updates.

Passes `test_bindless_uav_counter_sm51` on RADV.

### DIFF
--- a/libs/vkd3d-shader/spirv.c
+++ b/libs/vkd3d-shader/spirv.c
@@ -1710,6 +1710,7 @@ enum vkd3d_spirv_extension
     VKD3D_SPV_KHR_SHADER_DRAW_PARAMETERS        = 0x00000001,
     VKD3D_SPV_EXT_DEMOTE_TO_HELPER_INVOCATION   = 0x00000002,
     VKD3D_SPV_EXT_DESCRIPTOR_INDEXING           = 0x00000004,
+    VKD3D_SPV_KHR_PHYSICAL_STORAGE_BUFFER       = 0x00000008,
 };
 
 struct vkd3d_spirv_extension_info
@@ -1719,6 +1720,7 @@ struct vkd3d_spirv_extension_info
 }
 static const vkd3d_spirv_extensions[] =
 {
+    {VKD3D_SPV_KHR_PHYSICAL_STORAGE_BUFFER,     "SPV_KHR_physical_storage_buffer"},
     {VKD3D_SPV_KHR_SHADER_DRAW_PARAMETERS,      "SPV_KHR_shader_draw_parameters"},
     {VKD3D_SPV_EXT_DEMOTE_TO_HELPER_INVOCATION, "SPV_EXT_demote_to_helper_invocation"},
     {VKD3D_SPV_EXT_DESCRIPTOR_INDEXING,         "SPV_EXT_descriptor_indexing"},
@@ -1731,6 +1733,7 @@ struct vkd3d_spirv_capability_extension_mapping
 }
 static const vkd3d_spirv_capability_extensions[] =
 {
+    {SpvCapabilityPhysicalStorageBufferAddresses,         VKD3D_SPV_KHR_PHYSICAL_STORAGE_BUFFER},
     {SpvCapabilityDrawParameters,                         VKD3D_SPV_KHR_SHADER_DRAW_PARAMETERS},
     {SpvCapabilityDemoteToHelperInvocationEXT,            VKD3D_SPV_EXT_DEMOTE_TO_HELPER_INVOCATION},
     {SpvCapabilityRuntimeDescriptorArrayEXT,              VKD3D_SPV_EXT_DESCRIPTOR_INDEXING},
@@ -5026,7 +5029,7 @@ static const struct vkd3d_shader_global_binding *vkd3d_dxbc_compiler_get_global_
     const struct vkd3d_shader_descriptor_binding *binding_info = &binding->binding;
     struct vkd3d_spirv_builder *builder = &compiler->spirv_builder;
     struct vkd3d_shader_global_binding *current;
-    uint32_t array_type_id, type_id, var_id;
+    uint32_t type_id = 0, var_id = 0;
     unsigned int i;
 
     for (i = 0; i < compiler->global_binding_count; i++)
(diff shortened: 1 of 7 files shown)


libs/vkd3d/resource.c
```diff
@@ -3799,6 +3844,14 @@ static HRESULT d3d12_descriptor_heap_init(struct d3d12_descriptor_heap *descript
         }
     }
 
+    if (desc->Type == D3D12_DESCRIPTOR_HEAP_TYPE_CBV_SRV_UAV &&
```
HansKristian-Work: If the heap is not shader visible, can't we just use a plain vkd3d_malloc instead of mapping a Vulkan buffer?

libs/vkd3d/resource.c
```diff
@@ -3891,8 +3944,10 @@ void d3d12_descriptor_heap_cleanup(struct d3d12_descriptor_heap *descriptor_heap
     const struct vkd3d_vk_device_procs *vk_procs = &descriptor_heap->device->vk_procs;
     const struct d3d12_device *device = descriptor_heap->device;
 
-    VK_CALL(vkDestroyDescriptorPool(device->vk_device,
-            descriptor_heap->vk_descriptor_pool, NULL));
+    VK_CALL(vkDestroyBuffer(device->vk_device, descriptor_heap->uav_counters.vk_buffer, NULL));
+    VK_CALL(vkFreeMemory(device->vk_device, descriptor_heap->uav_counters.vk_memory, NULL));
```
HansKristian-Work: Do we need to unmap buffers before freeing it in Vulkan?

libs/vkd3d/resource.c
```diff
@@ -3048,18 +3071,28 @@ static void vkd3d_create_buffer_uav(struct d3d12_desc *descriptor, struct d3d12_
...
+        if (device->bindless_state.flags & VKD3D_BINDLESS_UAV_COUNTER)
         {
-            WARN("Failed to create counter buffer view.\n");
+            VkDeviceAddress address = vkd3d_get_buffer_device_address(device, counter_resource->u.vk_buffer);
+
             view->vk_counter_view = VK_NULL_HANDLE;
-            d3d12_desc_destroy(descriptor, device);
+            view->vk_counter_address = address + desc->u.Buffer.CounterOffsetInBytes;
+        }
+        else {
```
HansKristian-Work: Formatting.

libs/vkd3d/resource.c
```diff
@@ -3048,18 +3071,28 @@ static void vkd3d_create_buffer_uav(struct d3d12_desc *descriptor, struct d3d12_
...
-        if (!vkd3d_create_vk_buffer_view(device, counter_resource->u.vk_buffer, format,
-                desc->u.Buffer.CounterOffsetInBytes + resource->heap_offset, sizeof(uint32_t), &view->vk_counter_view))
+        if (device->bindless_state.flags & VKD3D_BINDLESS_UAV_COUNTER)
         {
-            WARN("Failed to create counter buffer view.\n");
+            VkDeviceAddress address = vkd3d_get_buffer_device_address(device, counter_resource->u.vk_buffer);
+
             view->vk_counter_view = VK_NULL_HANDLE;
-            d3d12_desc_destroy(descriptor, device);
+            view->vk_counter_address = address + desc->u.Buffer.CounterOffsetInBytes;
```
HansKristian-Work: Need to offset resource->heap_offset here in case the buffer is a placed resource.

libs/vkd3d/resource.c
```diff
@@ -3048,18 +3071,28 @@ static void vkd3d_create_buffer_uav(struct d3d12_desc *descriptor, struct d3d12_
...
+
+            if (!vkd3d_create_vk_buffer_view(device, counter_resource->u.vk_buffer, format,
+                    desc->u.Buffer.CounterOffsetInBytes + resource->heap_offset, sizeof(uint32_t), &view->vk_counter_view))
+            {
+                WARN("Failed to create counter buffer view.\n");
+                view->vk_counter_view = VK_NULL_HANDLE;
+                d3d12_desc_destroy(descriptor, device);
+            }
+
+            view->vk_counter_address = 0;
```
HansKristian-Work: Should write 0 to address as well if we're binding a NULL counter.

libs/vkd3d/resource.c
doitsujin: Actually, good point. We never read from that buffer anyway since the counter address is stored with the view, so I guess the proper solution would be to not create it in the first place.

libs/vkd3d-shader/spirv.c
```diff
@@ -5105,8 +5113,31 @@ static const struct vkd3d_shader_global_binding *vkd3d_dxbc_compiler_get_global_
...
+
+            pointer_struct_id = vkd3d_spirv_build_op_type_struct(builder, &array_type_id, 1);
+
+            vkd3d_spirv_build_op_member_decorate1(builder, pointer_struct_id, 0, SpvDecorationOffset, 0);
+            vkd3d_spirv_build_op_decorate(builder, pointer_struct_id, SpvDecorationBufferBlock, NULL, 0);
+            vkd3d_spirv_build_op_name(builder, pointer_struct_id, "uav_ctrs_t");
+
+            var_id = vkd3d_spirv_build_op_variable(builder, &builder->global_stream,
+                    vkd3d_spirv_get_op_type_pointer(builder, storage_class, pointer_struct_id),
+                    storage_class, 0);
```
HansKristian-Work: One missing piece here is that the SSBO variable should be declared with AliasedPointer decoration since the OpVariable holds physical pointers.

HansKristian-Work: See 2.18.2 in SPIR-V spec

libs/vkd3d/resource.c
doitsujin: No, `vkFreeMemory` says the following:
> If a memory object is mapped at the time it is freed, it is implicitly unmapped.

libs/vkd3d/resource.c
doitsujin: I'll just zero-initialize this in `vkd3d_view_create`, where we already set `vk_counter_view = VK_NULL_HANDLE` since we'd otherwise read undefined values in more cases.

PR #64, opened 2020-03-30
vkd3d: Fix feature enablement for extensions

doitsujin:
RenderDoc will disable `VK_KHR_buffer_device_address` on drivers that do not support the corresponding capture/replay feature, but would still fill in the feature structs, so we were still trying to use it. This would either crash RenderDoc or lead to unusable captures.


doitsujin:
Fixed.

PR #67, opened 2020-03-30
vkd3d: Expose ID3D12Device3

doitsujin:
Stubs for now, but we're going to need this anyway at some point.

Plans for the near-ish future:
- [x] Implement more feature queries
- [x] Implement `CreatePipelineState` so that we can support some newer rendering features at all (such as Depth Bounds). RE Engine games need this method.
- Implement Depth Bounds. RE Engine games may take advantage of this. This will be done as a separate PR.


libs/vkd3d/device.c
```diff
@@ -3528,29 +3549,80 @@ static D3D12_RESOURCE_BINDING_TIER d3d12_device_determine_resource_binding_tier(
...
+                    amd->wavefrontSize;
+
+        /* There doesn't seem to be a way to get the SIMD width,
+         * so just assume 64 lanes per CU. True on GCN and RDNA. */
+        options1->TotalLaneCount = compute_units * 64;
+    }
+    else if (device->vk_info.NV_shader_sm_builtins)
+    {
+        const VkPhysicalDeviceShaderSMBuiltinsPropertiesNV *nv = &device->device_info.shader_sm_builtins_properties;
+        options1->TotalLaneCount = nv->shaderSMCount * nv->shaderWarpsPerSM;
```
HansKristian-Work, line 3914: This needs to be multiplied by 32 I think. This is the number of warps, not threads.

libs/vkd3d/device.c
```diff
@@ -2836,6 +2836,25 @@ static HRESULT STDMETHODCALLTYPE d3d12_device_CheckFeatureSupport(d3d12_device_i
...
+
+            if (feature_data_size != sizeof(*data))
+            {
+                WARN("Invalid size %u.\n", feature_data_size);
+                return E_INVALIDARG;
+            }
+
+            /* FIXME MSDN says SINGLE_PSO is always supported, but we
+             * currently don't support the API, including GetCachedBlob. */
+            WARN("Shader cache features not supported.");
```
HansKristian-Work, line 2864: I think we can only support this when we have more dynamic state to play around with. App might ask for cached pipeline before we have actually compiled anything meaningful.

HansKristian-Work:
Getting a warning, probably benign, but annoying.
```
libs/vkd3d/state.c:2479:12: warning: ‘hr’ may be used uninitialized in this function [-Wmaybe-uninitialized]
 2479 |     return hr;
```

tests/d3d12.c
```diff
@@ -3174,6 +3174,556 @@ static void test_create_graphics_pipeline_state(void)
...
+
+        /* Assign root signature. To keep things simple, assume
+         * that the root signature is always the first element
+         * in each pipeline stream */
+        union d3d12_root_signature_subobject *rs_subobject = tests[i].stream_desc.pPipelineStateSubobjectStream;
+
+        if (rs_subobject && rs_subobject->type == D3D12_PIPELINE_STATE_SUBOBJECT_TYPE_ROOT_SIGNATURE)
+            rs_subobject->root_signature = root_signature;
+
+        hr = ID3D12Device2_CreatePipelineState(device2, &tests[i].stream_desc, &IID_ID3D12PipelineState, &pipeline_state);
```
HansKristian-Work, line 3709: Getting a warning here:
```
/home/maister/git/vkd3d-standalone/vkd3d/tests/d3d12.c:3709:106: warning: passing argument 4 of ‘ID3D12Device2_CreatePipelineState’ from incompatible pointer type [-Wincompatible-pointer-types]
 3709 |         hr = ID3D12Device2_CreatePipelineState(device2, &tests[i].stream_desc, &IID_ID3D12PipelineState, &pipeline_state);
      |                                                                                                          ^~~~~~~~~~~~~~~
      |                                                                                                          |
      |                                                                                                          ID3D12PipelineState ** {aka struct ID3D12PipelineState **}
```

HansKristian-Work: Needs cast to `void**`

libs/vkd3d/device.c
HansKristian-Work: On my 1660 Ti I get 24 for SM count and 32 for Warp count.

doitsujin: Hm, looking at more GPUInfo reports, it actually looks like we have a similar problem as with AMD in that this only tells us the number of warps in flight, not the number of threads per SM. On Pascal/Turing it looks like we want to multiply by 2, on Volta we don't want to multiply by anything, haven't checked Maxwell or Kepler. But yeah, this is inaccurate.

libs/vkd3d/device.c
```diff
@@ -3412,29 +3847,105 @@ static D3D12_RESOURCE_BINDING_TIER d3d12_device_determine_resource_binding_tier(
...
+        const VkPhysicalDeviceShaderCorePropertiesAMD *amd = &device->device_info.shader_core_properties;
+        const VkPhysicalDeviceShaderCoreProperties2AMD *amd2 = &device->device_info.shader_core_properties2;
+        uint32_t compute_units;
+        
+        if (device->vk_info.AMD_shader_core_properties2)
+            compute_units = amd2->activeComputeUnitCount;
+        else
+            compute_units = amd->shaderEngineCount *
+                    amd->computeUnitsPerShaderArray *
+                    amd->wavefrontSize;
```
HansKristian-Work, line 3904: Is wavefrontSize 64 here?

doitsujin: Ah whoops, this shouldn't even be in here. Derp from rebasing.

libs/vkd3d/device.c
HansKristian-Work: In that case we should not look at shaderWarpsPerSM, as that's occupancy, the more natural formulation would be something like `SMCount * subgroupSize * magic 2 factor for Turing`.

PR #68, opened 2020-03-30
Implement VK_KHR_timeline_semaphores

### LINKED ISSUE #65: Look into semaphore error spam
Games like WoW have a lot of spam like:
```
fixme:d3d12_command_queue_Wait: Failed to acquire Vulkan semaphore for fence 0x7f2fd72ffc60, value 0x2ed9, completed value 0x2ed8.
```
It does not seem to affect rendering, but we should investigate if there is some underlying bug.

HansKristian-Work:
Tested on NV with test suite, Metro, Control and WoW.

Queue submit out of order might not work as expected yet since we are currently serializing many virtual queues on top of one queue, but this feature does not seem to be used by much content out there. It would not work at any rate with current vkd3d as it stands anyways.

For now I'm keeping the old implementation around in case we need a fallback for certain drivers, but the goal is to remove the fallback path eventually.

Also, fix the test suite to not rely on non-monotonically increasing values. https://docs.microsoft.com/en-us/windows/win32/direct3d12/user-mode-heap-synchronization#fences specifies it as monotonically increasing, so not sure why tests don't follow that.

Fixes #65.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -945,17 +945,122 @@ static HRESULT d3d12_fence_add_vk_semaphore(struct d3d12_fence *fence,
     return hr;
 }
 
-static HRESULT d3d12_fence_signal_timeline_semaphore(struct d3d12_fence *fence, uint64_t value)
+static void d3d12_fence_signal_external_events_locked(struct d3d12_fence *fence)
+{
+    unsigned int i, j;
+
+    for (i = 0, j = 0; i < fence->event_count; ++i)
+    {
+        struct vkd3d_waiting_event *current = &fence->events[i];
+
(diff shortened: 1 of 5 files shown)


HansKristian-Work:
Apparently, rewinding fences on D3D12 is actually allowed on Win10 (but not documented), Win7 disallows it though ... Might have to rethink a bit how we implement this if we should care about this case.

SveSop:
> Apparently, rewinding fences on D3D12 is actually allowed on Win10 (but not documented), Win7 disallows it though ... Might have to rethink a bit how we implement this if we should care about this case.

So fence-rewinding is not available on Win7, and this could cause a problem for wine? Or is it more for backwards compatibility "in-case-of"?

PR #72, opened 2020-03-31
vkd3d: Require vulkan headers > 1.1.129

SveSop:
VK_KHR_buffer_device_address requires Vulkan headers > 1.1.129

Signed-off-by: Sveinar Søpler <cybermax@dexter.no>


SveSop:
Since i am on a roll checking for header stuff i mean... Should have done it with my first commit, but well.

PR #87, opened 2020-04-15
vkd3d: Check whether root descriptors are defined.

doitsujin:
Fixes one source of crashes in Monster Hunter World.


doitsujin:
Actually we might use `null` resources here instead of leaving the descriptors undefined. Any opinions?

HansKristian-Work:
If app dynamically accesses the descriptor, it should be active. When we get proper null descriptors, we can use those, I'm okay with this for now.

PR #89, opened 2020-04-16
vkd3d: Remove legacy fence/semaphore path.

HansKristian-Work:
Prepares for a rewrite of queue submission, the legacy path is never
run in practice and will likely break in subtle ways.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -19,7 +19,7 @@
 
 #include "vkd3d_private.h"
 
-static HRESULT d3d12_fence_signal(struct d3d12_fence *fence, uint64_t value, VkFence vk_fence);
+static HRESULT d3d12_fence_signal(struct d3d12_fence *fence, uint64_t value);
 
 HRESULT vkd3d_queue_create(struct d3d12_device *device,
         uint32_t family_index, const VkQueueFamilyProperties *properties, struct vkd3d_queue **queue)
@@ -38,18 +38,9 @@ HRESULT vkd3d_queue_create(struct d3d12_device *device,
         return hresult_from_errno(rc);
     }
 
-    object->completed_sequence_number = 0;
-    object->submitted_sequence_number = 0;
-
     object->vk_family_index = family_index;
     object->vk_queue_flags = properties->queueFlags;
     object->timestamp_bits = properties->timestampValidBits;
-
-    object->semaphores = NULL;
-    object->semaphores_size = 0;
-    object->semaphore_count = 0;
-
-    memset(object->old_vk_semaphores, 0, sizeof(object->old_vk_semaphores));
 
     VK_CALL(vkGetDeviceQueue(device->vk_device, family_index, 0, &object->vk_queue));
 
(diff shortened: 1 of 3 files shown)


libs/vkd3d/command.c
```diff
@@ -1305,7 +815,7 @@ static HRESULT STDMETHODCALLTYPE d3d12_fence_Signal(d3d12_fence_iface *iface, UI
     if (fence->timeline_semaphore)
         return d3d12_fence_signal_cpu_timeline_semaphore(fence, value);
     else
-        return d3d12_fence_signal(fence, value, VK_NULL_HANDLE);
+        return d3d12_fence_signal(fence, value);
```
doitsujin, line 818: Is there any situation where `fence->timeline_semaphore` can be `VK_NULL_HANDLE`?

Looking at `d3d12_fence_init`, we should always be creating a timeline semaphore if `timelineSemaphore` is supported, which is now a required feature.
```
    if (device->device_info.timeline_semaphore_features.timelineSemaphore)
        if (FAILED(hr = d3d12_fence_init_timeline(fence, device, initial_value)))
            return hr;
```

libs/vkd3d/command.c
```diff
@@ -814,137 +476,12 @@ static void d3d12_fence_destroy_vk_objects(struct d3d12_fence *fence)
...
-    for (i = 0; i < ARRAY_SIZE(fence->old_vk_fences); ++i)
-    {
-        if (fence->old_vk_fences[i])
-            VK_CALL(vkDestroyFence(device->vk_device, fence->old_vk_fences[i], NULL));
-        fence->old_vk_fences[i] = VK_NULL_HANDLE;
-    }
-
-    d3d12_fence_garbage_collect_vk_semaphores_locked(fence, true);
-
     if (fence->timeline_semaphore)
```
doitsujin, line 479: This check can be removed as well, even if the object is `null`.

libs/vkd3d/device.c
```diff
@@ -4514,6 +4514,14 @@ HRESULT d3d12_device_create(struct vkd3d_instance *instance,
...
     }
 
+    if (!object->vk_info.KHR_timeline_semaphore)
+    {
+        ERR("Timeline semaphores are not supported by this implementation. This is required for correct operation.\n");
+        d3d12_device_destroy(object);
+        vkd3d_free(object);
+        return E_NOTIMPL;
+    }
+
```
doitsujin, line 4524: This feature check should probably be moved into `vkd3d_init_device_caps` and return `E_INVALIDARG` instead of `E_NOTIMPL`, to be consistent with the feature level check.

libs/vkd3d/command.c
HansKristian-Work: Missed it.

HansKristian-Work: Should be fixed.

libs/vkd3d/command.c
HansKristian-Work: Should be fixed now.

libs/vkd3d/device.c
HansKristian-Work: Should be fixed now.

PR #90, opened 2020-04-16
vkd3d: Implement threaded submission queue.

HansKristian-Work:
D3D12 supports out-of-order signal and wait. So does Vulkan timeline
semaphores. However, in Vulkan we don't have an infinite amount of
virtual queues. We must potentially map multiple D3D12 queues on top of
Vulkan, which might lead to a deadlock when app attempts to
wait-before-signal if the two queues are mapped to the same physical
Vulkan queue.

In order to solve this, we need to hold back submissions until we know
it is safe to do so. To make this work in practice as simply as possible, each
ID3D12CommandQueue has its own submission thread, which will block on an
ID3D12Fence's pending timeline value for a Wait command. The main reason to use a
submission thread is that resolving this directly in
ID3D12CommandQueue::Signal is extremely tricky and potentially
needs recursively locking queues and fences.

Note that we only block on the pending wait value, not the actual wait
value, so there is no real CPU <-> GPU synchronization here. In the
common case, no submission thread will block.

The added benefit is that submits are async now, so main thread CPU
overhead might slightly decrease.

To play nice with DXGI swapchain, the external entry point for acquiring
the Vulkan queue needs to drain the submission thread and lock it to ensure
submissions happen in order.

Fixes hangs in The Division 1, which makes use of this D3D12 feature.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -20,6 +20,8 @@
 #include "vkd3d_private.h"
 
 static HRESULT d3d12_fence_signal(struct d3d12_fence *fence, uint64_t value);
+static void d3d12_command_queue_add_submission(struct d3d12_command_queue *queue,
+        const struct d3d12_command_queue_submission *sub);
 
 HRESULT vkd3d_queue_create(struct d3d12_device *device,
         uint32_t family_index, const VkQueueFamilyProperties *properties, struct vkd3d_queue **queue)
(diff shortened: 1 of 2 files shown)


doitsujin:
As discussed, this currently breaks some games:
- Resident Evil 3: Flickering
- Shadow of the Tomb Raider: Random crash / GPU hang

HansKristian-Work:
SOTTR issue only manifests on AMD for some reason.

libs/vkd3d/command.c
```diff
@@ -1515,6 +1517,23 @@ static HRESULT STDMETHODCALLTYPE d3d12_command_allocator_Reset(ID3D12CommandAllo
         TRACE("Resetting command list %p.\n", list);
     }
 
+    if ((pending = atomic_add_fetch(&allocator->outstanding_submissions_count, 0)) != 0)
```
doitsujin, line 1520: don't we have atomic load defined anywhere?

HansKristian-Work:
Found a workaround for yet another SotTR app-bug.

libs/vkd3d/command.c
HansKristian-Work: We could reuse atomic_load from spinlock.h I guess

doitsujin: We might want to clean that up at some point, this works as well for now.

PR #104, opened 2020-04-22
Implement shader-based image copies

doitsujin:
One half of the commits is basically a bunch of cleanups, the other half is the actual implementation.


libs/vkd3d/vkd3d_shaders.h
```diff
@@ -845,4 +845,293 @@ void main() {
...
+    0x00000023,0x0000001c,0x00070050,0x0000000d,0x00000026,0x0000001d,0x00000024,0x00000025,
+    0x0000001c,0x00050041,0x00000027,0x00000028,0x00000013,0x00000014,0x0003003e,0x00000028,
+    0x00000026,0x000100fd,0x00010038
+};
+
+const uint32_t gs_fullscreen_spv[] = {
+#if 0
+#version 450
+
+layout(triangles) in;
```
HansKristian-Work, line 961: Is GS really needed?

doitsujin: We need it to support copying multiple layers at once.

Actually, good point, I think the draw call only does one layer at a time right now.

libs/vkd3d/meta.c
```diff
@@ -335,20 +335,63 @@ VkExtent3D vkd3d_meta_get_clear_image_uav_workgroup_size(VkImageViewType view_ty
     }
 }
 
+static HRESULT vkd3d_meta_ops_common_init(struct vkd3d_meta_ops_common *meta_ops_common, struct d3d12_device *device)
+{
+    VkResult vr;
+
+    if (device->vk_info.EXT_shader_viewport_index_layer)
```
HansKristian-Work, line 778: Ok, now I see why GS was there.

libs/vkd3d/command.c
```diff
@@ -3474,87 +3474,231 @@ static HRESULT d3d12_command_list_allocate_transfer_buffer(struct d3d12_command_
...
+        vk_descriptor_write.pBufferInfo = NULL;
+        vk_descriptor_write.pTexelBufferView = NULL;
+
+        VK_CALL(vkUpdateDescriptorSets(list->device->vk_device, 1, &vk_descriptor_write, 0, NULL));
+
+        for (i = 0; i < ARRAY_SIZE(vk_image_barriers); i++)
+        {
+            vk_image_barriers[i].sType = VK_STRUCTURE_TYPE_IMAGE_MEMORY_BARRIER;
+            vk_image_barriers[i].pNext = NULL;
+            vk_image_barriers[i].srcAccessMask = 0;
```
HansKristian-Work, line 3634: Shouldn't this be VK_ACCESS_COLOR/DEPTH WRITE_BIT?

libs/vkd3d/command.c
```diff
@@ -3474,87 +3474,231 @@ static HRESULT d3d12_command_list_allocate_transfer_buffer(struct d3d12_command_
...
+            vk_image_barriers[0].oldLayout = VK_IMAGE_LAYOUT_UNDEFINED;
+
+        vk_image_barriers[1].oldLayout = src_layout;
+        vk_image_barriers[1].newLayout = VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL;
+        vk_image_barriers[1].dstAccessMask = VK_ACCESS_SHADER_READ_BIT;
+        vk_image_barriers[1].image = src_resource->u.vk_image;
+        vk_image_barriers[1].subresourceRange = vk_subresource_range_from_layers(&region->srcSubresource);
+
+        VK_CALL(vkCmdPipelineBarrier(list->vk_command_buffer,
+                VK_PIPELINE_STAGE_BOTTOM_OF_PIPE_BIT,
```
HansKristian-Work, line 3659: Shouldn't this be TRANSFER_BIT since that's where we expected to consume a copy in the first place?

libs/vkd3d/command.c
doitsujin: The first barrier only performs a layout transition so we shouldn't need it; we set the `*_ATTACHMENT_WRITE_BIT` for the second barrier.

libs/vkd3d/command.c
```diff
@@ -3474,87 +3474,231 @@ static HRESULT d3d12_command_list_allocate_transfer_buffer(struct d3d12_command_
...
+
+        vk_image_barriers[1].oldLayout = VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL;
+        vk_image_barriers[1].newLayout = src_layout;
+
+        VK_CALL(vkCmdPipelineBarrier(list->vk_command_buffer,
+                VK_PIPELINE_STAGE_FRAGMENT_SHADER_BIT |
+                VK_PIPELINE_STAGE_COLOR_ATTACHMENT_OUTPUT_BIT |
+                VK_PIPELINE_STAGE_EARLY_FRAGMENT_TESTS_BIT |
+                VK_PIPELINE_STAGE_LATE_FRAGMENT_TESTS_BIT,
+                VK_PIPELINE_STAGE_TOP_OF_PIPE_BIT,
```
HansKristian-Work, line 3692: Should probably be TRANSFER_BIT as well.

doitsujin: Is that still guaranteed to make the layout transition visible to whatever barrier the application might issue?

HansKristian-Work: The app will have to transition away from COPY_DEST state, in which case we get a transitive barrier.

HansKristian-Work: We block TRANSFER, and when app transitions away from COPY_DEST, it waits on TRANSFER, and we complete the barrier that way.

libs/vkd3d/command.c
HansKristian-Work: I think we need to set ATTACHMENT_WRITE_BIT after the layout transition into ATTACHMENT_OPTIMAL. Memory needs to be made visible.

HansKristian-Work: FWIW, this is dstAccessMask

libs/vkd3d/command.c

libs/vkd3d/command.c
doitsujin: In order to make it visibl we'd have to put it in `dstStageMask` though, and as far as I'm aware, putting write bits there has no effect.

Do we need the `READ` bits? We don't use blending, our depth test always passes and we don't do anything with stencil.

HansKristian-Work: If the renderpass is created with LOAD, then it accesses COLOR_ATTACHMENT_READ. For copy, we should probably use a DONT_CARE render pass.

HansKristian-Work: We're currently doing a TRANSFER -> FRAGMENT execution barrier here, with no dstAccessMask after the layout transition. I believe WRITE must be in dstAccessMask to make sure the writes are ordered (and it cannot write to a stale FB cache or something), but it's a corner case I haven't explored all that much. I'm aware WAR does not require srcAccessMask to be set, but not so sure about dstAccessMask when layout transitions are in the picture.

HansKristian-Work: The question is if memory must be visible if the only thing that happens are write operations.

doitsujin: Added both the read and write bits to be on the safe side, and also fixed that we weren't resetting the `dstAccessMask` for the second barrier.

libs/vkd3d/meta.c
```diff
@@ -324,20 +513,327 @@ VkExtent3D vkd3d_get_clear_image_uav_workgroup_size(VkImageViewType view_type)
...
+    memset(meta_copy_image_ops, 0, sizeof(*meta_copy_image_ops));
+
+    if ((rc = pthread_mutex_init(&meta_copy_image_ops->mutex, NULL)))
+    {
+        ERR("Failed to initialize mutex, error %d.\n", rc);
+        return hresult_from_errno(rc);
+    }
+
+    set_binding.binding = 0;
+    set_binding.descriptorType = VK_DESCRIPTOR_TYPE_STORAGE_IMAGE;
```
HansKristian-Work, line 533: Shouldn't this be sampled image? The shaders use texture2D etc, not image2D.

doitsujin: Yeah, typo.

doitsujin: Fixed.

doitsujin:
The copy_texture test already tests this partially, but I can add more down the line (just going to be really messy and I have absolutely no idea how to test e.g. the MSAA case without going insane).

HansKristian-Work:
FWIW, I don't consider it a high priority. But something to consider if we  run into bugs down the line.

PR #105, opened 2020-04-22
tests: Add basic DXR bringup test.

HansKristian-Work:
Creates a basic acceleration structure and asserts that the expected triangles and SBTs are invoked.


HansKristian-Work:
Passes on 1660 Ti on Windows. Don't have a "real" RTX card to test with.

PR #109, opened 2020-04-26
Barrier rework

doitsujin:
Fixes most validation errors related to image layouts, as well as some real-world issues like The Talos Priciple showing corruption with MSAA enabled.

Here's how it works:
- Based on the usage flags, a common layout is determined for each image resource.
- The common layout is always used for descriptors.
- For transfer operations, the images involved will be transitioned to the `TRANSFER_{SRC,DST}_OPTIMAL` layout before the transfer op itself, and back to their common layout afterwards.
- Color attachments are transitioned to `COLOR_ATTACHMENT_OPTIMAL` when beginning a render pass, and back to their common layout afterwards.
- Depth-stencil attachments are transitioned to the pipeline-specific DSV layout when beginning a render pass, and back to their common layout afterwards.
- `ResourceBarrier` does not perform any layout transitions, and instead batches everything into one single `VkMemoryBarrier`. The `COMMON` state is conservative and acts as a full barrier.

Caveats and things that still need to be done down the line:
- We should add a fast path for "suspending" the active render pass, so that beginning and ending queries does not introduce unnecessary stalls.
- ~~We cannot know the exact layout of the depth-stencil attachment at draw time when creating an SRV descriptor for it, so this is often going to be wrong.~~ Actually not an issue according to spec chapter 11.4.1.
- Initial state transitions should be performed by the command list that first uses a resource on the GPU timeline, rather than the CPU timeline. Not entirely sure how to tackle this.
- Destroying render targets that are bound to a command list is going to cause undefined behaviour.
- We should use an `UNDEFINED -> whatever` transition when overriding the entire destination image in transfer operations.

Unrelated:
- RTV/DSV clears need a rewrite.
- Render passes need some work; currently we're using `DONT_CARE` incorrectly for depth/stencil aspects.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -1751,7 +1751,7 @@ static void d3d12_command_list_emit_render_pass_transition(struct d3d12_command_
 
     dsv = &list->dsv;
 
-    if (dsv->view)
+    if (dsv->view && list->dsv_layout)
     {
         stage_mask |= vk_render_pass_barrier_from_view(dsv->view,
                 dsv->resource, mode, &vk_image_barriers[j++]);
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -2207,7 +2207,6 @@ static void d3d12_command_list_reset_state(struct d3d12_command_list *list,
 
     memset(list->rtvs, 0, sizeof(list->rtvs));
     memset(&list->dsv, 0, sizeof(list->dsv));
-    list->dsv_format = VK_FORMAT_UNDEFINED;
     list->fb_width = 0;
     list->fb_height = 0;
     list->fb_layer_count = 0;
@@ -2292,7 +2291,7 @@ static bool d3d12_command_list_has_depth_stencil_view(struct d3d12_command_list 
     assert(d3d12_pipeline_state_is_graphics(list->state));
     graphics = &list->state->u.graphics;
 
-    return graphics->dsv_format || (d3d12_pipeline_state_has_unknown_dsv_format(list->state) && list->dsv_format);
+    return graphics->dsv_format || (d3d12_pipeline_state_has_unknown_dsv_format(list->state) && list->dsv.format);
 }
 
 static void d3d12_command_list_get_fb_extent(struct d3d12_command_list *list,
@@ -2431,6 +2430,7 @@ static bool d3d12_command_list_update_graphics_pipeline(struct d3d12_command_lis
     const struct vkd3d_vk_device_procs *vk_procs = &list->device->vk_procs;
     VkRenderPass vk_render_pass;
     VkPipeline vk_pipeline;
+    VkFormat dsv_format;
 
     if (list->current_pipeline != VK_NULL_HANDLE)
         return true;
@@ -2441,8 +2441,10 @@ static bool d3d12_command_list_update_graphics_pipeline(struct d3d12_command_lis
         return false;
     }
 
+    dsv_format = list->dsv.format ? list->dsv.format->vk_format : VK_FORMAT_UNDEFINED;
+
     if (!(vk_pipeline = d3d12_pipeline_state_get_or_create_pipeline(list->state,
-            &list->dynamic_state, list->dsv_format, &vk_render_pass)))
+            &list->dynamic_state, dsv_format, &vk_render_pass)))
         return false;
 
     /* The render pass cache ensures that we use the same Vulkan render pass
@@ -4827,9 +4829,9 @@ static void STDMETHODCALLTYPE d3d12_command_list_OMSetRenderTargets(d3d12_comman
 {
     struct d3d12_command_list *list = impl_from_ID3D12GraphicsCommandList(iface);
     const VkPhysicalDeviceLimits *limits = &list->device->vk_info.device_limits;
+    VkFormat prev_dsv_format, next_dsv_format;
     const struct d3d12_rtv_desc *rtv_desc;
     const struct d3d12_dsv_desc *dsv_desc;
-    VkFormat prev_dsv_format;
     unsigned int i;
 
     TRACE("iface %p, render_target_descriptor_count %u, render_target_descriptors %p, "
@@ -4847,6 +4849,9 @@ static void STDMETHODCALLTYPE d3d12_command_list_OMSetRenderTargets(d3d12_comman
     list->fb_width = limits->maxFramebufferWidth;
     list->fb_height = limits->maxFramebufferHeight;
     list->fb_layer_count = limits->maxFramebufferLayers;
+
+    prev_dsv_format = list->dsv.format ? list->dsv.format->vk_format : VK_FORMAT_UNDEFINED;
+    next_dsv_format = VK_FORMAT_UNDEFINED;
 
     memset(list->rtvs, 0, sizeof(list->rtvs));
     memset(&list->dsv, 0, sizeof(list->dsv));
@@ -4881,8 +4886,6 @@ static void STDMETHODCALLTYPE d3d12_command_list_OMSetRenderTargets(d3d12_comman
         list->fb_layer_count = min(list->fb_layer_count, rtv_desc->layer_count);
     }
 
-    prev_dsv_format = list->dsv_format;
-    list->dsv_format = VK_FORMAT_UNDEFINED;
     if (depth_stencil_descriptor)
     {
         if ((dsv_desc = d3d12_dsv_desc_from_cpu_handle(*depth_stencil_descriptor))
@@ -4898,7 +4901,7 @@ static void STDMETHODCALLTYPE d3d12_command_list_OMSetRenderTargets(d3d12_comman
             list->fb_width = min(list->fb_width, dsv_desc->width);
             list->fb_height = min(list->fb_height, dsv_desc->height);
             list->fb_layer_count = min(list->fb_layer_count, dsv_desc->layer_count);
-            list->dsv_format = dsv_desc->format->vk_format;
+            next_dsv_format = dsv_desc->format->vk_format;
         }
         else
         {
@@ -4906,7 +4909,7 @@ static void STDMETHODCALLTYPE d3d12_command_list_OMSetRenderTargets(d3d12_comman
         }
     }
 
-    if (prev_dsv_format != list->dsv_format && d3d12_pipeline_state_has_unknown_dsv_format(list->state))
+    if (prev_dsv_format != next_dsv_format && d3d12_pipeline_state_has_unknown_dsv_format(list->state))
         d3d12_command_list_invalidate_current_pipeline(list);
 
     d3d12_command_list_invalidate_current_framebuffer(list);
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -1959,223 +1959,6 @@ static void vk_access_and_stage_flags_from_d3d12_resource_state(const struct d3d
         FIXME("Unhandled resource state %#x.\n", unhandled_state);
 }
 
-static bool vk_barrier_parameters_from_d3d12_resource_state(struct d3d12_device *device, unsigned int state,
-        unsigned int stencil_state, const struct d3d12_resource *resource, VkQueueFlags vk_queue_flags,
-        VkAccessFlags *access_mask, VkPipelineStageFlags *stage_flags, VkImageLayout *image_layout)
-{
-    bool is_swapchain_image = resource && (resource->flags & VKD3D_RESOURCE_PRESENT_STATE_TRANSITION);
-    VkPipelineStageFlags queue_shader_stages = 0;
-
-    if (vk_queue_flags & VK_QUEUE_GRAPHICS_BIT)
-    {
-        queue_shader_stages |= VK_PIPELINE_STAGE_VERTEX_SHADER_BIT
-                | VK_PIPELINE_STAGE_TESSELLATION_CONTROL_SHADER_BIT
-                | VK_PIPELINE_STAGE_TESSELLATION_EVALUATION_SHADER_BIT
-                | VK_PIPELINE_STAGE_GEOMETRY_SHADER_BIT
-                | VK_PIPELINE_STAGE_FRAGMENT_SHADER_BIT;
-    }
-    if (vk_queue_flags & VK_QUEUE_COMPUTE_BIT)
-        queue_shader_stages |= VK_PIPELINE_STAGE_COMPUTE_SHADER_BIT;
-
-    switch (state)
-    {
-        case D3D12_RESOURCE_STATE_COMMON: /* D3D12_RESOURCE_STATE_PRESENT */
-            /* The COMMON state is used for ownership transfer between
-             * DIRECT/COMPUTE and COPY queues. Additionally, a texture has to
-             * be in the COMMON state to be accessed by CPU. Moreover,
-             * resources can be implicitly promoted to other states out of the
-             * COMMON state, and the resource state can decay to the COMMON
-             * state when GPU finishes execution of a command list. */
-            if (is_swapchain_image)
-            {
-                if (resource->present_state == D3D12_RESOURCE_STATE_PRESENT)
-                {
-                    *access_mask = VK_ACCESS_MEMORY_READ_BIT;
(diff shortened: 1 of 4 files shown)


libs/vkd3d/resource.c
```diff
@@ -865,6 +865,20 @@ static bool vkd3d_is_linear_tiling_supported(const struct d3d12_device *device,
             && (image_info->samples & properties.sampleCounts);
 }
 
+static VkImageLayout vk_common_image_layout_from_d3d12_desc(const D3D12_RESOURCE_DESC *desc)
+{
+    if (desc->Flags & D3D12_RESOURCE_FLAG_ALLOW_UNORDERED_ACCESS)
+        return VK_IMAGE_LAYOUT_GENERAL;
+
+    if (desc->Flags & D3D12_RESOURCE_FLAG_DENY_SHADER_RESOURCE)
+        return VK_IMAGE_LAYOUT_DEPTH_STENCIL_ATTACHMENT_OPTIMAL;
```
HansKristian-Work, line 874: Why depth stencil and not color?

doitsujin: `D3D12_RESOURCE_FLAG_DENY_SHADER_RESOURCE` is only allowed in combination with `ALLOW_DEPTH_STENCIL` according to [MSDN](https://docs.microsoft.com/en-us/windows/win32/api/d3d12/ne-d3d12-d3d12_resource_flags).

libs/vkd3d/command.c
```diff
@@ -1681,6 +1681,90 @@ static void d3d12_command_list_invalidate_current_pipeline(struct d3d12_command_
...
+
+    if (view->format->vk_aspect_mask & VK_IMAGE_ASPECT_COLOR_BIT)
+    {
+        stages = VK_PIPELINE_STAGE_COLOR_ATTACHMENT_OUTPUT_BIT;
+        access = VK_ACCESS_COLOR_ATTACHMENT_WRITE_BIT | VK_ACCESS_COLOR_ATTACHMENT_READ_BIT;
+    }
+    else
+    {
+        stages = VK_PIPELINE_STAGE_EARLY_FRAGMENT_TESTS_BIT | VK_PIPELINE_STAGE_LATE_FRAGMENT_TESTS_BIT;
+        access = VK_ACCESS_DEPTH_STENCIL_ATTACHMENT_WRITE_BIT | VK_ACCESS_DEPTH_STENCIL_ATTACHMENT_READ_BIT;
```
HansKristian-Work, line 1707: ATTACHMENT_WRITE_BIT is not allowed if using READ_ONLY layout I think?

libs/vkd3d/command.c
```diff
@@ -2211,11 +2202,15 @@ static void d3d12_command_list_transition_resource_to_initial_state(struct d3d12
...
+    vk_access_and_stage_flags_from_d3d12_resource_state(list->device, resource->initial_state,
+            list->vk_queue_flags, &dst_stage_mask, &barrier.dstAccessMask);
+
     TRACE("Initial state %#x transition for resource %p (old layout %#x, new layout %#x).\n",
             resource->initial_state, resource, barrier.oldLayout, barrier.newLayout);
 
-    VK_CALL(vkCmdPipelineBarrier(list->vk_command_buffer, src_stage_mask, dst_stage_mask, 0,
-            0, NULL, 0, NULL, 1, &barrier));
+    VK_CALL(vkCmdPipelineBarrier(list->vk_command_buffer,
+            VK_PIPELINE_STAGE_BOTTOM_OF_PIPE_BIT, dst_stage_mask,
```
HansKristian-Work, line 1998: Can't the initial transition happen in TOP_OF_PIPE? Nothing on GPU should be able to access it before we make the transition.

doitsujin: Yes, this was meant to be `TOP_OF_PIPE`, not `BOTTOM`.

libs/vkd3d/command.c
```diff
@@ -4277,64 +4060,31 @@ static void STDMETHODCALLTYPE d3d12_command_list_SetPipelineState(d3d12_command_
...
+    VkPipelineStageFlags dst_stage_mask, src_stage_mask;
+    VkMemoryBarrier vk_memory_barrier;
     unsigned int i;
 
     TRACE("iface %p, barrier_count %u, barriers %p.\n", iface, barrier_count, barriers);
 
     d3d12_command_list_end_current_render_pass(list);
 
+    vk_memory_barrier.sType = VK_STRUCTURE_TYPE_MEMORY_BARRIER;
+    vk_memory_barrier.pNext = 0;
```
HansKristian-Work, line 4190: NULL for consistency.

libs/vkd3d/command.c
```diff
@@ -1953,223 +1953,6 @@ static void vk_access_and_stage_flags_from_d3d12_resource_state(const struct d3d
...
-    switch (state)
-    {
-        case D3D12_RESOURCE_STATE_COMMON: /* D3D12_RESOURCE_STATE_PRESENT */
-            /* The COMMON state is used for ownership transfer between
-             * DIRECT/COMPUTE and COPY queues. Additionally, a texture has to
-             * be in the COMMON state to be accessed by CPU. Moreover,
-             * resources can be implicitly promoted to other states out of the
-             * COMMON state, and the resource state can decay to the COMMON
-             * state when GPU finishes execution of a command list. */
-            if (is_swapchain_image)
```
HansKristian-Work, line 1782: Don't we need to handle swapchain images at least?

libs/vkd3d/command.c
doitsujin: Does it say that anyhwere in the spec? I can't really find anything, but it mentions e.g. that `STORE_OP_STORE` uses `ATTACHMENT_WRITE_BIT`, apparently regardless of the layout.

I can put the `WRITE` bit behind an `if (layout != DEPTH_STENCIL_READ_ONLY_OPTIMAL)` though.

libs/vkd3d/command.c
```diff
@@ -4981,11 +4981,11 @@ static void STDMETHODCALLTYPE d3d12_command_list_ClearDepthStencilView(d3d12_com
         attachment_desc.stencilLoadOp = VK_ATTACHMENT_LOAD_OP_DONT_CARE;
         attachment_desc.stencilStoreOp = VK_ATTACHMENT_STORE_OP_DONT_CARE;
     }
-    attachment_desc.initialLayout = VK_IMAGE_LAYOUT_DEPTH_STENCIL_ATTACHMENT_OPTIMAL;
-    attachment_desc.finalLayout = VK_IMAGE_LAYOUT_DEPTH_STENCIL_ATTACHMENT_OPTIMAL;
+    attachment_desc.initialLayout = dsv_desc->resource->common_layout;
```
HansKristian-Work, line 4989: Need EXTERNAL subpass dependencies here.

libs/vkd3d/resource.c
HansKristian-Work: Curious, but ok.

libs/vkd3d/command.c
doitsujin: I'll double-check what wine's DXGI does.

libs/vkd3d/command.c
HansKristian-Work: If it doesn't fail validation, I suppose it's fine. And yes, STORE_OP_STORE is technically needed, even for read-only resources.

libs/vkd3d/command.c
```diff
@@ -4862,27 +4866,22 @@ static void STDMETHODCALLTYPE d3d12_command_list_OMSetRenderTargets(d3d12_comman
...
             continue;
         }
 
         d3d12_command_list_track_resource_usage(list, rtv_desc->resource);
 
         /* In D3D12 CPU descriptors are consumed when a command is recorded. */
-        view = rtv_desc->view;
-        if (!d3d12_command_allocator_add_view(list->allocator, view))
-        {
+        if (!d3d12_command_allocator_add_view(list->allocator, rtv_desc->view))
```
HansKristian-Work, line 4819: Hm, what is actually happening here, do we create a new VkImageView for every OMSetRenderTargets?

doitsujin: We don't, `add_view` only ref-counts the view. Not sure if this is really necessary, but it was there before so I kept it.

libs/vkd3d/command.c
```diff
@@ -1839,6 +1839,120 @@ static void d3d12_command_list_invalidate_root_parameters(struct d3d12_command_l
...
+                break;
+
+            case D3D12_RESOURCE_STATE_UNORDERED_ACCESS:
+                *stages |= queue_shader_stages;
+                *access |= VK_ACCESS_SHADER_READ_BIT | VK_ACCESS_SHADER_WRITE_BIT;
+                break;
+
+            case D3D12_RESOURCE_STATE_DEPTH_WRITE:
+                *stages |= VK_PIPELINE_STAGE_EARLY_FRAGMENT_TESTS_BIT | VK_PIPELINE_STAGE_LATE_FRAGMENT_TESTS_BIT;
+                *access |= VK_ACCESS_DEPTH_STENCIL_ATTACHMENT_READ_BIT | VK_ACCESS_DEPTH_STENCIL_ATTACHMENT_WRITE_BIT;
```
HansKristian-Work, line 1901: Actually, for the resource states where we do implicit state transitions anyways, I think we can use access of 0. We only need an execution barrier to link up the incoming barrier for render pass or transfer operation.

HansKristian-Work: E.g. for a state transition into RENDER_TARGET we will see two barrier:

SOMETHING -> RENDER_TARGET (flush something, block COLOR_ATTACHMENT)
BeginRenderPass -> here we wait for COLOR_ATTACHMENT, perform layout transition).

We never need memory to be visible before we perform the layout transition.

libs/vkd3d/command.c
```diff
@@ -1839,6 +1839,120 @@ static void d3d12_command_list_invalidate_root_parameters(struct d3d12_command_l
...
+
+    while (state_mask)
+    {
+        uint32_t state = state_mask & -state_mask;
+
+        switch (state)
+        {
+            case D3D12_RESOURCE_STATE_VERTEX_AND_CONSTANT_BUFFER:
+                *stages |= queue_shader_stages | VK_PIPELINE_STAGE_VERTEX_INPUT_BIT;
+                *access |= VK_ACCESS_UNIFORM_READ_BIT | VK_ACCESS_VERTEX_ATTRIBUTE_READ_BIT;
```
HansKristian-Work, line 1878: One validation error I've seen in the past is that if we use this in the compute queue, we cannot use VERTEX_ATTRIBUTE_READ_BIT. Probably should check queue_flags for graphics related barriers.

doitsujin:
> Is it possible to split up the PR in a way which doesn't break anything on its own?

The moment we start changing image layout we need to change them everywhere in order to get things right, so probably not. Some individual commits may introduce regressions on actual hardware, but it's not like we were getting it right in the first place.

libs/vkd3d/resource.c
doitsujin: Done.

libs/vkd3d/command.c

libs/vkd3d/command.c

libs/vkd3d/command.c
HansKristian-Work: I assume we need it if the descriptors are meant to be consumed/copied in CPU timeline.

libs/vkd3d/command.c
doitsujin: Yeah, this is going to need special care; basically we'll have to transition the image layout like we did before and hope that we never receive a swap chain image for anything other than actual presentation. DXGI unfortunately uses D3D12 commands for barriers but raw Vulkan commands for the blit operation.

libs/vkd3d/command.c
doitsujin: Should be fixed.

libs/vkd3d/command.c
doitsujin: Done.

libs/vkd3d/command.c

libs/vkd3d/command.c
```diff
@@ -1839,6 +1839,120 @@ static void d3d12_command_list_invalidate_root_parameters(struct d3d12_command_l
...
+                {
+                    *stages |= VK_PIPELINE_STAGE_CONDITIONAL_RENDERING_BIT_EXT;
+                    *access |= VK_ACCESS_CONDITIONAL_RENDERING_READ_BIT_EXT;
+                }
+                break;
+
+            case D3D12_RESOURCE_STATE_COPY_DEST:
+            case D3D12_RESOURCE_STATE_RESOLVE_DEST:
+                *stages |= VK_PIPELINE_STAGE_TRANSFER_BIT;
+                *access |= VK_ACCESS_TRANSFER_WRITE_BIT;
```
HansKristian-Work, line 1939: Transfer ops are also implied barriers.

HansKristian-Work: Actually, not for buffers I suppose ...

libs/vkd3d/command.c
```diff
@@ -1839,6 +1839,120 @@ static void d3d12_command_list_invalidate_root_parameters(struct d3d12_command_l
...
+                break;
+
+            case D3D12_RESOURCE_STATE_COPY_SOURCE:
+            case D3D12_RESOURCE_STATE_RESOLVE_SOURCE:
+                *stages |= VK_PIPELINE_STAGE_TRANSFER_BIT;
+                *access |= VK_ACCESS_TRANSFER_READ_BIT;
```
HansKristian-Work, line 1945: But resolve source/dst is implied.

libs/vkd3d/command.c
```diff
@@ -3970,9 +3973,52 @@ static void STDMETHODCALLTYPE d3d12_command_list_ResolveSubresource(d3d12_comman
...
     VK_CALL(vkCmdResolveImage(list->vk_command_buffer, src_resource->u.vk_image,
             VK_IMAGE_LAYOUT_TRANSFER_SRC_OPTIMAL, dst_resource->u.vk_image,
             VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL, 1, &vk_image_resolve));
+
+    vk_image_barriers[0].srcAccessMask = VK_ACCESS_TRANSFER_WRITE_BIT;
+    vk_image_barriers[0].dstAccessMask = 0;
+    vk_image_barriers[0].oldLayout = dst_layout;
+    vk_image_barriers[0].newLayout = dst_resource->common_layout;
+
+    vk_image_barriers[1].srcAccessMask = VK_ACCESS_TRANSFER_READ_BIT;
```
HansKristian-Work, line 4019: READ barrier in src is redundant.

libs/vkd3d/command.c
HansKristian-Work: Is this implemented somewhere? It seems to be the only missing piece.

libs/vkd3d/command.c
HansKristian-Work: If the resource barrier is on an image, we can probably elide the access mask.

libs/vkd3d/command.c
doitsujin: Not yet.

HansKristian-Work: Ok, I'll do a final review when that is implemented.

libs/vkd3d/command.c
doitsujin: We don't take a resource parameter here. Should I add it back?

HansKristian-Work: I think so. The implementation will differ depending on resource type.

libs/vkd3d/command.c
doitsujin: Should be fixed.

libs/vkd3d/command.c
doitsujin: Removed.

libs/vkd3d/command.c
doitsujin: Done.

libs/vkd3d/command.c
doitsujin: Done, see last commit of the series. It's a bit dodgy since DXGI makes assumptions about the exact image layout.

PR #110, opened 2020-04-27
Suspend renderpasses

doitsujin:
Optimizes useless barriers away and fixes performance regressions introduced with the barrier rework (only SotTR seems to be really affected).

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -1768,7 +1768,7 @@ static void d3d12_command_list_emit_render_pass_transition(struct d3d12_command_
         j, vk_image_barriers));
 }
 
-static void d3d12_command_list_end_current_render_pass(struct d3d12_command_list *list)
+static void d3d12_command_list_end_current_render_pass(struct d3d12_command_list *list, bool suspend)
 {
     const struct vkd3d_vk_device_procs *vk_procs = &list->device->vk_procs;
 
@@ -1779,11 +1779,13 @@ static void d3d12_command_list_end_current_render_pass(struct d3d12_command_list
     }
 
     if (list->current_render_pass)
-    {
         VK_CALL(vkCmdEndRenderPass(list->vk_command_buffer));
-        d3d12_command_list_emit_render_pass_transition(list, VKD3D_RENDER_PASS_TRANSITION_MODE_END);
-    }
 
+    /* Don't emit barriers for temporary suspendion of the render pass */
+    if (!suspend && (list->current_render_pass || list->render_pass_suspended))
+        d3d12_command_list_emit_render_pass_transition(list, VKD3D_RENDER_PASS_TRANSITION_MODE_END);
+
+    list->render_pass_suspended = suspend && list->current_render_pass;
     list->current_render_pass = VK_NULL_HANDLE;
 
     if (list->xfb_enabled)
@@ -1805,7 +1807,7 @@ static void d3d12_command_list_end_current_render_pass(struct d3d12_command_list
 
 static void d3d12_command_list_invalidate_current_render_pass(struct d3d12_command_list *list)
 {
-    d3d12_command_list_end_current_render_pass(list);
+    d3d12_command_list_end_current_render_pass(list, false);
 }
 
 static void d3d12_command_list_invalidate_root_parameters(struct d3d12_command_list *list,
@@ -2010,7 +2012,7 @@ static void d3d12_command_list_track_resource_usage(struct d3d12_command_list *l
 {
     if (resource->flags & VKD3D_RESOURCE_INITIAL_STATE_TRANSITION)
     {
-        d3d12_command_list_end_current_render_pass(list);
+        d3d12_command_list_end_current_render_pass(list, true);
 
         d3d12_command_list_transition_resource_to_initial_state(list, resource);
         resource->flags &= ~VKD3D_RESOURCE_INITIAL_STATE_TRANSITION;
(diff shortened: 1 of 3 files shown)


libs/vkd3d/state.c
```diff
@@ -1189,36 +1194,36 @@ static HRESULT vkd3d_render_pass_cache_create_pass_locked(struct vkd3d_render_pa
...
         attachment_index++;
     }
 
+    dependencies[0].srcSubpass = VK_SUBPASS_EXTERNAL;
+    dependencies[0].dstSubpass = 0;
+    dependencies[0].srcStageMask = stages;
+    dependencies[0].dstStageMask = stages;
+    dependencies[0].srcAccessMask = 0;
+    dependencies[0].dstAccessMask = 0;
+    dependencies[0].dependencyFlags = 0;
```
HansKristian-Work, line 1217: BY_REGION as well probably?

libs/vkd3d/command.c
```diff
@@ -1779,11 +1779,13 @@ static void d3d12_command_list_end_current_render_pass(struct d3d12_command_list
...
     if (list->current_render_pass)
-    {
         VK_CALL(vkCmdEndRenderPass(list->vk_command_buffer));
+
+    /* Don't emit barriers for temporary suspendion of the render pass */
+    if (!suspend && (list->current_render_pass || list->render_pass_suspended))
         d3d12_command_list_emit_render_pass_transition(list, VKD3D_RENDER_PASS_TRANSITION_MODE_END);
-    }
 
+    list->render_pass_suspended = suspend && list->current_render_pass;
```
HansKristian-Work, line 1788: Review not to self: Do we perform suspended barriers on vkEndCommandBuffer?

doitsujin: We should, since `Close` calls `d3d12_command_list_end_current_render_pass(..., false)`.

HansKristian-Work: Yes

libs/vkd3d/state.c
doitsujin: Is `BY_REGION` meaningful at all outside of subpass self-dependencies?

Can do, at least AMD ignores this anyway, just not sure what exactly the semantics are supposed to be.

libs/vkd3d/command.c
```diff
@@ -4790,7 +4794,7 @@ static void STDMETHODCALLTYPE d3d12_command_list_SOSetTargets(d3d12_command_list
 
     TRACE("iface %p, start_slot %u, view_count %u, views %p.\n", iface, start_slot, view_count, views);
 
-    d3d12_command_list_end_current_render_pass(list);
+    d3d12_command_list_end_current_render_pass(list, true);
```
HansKristian-Work, line 4797: Does this need to end the render pass?

libs/vkd3d/state.c
HansKristian-Work: Yes, it should allow a correct IMR implementation to elide the barrier entirely. Can't hurt to add it, and gives a reason for drivers to actually take BY_REGION into consideration.

HansKristian-Work: src/dstAccessMask being 0 here is technically wrong, but it's the only known way currently to avoid full stalls on drivers. BY_REGION is supposed to be the fix for it, but we can't rely on that yet.

libs/vkd3d/command.c
```diff
@@ -1805,7 +1807,7 @@ static void d3d12_command_list_end_current_render_pass(struct d3d12_command_list
 
 static void d3d12_command_list_invalidate_current_render_pass(struct d3d12_command_list *list)
 {
-    d3d12_command_list_end_current_render_pass(list);
+    d3d12_command_list_end_current_render_pass(list, false);
```
HansKristian-Work, line 1810: Is this called from OMSetRenderTargets?

libs/vkd3d/command.c
```diff
@@ -3268,7 +3272,7 @@ static void STDMETHODCALLTYPE d3d12_command_list_CopyBufferRegion(d3d12_command_
     d3d12_command_list_track_resource_usage(list, dst_resource);
     d3d12_command_list_track_resource_usage(list, src_resource);
 
-    d3d12_command_list_end_current_render_pass(list);
+    d3d12_command_list_end_current_render_pass(list, true);
```
HansKristian-Work, line 3275: Is it possible to trigger CopyBufferToImage or ImageToBuffer here?

libs/vkd3d/command.c
doitsujin: I think this is being done so that we don't change transform feedback buffers while transform feedback is active.

If we were to remove this, we'd have to revisit Begin/EndTransformFeedback as well.

libs/vkd3d/command.c
doitsujin: Yes.

libs/vkd3d/command.c
HansKristian-Work: hm, ok.

libs/vkd3d/command.c
doitsujin: CopyBufferRegion strictly maps to `vkCmdCopyBuffer`, and aliasing between image and buffer resources is undefined behaviour in D3D12 as far as I'm aware.

libs/vkd3d/state.c
doitsujin: Alright, set `VK_DEPENDENCY_BY_REGION_BIT` and added a comment about the access masks.

PR #111, opened 2020-04-28
Clear rework

doitsujin:
Re-implements render target clears, with the goals to make it more robust and efficient:
- We no longer interrupt the current render pass when clearing an already active attachment, instead we'll just call `vkCmdClearAttachments`.
- When clearing a bound render target while no render pass is active, clears are deferred until we actually begin the render pass, so that we can emit `UNDEFINED`->`ATTACHMENT_OPTIMAL` transitions and use `vkCmdClearAttachments` for those as well. Most of the complexity comes from handling this case correctly.
- In other situations, we'll still fall back to using a separate render pass for clears, using `LOAD_OP_CLEAR` if possible.

**Edit**: Fixed commit messages.


libs/vkd3d/command.c
```diff
@@ -1681,6 +1681,284 @@ static void d3d12_command_list_invalidate_current_pipeline(struct d3d12_command_
...
+    {
+        full_rect = d3d12_get_view_rect(resource, view);
+        rect_count = 1;
+        rects = &full_rect;
+    }
+
+    /* We expect more than one clear rect to be very uncommon
+     * in practice, so make no effort to batch calls for now. */
+    vk_clear_attachment.aspectMask = clear_aspects;
+    vk_clear_attachment.colorAttachment = attachment_idx;
```
HansKristian-Work, line 1788: Does this work with depth-stencil?

libs/vkd3d/command.c
```diff
@@ -5068,6 +5346,44 @@ static void d3d12_command_list_clear(struct d3d12_command_list *list,
...
+    bool full_clear;
+    unsigned int i;
+
+    /* If one of the clear rectangles covers the entire image, we
+     * may be able to use a fast path and re-initialize the image */
+    full_rect = d3d12_get_view_rect(resource, view);
+    full_clear = !rect_count;
+
+    for (i = 0; i < rect_count && !full_clear; i++)
+        full_clear |= !memcmp(&rects[i], &full_rect, sizeof(full_rect));
```
HansKristian-Work, line 5388: Do we need to take scissor into account here?

libs/vkd3d/command.c
```diff
@@ -1681,6 +1681,284 @@ static void d3d12_command_list_invalidate_current_pipeline(struct d3d12_command_
...
+        return;
+    }
+
+    begin_info.sType = VK_STRUCTURE_TYPE_RENDER_PASS_BEGIN_INFO;
+    begin_info.pNext = NULL;
+    begin_info.renderPass = vk_render_pass;
+    begin_info.framebuffer = vk_framebuffer;
+    begin_info.renderArea.offset.x = 0;
+    begin_info.renderArea.offset.y = 0;
+    begin_info.renderArea.extent.width = extent.width;
```
HansKristian-Work, line 1933: We can still use clear. LOAD_OP_CLEAR only affects renderArea.

libs/vkd3d/command.c
doitsujin: Viewport/Scissor doesn't affect clears, that was a D3D9 thing.

libs/vkd3d/command.c
doitsujin: Yes, the color attachment index must be ignored if `clear_aspects` contains depth or stencil.

Could set this to 0, but even 0 isn't necessarily valid either (think depth-only passes).

libs/vkd3d/command.c
```diff
@@ -1681,6 +1681,284 @@ static void d3d12_command_list_invalidate_current_pipeline(struct d3d12_command_
...
+        subpass_desc.pColorAttachments = &attachment_ref;
+    }
+
+    dependencies[0].srcSubpass = VK_SUBPASS_EXTERNAL;
+    dependencies[0].dstSubpass = 0;
+    dependencies[0].srcStageMask = stages;
+    dependencies[0].dstStageMask = stages;
+    dependencies[0].srcAccessMask = 0;
+    dependencies[0].dstAccessMask = access;
+    dependencies[0].dependencyFlags = 0;
```
HansKristian-Work, line 1881: BY_REGION if initialLayout does not imply transition.

HansKristian-Work: Actually, no. initialLayout is common layout, nvm.

libs/vkd3d/command.c
HansKristian-Work: And GL fwiw, but ye. As expected then.

libs/vkd3d/command.c
HansKristian-Work: Could just add a comment to make it clear.

libs/vkd3d/command.c
```diff
@@ -1681,6 +1681,284 @@ static void d3d12_command_list_invalidate_current_pipeline(struct d3d12_command_
...
+         * discard the entire image, not just the layers to clear. */
+        if (clear_aspects == view->format->vk_aspect_mask &&
+                resource->desc.Dimension != D3D12_RESOURCE_DIMENSION_TEXTURE3D)
+            attachment_desc.initialLayout = VK_IMAGE_LAYOUT_UNDEFINED;
+    }
+
+    if (clear_aspects & (VK_IMAGE_ASPECT_DEPTH_BIT | VK_IMAGE_ASPECT_STENCIL_BIT))
+    {
+        stages = VK_PIPELINE_STAGE_EARLY_FRAGMENT_TESTS_BIT | VK_PIPELINE_STAGE_LATE_FRAGMENT_TESTS_BIT;
+        access = VK_ACCESS_DEPTH_STENCIL_ATTACHMENT_READ_BIT | VK_ACCESS_DEPTH_STENCIL_ATTACHMENT_WRITE_BIT;
```
HansKristian-Work, line 1864: Just using WRITE should be fine since CLEAR only accesses WRITE.

libs/vkd3d/command.c
```diff
@@ -1681,6 +1681,284 @@ static void d3d12_command_list_invalidate_current_pipeline(struct d3d12_command_
...
+        subpass_desc.pDepthStencilAttachment = &attachment_ref;
+    }
+    else
+    {
+        stages = VK_PIPELINE_STAGE_COLOR_ATTACHMENT_OUTPUT_BIT;
+        access = VK_ACCESS_COLOR_ATTACHMENT_READ_BIT | VK_ACCESS_COLOR_ATTACHMENT_WRITE_BIT;
```
HansKristian-Work, line 1870: Just write here as well.

libs/vkd3d/command.c
doitsujin: we're pretty much always transitioning from either `common_layout` or `UNDEFINED` to `ATTACHMENT_OPTIMAL` though.

HansKristian-Work: yup, so, BY_REGION is indeed not possible.

libs/vkd3d/command.c
doitsujin: Doable, but somewhat annoying since we can only really so it if we have *one* clear rect that *doesn't* cover the entire image, and we can't transition the image anyway. Is that really worth it over just doing `vkCmdClearAttachments` in that case?

HansKristian-Work: Right, for multiple rects, this gets rather annoying. Okay to do it like this I guess. Partial clear rect should be rare.

libs/vkd3d/command.c
```diff
@@ -3400,6 +3546,11 @@ static bool d3d12_command_list_begin_render_pass(struct d3d12_command_list *list
     vk_render_pass = list->pso_render_pass;
     assert(vk_render_pass);
 
+    /* Emit deferred clears that we cannot clear inside the
+     * render pass, e.g. when the attachments to clear are
+     * not used by the render pass. */
+    d3d12_command_list_emit_render_pass_clears(list, true);
```
HansKristian-Work, line 3552: How does this happen? If a clear is issued when that render target is not bound, shouldn't that get resolved immediately?

doitsujin: The problem is that the app may bind a render target with `OMSetRenderTargets` that isn't defined in the pipeline. We don't know this until we actually bind a pipeline.

The second problem is that we cannot clear a depth-stencil attachment the aspects to clear are read-only in the render pass, which we also do not know in advance.

HansKristian-Work: Right, so binding a pipeline can invalidate the render target and carve out only a subset?

HansKristian-Work: Probably a good place to place a comment about that as well. "not used by pipeline" seems more directed than "not used by render pass".

libs/vkd3d/command.c
```diff
@@ -3417,6 +3568,9 @@ static bool d3d12_command_list_begin_render_pass(struct d3d12_command_list *list
 
     list->current_render_pass = vk_render_pass;
 
+    /* Emit deferred clears with vkCmdClearAttachment */
+    d3d12_command_list_emit_render_pass_clears(list, false);
```
HansKristian-Work, line 3572: It is technically possible to use LOAD_OP_CLEAR here and rely on compatible render pass rules with PSOs?

doitsujin: Not without a major rework of how we obtain the render pass from the pipeline.

libs/vkd3d/command.c
HansKristian-Work: Actually, there might be another case where this is problematic.

What if attachments in a render pass have mismatched size, e.g. RT 0 is 64x64 and RT 1 is 128x128, we'll force renderArea of 64x64 in this case, but we actually wanted to clear the entire RT 1 as well. Images which are larger than the renderArea should probably be cleared separately as well.

libs/vkd3d/command.c
HansKristian-Work: Ok, good to know.

libs/vkd3d/command.c
HansKristian-Work: Ok, we're not using CLEAR in all cases, so nvm this.

libs/vkd3d/command.c
HansKristian-Work: Ok, we're not using CLEAR in all cases, so nvm this.

libs/vkd3d/command.c
doitsujin: I made this a bit more sophisticated, only uses `READ` if necessary now.

libs/vkd3d/command.c
doitsujin: Done.

libs/vkd3d/command.c
HansKristian-Work: Seems rather esoteric though.

PR #117, opened 2020-05-06
Implement Tiled resources Tier 1

doitsujin:
Implements reserved resources and basic sparse binding. Note that the path that allocates sparse metadata is largely untested, due to AMDVLK not requiring metadata in practice - it does however "work" when removing the early-out path in that it executes the sparse bind + queue wait idle commands as expected.

Things missing for full Tier 2 support, which is required for FL12_0:
- Shader instructions with feedback are not yet supported
- [CopyTiles](https://docs.microsoft.com/en-us/windows/win32/api/d3d12/nf-d3d12-id3d12graphicscommandlist-copytiles) is not yet implemented (technically a Tier 1 feature)
- Some Tier 2 guarantees correspond to Vulkan features and property bits that are currently ignored

**Edit:** Also, note that the commits appear out of order on github.


libs/vkd3d/resource.c
```diff
@@ -1155,6 +1155,112 @@ HRESULT vkd3d_get_image_allocation_info(struct d3d12_device *device,
...
+    const D3D12_RESOURCE_DESC *desc = &resource->desc;
+    unsigned int i, tile_count, packed_tiles;
+    uint32_t memory_requirement_count = 0;
+    VkExtent3D block_extent;
+
+    memset(vk_info, 0, sizeof(*vk_info));
+
+    if (desc->Dimension == D3D12_RESOURCE_DIMENSION_BUFFER)
+    {
+        tile_count = align(desc->Width, VKD3D_TILE_SIZE) / 65536;
```
HansKristian-Work, line 1173: Why state VKD3D_TILE_SIZE and 64k separately?

libs/vkd3d/command.c
```diff
@@ -6899,6 +6899,174 @@ static void d3d12_command_queue_execute(struct d3d12_command_queue *command_queu
...
+        for (i = 0; i < count; i++)
+        {
+            const struct vkd3d_sparse_memory_bind *bind = &bind_infos[i];
+
+            if (bind->dst_tile < first_packed_tile)
+                image_bind_count++;
+            else
+                opaque_bind_count++;
+
+            if (!(memory_binds = vkd3d_malloc(opaque_bind_count * sizeof(*memory_binds))) ||
```
HansKristian-Work, line 7191: This looks kinda sketchy. Why allocate in a loop with no frees?

doitsujin: Good catch, this is supposed to happen outside the loop.

libs/vkd3d/resource.c
doitsujin: Fixed.

libs/vkd3d/command.c

PR #126, opened 2020-05-27
vkd3d: Issue full barrier at the end of command lists.

doitsujin:
D3D12 apparently does this implicitly. Fixes rendering issues in the AMD COCOA demo on Polaris with RADV, which does not emit a barrier between the AO compute passes and the tone mapping pass in the next command buffer.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -2579,6 +2579,7 @@ static HRESULT STDMETHODCALLTYPE d3d12_command_list_Close(d3d12_command_list_ifa
 {
     struct d3d12_command_list *list = impl_from_ID3D12GraphicsCommandList(iface);
     const struct vkd3d_vk_device_procs *vk_procs;
+    VkMemoryBarrier memory_barrier;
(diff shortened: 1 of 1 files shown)


libs/vkd3d/command.c
```diff
@@ -2595,6 +2596,16 @@ static HRESULT STDMETHODCALLTYPE d3d12_command_list_Close(d3d12_command_list_ifa
     if (list->is_predicated)
         VK_CALL(vkCmdEndConditionalRenderingEXT(list->vk_command_buffer));
 
+    memory_barrier.sType = VK_STRUCTURE_TYPE_MEMORY_BARRIER;
+    memory_barrier.pNext = NULL;
+    memory_barrier.srcAccessMask = VK_ACCESS_MEMORY_WRITE_BIT;
```
HansKristian-Work, line 2601: BOTTOM_OF_PIPE cannot do memory barriers. Need ALL_COMMANDS_BIT on both if we're going to emit memory barriers.

Issue #124, opened 2020-05-25
Deus Ex: Mankind Divided crashes - vkd3d failed to allocate device memory


jqadev:
### Software information
@doitsujin according to your answer https://github.com/HansKristian-Work/vkd3d/issues/122#issuecomment-630757866 I found here https://www.protondb.com/app/449610 that I can download d3dcompiler_47 library from here https://lutris.net/files/tools/dll/d3dcompiler_47.dll
so I put it to `~/.steam/debian-installation/steamapps/common/Deus Ex Mankind Divided/retail` and the game started to work but it crashes when I press the space button to continue the story
![Screenshot from 2020-05-25 18-18-02](https://user-images.githubusercontent.com/20422423/82831534-f2a08a80-9eb8-11ea-8706-df650673d088.png)

### [System information (gist)](https://gist.github.com/jqadev/8a892c8f8b5145169b4d5ae8c142c7c9) 
- Distro: Ubuntu 20.04 LTS
- GPU: GeForce 930MX
- Driver: NVIDIA 440.64
- Proton version: 5.0-7

### Log files
- [steam-337000.zip](https://github.com/HansKristian-Work/vkd3d/files/4678193/steam-337000.zip) (collected using `PROTON_LOG=1 %command%`)

doitsujin:
> warn:vkd3d_allocate_device_memory: Failed to allocate device memory, vr -2.
warn:hresult_from_vk_result: Out of device memory.

Your GPU doesn't have enough VRAM, and vkd3d currently does not handle this at all.

jqadev:
@doitsujin DXVK handles that because I can run the game using it as I wrote here https://github.com/HansKristian-Work/vkd3d/issues/122#issue-620542247. 
Are there any plans to handle that in vkd3d as well? 
If no, please close this issue.

doitsujin:
> Are there any plans to handle that in vkd3d as well?

Yes, but it requires a rewrite of the memory allocation logic.

jqadev:
@doitsujin congrats on you great work here https://github.com/HansKristian-Work/vkd3d/pull/132 :+1: 
Do you have some CI/CD like this https://git.froggi.es/joshua/dxvk/-/jobs?
I would like to test master containing recently merged commits.

jqadev:
[...]

PR #132, opened 2020-06-03
Memory allocation rework

doitsujin:
Currently, when allocating a D3D12 heap for placed resources, we just pick a memory type, hope that all resources are going to be supported on it, and fail if the corresponding Vulkan heap is full on some drivers (see #124). Also, we unconditionally report `D3D12_RESOURCE_HEAP_TIER_2` even on Tier 1 hardware.

As a solution, figure out which memory types support which type of resource and use that information to find compatible memory types, and iterate over compatible memory types until the allocation succeeds.

This needs a bit more testing and I'll probably have to revisit the usage flags for the dummy resources we use to gather the memory type information, but generally seems to work on both my RX 480 and RTX 2060 systems.

### DIFF
--- a/libs/vkd3d/device.c
+++ b/libs/vkd3d/device.c
@@ -4622,6 +4622,9 @@ static HRESULT d3d12_device_init(struct d3d12_device *device,
     if (FAILED(hr = vkd3d_init_format_info(device)))
         goto out_stop_fence_worker;
 
+    if (FAILED(hr = vkd3d_memory_info_init(&device->memory_info, device)))
+        goto out_cleanup_format_info;
+
     if (FAILED(hr = vkd3d_init_null_resources(&device->null_resources, device)))
         goto out_cleanup_format_info;
 
--- a/libs/vkd3d/device.c
+++ b/libs/vkd3d/device.c
@@ -4264,6 +4264,29 @@ static D3D12_TILED_RESOURCES_TIER d3d12_device_determine_tiled_resources_tier(st
     return D3D12_TILED_RESOURCES_TIER_2;
 }
 
+static D3D12_RESOURCE_HEAP_TIER d3d12_device_determine_heap_tier(struct d3d12_device *device)
+{
+    const VkPhysicalDeviceLimits *limits = &device->device_info.properties2.properties.limits;
+    const VkPhysicalDeviceMemoryProperties *mem_properties = &device->memory_properties;
+    const struct vkd3d_memory_info *mem_info = &device->memory_info;
+    uint32_t i, host_visible_types = 0;
+
+    for (i = 0; i < mem_properties->memoryTypeCount; i++)
+    {
+        if (mem_properties->memoryTypes[i].propertyFlags & VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT)
+            host_visible_types |= 1u << i;
+    }
+
+    // Heap Tier 2 requires us to be able to create a heap that supports all resource
+    // categories at the same time, except RT/DS textures on UPLOAD/READBACK heaps.
+    if (limits->bufferImageGranularity > D3D12_DEFAULT_RESOURCE_PLACEMENT_ALIGNMENT ||
+            !(mem_info->buffer_type_mask & mem_info->sampled_type_mask & mem_info->rt_ds_type_mask) ||
+            !(mem_info->buffer_type_mask & mem_info->sampled_type_mask & host_visible_types))
+        return D3D12_RESOURCE_HEAP_TIER_1;
+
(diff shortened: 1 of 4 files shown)


libs/vkd3d/resource.c
```diff
@@ -5325,3 +5325,104 @@ void vkd3d_destroy_null_resources(struct vkd3d_null_resources *null_resources,
...
+    {
+        ERR("Failed to create dummy color image");
+        return hresult_from_vk_result(vr);
+    }
+
+    VK_CALL(vkGetImageMemoryRequirements(device->vk_device, image, &memory_requirements));
+    VK_CALL(vkDestroyImage(device->vk_device, image, NULL));
+    info->rt_ds_type_mask = memory_requirements.memoryTypeBits;
+
+    image_info.format = VK_FORMAT_D32_SFLOAT_S8_UINT;
```
HansKristian-Work, line 5466: This format is not guaranteed to be supported. Perhaps using D16 would be better since it's always supported.

libs/vkd3d/resource.c
```diff
@@ -5325,3 +5325,104 @@ void vkd3d_destroy_null_resources(struct vkd3d_null_resources *null_resources,
...
+    {
+        ERR("Failed to create dummy depth-stencil image");
+        return hresult_from_vk_result(vr);
+    }
+
+    VK_CALL(vkGetImageMemoryRequirements(device->vk_device, image, &memory_requirements));
+    VK_CALL(vkDestroyImage(device->vk_device, image, NULL));
+    info->rt_ds_type_mask &= memory_requirements.memoryTypeBits;
+
+    TRACE("Device supports buffers on memory types %#x.\n", info->buffer_type_mask);
```
HansKristian-Work, line 5482: Prepending with 0x might make trace easier to grasp.

libs/vkd3d/device.c
```diff
@@ -4264,6 +4264,27 @@ static D3D12_TILED_RESOURCES_TIER d3d12_device_determine_tiled_resources_tier(st
...
+    const struct vkd3d_memory_info *mem_info = &device->memory_info;
+    uint32_t i, host_visible_types = 0;
+
+    for (i = 0; i < mem_properties->memoryTypeCount; i++)
+    {
+        if (mem_properties->memoryTypes[i].propertyFlags & VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT)
+            host_visible_types |= 1u << i;
+    }
+
+    // Heap Tier 2 requires us to be able to create a heap that supports all resource
```
HansKristian-Work, line 4279: Should perhaps look at bufferImageGranularity as well. D3D12 would require 64K alignment, but Vulkan can technically return 128K here.

libs/vkd3d/resource.c
```diff
@@ -274,18 +275,31 @@ HRESULT vkd3d_allocate_buffer_memory(struct d3d12_device *device, VkBuffer vk_bu
...
+
+        flags_info.pNext = &dedicated_info;
     }
 
-    if (FAILED(hr = vkd3d_allocate_device_memory(device, heap_properties, heap_flags,
-            memory_requirements, dedicated_allocation, vk_memory, vk_memory_type)))
+    if (FAILED(hr = vkd3d_select_memory_flags(device, heap_properties, &type_flags)))
+        return hr;
+
+    if (FAILED(hr = vkd3d_allocate_memory(device, memory_requirements->size, type_flags,
```
HansKristian-Work, line 250: Doesn't this mean we always use pNext, even if MEMORY_ALLOCATE_FLAGS_INFO is not supported?

doitsujin: This is core in 1.1 which we made a hard requirement, so this should always be supported.

libs/vkd3d/resource.c
doitsujin: Right, but other than that, what do we do if this format is not supported and the D3D app wants it? It's a hard requirement in D3D12.

HansKristian-Work: In practice, I suppose this format will never *not* be supported. It might not be supported on mobile, but whatever.

libs/vkd3d/resource.c
HansKristian-Work: I figured it would only be supported for BDA, but I guess protected memory is in 1.1 as well.

libs/vkd3d/device.c
doitsujin: Done.

libs/vkd3d/resource.c
doitsujin: Also done.

HansKristian-Work:
Looks good!

PR #133, opened 2020-06-09
vkd3d: Serialize all queue submissions.

HansKristian-Work:
Gets rid of the full barrier on command buffer end.
Instead, do what D3D12 wants, which is to serialize all
ExecuteCommandLists. Simplify the existing timeline sempahore setup for
sparse queues and use it for all submissions.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -2579,7 +2579,6 @@ static HRESULT STDMETHODCALLTYPE d3d12_command_list_Close(d3d12_command_list_ifa
 {
     struct d3d12_command_list *list = impl_from_ID3D12GraphicsCommandList(iface);
     const struct vkd3d_vk_device_procs *vk_procs;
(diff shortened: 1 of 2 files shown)


libs/vkd3d/command.c
```diff
@@ -2596,18 +2596,6 @@ static HRESULT STDMETHODCALLTYPE d3d12_command_list_Close(d3d12_command_list_ifa
     if (list->is_predicated)
         VK_CALL(vkCmdEndConditionalRenderingEXT(list->vk_command_buffer));
 
-    // TODO We should emit a barrier between submissions,
-    // not between command lists within the same submission.
-    memory_barrier.sType = VK_STRUCTURE_TYPE_MEMORY_BARRIER;
```
doitsujin, line 2601: The declaration of `memory_barrier` should also be removed.

HansKristian-Work: Fixed.

PR #135, opened 2020-06-09
vkd3d-shader: Fix modifier handling.

doitsujin:
This is a set of flags, so we should treat it as such.

### DIFF
--- a/libs/vkd3d-shader/dxbc.c
+++ b/libs/vkd3d-shader/dxbc.c
@@ -1538,30 +1538,19 @@ static bool shader_sm4_read_param(struct vkd3d_sm4_data *priv, const DWORD **ptr
         }
         m = *(*ptr)++;
 
-        switch (m)
+        switch (m & 0xc0)
         {
-            case 0x41:
-                *modifier = VKD3DSPSM_NEG;
-                break;
-
-            case 0x81:
-                *modifier = VKD3DSPSM_ABS;
-                break;
-
-            case 0xc1:
-                *modifier = VKD3DSPSM_ABSNEG;
-                break;
-
-            case 0x20001:
-                param->modifier = VKD3DSPRM_NONUNIFORM;
-                *modifier = VKD3DSPSM_NONE;
-                break;
-
-            default:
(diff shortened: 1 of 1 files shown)


libs/vkd3d-shader/dxbc.c
```diff
@@ -1538,30 +1538,19 @@ static bool shader_sm4_read_param(struct vkd3d_sm4_data *priv, const DWORD **ptr
...
+        switch (m & 0xc0)
         {
-            case 0x41:
-                *modifier = VKD3DSPSM_NEG;
-                break;
+            case 0x00: *modifier = VKD3DSPSM_NONE; break;
+            case 0x40: *modifier = VKD3DSPSM_NEG; break;
+            case 0x81: *modifier = VKD3DSPSM_ABS; break;
+            case 0xc1: *modifier = VKD3DSPSM_ABSNEG; break;
+        };
```
HansKristian-Work, line 1547: weird semi-colon

libs/vkd3d-shader/dxbc.c
```diff
@@ -1538,30 +1538,19 @@ static bool shader_sm4_read_param(struct vkd3d_sm4_data *priv, const DWORD **ptr
...
 
-        switch (m)
+        switch (m & 0xc0)
         {
-            case 0x41:
-                *modifier = VKD3DSPSM_NEG;
-                break;
+            case 0x00: *modifier = VKD3DSPSM_NONE; break;
+            case 0x40: *modifier = VKD3DSPSM_NEG; break;
+            case 0x81: *modifier = VKD3DSPSM_ABS; break;
```
HansKristian-Work, line 1545: Can 0x81 or 0xc1 occur here?

doitsujin: Typo, this should be 0x80. 1 at the end can never occur because it's masked out.

doitsujin: Ideally we'd have the flags defined somewhere, but the headers currently don't do that...

PR #137, opened 2020-06-11
Move to Meson build system

Joshua-Ashton:
This moves VKD3D to build with the Meson build system.

This also builds dxil-spirv as a dependency, using Meson's CMake intergration.

This also makes vkd3d build fine on Windows, Native Linux and Cross-Building with MinGW64 :)

*(For cross-building vkd3d.dll with MinGW you will need to install mingw64-cmake in your package manager, fwiw chaotic-aur provides this package)*

### DIFF
--- a/build-win32.txt
+++ b/build-win32.txt
@@ -1,0 +1,19 @@
+[binaries]
+c = 'i686-w64-mingw32-gcc'
+cpp = 'i686-w64-mingw32-g++'
+ar = 'i686-w64-mingw32-ar'
+strip = 'i686-w64-mingw32-strip'
+cmake = 'i686-w64-mingw32-cmake'
+
+[properties]
+c_args=['-msse', '-msse2']
+cpp_args=['-msse', '-msse2']
+c_link_args = ['-static', '-static-libgcc']
+cpp_link_args = ['-static', '-static-libgcc', '-static-libstdc++']
+needs_exe_wrapper = true
+
+[host_machine]
+system = 'windows'
+cpu_family = 'x86'
+cpu = 'x86'
+endian = 'little'
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,6 @@ vkd3d-*.tar.xz
 .deps
 .dirstamp
 .libs
+build
+build.native
+build.cross
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,9 @@
+[submodule "subprojects/dxil-spirv"]
+	path = subprojects/dxil-spirv
+	url = https://github.com/HansKristian-Work/dxil-spirv
+[submodule "subprojects/Vulkan-Headers"]
+	path = subprojects/Vulkan-Headers
+	url = https://github.com/KhronosGroup/Vulkan-Headers
+[submodule "subprojects/SPIRV-Headers"]
+	path = subprojects/SPIRV-Headers
+	url = https://github.com/KhronosGroup/SPIRV-Headers
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,6 +1,6 @@
 ACLOCAL_AMFLAGS = -I m4
 AM_CFLAGS = @VKD3D_CFLAGS@
-AM_CPPFLAGS = -I$(srcdir)/include -I$(srcdir)/include/dummy -I$(srcdir)/include/private
+AM_CPPFLAGS = -I$(srcdir)/include -I$(srcdir)/include/dummy -I$(srcdir)/include/private -I$(srcdir)/subprojects/SPIRV-Headers/include -I$(srcdir)/subprojects/Vulkan-Headers/include
 AM_LDFLAGS = -no-undefined
 
 widl_headers = \
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,5 +1,5 @@
 ACLOCAL_AMFLAGS = -I m4
-AM_CFLAGS = @VKD3D_CFLAGS@
+AM_CFLAGS = @VKD3D_CFLAGS@ -DPACKAGE_VERSION=\"$(PACKAGE_VERSION)\" -D_GNU_SOURCE
 AM_CPPFLAGS = -I$(srcdir)/include -I$(srcdir)/include/dummy -I$(srcdir)/include/private -I$(srcdir)/subprojects/SPIRV-Headers/include -I$(srcdir)/subprojects/Vulkan-Headers/include
 AM_LDFLAGS = -no-undefined
 
--- a/build-win64.txt
+++ b/build-win64.txt
@@ -1,0 +1,17 @@
+[binaries]
+c = 'x86_64-w64-mingw32-gcc'
+cpp = 'x86_64-w64-mingw32-g++'
+ar = 'x86_64-w64-mingw32-ar'
+strip = 'x86_64-w64-mingw32-strip'
+cmake = 'x86_64-w64-mingw32-cmake'
+
+[properties]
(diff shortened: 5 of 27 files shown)


HansKristian-Work:
Tried building a bit. I noticed that we end up with two submodules of SPIRV-Headers. One in dxil-spirv and one in vkd3d. I tried to tweak dxil-spirv such that we could use a custom path for spirv-headers, but Meson complained:

```
|WARNING: CMake: path /tmp/vkd3d/subprojects/SPIRV-Headers/include/spirv/unified1 is inside the root project but not inside the subproject.
|WARNING:  --> Ignoring. This can lead to build errors.
|WARNING: CMake: path /tmp/vkd3d/subprojects/SPIRV-Headers/include/spirv/unified1 is inside the root project but not inside the subproject.
|WARNING:  --> Ignoring. This can lead to build errors.
|WARNING: CMake: path /tmp/vkd3d/subprojects/SPIRV-Headers/include/spirv/unified1 is inside the root project but not inside the subproject.
|WARNING:  --> Ignoring. This can lead to build errors.
|WARNING: CMake: path /tmp/vkd3d/subprojects/SPIRV-Headers/include/spirv/unified1 is inside the root project but not inside the subproject.
|WARNING:  --> Ignoring. This can lead to build errors.
|WARNING: CMake: path /tmp/vkd3d/subprojects/SPIRV-Headers/include/spirv/unified1 is inside the root project but not inside the subproject.
|WARNING:  --> Ignoring. This can lead to build errors.
|WARNING: CMake: path /tmp/vkd3d/subprojects/SPIRV-Headers/include/spirv/unified1 is inside the root project but not inside the subproject.
|WARNING:  --> Ignoring. This can lead to build errors.
|CMake project dxil-spirv has 8 build targets.
```
It ended up not building, not sure how to fix this. I'm not sure if there is a good way around this. This is more an annoyance than a serious issue though, but it means we need a recursive submodule update.

I was hacking around with this, to give you an idea:
```
diff --git a/third_party/CMakeLists.txt b/third_party/CMakeLists.txt
index e6afb3e..150295f 100644
--- a/third_party/CMakeLists.txt
+++ b/third_party/CMakeLists.txt
@@ -1,9 +1,16 @@
-add_subdirectory(spirv-headers EXCLUDE_FROM_ALL)
-set(SPIRV-Headers_SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/spirv-headers" CACHE STRING "SPIRV-Headers path" FORCE)
+
+if (DXIL_SPIRV_HEADERS)
+       message("Using custom SPIRV-Headers path.")
+       #add_subdirectory(${DXIL_SPIRV_HEADERS} spirv-headers EXCLUDE_FROM_ALL)
+    set(SPIRV-Headers_SOURCE_DIR "${DXIL_SPIRV_HEADERS}" CACHE STRING "SPIRV-Headers path" FORCE)
+else()
+       message("Using submodule SPIRV-Headers path.")
+       #add_subdirectory(spirv-headers EXCLUDE_FROM_ALL)
+    set(SPIRV-Headers_SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/spirv-headers" CACHE STRING "SPIRV-Headers path" FORCE)
+endif()
 
 add_library(dxil-spirv-headers INTERFACE)
-target_include_directories(dxil-spirv-headers INTERFACE
-        ${CMAKE_CURRENT_SOURCE_DIR}/spirv-headers/include/spirv/unified1)
+target_include_directories(dxil-spirv-headers INTERFACE "${SPIRV-Headers_SOURCE_DIR}/include/spirv/unified1")
 
 if (DXIL_SPIRV_CLI)
     add_subdirectory(SPIRV-Tools EXCLUDE_FROM_ALL)
@@ -24,8 +31,8 @@ if (CMAKE_COMPILER_IS_GNUCXX OR (${CMAKE_CXX_COMPILER_ID} MATCHES "Clang"))
 endif()
 
 target_include_directories(glslang-spirv-builder PUBLIC
-        ${CMAKE_CURRENT_SOURCE_DIR}/glslang-spirv)
-target_link_libraries(glslang-spirv-builder PUBLIC dxil-spirv-headers)
+               ${CMAKE_CURRENT_SOURCE_DIR}/glslang-spirv
+               ${SPIRV-Headers_SOURCE_DIR}/include/spirv/unified1)
 
 add_library(bc-decoder STATIC
         bc-decoder/llvm_bitreader.h bc-decoder/llvm_decoder.cpp
```

Another thing to fix would be the sonames, right now I get this for the install target:
```
Installing subprojects/dxil-spirv/libcm_dxil_spirv_c_shared.so to /tmp/foo/lib
Installing libs/vkd3d-shader/libvkd3d_shader.so to /tmp/foo/lib
Installing libs/vkd3d/libvkd3d.so to /tmp/foo/lib
Installing libs/vkd3d-utils/libvkd3d_utils.so to /tmp/foo/lib
```
These should get soname versioning and symlinking to be compatible with Wine. libcm_dxil_spirv_c_shared.so looks wrong as well.

HansKristian-Work:
Do we need the vulkan-1.lib stubs? We should be linking against the loader dynamically already I think.

build-win32.txt
```diff
@@ -0,0 +1,19 @@
+[binaries]
+c = 'i686-w64-mingw32-gcc'
+cpp = 'i686-w64-mingw32-g++'
+ar = 'i686-w64-mingw32-ar'
+strip = 'i686-w64-mingw32-strip'
+strip = 'i686-w64-mingw32-cmake'
```
Saancreed, line 6: Shouldn't this be `cmake = 'i686-w64-mingw32-cmake'`?

doitsujin:
The config.h files should really be generated dynamically, see https://mesonbuild.com/Configuration.html.

There's also a bunch of `#define`s that can be nuked since we're not using them at all.

doitsujin:
Another thing, when building as a Windows DLL we'd probably want this to build a d3d12.dll. Still need to implement the entry points, though.

Joshua-Ashton:
> Do we need the vulkan-1.lib stubs? We should be linking against the loader dynamically already I think.

For now, yes, because the demos and tests link against Vulkan directly without some loader.

We can, and I'd like to, eliminate this later down the line though.

build-win32.txt
Joshua-Ashton: Yes, I have fixed this now. Good spot.

HansKristian-Work:
> > Do we need the vulkan-1.lib stubs? We should be linking against the loader dynamically already I think.
> 
> For now, yes, because the demos and tests link against Vulkan directly without some loader.
> 
> We can, and I'd like to, eliminate this later down the line though.

Which targets require this? I really don't want to commit binary blobs to the repo if we can avoid it. We can clean that up in another PR which is committed before this one.

HansKristian-Work:
I have a pending PR which removes need to link directly against libvulkan. https://github.com/HansKristian-Work/vkd3d/pull/138

HansKristian-Work:
@Joshua-Ashton there are some refactoring commits here that would be good to have in general. Would it make sense to make separate PRs for them?

Joshua-Ashton:
Yeah sure, I'll move out some stuff to separate PRs.

HansKristian-Work:
There are some issues:
- No sonames for installed SOs.
- The DXIL library is named `libcm_dxil_spirv_c_shared.so` for some reason.

Joshua-Ashton:
I added versioning to vkd3d as per Discord discussion.

The DXIL library shouldn't matter given we build it and link to it.

PR #145, opened 2020-06-16
vkd3d: Refactor platform-specific code

Joshua-Ashton:
This PR moves the module handling code which was previously dumped in device.c and the code to retrieve the current executable path to its own file.

This also eliminates HAVE_DECL_PROGRAM_INVOCATION_NAME from config.h

It also implements vkd3d_get_program_name for Win32.


Makefile.am
```diff
@@ -110,6 +110,7 @@ libvkd3d_la_SOURCES = \
 	libs/vkd3d/command.c \
 	libs/vkd3d/device.c \
 	libs/vkd3d/meta.c \
+	libs/vkd3d/platform.c \
```
HansKristian-Work, line 113: New headers should also be added to Makefile.am

libs/vkd3d/platform.c
```diff
@@ -0,0 +1,137 @@
+#include "vkd3d_platform.h"
```
HansKristian-Work, line 1: Need copyright header.

Makefile.am
Joshua-Ashton: Wack. Okay.

Joshua-Ashton: We're missing the one for vkd3d_atomic.h in that case...

Not sure why the headers are included in the autotools sources given they're just going to be ignored but... I'll add another commit to make this consistent anyway

Joshua-Ashton: Done.

libs/vkd3d/platform.c
Joshua-Ashton: Done.

Makefile.am
HansKristian-Work: Ye, not sure either. Just following convention. It makes sense if there is some magical IDE tooling with automake (which I doubt).

PR #155, opened 2020-06-24
build: Fix MinGW build.

HansKristian-Work:
Revert removal of MinGW forceinline hacks.
They are necessary or MinGW fails to build.


doitsujin:
@Joshua-Ashton can you have a look?

Joshua-Ashton:
Hm, this seems to need it now... Not sure what was going on with my home PC where I needed to remove this...

PR #160, opened 2020-06-24
vkd3d-shader: Build statically

Joshua-Ashton:
I had to manually pull in the dependencies of `dxil-spirv` in Meson to link against; initially I thought that was because they weren't declared `PUBLIC` in CMake, but that didn't work either -- seems to be a bug with the Meson CMake integration.

This also tosses the `VKD3D_DEBUG_ENV_NAME` stuff that's really janky, it links back into vkd3d-common from the modules so you can set the env var to enable debugging per-module. Given this is static now we'd have two so that doesn't work anymore, so I just made it use `VKD3D_DEBUG` always which imo is much more sensible.


Joshua-Ashton:
Issue https://github.com/HansKristian-Work/vkd3d/issues/149

PR #166, opened 2020-06-25
Remove SPIRV-Tools

Saancreed:
Removes quite obsolete dependency on SPIRV-Tools, which, from what I was told, is unused anyway.

Affects only autotools–based builds, Meson never used it in the first place.


Saancreed:
Done.

PR #168, opened 2020-06-26
build: Enable warning level 2

Joshua-Ashton:
Fix some signed comparison warnings and enable warning level 2 in Meson
(Equivalent to -Wall -Wextra)

Should help spot some of the enum issues we've had in the past...

### DIFF
--- a/libs/vkd3d-common/utf8.c
+++ b/libs/vkd3d-common/utf8.c
@@ -103,6 +103,12 @@ static uint32_t vkd3d_utf16_read(const uint16_t **src)
     return 0x10000 + ((s[0] & 0x3ff) << 10) + (s[1] & 0x3ff);
 }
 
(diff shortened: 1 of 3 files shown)


demos/meson.build
```diff
@@ -22,10 +22,16 @@ executable('gears', 'gears.c', vkd3d_headers,
   dependencies        : demo_vkd3d_deps,
   install             : true,
   gui_app             : true,
-  override_options    : [ 'c_std='+vkd3d_c_std ])
+  override_options    : [
+    'c_std='         + vkd3d_c_std,
+    'warning_level=' + vkd3d_warning_level
+  ])
```
doitsujin, line 28: is there no way to set these globally? at least c_std should be a project option.

Joshua-Ashton: Ah hah, we can!

Joshua-Ashton: Done.

PR #176, opened 2020-07-01
Standalone D3D12 build

### LINKED ISSUE #151: Support standalone d3d12.dll build
Right now, we must build wined3d on Windows to be able to use vkd3d on Windows, which has a lot of issues associated with it, where RenderDoc breaks and almost no content runs on Windows. We also cannot implement certain APIs properly as we need to make Windows calls in vkd3d.

Goal is to use dxgi.dll from DXVK, but we will need to implement our custom swapchain in vkd3d.

Joshua-Ashton:
When building for Windows, specify `-Denable_standalone_d3d12=True` in Meson to enable.

Need to rebase against https://github.com/HansKristian-Work/vkd3d/pull/175 when that gets merged (contains those commits for now).

Closes #151


doitsujin:
Can you split `92fa100` "vkd3d: Add header variant for MinGW" into three commits so that the addition of the DXGI swap chain implementation as well as the addition of `d3d12.def` are separate?

libs/vkd3d/command.c
```diff
@@ -6821,6 +6832,9 @@ static ULONG STDMETHODCALLTYPE d3d12_command_queue_Release(ID3D12CommandQueue *i
         const struct vkd3d_vk_device_procs *vk_procs = &device->vk_procs;
 
         vkd3d_private_store_destroy(&command_queue->private_store);
+#ifdef VKD3D_BUILD_STANDALONE_D3D12
```
doitsujin, line 6835: Do we really need all these ifdefs?

libs/vkd3d/vulkan_procs.h
```diff
@@ -213,6 +213,26 @@ VK_DEVICE_EXT_PFN(vkCmdBindTransformFeedbackBuffersEXT)
 VK_DEVICE_EXT_PFN(vkCmdEndQueryIndexedEXT)
 VK_DEVICE_EXT_PFN(vkCmdEndTransformFeedbackEXT)
 
+#ifdef VKD3D_BUILD_STANDALONE_D3D12
```
doitsujin, line 216: Again, why the ifdef? Just always load them, and hide the win32-specific functions behind an *appropriate* ifdef (`#ifdef VK_KHR_win32_surface`).

libs/vkd3d/swapchain.c
```diff
@@ -0,0 +1,2252 @@
...
+static HRESULT d3d12_swapchain_factory_init(struct d3d12_command_queue *queue, struct d3d12_swapchain_factory *factory)
+{
+    factory->IWineDXGISwapChainFactory_iface.lpVtbl = &d3d12_swapchain_factory_vtbl;
+    factory->refcount = 1;
+    factory->queue = queue;
+
+    return S_OK;
+}
+
+HRESULT d3d12_swapchain_factory_create(struct d3d12_command_queue *queue, IWineDXGISwapChainFactory** factory)
```
doitsujin, line 2228: This whole thing is **way** overengineerd. Do you expect to add anything that warrants a memory allocation? Can't we just share the ref count with the queue?

I'd much rather have just the init function and nothing else, and make the struct itself a member of the queue rather than a pointer to it, this would also allow us to get rid of half the `#ifdefs` in common code.

Joshua-Ashton:
> Can you split 92fa100 "vkd3d: Add header variant for MinGW" into three commits so that the addition of the DXGI swap chain implementation as well as the addition of d3d12.def are separate?

Yeah, the commits are not split correctly rn.

libs/vkd3d/vulkan_procs.h
Joshua-Ashton: Resolved locally.

libs/vkd3d/swapchain.c
Joshua-Ashton: Resolved locally.

libs/vkd3d/command.c
Joshua-Ashton: Resolved locally.

libs/d3d12/main.c
```diff
@@ -0,0 +1,377 @@
...
+ */
+
+#define VK_NO_PROTOTYPES
+#include <vkd3d_win32.h>
+#include <stdio.h>
+
+#define ERR(fmt, ...) printf("vkd3d: " fmt, ##__VA_ARGS__)
+#define WARN(fmt, ...) printf("vkd3d: " fmt, ##__VA_ARGS__)
+#define TRACE(fmt, ...) printf("vkd3d: " fmt, ##__VA_ARGS__)
+#define FIXME(fmt, ...) printf("vkd3d: " fmt, ##__VA_ARGS__)
```
doitsujin, line 28: Is there a reason why we can't use the regular macros here? This seems like a lot of ugly and (hopefully) unnecessary code duplication, same for the `debugstr_guid` thing below.

libs/d3d12/main.c
```diff
@@ -0,0 +1,377 @@
...
+    if (!(thread_data = malloc(sizeof(*thread_data))))
+    {
+        ERR("Failed to allocate thread data.\n");
+        return NULL;
+    }
+
+    thread_data->main_pfn = main_pfn;
+    thread_data->data = data;
+
+    if (!(thread = CreateThread(NULL, 0, d3d12_thread_main, thread_data, 0, NULL)))
```
doitsujin, line 99: Is the extra wrapper needed due to calling conventions?

libs/d3d12/main.c
```diff
@@ -0,0 +1,377 @@
...
+    PFN_vkd3d_thread main_pfn;
+    void *data;
+};
+
+static DWORD WINAPI d3d12_thread_main(void *data)
+{
+    struct d3d12_thread_data *thread_data = data;
+
+    thread_data->main_pfn(thread_data->data);
+    free(thread_data);
```
doitsujin, line 81: Would probably be cleaner to do this on `join`.

libs/d3d12/main.c
```diff
@@ -0,0 +1,377 @@
...
+                vk_physical_device = vk_physical_devices[i];
+                break;
+            }
+        }
+    }
+
+    if (!vk_physical_device)
+    {
+        FIXME("Could not find Vulkan physical device for DXGI adapter.\n");
+        WARN("Using first available physical device...\n");
```
doitsujin, line 250: Maybe for some big brain fallback we could try to find a device with matching vendor/device ID, like DXVK does.

libs/vkd3d/swapchain.c
```diff
@@ -0,0 +1,2207 @@
...
+    allocate_info.sType = VK_STRUCTURE_TYPE_MEMORY_ALLOCATE_INFO;
+    allocate_info.pNext = NULL;
+    allocate_info.allocationSize = memory_size;
+
+    if (FAILED(hr = vk_select_memory_type(vk_procs, vk_physical_device,
+            memory_type_mask, VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT, &allocate_info.memoryTypeIndex)))
+        return hr;
+
+    assert(swapchain->vk_memory == VK_NULL_HANDLE);
+    if ((vr = vk_procs->vkAllocateMemory(vk_device, &allocate_info, NULL, &swapchain->vk_memory)) < 0)
```
doitsujin, line 503: Any reason not to use vkd3d's existing memory allocation code here?

I know it's a bit crap since it expects D3D12 heap properties, but you get stuff like dedicated allocations for free. At the very least, use those here.

libs/vkd3d/swapchain.c
```diff
@@ -0,0 +1,2207 @@
...
+    }
+    else
+    {
+        /* FIXME: handle DXGI_SCALING_ASPECT_RATIO_STRETCH. */
+        blit.dstOffsets[1].x = swapchain->vk_swapchain_width;
+        blit.dstOffsets[1].y = swapchain->vk_swapchain_height;
+    }
+    blit.dstOffsets[1].z = 1;
+
+    vk_procs->vkCmdBlitImage(vk_cmd_buffer,
```
doitsujin, line 609: We should rewrite this in the future to use an actual graphics pipleine since this frequently produces validation errors. This is good enough for now, but I'd like to see a `TODO` comment here.

libs/d3d12/main.c
Joshua-Ashton: Resolved locally.

libs/d3d12/main.c
Joshua-Ashton: Yes on x86 -- I'm not sure why they aren't just ``__stdcall`` inside VKD3D...

I'll add a comment about this and say that we can toss it later down the line when we don't have a need to be compatible with Wine's D3D12.

libs/d3d12/main.c
Joshua-Ashton: We cannot free it on join, we don't have that information there, only a thread handle.

libs/vkd3d/swapchain.c
Joshua-Ashton: This was all bastardized from DXVK and Wine's DXGI swapchains, can easily do that though 👍

libs/vkd3d/swapchain.c
Joshua-Ashton: Resolved locally.

libs/vkd3d/swapchain.c
Joshua-Ashton: Resolved locally.

libs/d3d12/main.c
Joshua-Ashton: Comment added.

libs/d3d12/main.c
Joshua-Ashton: Resolved locally.

libs/vkd3d/utils.c
```diff
@@ -635,6 +635,137 @@ HRESULT return_interface(void *iface, REFIID iface_iid,
     return hr;
 }
 
+const char* debug_dxgi_format(DXGI_FORMAT format)
```
HansKristian-Work, line 638: The style is `const char *foo` not `const char* foo`.

include/vkd3d_win32.h
```diff
@@ -0,0 +1,68 @@
...
+
+#define VK_USE_PLATFORM_WIN32_KHR
+#include <vulkan/vulkan.h>
+
+#include <dxgi1_6.h>
+
+#include <vkd3d_swapchain_factory.h>
+
+/* We already included regular DXGI...
+ * let's not redefine crap under a new header
```
HansKristian-Work, line 53: More neutral language here, please.

meson.build
```diff
@@ -4,13 +4,17 @@ project('vkd3d', ['c'], version : '1.1', meson_version : '>= 0.51', default_opti
 
 cpu_family = target_machine.cpu_family()
 
+vkd3d_compiler = meson.get_compiler('c')
+vkd3d_c_std    = 'c99'
+vkd3d_platform = target_machine.system()
+
+enable_tests             = get_option('enable_tests')
```
HansKristian-Work, line 12: weird spacing

meson.build
```diff
@@ -5,16 +5,26 @@ project('vkd3d', ['c'], version : '1.1', meson_version : '>= 0.51', default_opti
...
 vkd3d_compiler = meson.get_compiler('c')
+vkd3d_msvc     = vkd3d_compiler.get_id() == 'msvc'
 vkd3d_c_std    = 'c99'
 vkd3d_platform = target_machine.system()
 
 enable_tests             = get_option('enable_tests')
 enable_extras            = get_option('enable_extras')
+enable_standalone_d3d12  = get_option('enable_standalone_d3d12')
+
+if vkd3d_platform != 'windows' and enable_standalone_d3d12
```
HansKristian-Work, line 16: I generally had problems with this on MinGW. vkd3d_platform is detected as mingw64_nt or something like that in msys2.

include/vkd3d_swapchain_factory.h
```diff
@@ -0,0 +1,69 @@
...
+
+typedef interface IDXGIFactory IDXGIFactory;
+typedef struct DXGI_SWAP_CHAIN_DESC1 DXGI_SWAP_CHAIN_DESC1;
+typedef struct DXGI_SWAP_CHAIN_FULLSCREEN_DESC DXGI_SWAP_CHAIN_FULLSCREEN_DESC;
+typedef interface IDXGIOutput IDXGIOutput;
+typedef interface IDXGISwapChain1 IDXGISwapChain1;
+
+typedef interface IWineDXGISwapChainFactory IWineDXGISwapChainFactory;
+
+DEFINE_GUID(IID_IWineDXGISwapChainFactory, 0x53cb4ff0,0xc25a,0x4164,0xa8,0x91,0x0e,0x83,0xdb,0x0a,0x7a,0xac);
```
HansKristian-Work, line 31: Wouldn't this normally be part of an IDL header?

libs/vkd3d/swapchain.c
```diff
@@ -0,0 +1,2216 @@
+/*
+ * Copyright 2020 Joshua Ashton for Valve Software
```
HansKristian-Work, line 2: If this is mostly copied from Wine, shouldn't the copyright statements also be included?

libs/vkd3d/meson.build
```diff
@@ -30,6 +30,10 @@ vkd3d_src = [
   'vkd3d_main.c',
 ]
 
+if enable_standalone_d3d12
+  vkd3d_src += 'swapchain.c'
```
HansKristian-Work, line 34: What about the d3d12 stubs?

libs/vkd3d/swapchain.c
```diff
@@ -0,0 +1,2216 @@
...
+static HRESULT d3d12_swapchain_set_display_mode(struct d3d12_swapchain *swapchain,
+        IDXGIOutput* output, DXGI_MODE_DESC *mode)
+{
+    FIXME("TODO");
+    return S_OK;
+}
+
+static HRESULT d3d12_swapchain_set_fullscreen(struct d3d12_swapchain *swapchain)
+{
+    FIXME("TODO!");
```
HansKristian-Work, line 196: Does this work in Wine?

libs/vkd3d/swapchain.c
```diff
@@ -0,0 +1,2216 @@
...
+    unsigned int i;
+    VkResult vr;
+    HRESULT hr;
+
+    if (d3d12_swapchain_has_user_images(swapchain))
+        return S_OK;
+
+    memset(&image_info, 0, sizeof(image_info));
+    image_info.sType = VK_STRUCTURE_TYPE_IMAGE_CREATE_INFO;
+    image_info.flags = VK_IMAGE_CREATE_MUTABLE_FORMAT_BIT;
```
HansKristian-Work, line 442: This breaks DCC unless we use image_format_list, did Wine do this, is there a requirement that we can create arbitrarily aliased views in D3D12?

libs/vkd3d/swapchain.c
```diff
@@ -0,0 +1,2216 @@
...
+        dedicated_allocate_info.image  = swapchain->vk_images[i];
+        dedicated_allocate_info.buffer = VK_NULL_HANDLE;
+
+        if (FAILED(hr = vk_select_memory_type(vk_procs, vk_physical_device,
+            requirements.memoryRequirements.memoryTypeBits,
+            VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT, &memory_type_index)))
+            return hr;
+
+        allocate_info.sType           = VK_STRUCTURE_TYPE_MEMORY_ALLOCATE_INFO;
+        allocate_info.pNext           = dedicated_requirements.prefersDedicatedAllocation
```
HansKristian-Work, line 501: We're going to allocate once per image anyways, why bother to check for dedicated prefer? I'd suggest just always making a dedicated allocation here.

libs/vkd3d/swapchain.c
doitsujin: Wine did indeed do this, and we need to be able to cast formats within the same format family.

It's probably better to create these images using existing vkd3d functions (`vkd3d_image_resource_create_info`) since format compatibility will be handled by that.

libs/vkd3d/swapchain.c
```diff
@@ -0,0 +1,2216 @@
...
+
+        assert(swapchain->vk_memory[i] == VK_NULL_HANDLE);
+        if ((vr = vk_procs->vkAllocateMemory(vk_device, &allocate_info, NULL, &swapchain->vk_memory[i])) != VK_SUCCESS)
+        {
+            WARN("Failed to allocate device memory, vr %d.\n", vr);
+            swapchain->vk_memory[i] = VK_NULL_HANDLE;
+            return hresult_from_vk_result(vr);
+        }
+
+       if ((vr = vk_procs->vkBindImageMemory(vk_device, swapchain->vk_images[i],
```
HansKristian-Work, line 515: Weird formatting.

libs/d3d12/main.c
```diff
@@ -0,0 +1,392 @@
...
+#define VK_NO_PROTOTYPES
+#include "vkd3d_win32.h"
+#include "vkd3d_debug.h"
+
+static PFN_vkGetInstanceProcAddr load_vulkan(void)
+{
+    static HMODULE vulkan_module = NULL;
+
+    if (!vulkan_module)
+        vulkan_module = LoadLibraryA("vulkan-1.dll");
```
HansKristian-Work, line 30: This is not thread-safe.

libs/vkd3d/meson.build
HansKristian-Work: Was in a later commit, nvm.

libs/d3d12/main.c
HansKristian-Work: Perhaps just guard it with a static spinlock_t or something trivial.

meson.build
Joshua-Ashton: You **need** to use the cross-file, not just set CC=mingw-w64-gcc, etc.

include/vkd3d_swapchain_factory.h
Joshua-Ashton: It doesn't really work out because our DXGI stuff doesn't match up.

libs/vkd3d/utils.c
Joshua-Ashton: Resolved locally.

include/vkd3d_win32.h
Joshua-Ashton: Resolved locally.

meson.build
Joshua-Ashton: This lines up with enable_standalone_d3d12

include/vkd3d_swapchain_factory.h
Joshua-Ashton: I'll try it

meson.build
HansKristian-Work: Even when building on native Windows? From Linux cross-compiling, sure.

include/vkd3d_swapchain_factory.h
Joshua-Ashton: Resolved locally.

meson.build
Joshua-Ashton: You should still use the cross file, it *should* work -- also I'd highly recommend using WSL over msys2 for building like this.

libs/vkd3d/swapchain.c
Joshua-Ashton: Resolved locally.

libs/vkd3d/swapchain.c
Joshua-Ashton: Changed to use vkd3d resource creation.

libs/vkd3d/swapchain.c
Joshua-Ashton: Changed to use vkd3d resource creation.

libs/vkd3d/swapchain.c
Joshua-Ashton: Changed to use vkd3d resource creation.

libs/vkd3d/swapchain.c
Joshua-Ashton: Partially... but I have now done a proper and full implementation of fullscreen.

libs/d3d12/main.c
Joshua-Ashton: Resolved.

libs/vkd3d/swapchain.c
```diff
@@ -0,0 +1,2544 @@
...
+    resource_desc.SampleDesc.Quality = 0;
+    resource_desc.Layout = D3D12_TEXTURE_LAYOUT_UNKNOWN;
+    resource_desc.Flags = D3D12_RESOURCE_FLAG_ALLOW_RENDER_TARGET;
+
+    for (i = 0; i < swapchain->desc.BufferCount; i++)
+    {
+        if (FAILED(hr = d3d12_committed_resource_create(d3d12_swapchain_device(swapchain),
+            &heap_properties, D3D12_HEAP_FLAG_NONE,
+            &resource_desc, D3D12_RESOURCE_STATE_COPY_SOURCE,
+            NULL, &object)))
```
doitsujin, line 814: Formatting -- for multi-line function calls we have 8 spaces rather than 4, and this can probably be shortened to 3 lines as well rather than 4.

libs/vkd3d/swapchain.c
doitsujin: Would be nice to have that resolved non-locally as well.

HansKristian-Work: It is resolved below.

libs/vkd3d/swapchain.c
Joshua-Ashton: Done did it

PR #182, opened 2020-07-06
Revert "vkd3d: Add support for serializing root signature version 1.1"

HansKristian-Work:
This reverts commit 0384362065aa64650b01985195fde0f91401402d.

It is not allowed to use RS 1.1 serialization for the non-versioned
entry point. RS 1.1 serialization must use the versioned entry point.

Reverting this fixes the relevant test case in d3d12.c:12522.


HansKristian-Work:
@Joshua-Ashton in which scenario did you need D3D12SerializeRootSignature to support 1.1?

Joshua-Ashton:
> @Joshua-Ashton in which scenario did you need D3D12SerializeRootSignature to support 1.1?

After I made the thing use RS 1.0 for testing I forgot to change it back to Versioned root signature, so this was entirely my blunder...

PR #186, opened 2020-07-06
vkd3d: Correctly handle output + fix NULL old targets for fullscreen 

Joshua-Ashton:
The output here is actually for secure presentation and restricting a swapchain to a certain output.

Correctly handle NULL (desktop) targets that we used to have.

Fixes crashes with titles that use fullscreen via an initial fullscreen desc.

### DIFF
--- a/libs/vkd3d/swapchain.c
+++ b/libs/vkd3d/swapchain.c
@@ -384,7 +384,7 @@ static HRESULT d3d12_swapchain_set_display_mode(struct d3d12_swapchain *swapchai
         return hr;
     }
 
-    if (output != swapchain->target)
+    if (output != swapchain->target && swapchain->target)
     {
         if (FAILED(hr = d3d12_output_set_display_mode(swapchain->target, &swapchain->state.original_mode)))
         {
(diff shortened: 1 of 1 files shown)


Joshua-Ashton:
Actually, lemme do some extra cleanup.

libs/vkd3d/swapchain.c
```diff
@@ -2421,23 +2416,25 @@ static HRESULT d3d12_swapchain_init(struct d3d12_swapchain *swapchain, IDXGIFact
         }
     }
 
-    IDXGIFactory_AddRef(swapchain->factory = factory);
-
-    if (FAILED(hr = d3d12_swapchain_set_fullscreen(swapchain, target, TRUE)))
+    if (FAILED(d3d12_swapchain_set_fullscreen(swapchain, target, TRUE)))
```
doitsujin, line 2419: `FAILED(hr = d3d12_swapchain_set_fullscreen`?

libs/vkd3d/swapchain.c
```diff
@@ -2421,23 +2416,25 @@ static HRESULT d3d12_swapchain_init(struct d3d12_swapchain *swapchain, IDXGIFact
...
     {
-        WARN("Failed to go fullscreen.\n");
+        ERR("Failed to enter fullscreen.");
         return hr;
     }
 
-    if (!output && target)
+    if (!swapchain->fullscreen_desc.Windowed)
+        swapchain->target = target;
```
doitsujin, line 2426: Do we initialize swapchain->target anywhere if we don't take this code path?

Joshua-Ashton: It is set to 0 already given it is calloc'ed

libs/vkd3d/swapchain.c
Joshua-Ashton: It should, good catch.

PR #193, opened 2020-07-08
meson: generate pkgconfig files, install headers

telans:
Generates `/usr/lib64/pkgconfig/libvkd3d.pc` as:

```
prefix=/usr
libdir=${prefix}/lib64
includedir=${prefix}/include

Name: vkd3d
Description: The VKD3D 3D Graphics Library
Version: 1.1
Libs: -L${libdir} -lvkd3d
Cflags:-I${includedir}/vkd3d
```

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -77,6 +77,16 @@ vkd3d_version = vcs_tag(
 dxil_spirv = subproject('dxil-spirv')
 dxil_spirv_dep = dxil_spirv.get_variable('dxil_spirv_dep')
 
+if vkd3d_platform == 'linux'
+  pkg = import('pkgconfig')
+  vkd3d = library('vkd3d')
+  pkg.generate(vkd3d, filebase : 'libvkd3d', subdirs : 'vkd3d', description : 'The VKD3D 3D Graphics Library')
+  vkd3dutils = library('vkd3d-utils')
+  pkg.generate(vkd3dutils, filebase : 'libvkd3d-utils', subdirs : 'vkd3d', description : 'The VKD3D 3D Graphics Utility Library')
+
+  install_headers('include/vkd3d.h', 'include/vkd3d_sonames.h', 'include/vkd3d_types.h', 'include/vkd3d_utils.h', 'include/vkd3d_win32.h', 'include/vkd3d_windows.h', subdir : 'vkd3d')
+endif
+
 subdir('include')
 subdir('libs')
 


doitsujin:
Are these needed for anything?

Also, note that vkd3d-shader is no longer shared library but instead linked statically.

telans:
In my case, they're needed when installing vkd3d-proton as a system library through [my ebuild](https://github.com/telans/EBUILDS/tree/master/app-emulation/vkd3d-proton) if you expect to build wine against it

This also matches the generated pkgconfig files created by the existing makefile

HansKristian-Work:
I don't think we install the headers with Meson currently either.

telans:
Installs as 
```
>>> /usr/include/vkd3d/vkd3d_windows.h
>>> /usr/include/vkd3d/vkd3d_win32.h
```
etc

Joshua-Ashton:
I need to test this on Windows to make sure it doesn't break building there (given pkgconfig isnt a thing)

I am going to nap shortly so I will do it after that 👍

telans:
Well I'm having issues getting the generated header files installed. Since you can't `install` through `generator()`, the next best thing would be `custom_target()` but you can't seem to iterate through an array of input/output files with it

HansKristian-Work:
> 
> 
> Well I'm having issues getting the generated header files installed. Since you can't `install` through `generator()`, the next best thing would be `custom_target()` but you can't seem to iterate through an array of input/output files with it
> 
> Do these generated headers need to be installed? They don't look to be required when building wine/vkd3d
> 
> Cheers

Ok, if Wine does not require the headers, then I'm fine leaving them out. I assumed Wine would need them. Either way, I'll defer to @Joshua-Ashton to review and approve this PR.

Saancreed:
> I need to test this on Windows to make sure it doesn't break building there (given pkgconfig isnt a thing)

It does magically work when built in Wine at the very least, but I checked only standalone d3d12 build with VS2019.

```
c:/include
c:/include/vkd3d
c:/include/vkd3d/vkd3d_types.h
c:/include/vkd3d/vkd3d_windows.h
c:/include/vkd3d/vkd3d.h
c:/include/vkd3d/vkd3d_shader.h
c:/include/vkd3d/vkd3d_sonames.h
c:/include/vkd3d/vkd3d_win32.h
c:/include/vkd3d/vkd3d_utils.h
c:/bin
c:/bin/d3d12.dll
c:/bin/d3d12.pdb
c:/bin/vkd3d-utils.pdb
c:/bin/vkd3d-utils.dll
c:/lib64
c:/lib64/vkd3d-utils.lib
c:/lib64/d3d12.lib
c:/lib64/pkgconfig
c:/lib64/pkgconfig/libvkd3d-shader.pc
c:/lib64/pkgconfig/libvkd3d.pc
c:/lib64/pkgconfig/libvkd3d-utils.pc
```

Full build log is [here](https://gist.github.com/Saancreed/83d297e321322109ee7c3f2f8f2c0a58), in case anyone wants to take a look at 168 MSVC warnings :frog: (which are unrelated to changes introduced in this PR).

telans:
@Saancreed so should this be disabled on Windows? I'm unsure how the install dir is supposed to look there. (Is that installing /bin into the c drive anyway?)

Saancreed:
> so should this be disabled on Windows?

I just wanted to confirm that having `pkgconfig` stuff enabled even on Windows won't cause builds to fail. (with vs2019 backend and MSVC compiler at least).

But to answer your question: probably yes, since those files are useless on Windows anyway (same applies to headers when built in standalone d3d12 mode).

> Is that installing /bin into the c drive anyway?

Yup, that's because `c:` is the default value for `--prefix` for Meson on Windows. After this, installed stuff mostly mirrors what you'd get in `/usr/local` on Linux.

That said, I believe Joshie is about to take a look at this anyway, he should be able to tell you what to do here.

PR #194, opened 2020-07-08
Implement aliasing barriers

HansKristian-Work:


libs/vkd3d/command.c
```diff
@@ -4929,16 +4929,65 @@ static bool vk_image_memory_barrier_from_d3d12_transition(const struct d3d12_dev
     return vk_barrier->oldLayout != vk_barrier->newLayout;
 }
 
+static void vk_image_memory_barrier_for_after_aliasing_barrier(struct d3d12_device *device,
+        struct d3d12_resource *after, VkImageMemoryBarrier *vk_barrier)
+{
+    const struct vkd3d_format *vk_format = vkd3d_get_format(device, after->desc.Format, false);
```
HansKristian-Work, line 4935: The depth/stencil argument is a bit confusing, I think this is the right way to do this?

libs/vkd3d/command.c
```diff
@@ -5041,19 +5155,31 @@ static void STDMETHODCALLTYPE d3d12_command_list_ResourceBarrier(d3d12_command_l
...
 
+    /* Normalize pipeline stage / access flags. */
+    if (src_stage_mask & VK_PIPELINE_STAGE_ALL_COMMANDS_BIT)
+        src_stage_mask = VK_PIPELINE_STAGE_ALL_COMMANDS_BIT;
+    if (dst_stage_mask & VK_PIPELINE_STAGE_ALL_COMMANDS_BIT)
+        dst_stage_mask = VK_PIPELINE_STAGE_ALL_COMMANDS_BIT;
+    if (vk_memory_barrier.srcAccessMask & VK_ACCESS_MEMORY_WRITE_BIT)
+        vk_memory_barrier.srcAccessMask = VK_ACCESS_MEMORY_WRITE_BIT;
+    if (vk_memory_barrier.dstAccessMask & VK_ACCESS_MEMORY_WRITE_BIT)
+        vk_memory_barrier.dstAccessMask = VK_ACCESS_MEMORY_READ_BIT | VK_ACCESS_MEMORY_WRITE_BIT;
```
doitsujin, line 5166: RADV doesn't support `VK_ACCESS_MEMORY_*_BIT`, so we'd break existing barriers by using these.

In general I wonder if we really need to do the normalization at all, it might not be pretty to have something like `ALL_COMMANDS | TRANSFER` but it should be legal and work fine regardless.

libs/vkd3d/command.c
```diff
@@ -4929,16 +4929,65 @@ static bool vk_image_memory_barrier_from_d3d12_transition(const struct d3d12_dev
...
+    VkImageMemoryBarrier *vk_image_barriers;
     VkImageMemoryBarrier vk_image_barrier;
     VkMemoryBarrier vk_memory_barrier;
-    unsigned int i;
+    unsigned int num_image_barriers;
+    unsigned int i, j;
+
+    num_image_barriers = 0;
+    if (barrier_count > ARRAY_SIZE(vk_image_barriers_stack))
+        vk_image_barriers = vkd3d_malloc(barrier_count * sizeof(*vk_image_barriers));
```
doitsujin, line 4988: Not a big fan of the allocation here; can't we just emit image barriers one by one? They should be very rare in practice.

tests/d3d12.c
```diff
@@ -44639,6 +44639,158 @@ static void test_texture_feedback_instructions_dxil(void)
     test_texture_feedback_instructions(true);
 }
 
+static void test_aliasing_barrier(void)
```
doitsujin, line 44642: What exactly do we count as "success" here? Not crashing / no validation errors?

libs/vkd3d/command.c
doitsujin: Honestly I'm not sure, ideally we'd store the vkd3d_format pointer with the resource....

tests/d3d12.c
HansKristian-Work: Yes, there isn't much meaningful we can test for here otherwise.

libs/vkd3d/command.c
HansKristian-Work: That's a ... particular driver bug. Spec is very clear on this. Is there a bug report for this?

doitsujin: Not that I'm aware of; I stumbled across that a while ago but forgot about it.

libs/vkd3d/command.c
HansKristian-Work: Looks correct, the depth stencil variant is there to have different typeless overrides.

libs/vkd3d/command.c

libs/vkd3d/command.c

tests/d3d12.c
HansKristian-Work: Added a comment to clarify intent of the test.

PR #198, opened 2020-07-09
Logging improvements

HansKristian-Work:


libs/d3d12/main.c
```diff
@@ -18,6 +18,8 @@
  *
  */
 
+#define VKD3D_DBG_CHANNEL VKD3D_DBG_CHANNEL_API
```
HansKristian-Work, line 21: Not super happy about doing this, the alternative is using private compiler defines per submodule or something.

PR #203, opened 2020-07-14
Make MSVC compile clean

HansKristian-Work:
Adds some warning disable flags to vkd3d (as well as dxil-spirv) to remove the more useless warnings MSVC emits. The more relevant warnings, GCC does not seem to catch with our current flags, so fixed them as well.

MSVC now compiles clean, so we should be able to catch actual stupid bugs now without being boggled down by 200+ warnings.

### DIFF
--- a/libs/vkd3d-shader/dxbc.c
+++ b/libs/vkd3d-shader/dxbc.c
@@ -1937,11 +1937,13 @@ static void read_dword(const char **ptr, DWORD *d)
     *ptr += sizeof(*d);
 }
 
-static void read_uint32(const char **ptr, uint32_t *u)
+static void read_uint32_(const char **ptr, void *u)
 {
-    memcpy(u, *ptr, sizeof(*u));
-    *ptr += sizeof(*u);
+    memcpy(u, *ptr, sizeof(uint32_t));
+    *ptr += sizeof(uint32_t);
 }
+
+#define read_uint32(ptr, u) do { STATIC_ASSERT(sizeof(*(u)) == sizeof(uint32_t)); read_uint32_(ptr, u); } while(0)
 
 static void read_float(const char **ptr, float *f)
 {
--- a/libs/vkd3d/device.c
+++ b/libs/vkd3d/device.c
@@ -528,7 +528,7 @@ static HRESULT vkd3d_instance_init(struct vkd3d_instance *instance,
     vkd3d_free(user_extension_supported);
 
     vr = vk_global_procs->vkCreateInstance(&instance_info, NULL, &vk_instance);
-    vkd3d_free(extensions);
+    vkd3d_free((void *)extensions);
     if (vr < 0)
     {
         ERR("Failed to create Vulkan instance, vr %d.\n", vr);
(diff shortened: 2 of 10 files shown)


libs/vkd3d-shader/dxbc.c
```diff
@@ -2110,13 +2110,13 @@ static int shader_parse_signature(DWORD tag, const char *data, DWORD data_size,
             return VKD3D_ERROR_INVALID_ARGUMENT;
         }
         read_uint32(&ptr, &e[i].semantic_index);
-        read_uint32(&ptr, &e[i].sysval_semantic);
```
doitsujin, line 2113: I wonder if we could do this more easily by just making `read_uint32` take `void*`. Not pretty either, but we're already passing random stuff to it and have to make sure it's actually four bytes in size.

libs/vkd3d/device.c
```diff
@@ -528,7 +528,7 @@ static HRESULT vkd3d_instance_init(struct vkd3d_instance *instance,
     vkd3d_free(user_extension_supported);
 
     vr = vk_global_procs->vkCreateInstance(&instance_info, NULL, &vk_instance);
-    vkd3d_free(extensions);
+    vkd3d_free((char **)extensions);
```
doitsujin, line 531: What's the problem here exactly, const-ness?

meson.build
```diff
@@ -59,7 +59,15 @@ endif
...
-    '-Wno-unused-parameter']),
+    '-Wno-unused-parameter',
+    '/wd4244', # Narrowing conversion
+    '/wd4101', # Unreferenced local variable
+    '/wd4267', # Another narrowing conversion
+    '/wd4996', # Secure no warnings
+    '/wd4334', # Result of 32-bit shift cast to 64-bit
+    '/wd4146', # Unary minus on unsigned
+    '/wd4305', # Truncation from double to float
+    ]),
```
doitsujin, line 70: Do we pass all of these to gcc as well and does that work?

HansKristian-Work: It's filtered by Meson based on what works.

libs/vkd3d-shader/dxbc.c
HansKristian-Work: ye, `void *` should work as well, might be cleaner.

libs/vkd3d/device.c
HansKristian-Work: const, yes. I don't like removing const warnings, since those can be quite serious.

HansKristian-Work: We could also make vkd3d_free take `const void*`, but that feels a bit icky too.

meson.build
HansKristian-Work: Similarly, GNU-style flags are filtered out when using MSVC.

doitsujin: Ah, nice.

libs/vkd3d-shader/dxbc.c
Joshua-Ashton: Not having templates is so annoying.

Would be nice to just read a sizeof(T) or static_assert(sizeof(T) == sizeof(uint32_t))

libs/vkd3d/device.c
Joshua-Ashton: I think you can eliminate this with `char *const *extensions;` or something similar.

The const is the thing in the wrong place.

Joshua-Ashton:
Other than what I commented on it looks good.

I hope we can eventually get to a point where we don't need to disable any warnings.

libs/vkd3d/device.c
HansKristian-Work: The const is in the right place I think, because ppEnabledExtensions is `const char * const *extensions`, and you would get a similar warning if you were to assign `char * const` to `const char *`.

PR #205, opened 2020-07-14
build: Drop minimum meson version

aeikum:
Debian 10 ships 0.49.

Signed-off-by: Andrew Eikum <aeikum@codeweavers.com>


rbernon:
I think there's the same kind of thing in dxil-spirv.

aeikum:
Good catch, opened HansKristian-Work/dxil-spirv#7.

PR #207, opened 2020-07-15
Implement a simple CPU profiling setup.

HansKristian-Work:


include/private/vkd3d_profiling.h
```diff
@@ -0,0 +1,88 @@
...
+    return ts.tv_sec * 1000000000ll + ts.tv_nsec;
+#endif
+}
+
+#define VKD3D_REGION_DECL(name) \
+    static uint32_t _vkd3d_region_latch_##name; \
+    static spinlock_t _vkd3d_region_lock_##name; \
+    uint64_t _vkd3d_region_begin_tick_##name; \
+    uint64_t _vkd3d_region_end_tick_##name; \
+    unsigned int _vkd3d_region_index_##name
```
Joshua-Ashton, line 62: Struct? :-)

libs/vkd3d/command.c
```diff
@@ -7142,7 +7151,16 @@ static HRESULT d3d12_command_list_init(struct d3d12_command_list *list, struct d
     HRESULT hr;
 
     memset(list, 0, sizeof(*list));
+
+#ifdef VKD3D_ENABLE_PROFILING
+    if (vkd3d_uses_profiling())
+        list->ID3D12GraphicsCommandList_iface.lpVtbl = &d3d12_command_list_vtbl_profiled;
+    else
+        list->ID3D12GraphicsCommandList_iface.lpVtbl = &d3d12_command_list_vtbl;
```
Joshua-Ashton, line 7159: Maybe just make vkd3d_use_profiling be some inline thing that returns 0 when not defined?

include/private/vkd3d_profiling.h
HansKristian-Work: Gets a bit messy since it's a mix of static and non-static variables here. It's wrapped in a macro anyways, so don't see the gain of splitting this up into a struct.

libs/vkd3d/command.c
HansKristian-Work: Possible, but that means having to build in the profiling vtables and entry points always, which seems a bit useless.

include/private/vkd3d_profiling.h
Joshua-Ashton: Hmmm, understandable in that case then.

libs/vkd3d/command.c
Joshua-Ashton: If only we had constexpr :(

PR #208, opened 2020-07-15
vkd3d: Implement DiscardResource.

doitsujin:
Tries to make use of our existing render pass transitions where possible, and otherwise just emits a barrier transitioning away from ``UNDEFINED``.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -2092,6 +2092,7 @@ static void d3d12_command_list_clear_attachment_deferred(struct d3d12_command_li
      * depth-only clear does not override a stencil-only clear. */
     clear_state->attachment_mask |= 1u << attachment_idx;
     attachment->aspect_mask |= clear_aspects;
+    attachment->discard_mask &= ~clear_aspects;
 
     if (clear_aspects & VK_IMAGE_ASPECT_COLOR_BIT)
         attachment->value.color = clear_value->color;
@@ -2121,6 +2122,7 @@ static bool d3d12_command_list_has_depth_stencil_view(struct d3d12_command_list 
 
 static bool d3d12_command_list_has_render_pass_dsv_clear(struct d3d12_command_list *list)
 {
+    struct vkd3d_clear_attachment *clear_attachment = &list->clear_state.attachments[D3D12_SIMULTANEOUS_RENDER_TARGET_COUNT];
     VkImageAspectFlags clear_aspects, write_aspects;
 
     if (!d3d12_pipeline_state_is_graphics(list->state))
@@ -2135,13 +2137,14 @@ static bool d3d12_command_list_has_render_pass_dsv_clear(struct d3d12_command_li
     /* If any of the aspects to clear are read-only in the render
      * pass, we have to perform the DSV clear in a separate pass. */
     write_aspects = vk_writable_aspects_from_image_layout(list->dsv_layout);
-    clear_aspects = list->clear_state.attachments[D3D12_SIMULTANEOUS_RENDER_TARGET_COUNT].aspect_mask;
+    clear_aspects = clear_attachment->aspect_mask | clear_attachment->discard_mask;
     return (write_aspects & clear_aspects) == clear_aspects;
 }
 
 static void d3d12_command_list_emit_deferred_clear(struct d3d12_command_list *list, unsigned int attachment_idx)
 {
     struct vkd3d_clear_attachment *clear_attachment = &list->clear_state.attachments[attachment_idx];
+    VkImageSubresourceLayers vk_subresource_layers;
     struct d3d12_resource *resource;
     struct vkd3d_view *view;
 
@@ -2158,18 +2161,24 @@ static void d3d12_command_list_emit_deferred_clear(struct d3d12_command_list *li
 
     if (list->current_render_pass)
     {
-        d3d12_command_list_clear_attachment_inline(list, resource,
-                view, attachment_idx, clear_attachment->aspect_mask,
-                &clear_attachment->value, 0, NULL);
+        if (clear_attachment->aspect_mask)
+        {
+            d3d12_command_list_clear_attachment_inline(list, resource,
+                    view, attachment_idx, clear_attachment->aspect_mask,
+                    &clear_attachment->value, 0, NULL);
+        }
     }
     else
     {
-        d3d12_command_list_clear_attachment_pass(list, resource, view,
-                clear_attachment->aspect_mask, &clear_attachment->value,
-                0, NULL);
+        if (clear_attachment->aspect_mask)
+        {
+            d3d12_command_list_clear_attachment_pass(list, resource, view,
+                    clear_attachment->aspect_mask, &clear_attachment->value, 0, NULL);
+        }
     }
 
     clear_attachment->aspect_mask = 0;
+    clear_attachment->discard_mask = 0;
 }
 
 static void d3d12_command_list_emit_render_pass_clears(struct d3d12_command_list *list, bool invert_mask)
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -1809,13 +1809,13 @@ static void d3d12_command_list_invalidate_current_pipeline(struct d3d12_command_
 static bool d3d12_command_list_create_framebuffer(struct d3d12_command_list *list, VkRenderPass render_pass,
         uint32_t view_count, const VkImageView *views, VkExtent3D extent, VkFramebuffer *vk_framebuffer);
 
-static D3D12_RECT d3d12_get_view_rect(struct d3d12_resource *resource, struct vkd3d_view *view)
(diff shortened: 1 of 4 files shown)


libs/vkd3d/resource.c
```diff
@@ -1960,6 +1960,31 @@ struct d3d12_resource *unsafe_impl_from_ID3D12Resource(ID3D12Resource *iface)
...
+    const struct vkd3d_format *format = vkd3d_format_from_d3d12_resource_desc(resource->device, &resource->desc, 0);
+    VkImageSubresource subresource;
+
+    subresource.aspectMask = format->vk_aspect_mask;
+    subresource.mipLevel = subresource_idx % resource->desc.MipLevels;
+    subresource.arrayLayer = (subresource_idx / resource->desc.MipLevels) % resource->desc.DepthOrArraySize;
+
+    if (!all_aspects)
+    {
+        uint32_t i, plane_idx = subresource_idx / (resource->desc.MipLevels * resource->desc.DepthOrArraySize);
```
HansKristian-Work, line 1974: This seems a little sketchy for 3D images. Can they enter this path somehow?

libs/vkd3d/resource.c
```diff
@@ -1960,6 +1960,31 @@ struct d3d12_resource *unsafe_impl_from_ID3D12Resource(ID3D12Resource *iface)
...
+
+        /* Clear the lower 'plane_idx' bits, then isolate the lowest bit.
+         * This works because aspect numbering between D3D12 and Vulkan
+         * is the same. */
+        for (i = 0; i < plane_idx; i++)
```
HansKristian-Work, line 1979: It is a bit hard to understand (without knowing the full implementation) why this is done.

libs/vkd3d/command.c
```diff
@@ -1857,39 +1857,46 @@ static VkImageAspectFlags vk_writable_aspects_from_image_layout(VkImageLayout vk
...
-        const struct d3d12_resource *resource, const struct vkd3d_view *view)
+static int d3d12_command_list_find_attachment_view(struct d3d12_command_list *list,
+        const struct d3d12_resource *resource, const VkImageSubresourceLayers *subresource)
 {
     unsigned int i;
 
     if (list->dsv.resource == resource)
     {
-        const struct vkd3d_view *dsv = list->dsv.view;
+        struct vkd3d_view *dsv = list->dsv.view;
```
HansKristian-Work, line 1867: Why drop the const?

libs/vkd3d/command.c
```diff
@@ -1857,39 +1857,46 @@ static VkImageAspectFlags vk_writable_aspects_from_image_layout(VkImageLayout vk
...
+                dsv->info.texture.layer_idx == subresource->baseArrayLayer &&
+                dsv->info.texture.layer_count == subresource->layerCount)
             return D3D12_SIMULTANEOUS_RENDER_TARGET_COUNT;
     }
     else
     {
         for (i = 0; i < D3D12_SIMULTANEOUS_RENDER_TARGET_COUNT; i++)
         {
-            const struct vkd3d_view *rtv = list->rtvs[i].view;
+            struct vkd3d_view *rtv = list->rtvs[i].view;
```
HansKristian-Work, line 1878: Ditto about const.

libs/vkd3d/command.c
```diff
@@ -6287,7 +6355,77 @@ static void STDMETHODCALLTYPE d3d12_command_list_ClearUnorderedAccessViewFloat(d
...
+        return;
+    }
+
+    /* We can't meaningfully discard sub-regions of an image. If rects
+     * are specified, all specified subresources must have the same
+     * dimensions, so just base this off the first subresource */
+    if (!(full_discard = !region->NumRects))
+    {
+        vk_subresource = d3d12_resource_get_vk_subresource(texture, region->FirstSubresource, false);
+        full_rect = d3d12_get_image_rect(texture, vk_subresource.mipLevel);
```
HansKristian-Work, line 6396: I guess the reason you're not checking all subresources is the d3d12 requirement that all subresources touched by Discard must have the same dimensions?

libs/vkd3d/command.c
```diff
@@ -6287,7 +6355,77 @@ static void STDMETHODCALLTYPE d3d12_command_list_ClearUnorderedAccessViewFloat(d
...
+
+        for (i = 0; i < region->NumRects && !full_discard; i++)
```
HansKristian-Work, line 6398: This seems a bit fishy, I would expect to see full_discard be initialized to true here, and if there is any non-full discard, it nopes out.

HansKristian-Work: The way I read the code now, it's considered a full discard if just one rect is full.

libs/vkd3d/command.c
```diff
@@ -2111,6 +2111,66 @@ static void d3d12_command_list_clear_attachment_deferred(struct d3d12_command_li
...
+    const struct vkd3d_format *format;
+    VkImageMemoryBarrier barrier;
+    VkPipelineStageFlags stages;
+    VkAccessFlags access;
+    VkImageLayout layout;
+
+    if (resource->desc.Flags & D3D12_RESOURCE_FLAG_ALLOW_RENDER_TARGET)
+    {
+        stages = VK_PIPELINE_STAGE_COLOR_ATTACHMENT_OUTPUT_BIT;
+        access = VK_ACCESS_COLOR_ATTACHMENT_WRITE_BIT;
```
HansKristian-Work, line 2127: What about read bits?

libs/vkd3d/command.c
```diff
@@ -2111,6 +2111,66 @@ static void d3d12_command_list_clear_attachment_deferred(struct d3d12_command_li
...
+
+    /* Transitioning only one aspect of a depth-stencil image is not allowed */
+    format = vkd3d_format_from_d3d12_resource_desc(list->device, &resource->desc, 0);
+
+    if (format->vk_aspect_mask != subresource->aspectMask)
+        return;
+
+    barrier.sType = VK_STRUCTURE_TYPE_IMAGE_MEMORY_BARRIER;
+    barrier.pNext = NULL;
+    barrier.srcAccessMask = 0;
```
HansKristian-Work, line 2150: If we're in a suspended render pass, is it safe to use 0 here for srcAccessMask?

libs/vkd3d/command.c
doitsujin: Yes.

libs/vkd3d/command.c
doitsujin: > it's considered a full discard if just one rect is full.

Yes, that's the idea and also done this way for clears. Discarding sub-rects doesn't have any effect when the entire subresource is cleared.

libs/vkd3d/command.c
doitsujin: Do we need those? Reads are undefined after discard.

libs/vkd3d/command.c
doitsujin: We're discarding old content, so we have no reason to flush it.

HansKristian-Work: Technically, memory needs to be available to be able to safely do a layout transition, even from UNDEFINED. I guess the possible hazard would be that this chain of events could get mangled:

- Write something (A)
- Discard
- Write something (B)

The problem scenario is if there are some pending writes in A, which are only resolved after B writes resolve.

libs/vkd3d/command.c
HansKristian-Work: I suppose we can drop the read.

libs/vkd3d/command.c
HansKristian-Work: Oh, wait. All subrects apply to all subresources?

libs/vkd3d/command.c
HansKristian-Work: This seems a bit pedantic though. I'll be okay doing 0 here if it's commented why it's 0.

libs/vkd3d/command.c
doitsujin: Yes, that's why the API requires to have all subresources to have the same dimension if discard rects are used.

libs/vkd3d/command.c
doitsujin: Re-added; this was part of a temporary change and is no longer needed.

libs/vkd3d/command.c
doitsujin: Same.

libs/vkd3d/resource.c
doitsujin: Changed the comment.

libs/vkd3d/resource.c
doitsujin: Yeah, fixed.

libs/vkd3d/resource.c
HansKristian-Work: Ok, I think I get it now. Clever bithack indeed.

PR #211, opened 2020-07-16
vkd3d: Implement calibrated timestamps

doitsujin:
Requires QueryPerformanceCounter to work properly, so no support on native Linux builds at the moment.


libs/vkd3d/device.c
```diff
@@ -637,6 +638,50 @@ VkInstance vkd3d_instance_get_vk_instance(struct vkd3d_instance *instance)
...
+    VkPhysicalDevice physical_device = device->vk_physical_device;
+    uint32_t i, domain_count = 0;
+    VkTimeDomainEXT *domains;
+    uint32_t result = 0;
+    VkResult vr;
+
+    if ((vr = VK_CALL(vkGetPhysicalDeviceCalibrateableTimeDomainsEXT(physical_device, &domain_count, NULL))) < 0)
+    {
+        ERR("Failed to enumerate time domains, vr %d.\n", vr);
+        return false;
```
HansKristian-Work, line 653: false?

doitsujin: fixed, this used to return bool at some point.

PR #215, opened 2020-07-20
programs: Fix building vkd3d-compiler.

doitsujin:
pthread_once is required everywhere now, so we have to include the threads dependency.

### DIFF
--- a/programs/vkd3d-compiler/meson.build
+++ b/programs/vkd3d-compiler/meson.build
@@ -1,5 +1,5 @@
 executable('vkd3d-compiler', 'main.c', vkd3d_headers,
-  dependencies        : vkd3d_shader_dep,
+  dependencies        : [ vkd3d_shader_dep, threads_dep ],
   include_directories : vkd3d_private_includes,
   install             : true,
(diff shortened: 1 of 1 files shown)


programs/vkd3d-compiler/meson.build
```diff
@@ -1,5 +1,5 @@
 executable('vkd3d-compiler', 'main.c', vkd3d_headers,
-  dependencies        : vkd3d_shader_dep,
+  dependencies        : [ vkd3d_shader_dep, threads_dep],
```
Joshua-Ashton, line 2: Spacing.

```suggestion
  dependencies        : [ vkd3d_shader_dep, threads_dep ],
```

doitsujin: ffs

doitsujin: fixed.

Joshua-Ashton: I did it as a suggestion so you can click commit suggestion and I can click squash.

doitsujin: Ah, well. Fixed anyway.

PR #222, opened 2020-07-26
meson: generate generatable sources

wine-pardon:
PLEASE NOTE:
- `winebuild` included because of `widl` and `wine` requirement.
  Actually better to drop those dependencies and include all generated files somewhere in `./generated/`,
  along with their sources.
- Optional `glslangValidator -Os` flag used, presence in all distributions not guaranteed.
  No special purpose, just forgot to remove after testing. Can save a few KB per arch, or break something, or both.
- shader  headers generator it's  a saturday night "stack overflow".
  Can be better than this, but this better than now.

Tested on i686-w64-mingw32 standalone build and demos.

---
Mostly cosmetic changes for the build directory,
nothing special, maybe a few MB disk space saved,
or some hidden 32bit errors fixed (because of per-arch generators?).

Basically
* header files geherated only once
* .def generated per-arch
  (data for the .spec file was collected from different sources
   by "googling", maybe not 100% correct)
* shaders also added to generated vkd3d_shaders.h
  (original one can be safely removed)


wine-pardon:
Superseded by: #225

PR #230, opened 2020-07-29
Bindless UAV counter rework

doitsujin:
Title. Instead of relying on the packed descriptor set, we instead add another bindless set that contains storage texel buffer views . No raw storage buffers due to alignment requirements on some hardware, and I didn't feel like adding *even more* code paths to what is already a fallback.

Passes existing tests and otherwise seems to work in SotTR, Metro, RE2, Hitman 2, Death Stranding.


libs/vkd3d-shader/spirv.c
```diff
@@ -5766,9 +5769,12 @@ static void vkd3d_dxbc_compiler_emit_resource_declaration(struct vkd3d_dxbc_comp
 
         if (counter_binding && (counter_binding->flags & VKD3D_SHADER_BINDING_FLAG_BINDLESS))
         {
+            SpvStorageClass storage_class = (counter_binding->flags & VKD3D_SHADER_BINDING_FLAG_RAW_VA)
+                    ? SpvStorageClassUniform : SpvStorageClassUniformConstant;
```
HansKristian-Work, line 5773: Why not StorageClassStorageBuffer? Is it because we're targeting SPIR-V 1.0 here?

libs/vkd3d/resource.c
```diff
@@ -2821,19 +2821,43 @@ static bool vk_descriptor_type_is_buffer(VkDescriptorType type)
...
+        set_index = d3d12_descriptor_heap_uav_counter_set_index();
+
+        if (dst->heap->uav_counters.data)
+        {
+            dst->heap->uav_counters.data[descriptor_index] = dst->info.view
+                    ? dst->info.view->vk_counter_address : 0;
+        }
+        else if ((vk_descriptor_set = dst->heap->vk_descriptor_sets[set_index]))
+        {
+            VkBufferView null_view = VK_NULL_HANDLE;
```
HansKristian-Work, line 2845: Do we require robustness2 now?

doitsujin: We always have a non-NULL view if we don't have robustness2.

libs/vkd3d-shader/spirv.c
doitsujin: It's consistent with what we're doing elsewhere; moving to "proper" 1.3 (which we should probably do at some point) requires more work.

libs/vkd3d/resource.c
doitsujin: That said, I wanted to change this piece of code slightly.

HansKristian-Work:
test_bindless_uav_counter_sm51 does not work for me on RADV with fallback path.

libs/vkd3d-shader/spirv.c
HansKristian-Work: Ok, no worries in that case.

PR #236, opened 2020-07-30
build: Use widl from mingw-tools.

jp7677:
This removes the wine dependency for cross-builds but adds a dependency for mingw-tools instead.

Signed-off-by: Jens Peters <jp7677@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,11 +27,11 @@ in order to pull in all the submodules which are needed for building.
 ## Building VKD3D
 
 ### Requirements:
-- [wine](https://www.winehq.org/) (for `widl`)
+- [wine](https://www.winehq.org/) (for `widl`) [for native builds]
   - On Windows this may be substituted for [Strawberry Perl](http://strawberryperl.com/) as it ships `widl` and is easy to find and install -- although this dependency may be eliminated in the future.
 - [Meson](http://mesonbuild.com/) build system (at least version 0.51)
 - [glslang](https://github.com/KhronosGroup/glslang) compiler
-- [Mingw-w64](http://mingw-w64.org/) compiler and headers (at least version 7.0) [for cross-builds for d3d12.dll which are default]
+- [Mingw-w64](http://mingw-w64.org/) compiler, headers and tools (at least version 7.0) [for cross-builds for d3d12.dll which are default]
 
 ### Building:
 #### The simple way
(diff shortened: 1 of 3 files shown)


jp7677:
Please test this carefully. It works fine here on Linux for cross-builds with mingw7, but to be honest I haven't tested native Linux builds nor building on Windows.

jp7677:
I struggled a bit with the Readme. Please let me know if you prefer a different wording.

jp7677:
Thanks a lot for the fast feedback. Yes, will do that. Is building Proton at the `proton_5.0-next` branch with this change sufficient?

HansKristian-Work:
I think so, if it's building vkd3d as a DLL, that's the build you want to test. You'll see in the build, vkd3d target depends on widl build, which is then used through PATH, which is why the 'widl' path name itself is significant.

jp7677:
I'm currently lookingat https://github.com/ValveSoftware/Proton/blob/proton_5.0-next/build/makefile_base.mak#L1677
Since Proton already indeed uses cross-builds, the executable name will be overwritten to `x86_64-w64-mingw32-widl` when search for it and thus having just `widl` in the PATH will be no longer be relevant/wont be used.
The question then is, does the mingw build within Proton also builds/installs mingw-tools? Currently looking for clues...

jp7677:
Well, no, seems the mingw-tools are not build, see https://github.com/ValveSoftware/Proton/blob/proton_5.0-next/build-mingw-w64.sh#L193
So without even trying, I'm pretty certain that this will fail for Proton :(. So for doing this change, the Proton build system needs to include the mingw-tools first. On the pro-side, the custom widl handling could then be removed.

jp7677:
@HansKristian-Work @Joshua-Ashton I've opened a pull request for Proton (https://github.com/ValveSoftware/Proton/pull/4110) that world fix the build steps after breaking the proton build with PR. if you consider this PR, then the other has to be merged too.

wine-pardon:
Try 
```
idl_compiler  = find_program([
  'widl',
  'x86_64-w64-mingw32-widl',
  'midl-or-what',
  '/path/to/alternative-idl-gen',
])
```
No changes to Proton will required then.

jp7677:
@wine-pardon thanks for the suggestion. This might indeed be another way, though let’s see first how the Proton people respond to my PR over there. In my opinion (well, obviously :) ) the Proton PR makes the build system there also somewhat nicer.

jp7677:
@HansKristian-Work please see the response to the Proton PR https://github.com/ValveSoftware/Proton/pull/4110#issuecomment-667976179
In short, green light from there to go ahead with this PR (though I would still recommend to test this briefly on your setup, I don’t want to be _the_ one person that had broken all your local setup’s ;) ).

jp7677:
Thanks a lot @HansKristian-Work for merging and all your hard work here for improving VKD3D-Proton!

PR #237, opened 2020-07-30
build: Use the commit hash as fallback for the built-in version.

jp7677:
The `git describe` command currently fails because there are no tags yet. Using `--always` lets it fall back to the current commit hash.

Signed-off-by: Jens Peters <jp7677@gmail.com>

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -83,7 +83,7 @@ if cpu_family == 'x86'
 endif
 
 vkd3d_version = vcs_tag(
-  command : ['git', 'describe', '--dirty=+'],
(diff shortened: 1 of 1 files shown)


jp7677:
Currently the build-in version always looks like `const char vkd3d_build[] = "1.1";` because `git describe` fails because the repository has no tags yet. The build system then falls back to the project version.
With this PR the generated version would look like `const char vkd3d_build[] = "d3761c2+";`.

jp7677:
Obviously `git describe` does the job once there is a tag, so rejecting this PR and just waiting would also perfectly fine.

jp7677:
Thanks!

PR #240, opened 2020-07-30
vkd3d: Load VK_LAYER_KHRONOS_validation if we enable debug mode.

HansKristian-Work:


libs/vkd3d/device.c
```diff
@@ -528,6 +528,33 @@ static HRESULT vkd3d_instance_init(struct vkd3d_instance *instance,
     instance_info.ppEnabledExtensionNames = extensions;
     vkd3d_free(user_extension_supported);
 
+    if (instance->config_flags & VKD3D_CONFIG_FLAG_VULKAN_DEBUG)
+    {
+        static const char *debug_layer_name = "VK_LAYER_KHRONOS_validation";
```
Joshua-Ashton, line 533: I would just put this at the top of the function to avoid this needing to be static.

Referencing out of scope static things is really gross, as well as just having random statics.

HansKristian-Work: Fixed.

PR #241, opened 2020-07-30
tests: Fix uninitialized Layout member.

HansKristian-Work:
C is fine ...


HansKristian-Work:
`-Wmissing-field-initializers` perhaps?

HansKristian-Work:
GCC is incapable of warning against this:
```
#include <stdio.h>

struct foo
{
	int a, b, c;
};

__attribute__((noinline))
void printer(const struct foo *f)
{
	printf("Meep: %d, %d, %d.\n", f->a, f->b, f->c);
}

int main(void)
{
	struct foo boo;
	boo.a = 10;
	boo.b = 20;
	printer(&boo);
}
```
It can catch it if it's inlined, sigh ...

PR #242, opened 2020-07-30
meson: Target native d3d12 for Windows test binaries.

HansKristian-Work:

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -103,10 +103,6 @@ endif
 subdir('include')
 subdir('libs')
 
-if enable_tests
-  subdir('tests')
-endif
-
 if enable_extras
   if vkd3d_platform == 'linux'
     lib_m            = vkd3d_compiler.find_library('m')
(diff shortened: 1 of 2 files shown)


Joshua-Ashton:
Maybe only do this for ``enable_standalone_d3d12`` = True?

HansKristian-Work:
It actually works out nicely without standalone_d3d12 since it ends up linking against vkd3d-utils in that case.

HansKristian-Work:
actually, need to hoist out the conditional path in enable_extras ...

Joshua-Ashton:
> It actually works out nicely without standalone_d3d12 since it ends up linking against vkd3d-utils in that case.

I meant when building for Windows without standalone_d3d12.

I would add the DXGI always if the platform is Windows and substitute vkd3d and vkd3d_utils for lib_d3d12 depending on the standalone_d3d12 option.

HansKristian-Work:
Actually, I think I need to make use of the VKD3D_FORCE_UTILS_WRAPPER thingy to make this work since we cannot mix and match DXGI with utils wrapper, I'll close the PR for now.

PR #245, opened 2020-08-01
gitignore: Ignore additional build artifacts

FaithLV:
This PR adds `.dll` and `builddir/*` to .gitignore

Signed-off-by: Valters <valters.tomsons@protonmail.com>


Joshua-Ashton:
Why `builddir`? Why are you calling your build folder that when there are plenty of other options in the .gitignore already?

All the DLLs should also be built inside the meson build folder so I don't see why that's needed.

FaithLV:
@Joshua-Ashton I'm using VSCode which defaults meson builder folder to `builddir`

Joshua-Ashton:
You shouldn't use that intergration because of the terrible default dir and how painful it is to specify definitions and options.

The Meson intergration for VSCode is also horifically broken in general last time I used it.

DadSchoorse:
>Why builddir? Why are you calling your build folder that when there are plenty of other options in the .gitignore already?
>
>All the DLLs should also be built inside the meson build folder so I don't see why that's needed. 

`builddir` is what meson recommends nowadays.
https://github.com/mesonbuild/meson/commit/f3be687cbbfd1ff25c7ec4ffe8ba4adfee9a6579

PR #142, opened 2020-06-15
Make nameless unions really nameless

Joshua-Ashton:
This makes it consistent across tests and vkd3d.

PR #246, opened 2020-08-07
Descriptor rework, Part 1

doitsujin:
- Eliminates reference counting views as well as the per-descriptor spin lock, which speeds up descriptor copies by a factor of 5 on RADV according to the microbenchmark.
- Views are now stored in an open addresing hash map as part of the resource they are created from.
- Non-static samplers are now stored globally per device, this also avoids duplicates.

Tested with Shadow of the Tomb Raider, Death Stranding, Control, Monster Hunter World.

Part 2 is going to refactor the actual Vulkan descriptor updates so that we can trim down the `d3d12_desc` struct, which may reduce overhead a bit further.

### DIFF
--- a/include/private/hashmap.h
+++ b/include/private/hashmap.h
@@ -1,0 +1,215 @@
+/*
+ * Hash map support
+ *
+ * Copyright 2020 Philip Rebohle for Valve Corporation
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
+ */
+
+#ifndef __VKD3D_HASHMAP_H
+#define __VKD3D_HASHMAP_H
+
+#include <stddef.h>
+
+#include "memory.h"
+
+enum hash_map_entry_flag
+{
+    HASH_MAP_ENTRY_OCCUPIED = (1 << 0),
+};
+
+struct hash_map_entry
+{
+    uint32_t hash_value;
+    uint32_t flags;
+};
+
+typedef uint32_t (*pfn_hash_func)(const void* key);
+typedef bool (*pfn_hash_compare_func)(const void *key, const struct hash_map_entry *entry);
+
+/* Open-addressing hash table */
+struct hash_map
+{
+    pfn_hash_func hash_func;
+    pfn_hash_compare_func compare_func;
+    void *entries;
+    size_t entry_size;
+    uint32_t entry_count;
+    uint32_t used_count;
+};
+
+static inline void *hash_map_ptr_offset(void *ptr, size_t offset)
+{
+    return ((char*)ptr) + offset;
+}
+
+static inline struct hash_map_entry *hash_map_get_entry(struct hash_map *hash_map, uint32_t entry_idx)
+{
+    return hash_map_ptr_offset(hash_map->entries, hash_map->entry_size * entry_idx);
+}
+
+static inline uint32_t hash_map_get_entry_idx(struct hash_map *hash_map, uint32_t hash_value)
+{
+    return hash_value % hash_map->entry_count;
+}
+
+static inline uint32_t hash_map_next_entry_idx(struct hash_map *hash_map, uint32_t entry_idx)
+{
+    uint32_t next_idx = entry_idx + 1;
+    return next_idx < hash_map->entry_count ? next_idx : 0;
+}
+
+static inline uint32_t hash_map_next_size(uint32_t old_size)
+{
+    /* This yields a sequence of primes and numbers with two
+     * relatively large prime factors for any reasonable hash
+     * table size */
+    return old_size ? (old_size * 2 + 5) : 37;
+}
+
+static inline bool hash_map_grow(struct hash_map *hash_map)
+{
+    uint32_t i, old_count, new_count;
+    void *new_entries, *old_entries;
+
+    old_count = hash_map->entry_count;
+    old_entries = hash_map->entries;
+
+    new_count = hash_map_next_size(hash_map->entry_count);
+
+    if (!(new_entries = vkd3d_calloc(new_count, hash_map->entry_size)))
+        return false;
+
+    hash_map->entry_count = new_count;
+    hash_map->entries = new_entries;
+
+    for (i = 0; i < old_count; i++)
+    {
+        /* Relocate existing entries one by one */
+        struct hash_map_entry *old_entry = hash_map_ptr_offset(old_entries, i * hash_map->entry_size);
+
+        if (old_entry->flags & HASH_MAP_ENTRY_OCCUPIED)
+        {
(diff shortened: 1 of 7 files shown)


include/private/hashmap.h
```diff
@@ -0,0 +1,197 @@
...
+    pfn_hash_compare_func compare_func;
+    void *entries;
+    size_t entry_size;
+    uint32_t entry_count;
+    uint32_t used_count;
+};
+
+static inline struct hash_map_entry *hash_map_get_entry(struct hash_map *hash_map, uint32_t entry_idx)
+{
+    return hash_map->entries + hash_map->entry_size * entry_idx;
```
Saancreed, line 55: This breaks MSVC build because it disallows pointer arithmetic on `void *`. The exact error is:
```
       "Z:\home\saancreed\Repositories\vkd3d\build\vkd3d.sln" (default target) 
       (1) ->
       "Z:\home\saancreed\Repositories\vkd3d\build\libs\vkd3d\eba9057@@vkd3d@st
       a.vcxproj" (default target) (7) ->
       (ClCompile target) -> 
         Z:\home\saancreed\Repositories\vkd3d\include\private\hashmap.h(55,64):
        error C2036: 'void *': unknown size [Z:\home\saancreed\Repositories\vkd
       3d\build\libs\vkd3d\eba9057@@vkd3d@sta.vcxproj]
```

include/private/hashmap.h
```diff
@@ -0,0 +1,197 @@
...
+    if (!(new_entries = vkd3d_calloc(new_count, hash_map->entry_size)))
+        return false;
+
+    hash_map->entry_count = new_count;
+    hash_map->entries = new_entries;
+
+    for (i = 0; i < old_count; i++)
+    {
+        /* Relocate existing entries one by one */
+        struct hash_map_entry *old_entry = old_entries + i * hash_map->entry_size;
```
Saancreed, line 90: Also here:
```
         Z:\home\saancreed\Repositories\vkd3d\include\private\hashmap.h(90,82):
        error C2036: 'void *': unknown size [Z:\home\saancreed\Repositories\vkd
       3d\build\libs\vkd3d\eba9057@@vkd3d@sta.vcxproj]
```

libs/vkd3d/resource.c
```diff
@@ -1211,6 +1212,227 @@ HRESULT vkd3d_get_image_allocation_info(struct d3d12_device *device,
...
 
+struct vkd3d_view_key
+{
+    enum vkd3d_view_type view_type;
+    union
+    {
+        struct vkd3d_buffer_view_desc buffer;
+        struct vkd3d_texture_view_desc texture;
+        D3D12_SAMPLER_DESC sampler;
+    } u;
```
Saancreed, line 1223: Aren't nameless unions supposed to be actually nameless now? That was my impression after PR #142.

Joshua-Ashton: Doesn't matter for stuff internally, that was just for the D3D12 headers.

include/private/hashmap.h
HansKristian-Work: Yes, pointer arithmetic on `void *` is a GNU extension.

include/private/hashmap.h
```diff
@@ -0,0 +1,197 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
+ */
+
+#ifndef __VKD3D_HASHMAP_H
```
HansKristian-Work, line 21: Is this implementation based on existing, well tested code? Not sure how detailed I should be in reviewing this.

include/private/hashmap.h
```diff
@@ -0,0 +1,197 @@
...
+static inline void hash_map_clear(struct hash_map *hash_map)
+{
+    vkd3d_free(hash_map->entries);
+    hash_map->entries = NULL;
+    hash_map->entry_count = 0;
+    hash_map->used_count = 0;
+}
+
+static inline uint32_t hash_combine(uint32_t old_hash, uint32_t new_hash) {
+  return old_hash ^ (new_hash + 0x9e3779b9 + (old_hash << 6) + (old_hash >> 2));
```
HansKristian-Work, line 194: Formatting of the function is off.

include/private/hashmap.h
```diff
@@ -0,0 +1,197 @@
...
+    hash_map->entries = new_entries;
+
+    for (i = 0; i < old_count; i++)
+    {
+        /* Relocate existing entries one by one */
+        struct hash_map_entry *old_entry = old_entries + i * hash_map->entry_size;
+
+        if (old_entry->flags & HASH_MAP_ENTRY_OCCUPIED)
+        {
+            memcpy(hash_map_get_entry(hash_map, hash_map_get_entry_idx(hash_map, old_entry->hash_value)),
```
HansKristian-Work, line 94: Does this work? Can't there be collisions?

libs/vkd3d/resource.c
```diff
@@ -1211,6 +1212,132 @@ HRESULT vkd3d_get_image_allocation_info(struct d3d12_device *device,
...
+{
+    const struct vkd3d_view_key *k = key;
+    uint64_t resource_handle;
+    uint32_t hash;
+
+    switch (k->view_type)
+    {
+        case VKD3D_VIEW_TYPE_BUFFER:
+            resource_handle = (uint64_t)k->u.buffer.buffer;
+            hash = hash_combine(resource_handle, resource_handle >> 32);
```
HansKristian-Work, line 1243: Should probably cast the resource_handle to uint32_t for consistency with the other casting below?

libs/vkd3d/resource.c
```diff
@@ -1263,6 +1264,22 @@ uint32_t vkd3d_view_entry_hash(const void *key)
...
 
+        case VKD3D_VIEW_TYPE_SAMPLER:
+            hash = (uint32_t)k->u.sampler.Filter;
+            hash = hash_combine(hash, (uint32_t)k->u.sampler.AddressU);
+            hash = hash_combine(hash, (uint32_t)k->u.sampler.AddressV);
+            hash = hash_combine(hash, (uint32_t)k->u.sampler.AddressW);
+            hash = hash_combine(hash, float_bits_to_uint32(k->u.sampler.MipLODBias));
+            hash = hash_combine(hash, (uint32_t)k->u.sampler.MaxAnisotropy);
+            hash = hash_combine(hash, (uint32_t)k->u.sampler.ComparisonFunc);
+            hash = hash_combine(hash, float_bits_to_uint32(k->u.sampler.BorderColor[0]));
```
HansKristian-Work, line 1275: Might be good to ignore BorderColor if app leaves it uninitialized and no address uses border color.

include/private/hashmap.h
doitsujin: It's a custom implementation, and doesn't even have a `remove` function or anything like that because we're not using it.

include/private/hashmap.h
doitsujin: Good point, will fix.

libs/vkd3d/resource.c
doitsujin: Fixed.

include/private/hashmap.h
doitsujin: Fixed.

include/private/hashmap.h
doitsujin: Fixed.

libs/vkd3d/resource.c
doitsujin: Ended up fixing this differently by adding a simple hash function for 64-bit numbers.

include/private/hashmap.h
doitsujin: Fixed.

HansKristian-Work (approved):

PR #249, opened 2020-08-19
Implement external host memory import.

HansKristian-Work:


libs/vkd3d/device.c
```diff
@@ -3888,19 +3897,78 @@ static HRESULT STDMETHODCALLTYPE d3d12_device_CreatePipelineState(d3d12_device_i
...
+    TRACE("iface %p, file_mapping %p, riid %s, heap %p\n",
+          iface, file_mapping, debugstr_guid(riid), heap);
+
+    /* 0 size maps everything. */
+    addr = MapViewOfFile(file_mapping, FILE_MAP_ALL_ACCESS, 0, 0, 0);
+    if (!addr)
+        return E_INVALIDARG;
 
+    hr = d3d12_device_OpenExistingHeapFromAddress(iface, addr, riid, heap);
+    UnmapViewOfFile(addr);
```
doitsujin, line 3966: Is it safe to unmap here? Wouldn't this effectively free the virtual address range?

HansKristian-Work: It should be, the physical allocation should be ref-counted in the OS.

HansKristian-Work: Changed it to unmap at the end to be 100% sure about behavior.

doitsujin (approved):

PR #255, opened 2020-08-26
Static sampler rework

doitsujin:
Deduplicates static samplers and uses a global descriptor pool for the static sampler sets used in root signatures, so we don't have to allocate them at bind time.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -5173,13 +5173,8 @@ static void d3d12_command_list_set_root_signature(struct d3d12_command_list *lis
     bindings->root_signature = root_signature;
     bindings->static_sampler_set = VK_NULL_HANDLE;
 
-    if (root_signature && root_signature->vk_sampler_descriptor_layout)
-    {
-        /* FIXME allocate static sampler sets globally */
-        bindings->static_sampler_set = d3d12_command_allocator_allocate_descriptor_set(
-                list->allocator, root_signature->vk_sampler_descriptor_layout,
-                VKD3D_DESCRIPTOR_POOL_TYPE_IMMUTABLE_SAMPLER);
-    }
+    if (root_signature && root_signature->vk_sampler_set)
+        bindings->static_sampler_set = root_signature->vk_sampler_set;
 
     d3d12_command_list_invalidate_root_parameters(list, bind_point, true);
 }
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -1249,10 +1249,7 @@ static VkDescriptorPool d3d12_command_allocator_allocate_descriptor_pool(
 {
     static const VkDescriptorPoolSize pool_sizes[] =
     {
-        /* Must be first in the array. */
-        /* Need at least 2048 so we can allocate an immutable sampler set. */
         {VK_DESCRIPTOR_TYPE_SAMPLER, 2048},
-
         {VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER, 1024},
         {VK_DESCRIPTOR_TYPE_UNIFORM_TEXEL_BUFFER, 1024},
         {VK_DESCRIPTOR_TYPE_SAMPLED_IMAGE, 1024},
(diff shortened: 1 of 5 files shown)


libs/vkd3d/resource.c
```diff
@@ -1727,6 +1734,98 @@ HRESULT vkd3d_sampler_state_create_static_sampler(struct vkd3d_sampler_state *st
...
+    const struct vkd3d_vk_device_procs *vk_procs = &device->vk_procs;
+    VkDescriptorPoolCreateInfo pool_info;
+    VkDescriptorPoolSize pool_size;
+
+    pool_size.type = VK_DESCRIPTOR_TYPE_SAMPLER;
+    pool_size.descriptorCount = 16384;
+
+    pool_info.sType = VK_STRUCTURE_TYPE_DESCRIPTOR_POOL_CREATE_INFO;
+    pool_info.pNext = NULL;
+    pool_info.flags = VK_DESCRIPTOR_POOL_CREATE_FREE_DESCRIPTOR_SET_BIT;
```
HansKristian-Work, line 1748: This seems a bit risque if implementations do not have a good FREE_DESCRIPTOR_SET implementation. Would it be terrible if each root signature owns its own static pool/set?

libs/vkd3d/resource.c
```diff
@@ -1727,6 +1734,98 @@ HRESULT vkd3d_sampler_state_create_static_sampler(struct vkd3d_sampler_state *st
...
+        ERR("Failed to lock mutex, rc %d.\n", rc);
+        return hresult_from_errno(rc);
+    }
+
+    alloc_info.sType = VK_STRUCTURE_TYPE_DESCRIPTOR_SET_ALLOCATE_INFO;
+    alloc_info.pNext = NULL;
+    alloc_info.descriptorSetCount = 1;
+    alloc_info.pSetLayouts = &vk_layout;
+
+    for (i = 1; i <= state->vk_descriptor_pool_count; i++)
```
HansKristian-Work, line 1777: This loop is kinda confusing. I don't understand why it starts at 1 and uses <=.

libs/vkd3d/command.c
```diff
@@ -5173,13 +5173,8 @@ static void d3d12_command_list_set_root_signature(struct d3d12_command_list *lis
     bindings->root_signature = root_signature;
     bindings->static_sampler_set = VK_NULL_HANDLE;
 
-    if (root_signature && root_signature->vk_sampler_descriptor_layout)
-    {
-        /* FIXME allocate static sampler sets globally */
-        bindings->static_sampler_set = d3d12_command_allocator_allocate_descriptor_set(
-                list->allocator, root_signature->vk_sampler_descriptor_layout,
-                VKD3D_DESCRIPTOR_POOL_TYPE_IMMUTABLE_SAMPLER);
```
HansKristian-Work, line 5181: Should probably remove the enum for POOL_TYPE_IMMUTABLE_SAMPLER as well?

HansKristian-Work (changes requested):
Comments. Looks good otherwise.

libs/vkd3d/resource.c
doitsujin: We can have a large number of root signatures and each descriptor pool is a memory object to track (for like 1-2 sampler descriptors most of the time); with this we'll still pack a large number of sets into a single pool even if the driver in question never reuses memory.

libs/vkd3d/resource.c
doitsujin: Basically the assumption is that the most recently created pool is the easiest one to allocate a set from, so we iterate in reverse order. I can add a comment.

libs/vkd3d/resource.c
HansKristian-Work: I suppose. My only concern is if application allocates and frees root signatures all the time, and FREE_DESCRIPTOR_SET_BIT is not able to recycle memory internally, but I guess this is so unlikely to be a problem that it's irrelevant for our use case?

libs/vkd3d/resource.c
HansKristian-Work: I figured it was accessing the array in reverse after staring at it for a bit longer, would be easier to understand if i is decrementing i, or a comment.

doitsujin: Actually, since this isn't perf-critical anyway, I'll just change it to iterate normally.

libs/vkd3d/resource.c
doitsujin: I'd argue that if we ever find such a driver, we should push for that getting fixed instead.

HansKristian-Work: Fair enough.

HansKristian-Work (approved):

libs/vkd3d/command.c
doitsujin: Fixed.

PR #256, opened 2020-08-26
vkd3d: Rewrite timeline semaphore handling.

HansKristian-Work:
Need to handle large (> 4G) jumps in timeline value, which is not
supported by all implementations.

There is no good way to handle that, so rewrite and clean up timeline
semaphore handling by separating the timeline into a virtual timeline
(which can rewind and jump around arbitrarely) and a physical timeline
which increments by one each time.


HansKristian-Work:
Tested on SOTTR, RE2, Metro, DS (Windows). Seems to work fine. No regressions in test suite.

doitsujin (approved):
Tested in some more games (RE3, MHW), works fine, looks good.

PR #260, opened 2020-08-28
Implement WriteBufferImmediate

doitsujin:
Title. Passes the already existing test for WriteBufferImmediate and silences log spam in Horizon Zero Dawn. Perf impact in that game seems acceptable even with the fallback path.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -6670,6 +6670,7 @@ static void STDMETHODCALLTYPE d3d12_command_list_WriteBufferImmediate(d3d12_comm
         const D3D12_WRITEBUFFERIMMEDIATE_MODE *modes)
 {
     struct d3d12_command_list *list = impl_from_ID3D12GraphicsCommandList(iface);
+    VkPipelineStageFlags wait_stage_mask = VK_PIPELINE_STAGE_TOP_OF_PIPE_BIT;
     const struct vkd3d_vk_device_procs *vk_procs = &list->device->vk_procs;
     VkPipelineStageFlagBits stage = VK_PIPELINE_STAGE_TOP_OF_PIPE_BIT;
     struct d3d12_resource *resource;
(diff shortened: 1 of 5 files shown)


libs/vkd3d/command.c
```diff
@@ -6645,20 +6645,67 @@ static void STDMETHODCALLTYPE d3d12_command_list_SetViewInstanceMask(d3d12_comma
...
         resource = vkd3d_gpu_va_allocator_dereference(&list->device->gpu_va_allocator, parameters[i].Dest);
-        d3d12_command_list_track_resource_usage(list, resource);
+
+        if (!resource || !d3d12_resource_is_buffer(resource))
+        {
+            d3d12_command_list_mark_as_invalid(list, "Invalid target address %p.\n", parameters[i].Dest);
+            return;
+        }
+
+        offset = parameters[i].Dest - resource->gpu_address + resource->heap_offset;
```
HansKristian-Work, line 6692: We should not add heap_offset here I think. The resource we obtain from dereferencing VA is the heap itself, not the resource.

libs/vkd3d/command.c
```diff
@@ -6645,20 +6645,67 @@ static void STDMETHODCALLTYPE d3d12_command_list_SetViewInstanceMask(d3d12_comma
...
+        stage = VK_PIPELINE_STAGE_TOP_OF_PIPE_BIT;
```
HansKristian-Work, line 6693: Seems a bit odd to redundantly set the variable here.

HansKristian-Work (changes requested):
Comments.

libs/vkd3d/command.c
doitsujin: `modes` is an optional parameter. I could move the assignment out of the loop though.

libs/vkd3d/command.c
```diff
@@ -6704,7 +6705,17 @@ static void STDMETHODCALLTYPE d3d12_command_list_WriteBufferImmediate(d3d12_comm
         }
         else
         {
-            FIXME_ONCE("VK_AMD_buffer_marker not supported by device.\n");
+            d3d12_command_list_end_current_render_pass(list, true);
```
HansKristian-Work, line 6708: Is it a problem if this is spammed multiple times if count > 1?

doitsujin: No, we'll just do nothing in that case.

doitsujin: fwiw this already happens when calling any other method that has to end the current render pass.

libs/vkd3d/command.c
doitsujin: I checked the code and it does look like we only ever allocate VAs for committed buffer resources, so yeah, heap_offset will always be 0.

libs/vkd3d/command.c
HansKristian-Work: Figured, just wanted to make sure.

HansKristian-Work (approved):
Looks good!

PR #261, opened 2020-08-31
vkd3d: Zero out descriptor heaps on allocation.

HansKristian-Work:
Only bother if nullDescriptors are supported.


doitsujin:
Looks good, although to keep things a bit more readable I'd prefer the actual init code to be moved to a separate function, so that we can do something like
```
    if (device->device_info.robustness2_features.nullDescriptor &&
            binding->vk_descriptor_type != VK_DESCRIPTOR_TYPE_SAMPLER)
        frog();
```

HansKristian-Work:
Hoisted out to separate function.

doitsujin (approved):

PR #262, opened 2020-09-02
vkd3d: Emit host memory barrier as needed.

HansKristian-Work:
There is no resource state associated with this, so emit the barrier at
the end of a command buffer based on trivial tracking.


doitsujin:
Do we need to handle `ResolveQueryData` as well?

HansKristian-Work:
Fixed.

doitsujin (approved):

PR #265, opened 2020-09-08
Implement pipeline libraries

doitsujin:
Helps avoid the shader compile screen in Horizon Zero Dawn.

This implementation is rather basic, and only stores the contents of a VkPipelineCache that is created for each pipeline. State validation is not performed, however we do handle the case when vkd3d itself or the underlying graphics driver were updated, in order to notify the game that the cache is out of date.

Unfortunately this means that making commits to vkd3d will invalidate cache blobs, but I don't really see a good way around that since we need to do this in case of shader compiler changes.


libs/vkd3d/meson.build
```diff
@@ -20,6 +20,7 @@ vkd3d_shaders =[
 ]
 
 vkd3d_src = [
+  'cache.c',
```
HansKristian-Work, line 23: Makefile.am as well (I look forward to when we nix autotools).

libs/vkd3d/cache.c
```diff
@@ -20,6 +20,104 @@
...
+    VkResult vr;
+
+    if (!state->CachedBlobSizeInBytes)
+    {
+        vr = vkd3d_create_pipeline_cache(device, 0, NULL, cache);
+        return hresult_from_vk_result(vr);
+    }
+
+    /* Avoid E_INVALIDARG with an invalid header size, since that may confuse some games */
+    if (state->CachedBlobSizeInBytes < sizeof(*blob) || blob->version != VKD3D_CACHE_BLOB_VERSION)
```
HansKristian-Work, line 65: Seems like it would be a good idea to use a more specific "magic" number here to avoid any potential mismatch with a D3D12 blob coming from a different implementation.

libs/vkd3d/cache.c
```diff
@@ -20,6 +20,104 @@
...
+#define VKD3D_CACHE_BUILD_SIZE 8
+
+struct vkd3d_pipeline_blob
+{
+    uint32_t version;
+    uint32_t vendor_id;
+    uint32_t device_id;
+    char vkd3d_build[VKD3D_CACHE_BUILD_SIZE];
+    uint8_t cache_uuid[VK_UUID_SIZE];
+    char vk_blob[];
```
HansKristian-Work, line 48: uint8_t[] seems a bit more direct.

include/private/vkd3d_common.h
```diff
@@ -250,4 +250,22 @@ static inline uint32_t float_bits_to_uint32(float f)
...
 
+static inline size_t vkd3d_wcslen(const WCHAR *wstr, size_t wchar_size)
+{
+    const char *data = (const char*)wstr;
+    uint32_t curr_char;
+    size_t length = 0;
+
+    while (true)
+    {
+        memset(&curr_char, 0, sizeof(curr_char));
```
HansKristian-Work, line 261: Why memset a uint32_t?

HansKristian-Work: Seems easier to have a uint16_t variant and a uint32_t variant than dealing with memset/memcmp.

libs/vkd3d/cache.c
```diff
@@ -118,12 +118,125 @@ VkResult vkd3d_serialize_pipeline_state(const struct d3d12_pipeline_state *state
...
+
+    return k->name_length == e->key.name_length &&
+            !memcmp(k->name, e->key.name, k->name_length);
+}
+
+struct vkd3d_serialized_pipeline
+{
+    uint32_t name_length;
+    uint32_t blob_length;
+    char data[];
```
HansKristian-Work, line 171: uint8_t?

libs/vkd3d/cache.c
```diff
@@ -118,12 +118,125 @@ VkResult vkd3d_serialize_pipeline_state(const struct d3d12_pipeline_state *state
...
+};
+
+#define VKD3D_PIPELINE_LIBRARY_VERSION 1
+
+struct vkd3d_serialized_pipeline_library
+{
+    uint32_t version;
```
HansKristian-Work, line 178: A more specialized magic number would be nice.

libs/vkd3d/cache.c
```diff
@@ -118,12 +118,125 @@ VkResult vkd3d_serialize_pipeline_state(const struct d3d12_pipeline_state *state
...
+    uint32_t vendor_id;
+    uint32_t device_id;
+    uint32_t pipeline_count;
+    char vkd3d_build[VKD3D_CACHE_BUILD_SIZE];
+    uint8_t cache_uuid[VK_UUID_SIZE];
+    char data[];
```
HansKristian-Work, line 184: uint8_t?

libs/vkd3d/cache.c
```diff
@@ -307,22 +624,92 @@ static CONST_VTBL struct ID3D12PipelineLibrary1Vtbl d3d12_pipeline_library_vtbl
...
+    const VkPhysicalDeviceProperties *device_properties = &device->device_info.properties2.properties;
+    const struct vkd3d_serialized_pipeline_library *header = blob;
+    const char *end = header->data + blob_length - sizeof(*header);
+    const char *cur = header->data;
+    uint32_t i;
+
+    /* Same logic as for pipeline blobs, indicate that the app needs
+     * to rebuild the pipeline library in case vkd3d itself or the
+     * underlying device/driver changed */
+    if (blob_length < sizeof(*header) || header->version != VKD3D_PIPELINE_LIBRARY_VERSION)
```
HansKristian-Work, line 639: Possible to move this logic to a common file, since it seems to use the same header?

doitsujin: It's not quite the same header, the pipeline library one also stores the pipeline count.

HansKristian-Work (changes requested):
Mostly minor nits. Looks pretty solid overall!

libs/vkd3d/cache.c
doitsujin: Now using `MAKE_MAGIC(...)` to generate something more specific that's still versioned.

libs/vkd3d/cache.c
doitsujin: Fixed.

include/private/vkd3d_common.h
doitsujin: Fixed.

libs/vkd3d/cache.c
doitsujin: See above, fixed.

libs/vkd3d/cache.c
doitsujin: Fixed.

libs/vkd3d/meson.build
doitsujin: Fixed.

libs/vkd3d/cache.c
doitsujin: Fixed.

doitsujin:
Actually we have a useless log message somewhere, let me remove that.

HansKristian-Work (approved):
LGTM!

PR #272, opened 2020-09-11
vkd3d: Remove bindless CBV workaround.

HansKristian-Work:
Not relevant anymore, and should see good uplift on NV depending on
content.


doitsujin (approved):
Tested this a fair bit on AMD already and it seems to be working perfectly fine, so we shouldn't have any awkward bugs in our UBO code path.

PR #274, opened 2020-09-29
Descriptor rework, Part 2

doitsujin:
Moves descriptor update code to the respective `d3d12_desc_create_*` functions and removes a whole bunch of old code we no longer need, which is also going to make it easier to support multiple descriptor types for any given view type.

Tested briefly with RE3, Control, Death Stranding, and Horizon Zero Dawn.


HansKristian-Work (approved):
Looks clean. Nice!

PR #279, opened 2020-10-01
Migrate to `-fvisibility=hidden`

Joshua-Ashton:


libs/vkd3d-utils/vkd3d_utils_main.c
```diff
@@ -21,14 +21,14 @@
 #include "vkd3d_common.h"
 #include "vkd3d_utils_private.h"
 
-HRESULT WINAPI D3D12GetDebugInterface(REFIID iid, void **debug)
+VKD3D_UTILS_EXPORT HRESULT WINAPI D3D12GetDebugInterface(REFIID iid, void **debug)
```
HansKristian-Work, line 24: Does this actually make these symbols visible? The Meson build just defines VKD3D_UTILS_EXPORT to nothing.

include/vkd3d.h
```diff
@@ -33,6 +33,24 @@
...
 #endif  /* VKD3D_NO_VULKAN_H */
 
+#if defined(__GNUC__)
+# define DECLSPEC_VISIBLE __attribute__((visibility("default")))
+#else
+# define DECLSPEC_VISIBLE
+#endif
+
+#if defined(_WIN32) && !defined(VKD3D_BUILD_STANDALONE_D3D12)
+# ifdef VKD3D_EXPORTS
```
HansKristian-Work, line 43: What is VKD3D_EXPORTS?

libs/vkd3d/meson.build
```diff
@@ -45,8 +45,8 @@ if not enable_standalone_d3d12
     dependencies        : [ vkd3d_common_dep, vkd3d_shader_dep ] + vkd3d_extra_libs,
     include_directories : vkd3d_private_includes,
     install             : true,
-    vs_module_defs      : 'vkd3d.def',
     version             : '1.1.0',
+    c_args              : '-DVKD3D_EXPORTS',
```
HansKristian-Work, line 49: Ok, it's defined here ... Any reason why we would be building without VKD3D_EXPORTS, does it need to be a global define at all?

HansKristian-Work (commented):
Comments.

libs/vkd3d-utils/vkd3d_utils_main.c
Joshua-Ashton: Difference between ``VKD3D_UTILS_EXPORTS`` and ``VKD3D_UTILS_EXPORT``.

This is a common Windows DLL export pattern.

libs/vkd3d/meson.build
Joshua-Ashton: We need to define it when building as a DLL as opposed to importing it because of the differences between ``dllexport`` and ``dllimport`` declspec on Windows.

When building as an SO, this doesn't do anything -- it will just always mark the symbols as visible.

HansKristian-Work (approved):

HansKristian-Work:
Built locally on Lin/Win and only relevant symbols are exported as expected.

PR #282, opened 2020-10-02
Keep track of resource's `vkd3d_format`

Joshua-Ashton:
Eliminates a lot of the format lookups that kept happening in code that shouldn't probably be looking up formats.

NULL checks that have been removed were impossible anyway as there was already a `is_buffer` or equivalent check and it's impossible to have texture resources with NULL formats due to the desc validation.


Joshua-Ashton:
Tested with Gears, RE2 and HZD. Works fine.

Let me know anything else I should be testing with. Opening it up for review.

HansKristian-Work (approved):
Looks like a good cleanup to me.

PR #283, opened 2020-10-04
build: Add install script to packaged releases

Joshua-Ashton:

### DIFF
--- a/setup_vkd3d.sh
+++ b/setup_vkd3d.sh
@@ -1,0 +1,183 @@
+#!/bin/bash
+
+# default directories
+vkd3d_lib32=${vkd3d_lib32:-"x86"}
+vkd3d_lib64=${vkd3d_lib64:-"x64"}
+
+# figure out where we are
+basedir=$(dirname "$(readlink -f "$0")")
+
+# figure out which action to perform
+action="$1"
+
+case "$action" in
+install)
+  ;;
+uninstall)
+  ;;
+*)
+  echo "Unrecognized action: $action"
+  echo "Usage: $0 [install|uninstall] [--symlink]"
+  exit 1
+esac
+
+# process arguments
+shift
+
+file_cmd="cp -v"
+
+while (($# > 0)); do
+  case "$1" in
+  "--symlink")
+    file_cmd="ln -s -v"
+    ;;
+  esac
+  shift
+done
+
+# check wine prefix before invoking wine, so that we
+# don't accidentally create one if the user screws up
+if [ -n "$WINEPREFIX" ] && ! [ -f "$WINEPREFIX/system.reg" ]; then
+  echo "$WINEPREFIX:"' Not a valid wine prefix.' >&2
+  exit 1
+fi
+
+# find wine executable
+export WINEDEBUG=-all
+
+wine="wine"
+wine64="wine64"
+wineboot="wineboot"
+
+# $PATH is the way for user to control where wine is located (including custom Wine versions).
+# Pure 64-bit Wine (non Wow64) requries skipping 32-bit steps.
+# In such case, wine64 and winebooot will be present, but wine binary will be missing,
+# however it can be present in other PATHs, so it shouldn't be used, to avoid versions mixing.
+wine_path=$(dirname "$(which $wineboot)")
+wow64=true
+if ! [ -f "$wine_path/$wine" ]; then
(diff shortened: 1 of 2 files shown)


setup_vkd3d.sh
```diff
@@ -0,0 +1,191 @@
...
+}
+
+# copy or link vkd3d dll, back up original file
+installFile() {
+  dstfile="${1}/${3}.dll"
+  srcfile="${basedir}/${2}/${3}.dll"
+
+  if [ -f "${srcfile}.so" ]; then
+    srcfile="${srcfile}.so"
+  fi
```
Saancreed, line 110: I remember this to be a fallback path for DXVK `.dll.so` builds, but afaik vkd3d doesn't produce any such files, making this unnecessary and safe to remove (especially since this script is installed only when building `d3d12.dll`).
```suggestion
```

setup_vkd3d.sh
```diff
@@ -0,0 +1,191 @@
...
+}
+
+# remove vkd3d dll, restore original file
+uninstallFile() {
+  dstfile="${1}/${3}.dll"
+  srcfile="${basedir}/${2}/${3}.dll"
+
+  if [ -f "${srcfile}.so" ]; then
+    srcfile="${srcfile}.so"
+  fi
```
Saancreed, line 140: Also here.
```suggestion
```

setup_vkd3d.sh
```diff
@@ -0,0 +1,191 @@
+#!/bin/bash
+
+# default directories
+vkd3d_lib32=${vkd3d_lib32:-"x86"}
+vkd3d_lib64=${vkd3d_lib64:-"x64"}
+
+# figure out where we are
+basedir=$(dirname "$(readlink -f $0)")
```
Saancreed, line 8: Quote `$0` to prevent word splitting.
```suggestion
basedir=$(dirname "$(readlink -f "$0")")
```

Joshua-Ashton: This doesn't make any difference on my end on folders with spaces.

Saancreed: It certainly does on mine…

```
$ install -D /dev/stdin '/tmp/name with spaces/test.sh' <<'EOF'
#!/bin/bash
basedir=$(dirname "$(readlink -f $0)")
echo "First:  $basedir"
basedir=$(dirname "$(readlink -f "$0")")
echo "Second: $basedir"
EOF
$ '/tmp/name with spaces/test.sh'
First:  /tmp/name
/home/saancreed
Second: /tmp/name with spaces

Comment (Joshua-Ashton) [reply setup_vkd3d.sh:8]:
```
[joshua@lilypad frog with spaces]$ basedir=$(dirname "$(readlink -f $0)")
[joshua@lilypad frog with spaces]$ echo $basedir
/home/joshua/frog with spaces
[joshua@lilypad frog with spaces]$ 
```
And the script works fine if I run it in the "frog with spaces" folder.

Comment (Joshua-Ashton) [reply setup_vkd3d.sh:8]: Not sure sure how your thing works anyway given the double quotes. (although I am not super up on my bash)

Wouldn't you want `\"`?

Comment (Saancreed) [reply setup_vkd3d.sh:8]: > Wouldn't you want `\"`?

No, just like you don't want `\"` around `$(readlink …)`. These quotes are used for escaping spaces in `$0`, and using `\"` would cause them to be passed to `readlink`, so it would see the following parameters:

* `-f`
* `"/home/joshua/frog`
* `with`
* `spaces"`

When in doubt, ask `shellcheck`.

```
In setup_vkd3d.sh line 8:
basedir=$(dirname "$(readlink -f $0)")
                                 ^-- SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
basedir=$(dirname "$(readlink -f "$0")")
```

Also, I have no idea why your shell _doesn't_ perform word splitting on `$0` like it should, but maybe Hans-Kristian or Philip will.

Joshua-Ashton: I'm using the bash straight from Arch Linux, we all use Arch (I think) so I think it'd work for all of us.

I'll take another look at this later today -- shellcheck looks cool, never heard of it before.

Joshua-Ashton: Resolved.

Joshua-Ashton:
This should be okay to merge now.

PR #285, opened 2020-10-05
Fix various issues with fullscreen swapchain.

HansKristian-Work:


doitsujin:
It might be necessary to make the swap chain mutex recursive, since DXGI calls are re-entrant to a degree. At least for D3D11, there are games calling ResizeBuffers as they handle window resize event, which in turn can be triggered by ResizeTarget, so we'd be calling ResizeBuffers from within ResizeTarget and cause a deadlock.

HansKristian-Work:
Hm ... Should we use CRITICAL_SECTION then since it's win32-only code and CS is reentrant?

doitsujin:
Yeah, sounds alright.

HansKristian-Work:
Fixed.

libs/vkd3d/swapchain.c
```diff
@@ -385,6 +385,12 @@ static HRESULT d3d12_swapchain_set_display_mode(struct d3d12_swapchain *swapchai
 {
     DXGI_MODE_DESC matching_mode;
     HRESULT hr;
+
+    /* Workaround. FindClosestMatchingMode can take UNKNOWN format, but neither Wine nor DXVK DXGI seems to like it.
+     * Just opt for BGRA8 to workaround the issue. */
+    if (mode->Format == DXGI_FORMAT_UNKNOWN)
+        mode->Format = DXGI_FORMAT_B8G8R8A8_UNORM;
```
doitsujin, line 392: I think the correct fix here would be to use `swapchain->desc.Format` instead if the given format is `UNKNOWN`. At least that's what DXVK does.

Passing `UNKNOWN` to FindClosestMatchingMode like this is indeed intended to fail.

doitsujin (changes requested):

libs/vkd3d/swapchain.c
HansKristian-Work: Fixed.

Joshua-Ashton (approved):
LGTM

PR #286, opened 2020-10-05
Swapchain blitting rewrite - part 1

HansKristian-Work:
The first step is to add various pipelines to the meta system. It will be used by swapchain module in later parts.

### DIFF
--- a/libs/vkd3d/meta.c
+++ b/libs/vkd3d/meta.c
@@ -616,6 +616,95 @@ void vkd3d_copy_image_ops_cleanup(struct vkd3d_copy_image_ops *meta_copy_image_o
     vkd3d_free(meta_copy_image_ops->pipelines);
 }
 
+static VkResult vkd3d_meta_create_swapchain_render_pass(struct d3d12_device *device,
+        const struct vkd3d_swapchain_pipeline_key *key, VkRenderPass *render_pass)
+{
+    const struct vkd3d_vk_device_procs *vk_procs = &device->vk_procs;
+    VkRenderPassCreateInfo render_pass_info;
+    VkAttachmentDescription attachment_desc;
+    VkAttachmentReference attachment_ref;
+    VkSubpassDescription subpass_desc;
+    VkSubpassDependency subpass_dep;
+
+    render_pass_info.sType = VK_STRUCTURE_TYPE_RENDER_PASS_CREATE_INFO;
+    render_pass_info.pNext = NULL;
+    render_pass_info.flags = 0;
+    render_pass_info.attachmentCount = 1;
+    render_pass_info.pAttachments = &attachment_desc;
+    render_pass_info.subpassCount = 1;
+    render_pass_info.pSubpasses = &subpass_desc;
+    render_pass_info.dependencyCount = 1;
+    render_pass_info.pDependencies = &subpass_dep;
+
+    attachment_desc.loadOp = key->load_op;
+    attachment_desc.storeOp = VK_ATTACHMENT_STORE_OP_STORE;
+    attachment_desc.stencilLoadOp = VK_ATTACHMENT_LOAD_OP_DONT_CARE;
+    attachment_desc.stencilStoreOp = VK_ATTACHMENT_STORE_OP_DONT_CARE;
+    attachment_desc.format = key->format;
+    attachment_desc.samples = VK_SAMPLE_COUNT_1_BIT;
+    attachment_desc.initialLayout = VK_IMAGE_LAYOUT_UNDEFINED;
+    attachment_desc.finalLayout = VK_IMAGE_LAYOUT_PRESENT_SRC_KHR;
+    attachment_desc.flags = 0;
+
+    attachment_ref.attachment = 0;
+    attachment_ref.layout = VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL;
+
+    memset(&subpass_desc, 0, sizeof(subpass_desc));
+    subpass_desc.colorAttachmentCount = 1;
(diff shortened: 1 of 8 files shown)


libs/vkd3d/meta.c
```diff
@@ -53,6 +53,32 @@ static VkResult vkd3d_meta_create_descriptor_set_layout(struct d3d12_device *dev
...
+    sampler_info.compareEnable = VK_FALSE;
+    sampler_info.compareOp = VK_COMPARE_OP_ALWAYS;
+    sampler_info.unnormalizedCoordinates = VK_FALSE;
+    sampler_info.minLod = 0.0f;
+    sampler_info.maxLod = VK_LOD_CLAMP_NONE;
+    sampler_info.minFilter = filter;
+    sampler_info.magFilter = filter;
+    sampler_info.borderColor = VK_BORDER_COLOR_FLOAT_TRANSPARENT_BLACK;
+    sampler_info.mipLodBias = 0.0f;
+    return VK_CALL(vkCreateSampler(device->vk_device, &sampler_info, NULL, vk_sampler));
```
doitsujin, line 79: could we use the global device sampler map to create these?

If initialization order is an issue then we can keep it as it is.

libs/vkd3d/meta.c
```diff
@@ -620,6 +620,91 @@ void vkd3d_copy_image_ops_cleanup(struct vkd3d_copy_image_ops *meta_copy_image_o
...
+    render_pass_info.pNext = NULL;
+    render_pass_info.flags = 0;
+    render_pass_info.attachmentCount = 1;
+    render_pass_info.pAttachments = &attachment_desc;
+    render_pass_info.subpassCount = 1;
+    render_pass_info.pSubpasses = &subpass_desc;
+    render_pass_info.dependencyCount = 1;
+    render_pass_info.pDependencies = &subpass_dep;
+
+    attachment_desc.loadOp = VK_ATTACHMENT_LOAD_OP_CLEAR;
```
doitsujin, line 643: Do we really need to clear? DXVK uses `DONT_CARE` here.

libs/vkd3d/meta.c
```diff
@@ -620,6 +620,91 @@ void vkd3d_copy_image_ops_cleanup(struct vkd3d_copy_image_ops *meta_copy_image_o
...
+        ERR("Failed to create render pass, vr %d.\n", vr);
+        return hresult_from_vk_result(vr);
+    }
+
+    memset(&cb_state, 0, sizeof(cb_state));
+    memset(&blend_att, 0, sizeof(blend_att));
+    cb_state.sType = VK_STRUCTURE_TYPE_PIPELINE_COLOR_BLEND_STATE_CREATE_INFO;
+    cb_state.attachmentCount = 1;
+    cb_state.pAttachments = &blend_att;
+    blend_att.colorWriteMask = 0xf;
```
doitsujin, line 692: `R_BIT | G_BIT | B_BIT | A_BIT`? I know it's annoying to type but I don't like magic numbers.

We could also make a `#define` for that somewhere since we run into this case quite a lot.

libs/vkd3d/shaders/cs_swapchain_fullscreen.comp
```diff
@@ -0,0 +1,21 @@
+#version 450
+layout(local_size_x = 8, local_size_y = 8) in;
+
+layout(set = 1, binding = 0) writeonly uniform image2D FragColor;
+layout(set = 0, binding = 0) uniform sampler2D Tex;
+
+layout(push_constant, std430) uniform Registers
+{
+  uvec2 output_resolution;
```
doitsujin, line 9: Could we query this with `imageSize` or is there a scenario where the image size and output resolution would differ?

doitsujin (changes requested):
I'm not a big fan of adding a compute path already when we currently cannot use compute queues at all, and without seeing how it's going to be used, I think this should wait until we have an actual use case in order to prevent unintended bit rot.

As for the graphics path, just a few nits, see comments.

libs/vkd3d/shaders/cs_swapchain_fullscreen.comp
HansKristian-Work: Scaling as mentioned earlier.

libs/vkd3d/meta.c
HansKristian-Work: We could do that I guess.

HansKristian-Work:
I modified the gears demo to use compute present FWIW. That might serve as a test to avoid bitrot perhaps?

libs/vkd3d/meta.c
Joshua-Ashton: Right now scaling is not implemented properly, I will fix it later.

Maybe conditionally use DONT_CARE depending on whether we are NONE or STRETCH.
-> note: STRETCH and ASPECT_RATIO_STRETCH are the same for Win32 swapchains.

libs/vkd3d/meta.c
HansKristian-Work: Fixed.

libs/vkd3d/meta.c
HansKristian-Work: Fixed.

libs/vkd3d/meta.c
HansKristian-Work: Added a load_op member in the key.

libs/vkd3d/shaders/cs_swapchain_fullscreen.comp
HansKristian-Work: Removed from PR.

doitsujin (approved):

PR #287, opened 2020-10-06
vkd3d: Rewrite initial resource state tracking

HansKristian-Work:
For correctness, we will need to defer any initial resource state
    handling to the queue timeline. Here, we will build an UNDEFINED ->
    common layout barrier if (and only if):
    
    - The resource is marked to care about initial layout transition.
    - We are the first queue thread to observe that initial_transition
      member is 1 (atomic exchange).
    - The first use of the resource was not marked to be a discard.
      E.g., if the first use of the resource is an alias barrier, we must
      not emit an early barrier. The only we should do here is to clear the
      initial_transition member, and leave it like that.
    
    A command list maintains a list of d3d12_resources which *might* need a
    transition. For the first frame a resource is used (or so), it will not
    have the flag cleared yet, so multiple command lists might add the
    d3d12_resource to its own transition list. This is fine, as the queue
    will resolve it.
    
    If multiple queues see the same initial transition, there might be
    shenanigans, but the application must ensure there is either a
    submission boundary or fence boundary between the uses. Any initial
    layout transition will only be submitted after a Wait() is observed, as
    submission of the transition command buffer will be in-order with other
    submissions.

### DIFF
--- a/include/vkd3d.h
+++ b/include/vkd3d.h
@@ -151,7 +151,6 @@ struct vkd3d_optional_device_extensions_info
 };
 
 /* vkd3d_image_resource_create_info flags */
-#define VKD3D_RESOURCE_INITIAL_STATE_TRANSITION 0x00000001
 #define VKD3D_RESOURCE_PRESENT_STATE_TRANSITION 0x00000002
 
 struct vkd3d_image_resource_create_info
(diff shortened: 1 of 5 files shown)


HansKristian-Work:
Did some testing and seems to work fine in RE2, HZD and Metro with an emphasis on flipping fullscreen.

doitsujin (approved):
LGTM.

PR #290, opened 2020-10-08
Support typed UAV reads without explicit format.

HansKristian-Work:
Fixes red halos in Death Stranding on RADV. We declared a 4-component texture as r32f, which triggered some unwanted optimizations to only write the red channel.

DXIL is TODO. There's a lot of pending SSBO work that I don't want to pull in right away.


libs/vkd3d-shader/spirv.c
```diff
@@ -5670,15 +5683,16 @@ static uint32_t vkd3d_dxbc_compiler_get_image_type_id(struct vkd3d_dxbc_compiler
     uint32_t sampled_type_id;
     unsigned int uav_flags;
     SpvImageFormat format;
```
Joshua-Ashton, line 5685: Uninitialized?

Joshua-Ashton: nvm

Joshua-Ashton (approved):

PR #291, opened 2020-10-12
Use descriptor cookies to avoid comparing against stale pointers

HansKristian-Work:
We cannot compare resource pointers or view pointers,
since the pointers might have been recycled.
This leads to a scenario where we're not updating descriptors we're
supposed to, and the GPU reads a stale descriptor.

Fixes a GPU hang in Death Stranding (and possibly lots of other weird
crashes as well).


doitsujin (approved):

doitsujin:
Ah, actually - shouldn't we update `dst_metadata->cookie` in descriptor copies as well?

Issue #266, opened 2020-09-08
RE2: VkBufferView creation can overflow available memory


HansKristian-Work:
The hashmap approach has some flaws if apps create objects which cannot be reused. See https://github.com/HansKristian-Work/vkd3d-proton/issues/259

K0bin:
Here's the kind of buffer views that cause the problems in RE2:
`create buffer view dxgi format: 0, structure stride: 4, element size: 4, size: 96, offset: 0 flags: 0` (logged in 
`vkd3d_create_buffer_view_for_resource`)

fsyy:
in this thread they say this patch is the cause for crashing Horizon Zero Dawn every 10-20 minutes

https://github.com/ValveSoftware/Proton/issues/4125#issuecomment-691735117
https://github.com/ValveSoftware/Proton/issues/4125#issuecomment-691812754

didn't try it myself but wanted to mention it here.

RoyShapiro:
Here to confirm that reverting the following commits:

daf9f5c69fb69ab87672e61ee6c71ec2fb16d218
5a9d132b20de854f751d4c606c9546e6c34f5c4c
73d578e5abe5658bd8f9cca330a2f7a8f48e0465
684c658e22930f3f77488f77afb590d6889920a4

as suggested by @intersectRaven in the comment linked in the previous post has allowed me to progress much further in Resident Evil 2's The 4th Survivor mode, specifically to the point where I was genuinely defeated by the enemies and then manually quit the game successfully. That is, I did not encounter the freeze we are discussing here. Previously (without reverting the commits) I could reliably reproduce the freeze every time within 5-7 minutes of gameplay at max.

I did, however, notice, that during gameplay the fps has steadily decreased from around 50 fps in the main menu on launch to around 30 fps in the main menu when I quit after gameplay.

RoyShapiro:
@Emanem recently did an analysis of this issue, it may help: https://github.com/ValveSoftware/Proton/issues/4125#issuecomment-713530577
Also, I've tried compiling with SSBO enabled, both the WIP and the current master, and in RE2 I get the same, visualy, kind of crash as I did before, only _faster_. The framerate has also suffered significantly (lost about 15-20 fps compared to non-SSBO version), and, in the WIP version (https://github.com/HansKristian-Work/vkd3d-proton/pull/298) , I was getting objects, including parts of the level, appearing and disappearing out of view at odd angles.
[...]

PR #295, opened 2020-10-16
SSBO support for raw and structured buffers

doitsujin:
Adds a path to use untyped storage buffers for raw/structured buffers, rather than texel buffers, in an attempt to improve the situation for #266. This decision is made on a per-view basis depending on alignment requirements.

Currently disabled as it regresses some DXIL tests. Tested in Control, Shadow of the Tomb Raider, Ghostrunner Demo, and Resident Evil 3.


libs/vkd3d-shader/spirv.c
```diff
@@ -2389,7 +2390,7 @@ static const struct vkd3d_shader_resource_binding *vkd3d_dxbc_compiler_get_resou
     {
         const struct vkd3d_shader_resource_binding *current = &shader_interface->bindings[i];
 
-        if (!(current->flags & binding_flag))
+        if ((current->flags & binding_flags) != binding_flags)
```
HansKristian-Work, line 2401: Hm, how would this match against "non-SSBO" case? E.g. if we have SHADER_BINDING_FLAG_BUFFER | RAW_SSBO and we query for plain FLAG_BUFFER, we would get a false match I think?

libs/vkd3d-shader/spirv.c
```diff
@@ -5687,6 +5716,15 @@ static uint32_t vkd3d_dxbc_compiler_get_image_type_id(struct vkd3d_dxbc_compiler
             depth, resource_type_info->arrayed, resource_type_info->ms, is_uav ? 2 : 1, format);
 }
 
+static bool vkd3d_dxbc_compiler_use_ssbo(struct vkd3d_dxbc_compiler *compiler,
+        unsigned int structure_stride, bool raw)
+{
+    /* Extract lowest bit from stride to get alignment
+     * in case the stride is not a power of two. */
+    unsigned int dword_alignment = raw ? 4 : (structure_stride & -structure_stride);
```
HansKristian-Work, line 5733: Is structure stride in dwords in DXBC? It's bytes in DXIL at least.

libs/vkd3d-shader/spirv.c
```diff
@@ -5746,18 +5794,37 @@ static void vkd3d_dxbc_compiler_emit_resource_declaration(struct vkd3d_dxbc_comp
...
+        {
+            uint32_t array_type_id = vkd3d_spirv_build_op_type_runtime_array(builder,
+                    vkd3d_spirv_get_type_id(builder, VKD3D_TYPE_UINT, 1));
+            vkd3d_spirv_build_op_decorate1(builder, array_type_id, SpvDecorationArrayStride, 4);
+
+            type_id = vkd3d_spirv_build_op_type_struct(builder, &array_type_id, 1);
+            vkd3d_spirv_build_op_decorate(builder, type_id, SpvDecorationBufferBlock, NULL, 0);
+            vkd3d_spirv_build_op_member_decorate1(builder, type_id, 0, SpvDecorationOffset, 0);
+
+            if (!is_uav)
```
HansKristian-Work, line 5816: Shouldn't this target the SSBO variable itself, not the member type?

libs/vkd3d-shader/spirv.c
doitsujin: It is, yes.

libs/vkd3d-shader/spirv.c
doitsujin: Yeh, this should probably test for equality rather than ANDing flags.

But this aspect of the dxbc compiler is so broken that I'm not really sure, needs testing. It currently works because we always define raw SSBO bindings after any others.

libs/vkd3d-shader/spirv.c
doitsujin: Offset has to be applied to a member anyway, and BufferBlock is applied to the struct type, not the variable. This matches glslang behaviour.

libs/vkd3d-shader/spirv.c
doitsujin: At least that's what it seems to be like. I tested this with a `StructuredBuffer<uint>` and we're getting 1 for `structure_stride`.

doitsujin: This is why:
```c
    vkd3d_dxbc_compiler_emit_resource_declaration(compiler, reg,
            VKD3D_SHADER_RESOURCE_BUFFER, VKD3D_DATA_UINT, stride / 4, false);
```

libs/vkd3d-shader/spirv.c
```diff
@@ -5187,7 +5216,7 @@ static const struct vkd3d_shader_global_binding *vkd3d_dxbc_compiler_get_global_
 
     vkd3d_dxbc_compiler_emit_descriptor_binding(compiler, var_id, binding_info);
 
-    if (flags & VKD3D_SHADER_GLOBAL_BINDING_WRITE_ONLY)
+    if ((flags & VKD3D_SHADER_GLOBAL_BINDING_WRITE_ONLY) && !(flags & VKD3D_SHADER_GLOBAL_BINDING_RAW_SSBO))
         vkd3d_spirv_build_op_decorate(builder, var_id, SpvDecorationNonReadable, NULL, 0);
```
HansKristian-Work, line 5228: Should decorate SRV SSBO as NonWritable here on var_id.

libs/vkd3d-shader/spirv.c
```diff
@@ -5069,6 +5080,24 @@ static const struct vkd3d_shader_global_binding *vkd3d_dxbc_compiler_get_global_
...
+    {
+        uint32_t array_type_id = vkd3d_spirv_build_op_type_runtime_array(builder,
+                vkd3d_spirv_get_type_id(builder, VKD3D_TYPE_UINT, 1));
+        vkd3d_spirv_build_op_decorate1(builder, array_type_id, SpvDecorationArrayStride, 4);
+
+        type_id = vkd3d_spirv_build_op_type_struct(builder, &array_type_id, 1);
+        vkd3d_spirv_build_op_decorate(builder, type_id, SpvDecorationBufferBlock, NULL, 0);
+        vkd3d_spirv_build_op_member_decorate1(builder, type_id, 0, SpvDecorationOffset, 0);
+
+        if (data_type == VKD3D_DATA_RESOURCE)
```
HansKristian-Work, line 5101: I don't think we need to decorate the type itself when we decorate the variable declaration.

libs/vkd3d-shader/spirv.c
HansKristian-Work: Ok. Weird ...

libs/vkd3d-shader/spirv.c
HansKristian-Work: Hm ... OK. Vulkan spec mentions that "If fragmentStoresAndAtomics is not enabled, then all storage image, storage texel buffer, and storage buffer variables in the fragment stage must be decorated with the NonWritable decoration." So in that case it needs to be set on the variable, but sure, this shouldn't really matter for us.

libs/vkd3d-shader/spirv.c
doitsujin: That is (again) not what glslang does, which decorates the struct member. We're already doing that.

libs/vkd3d-shader/spirv.c
```diff
@@ -8847,7 +8854,14 @@ static void vkd3d_dxbc_compiler_emit_bufinfo(struct vkd3d_dxbc_compiler *compile
     vkd3d_dxbc_compiler_prepare_image(compiler, &image, &src->reg, NULL, VKD3D_IMAGE_FLAG_NONE);
 
     type_id = vkd3d_spirv_get_type_id(builder, VKD3D_TYPE_UINT, 1);
-    val_id = vkd3d_spirv_build_op_image_query_size(builder, type_id, image.image_id);
+
+    if (image.ssbo)
+    {
```
HansKristian-Work, line 8859: Probably doesn't need braces here?

libs/vkd3d/vkd3d_private.h
```diff
@@ -54,7 +54,7 @@
 #define VKD3D_MAX_VK_SYNC_OBJECTS         4u
 
 #define VKD3D_MAX_DESCRIPTOR_SETS 9u
```
HansKristian-Work, line 56: Does this need to be bumped to 10?

libs/vkd3d-shader/spirv.c
HansKristian-Work: Could leave it and add a comment about it at least?

libs/vkd3d-shader/spirv.c
HansKristian-Work: Ok.

libs/vkd3d/vkd3d_private.h
doitsujin: Yes, will fix.

libs/vkd3d-shader/spirv.c
doitsujin: Checking for equality does not work, everything just crashes and burns. Not sure why.

Can add a comment.

doitsujin: Ah, actually - I guess we should just ignore certain flags, such as `VKD3D_SHADER_BINDING_FLAG_BINDLESS`, when doing this check.

libs/vkd3d/vkd3d_private.h
```diff
@@ -1569,12 +1569,22 @@ enum vkd3d_bindless_flags
     VKD3D_BINDLESS_RAW_SSBO     = (1u << 6),
 };
 
+enum vkd3d_bindless_set_flag
+{
+    VKD3D_BINDLESS_SET_SAMPLER  = (1u << 0),
```
HansKristian-Work, line 1575: I like this approach. Does this mean for generic/mutable descriptor types that we will be able to have a single set with CBV | SRV | UAV | IMAGE | BUFFER ?

doitsujin: Yes, that's the idea.

HansKristian-Work (changes requested):
Ok, gone over the commits 3-4 times now and I think I get it. I like the general direction. I think it's mostly smaller nits.

libs/vkd3d/vkd3d_private.h
doitsujin: Done.

libs/vkd3d-shader/spirv.c
doitsujin: Leftover; removed.

libs/vkd3d-shader/spirv.c
doitsujin: Fixed.

libs/vkd3d/state.c
```diff
@@ -3423,11 +3423,9 @@ static uint32_t vkd3d_bindless_state_get_bindless_flags(struct d3d12_device *dev
         flags |= VKD3D_BINDLESS_CBV | VKD3D_BINDLESS_CBV_AS_SSBO;
 
     /* TODO enable */
-    #if 0
```
HansKristian-Work: hm?

doitsujin: Accident, fixed.

HansKristian-Work (approved):
LGTM now. Great stuff! :D

PR #303, opened 2020-10-19
vkd3d-shader: Remove d3d9 clutter

Joshua-Ashton:
This is all dead code doing absolutely nothing at all.


doitsujin:
Now I don't mind the pruning of useless opcodes (even though there's no real reason to do it either), but

> There are no other modifiers than saturate in DXBC.

isn't true, there are modifiers for 16-bit stuff. The encoding might be different, but I'd still like to keep them around as modifier flags.

Joshua-Ashton:
> Now I don't mind the pruning of useless opcodes (even though there's no real reason to do it either), but
> 
> > There are no other modifiers than saturate in DXBC.
> 
> isn't true, there are modifiers for 16-bit stuff. The encoding might be different, and we currently don't implement these, but I'd still like to keep the option around to implement these as modifiers.
> 
> First two commits look good, would drop the third one.

Ah yes, forgot about those (being on dsts specifically) -- I was mistaken here because this is an instruction modifier being passed through to a destination register for some reason...

Joshua-Ashton:
Updated branch with that change removed.

PR #307, opened 2020-10-21
vkd3d-shader: Implement 64-bit floats

Joshua-Ashton:
Implement support for 64 bit (or double-precision) floating point numbers in vkd3d-shader.


libs/vkd3d-shader/spirv.c
```diff
@@ -4486,6 +4486,41 @@ static unsigned int vkd3d_dxbc_compiler_get_output_variable_index(
...
+        unsigned int component_count = vkd3d_write_mask_component_count(mask);
+        if (component_count >= 3)
+            return VKD3DSP_WRITEMASK_0 | VKD3DSP_WRITEMASK_1;
+        else if (component_count >= 1)
+            return VKD3DSP_WRITEMASK_0;
+        else
+            return 0;
+    }
+
+    return mask;
```
doitsujin, line 4550: The logic here seems wrong. Double instructions may operate on only the `.zw` components, in which case we would have to compress the mask to `VKD3DSP_WRITEMASK_1`.

libs/vkd3d-shader/spirv.c
```diff
@@ -4486,6 +4486,41 @@ static unsigned int vkd3d_dxbc_compiler_get_output_variable_index(
...
 
+static DWORD vkd3d_dxbc_compiler_double_source_mask_fixup(enum vkd3d_data_type src, enum vkd3d_data_type dst, DWORD mask)
+{
+    /* In DXBC, doubles take up a .xy whereas a normal float/int
+     * takes up a .x or a .y, etc.
+     * This maps a .xy writemask to a .x when going between
+     * double -> float/int and vice-versa for float/int -> double.
+     */
+
+    if (src == dst)
```
doitsujin, line 4526: kind of redundant

doitsujin (changes requested):
Couple of things:
- Commit message of `vkd3d: Implement double data type` should be prefixed `vkd3d-shader` as well.
- Trivial commits could be squashed into one (DMOV, DADD/DMUL/DDIV, DMAX/DMIN/DFMA)
- Do operand modifiers work correctly? DXBC does actually support this, we should have a test for saturate, negate and abs and imlpement it as necessary.

libs/vkd3d-shader/spirv.c
Joshua-Ashton: Double instructions can operate on .xy too.

The actual elements of the mask here don't matter anyway, it's ``popcount``'ed after so just the number of bits count.

This is just the mask for loading src regs.

libs/vkd3d-shader/spirv.c
Joshua-Ashton: Not when we call this in regular movc which could have this.

You could theoretically insert this into any normal function.

libs/vkd3d-shader/spirv.c
```diff
@@ -5344,6 +5345,12 @@ static void vkd3d_dxbc_compiler_emit_dcl_global_flags(struct vkd3d_dxbc_compiler
         flags &= ~VKD3DSGF_FORCE_EARLY_DEPTH_STENCIL;
     }
 
+    if ((flags & VKD3DSGF_ENABLE_DOUBLE_PRECISION_FLOAT_OPS) || (flags & VKD3DSGF_ENABLE_11_1_DOUBLE_EXTENSIONS))
```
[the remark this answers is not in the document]
Joshua-Ashton: Can do

Joshua-Ashton:
> Do operand modifiers work correctly? DXBC does actually support this, we should have a test for saturate, negate and abs and imlpement it as necessary.

I think so, but I never actually tested this.
I will add a test for this. :)

libs/vkd3d-shader/spirv.c
Joshua-Ashton: Or any ops that have DOUBLE == DOUBLE.

The alternative here would be SRC == DOUBLE && DST != DOUBLE and vice versa.

Joshua-Ashton:
> Trivial commits could be squashed into one (DMOV, DADD/DMUL/DDIV, DMAX/DMIN/DFMA)
I committed these as they go into their families, but they could be merged

libs/vkd3d-shader/spirv.c
doitsujin: The point is that we take the same path anyway when falling through the `if`s below.

libs/vkd3d-shader/spirv.c
doitsujin: This isn't really clear from the code, and if we need a bit count rather than a mask, wouldn't it be better to return the bit count instead?

Joshua-Ashton: The src reg loading thing takes in the writemask, similarly to DXVK.

Joshua-Ashton: Consider it akin to https://github.com/doitsujin/dxvk/blob/master/src/dxbc/dxbc_compiler.cpp#L2947

Also I explained this in the comment in that function also.

libs/vkd3d-shader/spirv.c
Joshua-Ashton: That's not true when src == DOUBLE and dst == DOUBLE

libs/vkd3d-shader/spirv.c
doitsujin: The comment explains what this does, but not why it is the correct thing to do, and I'm not convinced that it is. Take this for example:

```
    condition_mask = vkd3d_dxbc_compiler_double_source_mask_fixup(src[0].reg.data_type, dst->reg.data_type, dst->write_mask);
    condition_id = vkd3d_dxbc_compiler_emit_load_src(compiler, &src[0], condition_mask);
```
Now if the write mask is `.zw`, we load the *first* 32-bit component of the condition vector, which is wrong since we'd need to read the second one.

Joshua-Ashton: > Now if the write mask is .zw, the resulting condition mask is VKD3DSP_WRITEMASK_0, which means that we load the first 32-bit component of the condition vector, which is wrong since we'd need to load the second one.

We'd need to load the first one regardless of whether the first element is selected or not.

``dadd r0.xyzw r1.xyzw r2.xyzw``
means ``r0.xyzw = r1.xyzw + r2.xyzw``

``dadd r0.zw r1.xyzw r2.xyzw``
means ``r0.zw = r1.xy + r2.xy``

doitsujin: But this isn't about dadd, this is about dmovc.

Joshua-Ashton: Hmmmmmmmmmmm, I will do more testing on this.

doitsujin: fwiw it seems like a pain in the arse to convince fxc to generate fp64 code with .zw dst masks (although not necessarily impossible), but i'd think something like
```c
uint32_t out_mask = 0;
if (mask & VKD3DSP_WRITEMASK_0)
  out_mask |= VKD3DSP_WRITEMASK_0;
if (mask & VKD3DSP_WRITEMASK_2)
  out_mask |= VKD3DSP_WRITEMASK_1;
```
would be correct for the dmovc case.

I'm not so sure about the float conversions, the MSDN docs are weird for those. But `dtof` in particular is special because you can (in theory) have something like `dtof r0.w, r1.xyzw`, in which case I think we should be converting `r1.xy` anyway? At least I think that's what the logic behind the DXVK code was.

doitsujin: At least for `ftod`, this logic is correct:
- In the scalar case, fxc only emits single-component swizzles anyway (i.e. `ftod r0.zw, r1.x`), which basically tells us that we'd need to look at the first component even if for some reason we end up getting a four-component swizzle.
- In the vec2 case, we get something like `ftod r0.xyzw, r1.yxyy`, so we only look at the first two components of the source operand.

Looking at [this](https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dtof--sm5---asm-), it does sound like `dtof` basically does the reverse, so DXVK's (and by extension, this) logic is indeed correct for the float conversion case.

HansKristian-Work (commented):
I don't have any particular issues beyond what @doitsujin has reviewed.

libs/vkd3d-shader/spirv.c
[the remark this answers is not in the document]
Joshua-Ashton: Resolved locally.

libs/vkd3d-shader/spirv.c
Joshua-Ashton: So does this overall logic seem okay?

I think this satisfies all the citeria:

```c
static DWORD vkd3d_dxbc_compiler_double_source_mask_fixup(enum vkd3d_data_type src, enum vkd3d_data_type dst, DWORD mask)
{
    /* In DXBC, doubles take up a .xy whereas a normal float/int
     * takes up a .x or a .y, etc.
     * This maps a .xy writemask to a .x when going between
     * double -> float/int and vice-versa for float/int -> double.
     */

    if (src == dst)
        return mask;

    if (src == VKD3D_DATA_DOUBLE)
    {
        unsigned int component_count = vkd3d_write_mask_component_count(mask);
        if (component_count >= 2)
            return VKD3DSP_WRITEMASK_0 | VKD3DSP_WRITEMASK_1 | VKD3DSP_WRITEMASK_2 | VKD3DSP_WRITEMASK_3;
        else if (component_count == 1)
            return VKD3DSP_WRITEMASK_0 | VKD3DSP_WRITEMASK_1;
        else
            return 0;
    }
    else if (dst == VKD3D_DATA_DOUBLE)
    {
        DWORD out_mask = 0;
        if (mask & VKD3DSP_WRITEMASK_0)
            out_mask |= VKD3DSP_WRITEMASK_0;
        if (mask & VKD3DSP_WRITEMASK_2)
            out_mask |= VKD3DSP_WRITEMASK_1;
        return out_mask;
    }

    return mask;
}
```

Joshua-Ashton:
Good catch on the operand modifiers, I have implemented those (in `Implement double data type`) and added a couple of tests for it :)

libs/vkd3d-shader/spirv.c
Joshua-Ashton: Committed the above for now.

doitsujin: This breaks ftod and friends whereas the old logic was correct.

Can't you just use your original logic for those and treat cmovd as a special case? It doesn't make much sense to unify them in a single function since it behaves differently.

Joshua-Ashton: Done.

PR #313, opened 2020-10-22
Move UAV counter address binding to bindless descriptor sets.

doitsujin:
Now that bindless is a hard requirement for all D3D12 descriptor types, and after the introduction of binding set flags with the Raw SSBO work, we can just pick an arbitrary set to cram extra bindings into, such as the UAV counter address buffer, and soon the offset buffer for raw SSBOs. This way  we don't have to track and bind these at draw time.

### DIFF
--- a/libs/vkd3d/resource.c
+++ b/libs/vkd3d/resource.c
@@ -3467,11 +3467,11 @@ void d3d12_desc_copy(struct d3d12_desc *dst, struct d3d12_desc *src,
             vk_copy = &vk_copies[copy_count++];
             vk_copy->sType = VK_STRUCTURE_TYPE_COPY_DESCRIPTOR_SET;
             vk_copy->pNext = NULL;
-            vk_copy->srcSet = src->heap->vk_descriptor_sets[metadata.set_index];
-            vk_copy->srcBinding = 0;
+            vk_copy->srcSet = src->heap->vk_descriptor_sets[metadata.binding.set];
+            vk_copy->srcBinding = metadata.binding.binding;
             vk_copy->srcArrayElement = src->heap_offset;
-            vk_copy->dstSet = dst->heap->vk_descriptor_sets[metadata.set_index];
-            vk_copy->dstBinding = 0;
+            vk_copy->dstSet = dst->heap->vk_descriptor_sets[metadata.binding.set];
+            vk_copy->dstBinding = metadata.binding.binding;
             vk_copy->dstArrayElement = dst->heap_offset;
             vk_copy->descriptorCount = 1;
         }
@@ -3485,17 +3485,17 @@ void d3d12_desc_copy(struct d3d12_desc *dst, struct d3d12_desc *src,
             }
             else
             {
-                uint32_t set_index = vkd3d_bindless_state_find_set(&device->bindless_state,
-                        VKD3D_BINDLESS_SET_UAV | VKD3D_BINDLESS_SET_COUNTER);
+                struct vkd3d_descriptor_binding binding = vkd3d_bindless_state_find_set(
+                        &device->bindless_state, VKD3D_BINDLESS_SET_UAV | VKD3D_BINDLESS_SET_COUNTER);
 
                 vk_copy = &vk_copies[copy_count++];
                 vk_copy->sType = VK_STRUCTURE_TYPE_COPY_DESCRIPTOR_SET;
                 vk_copy->pNext = NULL;
-                vk_copy->srcSet = src->heap->vk_descriptor_sets[set_index];
-                vk_copy->srcBinding = 0;
+                vk_copy->srcSet = src->heap->vk_descriptor_sets[binding.set];
+                vk_copy->srcBinding = binding.binding;
                 vk_copy->srcArrayElement = src->heap_offset;
-                vk_copy->dstSet = dst->heap->vk_descriptor_sets[set_index];
-                vk_copy->dstBinding = 0;
+                vk_copy->dstSet = dst->heap->vk_descriptor_sets[binding.set];
+                vk_copy->dstBinding = binding.binding;
                 vk_copy->dstArrayElement = dst->heap_offset;
                 vk_copy->descriptorCount = 1;
             }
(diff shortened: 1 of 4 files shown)


libs/vkd3d/resource.c
```diff
@@ -5265,9 +5266,19 @@ static HRESULT d3d12_descriptor_heap_create_descriptor_pool(struct d3d12_descrip
...
+                ssbo_pool = vk_pool_size;
         }
+
+        ssbo_count += vkd3d_popcount(set_info->flags & VKD3D_BINDLESS_SET_EXTRA_MASK);
     }
 
+    if (ssbo_count && !ssbo_pool)
+        ssbo_pool = &vk_pool_sizes[pool_count++];
+
+    ssbo_pool->descriptorCount += ssbo_count;
```
HansKristian-Work, line 5280: What if ssbo_pool is NULL?

libs/vkd3d/resource.c
```diff
@@ -5427,6 +5440,66 @@ static HRESULT d3d12_descriptor_heap_create_uav_counter_buffer(struct d3d12_desc
     return S_OK;
 }
 
+static void d3d12_descriptor_heap_update_extra_bindings(struct d3d12_descriptor_heap *descriptor_heap,
+        struct d3d12_device *device)
+{
+    const struct vkd3d_vk_device_procs *vk_procs = &device->vk_procs;
+    unsigned int i, binding_index, set_index = 0, write_count = 0;
+    VkDescriptorBufferInfo vk_buffer_info[8];
```
HansKristian-Work, line 5448: Should probably have a define instead of magic value of 8.

libs/vkd3d/state.c
```diff
@@ -3353,44 +3353,59 @@ static HRESULT vkd3d_bindless_state_add_binding(struct vkd3d_bindless_state *bin
     VkDescriptorSetLayoutBindingFlagsCreateInfoEXT vk_binding_flags_info;
     const struct vkd3d_vk_device_procs *vk_procs = &device->vk_procs;
     VkDescriptorSetLayoutCreateInfo vk_set_layout_info;
-    VkDescriptorSetLayoutBinding vk_binding_info;
-    VkDescriptorBindingFlagsEXT vk_binding_flags;
+    VkDescriptorSetLayoutBinding vk_binding_info[9];
```
HansKristian-Work, line 3340: Why 9?

libs/vkd3d/state.c
```diff
@@ -3353,44 +3353,59 @@ static HRESULT vkd3d_bindless_state_add_binding(struct vkd3d_bindless_state *bin
...
+    unsigned int i;
     VkResult vr;
 
     set_info->vk_descriptor_type = vk_descriptor_type;
     set_info->heap_type = flags & VKD3D_BINDLESS_SET_SAMPLER
             ? D3D12_DESCRIPTOR_HEAP_TYPE_SAMPLER
             : D3D12_DESCRIPTOR_HEAP_TYPE_CBV_SRV_UAV;
     set_info->flags = flags;
-    set_info->binding_index = 0;
+    set_info->binding_index = vkd3d_popcount(flags & VKD3D_BINDLESS_SET_EXTRA_MASK);
```
HansKristian-Work, line 3351: Will flags be 0 for non-SSBO descriptor types?

libs/vkd3d/state.c
```diff
@@ -3466,12 +3482,15 @@ HRESULT vkd3d_bindless_state_init(struct vkd3d_bindless_state *bindless_state,
...
+    if (bindless_state->flags & VKD3D_RAW_VA_UAV_COUNTER)
+        extra_bindings |= VKD3D_BINDLESS_SET_EXTRA_UAV_COUNTER_BUFFER;
+
     if (FAILED(hr = vkd3d_bindless_state_add_binding(bindless_state, device,
             VKD3D_BINDLESS_SET_SAMPLER, VK_DESCRIPTOR_TYPE_SAMPLER)))
         goto fail;
 
     if (FAILED(hr = vkd3d_bindless_state_add_binding(bindless_state, device,
-            VKD3D_BINDLESS_SET_CBV,
+            VKD3D_BINDLESS_SET_CBV | extra_bindings,
```
HansKristian-Work, line 3477: Why here and not VKD3D_BINDLESS_SET_UAV | BUFFER? Ideally we shouldn't have to bind the CBV set only for the sake of using UAVs.

HansKristian-Work (changes requested):
Comments.

libs/vkd3d/state.c
doitsujin: What do you mean exactly?

`flags & VKD3D_BINDLESS_SET_EXTRA_MASK` will be 0 whenever there are no extra bindings set. `flags` is never going to be 0 since that'll contain `VKD3D_BINDLESS_SET_UAV` and friends.

HansKristian-Work: I mean if `flags & VKD3D_BINDLESS_SET_EXTRA_MASK` is only non-null for one descriptor type?

HansKristian-Work: Ok, I think I see now.

libs/vkd3d/state.c
doitsujin: Putting it into `VKD3D_BINDLESS_SET_UAV | BUFFER` actually seems worse if the SSBO path is enabled since we'd always bind the texel buffer set, which is less likely to be used in a shader that uses UAV counters than the CBV set.

Of course we can conditionally add it either to the `UAV | BUFFER` or `UAV | RAW_SSBO` set depending on whether the SSBO path is enabled, but that's a lot of extra logic just to decide where to put the binding and *potentially* save one SGPR, not really sure if that's really worth it.

libs/vkd3d/resource.c
doitsujin: Fixed; the code above also needs to initialize the pool which I've also fixed.

libs/vkd3d/resource.c
doitsujin: Added `VKD3D_BINDLESS_SET_MAX_EXTRA_BINDINGS`.

libs/vkd3d/state.c
doitsujin: `VKD3D_BINDLESS_SET_MAX_EXTRA_BINDINGS + 1`, the last one being the descriptor array.

libs/vkd3d/state.c
HansKristian-Work: Ok.

HansKristian-Work (approved):
LGTM.

PR #318, opened 2020-10-27
Implement offset buffers

### LINKED ISSUE #266: RE2: VkBufferView creation can overflow available memory

doitsujin:
Cursed workaround for SSBO alignment requirements on Nvidia. Should fix #266 for good.

At least the last commit should probably not be merged before DXIL support lands, since it may break games using DXIL shaders.

### DIFF
--- a/include/private/hashmap.h
+++ b/include/private/hashmap.h
@@ -50,14 +50,9 @@ struct hash_map
     uint32_t used_count;
 };
 
-static inline void *hash_map_ptr_offset(void *ptr, size_t offset)
-{
-    return ((char*)ptr) + offset;
-}
-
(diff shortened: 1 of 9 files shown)


HansKristian-Work (commented):
Looks pretty good! I'll implement the DXIL side tomorrow and test.

I suppose we will slide in DXIL side right before the last commit `vkd3d: Enable offset buffer as necessary`?

doitsujin:
Sounds like a plan.

HansKristian-Work (approved):
No obvious regressions on NV at least. HZD, DS and RE2 renders correctly for me.

PR #320, opened 2020-10-27
build: Use separate version string and build hash

Saancreed:
Version string is used in logging for information purposes, but pipelines blobs and libraries use `uint64_t`–based commit hash. Using fixed–size integer silences warnings about string length and makes storing build info a little more space–efficient.

The hash is obtained separately from version string and is shifted to the left by 4 bits if the working tree is dirty.

Supersedes PR #284.


doitsujin (approved):
LGTM. Not the biggest fan of having *two* headers for almost the same purpose, but this seems like the easiest solution.

PR #333, opened 2020-10-28
Update build-script

SveSop:
Change to delete libvkd3d-proton-utils-2.dll


SveSop:
Yeah.. someone was faster :+1:

Issue #335, opened 2020-10-28
Rename enable_standalone_d3d12


Joshua-Ashton:
I always forget whether its enable_d3d12_standalone or enable_standalone_d3d12 -- we should probably just make this enable_d3d12.

K0bin:
Why not just `standalone`?

Joshua-Ashton:
I don't think that's necessarily clear enough because technically vkd3d native, etc *is* standalone. You can use it without anything.

This makes it simpler as it just says ``enable_d3d12`` which implies wouldn't be built otherwise -- which is true.

Saancreed:
How about defaulting this options to `True` if target machine is Windows instead?

HansKristian-Work:
Yes. I agree on that. There is very little reason to not build the standalone dll when targeting Windows.

PR #337, opened 2020-10-28
build: Rename enable_standalone_d3d12 to enable_d3d12

Joshua-Ashton:
The old name was way too long if we want to quickly test builds, and it was easy to mistake.

Closes: #335


HansKristian-Work:
This will break Proton build scripts. How annoying is it to keep the old name as well for compat?

Issue #336, opened 2020-10-28
Remove dependency on x86-linux-gnu cross file


Joshua-Ashton:
Not everywhere ships this, and we can eliminate it's dependency easily.

PR #338, opened 2020-10-28
build: Eliminate need for x86-linux-gnu cross file

Joshua-Ashton:
Replace this with our own build-linux32.txt

Closes: #336

### DIFF
--- a/build-linux32.txt
+++ b/build-linux32.txt
@@ -1,0 +1,24 @@
+[binaries]
+c = 'gcc'
+cpp = 'g++'
+ar = 'gcc-ar'
+strip = 'strip'
+
+[properties]
+c_args = ['-m32']
+c_link_args = ['-m32']
+cpp_args = ['-m32']
+cpp_link_args = ['-m32']
+pkg_config_path = ['/usr/lib32/pkgconfig']
+pkg_config_libdir = ['/usr/lib32/pkgconfig']
+
+[paths]
+prefix = '/usr'
+libdir = 'lib32'
+bindir = 'bin'
+
+[host_machine]
+system = 'linux'
+cpu_family = 'x86'
+cpu = 'i686'
+endian = 'little'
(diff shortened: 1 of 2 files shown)


build-linux32.txt
```diff
@@ -0,0 +1,24 @@
...
+cpp = 'g++'
+ar = 'gcc-ar'
+strip = 'strip'
+
+[properties]
+c_args = ['-m32']
+c_link_args = ['-m32']
+cpp_args = ['-m32']
+cpp_link_args = ['-m32']
+pkg_config_path = ['/usr/lib32/pkgconfig']
```
DadSchoorse, line 12: This isn't true on all distros.

DadSchoorse:
Why not instruct people to use `CC="gcc -m32" CXX="g++ -m32" PKG_CONFIG="i686-pc-linux-gnu-pkg-config" meson ...`  with the correct `--libdir` for their distro instead of using a cross file?

Joshua-Ashton:
Do we even need the package config stuff anymore? @HansKristian-Work

Joshua-Ashton:
> Why not instruct people to use `CC="gcc -m32" CXX="g++ -m32" PKG_CONFIG="i686-pc-linux-gnu-pkg-config" meson ...` with the correct `--libdir` for their distro instead of using a cross file?

We could do that in the package-release script when you do ``--native`` perhaps and update the readme for it.

HansKristian-Work:
@Joshua-Ashton I don't see a reason to remove the pkgconfig. If we're installing any .so/.a files we should also install a pkg-config for it.

build-linux32.txt
HansKristian-Work: For weird distros it should be fine to bring your own toolchain file or use good old envvars.

HansKristian-Work (approved):
LGTM.

PR #225, opened 2020-07-27
meson: Fix header files generator

wine-pardon:
I kept only good part from #222 .

---
During the build process .idl header files will be geherated only once.

Signed-off-by: wine-pardon <68657201+wine-pardon@users.noreply.github.com>

PR #339, opened 2020-10-28
build: Only generate headers once

Joshua-Ashton:
This makes headers a dependency rather than a generator target.
This also means we get proper dependency tracking of them between projects.

Supercedes: #225


HansKristian-Work (approved):
LGTM.

PR #342, opened 2020-10-29
vkd3d-shader: Fix hard-coded UAV and sampler limits

### LINKED ISSUE #326: Fix 64 UAV limit in vkd3d-shader
We should use a hash map rather than a flat fixed-size array to store UAV metadata.

doitsujin:
Title, basically.

Closes #326.


HansKristian-Work (approved):
LGTM.

PR #349, opened 2020-10-30
Implement a workaround for typed buffer view memory explosion

### LINKED ISSUE #345: Workaround typed SRV/UAV buffer memory exhaustion when using VkBufferView
It is still theoretically possible to exhaust some resources when using typed buffer views too aggressively (SSBO path only fixes the raw buffer situation). Investigate if this affects HZD.

HansKristian-Work:
The overall idea is to bind typed views at aligned offsets and ranges, and rely on the offset buffer framework to make this work. This way we guarantee a reasonable upper bound for number of unique views that might exist per ID3D12Resource.

Fix #345.

Fixes this OOM condition on HZD and Death Stranding (and possibly others) on NVIDIA based on basic testing. There is no logging about ~1k views being created, where it would create at least 150k views before after a few minutes.


doitsujin:
We need to check for `VKD3D_TYPED_OFFSET_BUFFER` in `d3d12_descriptor_heap_init_data_buffer` as well, or otherwise we won't create the offset buffer on AMD cards which currently leads to crashes.

HansKristian-Work:
@doitsujin Fixed on AMD. Tested AMD/NV test suite, RE2, HZD, DS. Renders fine.

libs/vkd3d/command.c
```diff
@@ -5796,16 +5796,25 @@ static void d3d12_command_list_clear_uav(struct d3d12_command_list *list, const
...
     {
+        const struct vkd3d_bound_buffer_range *ranges = desc->heap->buffer_ranges.host_ptr;
+
         if (args->has_view)
         {
-            VkDeviceSize byte_count = args->u.view->format->byte_count
-                    ? args->u.view->format->byte_count
-                    : sizeof(uint32_t);  /* structured buffer */
-            full_rect.right = args->u.view->info.buffer.size / byte_count;
+            if ((list->device->bindless_state.flags & VKD3D_TYPED_OFFSET_BUFFER) != 0)
```
doitsujin, line 5803: Nit: we never really do the `!= 0` thing when testing bit masks, could be removed here as well for consistency.

libs/vkd3d/resource.c
```diff
@@ -4004,6 +4004,20 @@ static void vkd3d_buffer_view_get_bound_range_ssbo(struct d3d12_desc *descriptor
     }
 }
 
+static bool vkd3d_buffer_view_get_bound_range_typed(struct d3d12_desc *descriptor,
```
doitsujin, line 4007: I'm not a big fan of the name here since it does something different from `get_bound_range_ssbo` - this creates an actual view object, the other one really just computes some numbers.

libs/vkd3d-shader/spirv.c
```diff
@@ -8514,6 +8521,39 @@ static uint32_t vkd3d_dxbc_compiler_adjust_ssbo_offset(struct vkd3d_dxbc_compile
...
+    length_id = vkd3d_spirv_build_op_composite_extract1(builder, uint_type_id, bounds_id, 1);
+
+    /* cond = offset < length */
+    cond_id = vkd3d_spirv_build_op_uless_than(builder, bool_type_id, coordinate_id, length_id);
+
+    /* In case of out-of-bounds access, set offset to a number that we
+     * expect to be out-of-bounds of the actual Vulkan resource as well. */
+    coordinate_id = vkd3d_spirv_build_op_select(builder, uint_type_id, cond_id,
+            vkd3d_spirv_build_op_iadd(builder, uint_type_id, coordinate_id, offset_id),
+            vkd3d_dxbc_compiler_get_constant_uint(compiler, 0xffffffffu));
```
doitsujin, line 8552: Do we know if this could potentially overflow and cause issues depending on view format?

doitsujin (commented):
Couple of nits, looks good otherwise.

libs/vkd3d-shader/spirv.c
HansKristian-Work: It passes the OOB tests on both AMD and NV. There cannot be address computation in the shader with typed views.

libs/vkd3d/command.c

libs/vkd3d/resource.c

doitsujin (approved):

PR #350, opened 2020-10-30
Rework RTV/DSV descriptors

### LINKED ISSUE #325: Use view cache for RTV/DSV descriptors
Basically, do the same things we do for SRV/UAV descriptors and get rid of ref-counted views.

**Edit:** And implement RTV/DSV copies based on that.

doitsujin:
Makes use of the per-resource view map for render targets and implements some missing functionality for these descriptor types, like NULL descriptors (trivial) and descriptor copies.

Closes #325.


HansKristian-Work (approved):
Nice! LGTM.

PR #354, opened 2020-10-30
vkd3d: Init hashmap in vkd3d_create_image_resource

Joshua-Ashton:
Closes: #353


doitsujin:
Do we want to keep the native demos around even though parts of the API implementation don't even work with native builds?

HansKristian-Work:
The extra maintenance to have them is fairly minimal, and it's useful to sanity check with them. The APIs that don't work in native builds are quite obscure to be fair ...

Joshua-Ashton:
I think we should keep the native demos around because afaik we aren't planning to (and imo, shouldn't) deprecate/toss out the ability to compile native.

It's a good sanity check also.

HansKristian-Work (approved):
LGTM.

PR #361, opened 2020-11-05
Fix pkgconfig + fix missing newlines at end of files + remove trailing whitespace

Joshua-Ashton:
Supercedes: #218


HansKristian-Work (approved):
Not sure if I like doing whitespace fix commits, but guess it's fine. Not the first time.

PR #362, opened 2020-11-05
build: fix build_arch() in package-release.sh so that quoted input ar…

imaami:
…guments behave normally

Signed-off-by: Juuso Alasuutari <juuso.alasuutari@gmail.com>

### DIFF
--- a/package-release.sh
+++ b/package-release.sh
@@ -45,28 +45,30 @@ while [ $# -gt 0 ]; do
 done
 
 function build_arch {
+  local arch="$1"
+  shift
+
   cd "$VKD3D_SRC_DIR"
 
-  # shellcheck disable=SC2086
-  meson $2                           \
+  meson "$@"                         \
         --buildtype "release"        \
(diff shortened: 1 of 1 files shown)


imaami:
This makes the shellcheck exception for [SC2086](https://github.com/koalaman/shellcheck/wiki/SC2086) unnecessary as well.

package-release.sh
```diff
@@ -45,28 +45,30 @@ while [ $# -gt 0 ]; do
 done
 
 function build_arch {
+  local _1="$1"
```
Joshua-Ashton, line 48: Can you give it a better name than _1?

local arch="$1"
shift
local cross_args="$@"

Joshua-Ashton (changes requested):

package-release.sh
imaami: Sure thing, although the latter can't be a normal variable as `"$@"` is an array expansion. I'd politely argue that `"$@"` should remain as-is because it's a common shellscript idiom.

PR #367, opened 2020-11-06
MHW fixes

doitsujin:
Fixes various validation errors and a crash at startup.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -4061,13 +4061,13 @@ cleanup:
     vk_image_barriers[0].oldLayout = dst_layout;
     vk_image_barriers[0].newLayout = dst_resource->common_layout;
 
-    vk_image_barriers[1].srcAccessMask = src_access;
+    vk_image_barriers[1].srcAccessMask = 0;
     vk_image_barriers[1].dstAccessMask = 0;
     vk_image_barriers[1].oldLayout = src_layout;
     vk_image_barriers[1].newLayout = src_resource->common_layout;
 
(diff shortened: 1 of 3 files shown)


libs/vkd3d/command.c
```diff
@@ -4061,13 +4061,13 @@ static void d3d12_command_list_copy_image(struct d3d12_command_list *list,
     vk_image_barriers[0].oldLayout = dst_layout;
     vk_image_barriers[0].newLayout = dst_resource->common_layout;
 
-    vk_image_barriers[1].srcAccessMask = src_access;
-    vk_image_barriers[1].dstAccessMask = 0;
+    vk_image_barriers[1].srcAccessMask = 0;
+    vk_image_barriers[1].dstAccessMask = VK_ACCESS_TRANSFER_READ_BIT;
```
HansKristian-Work, line 4065: Is VK_ACCESS_TRANSFER_READ_BIT even meaningful in common_layout?

HansKristian-Work (changes requested):
Minor nit, otherwise fine.

libs/vkd3d/command.c
doitsujin: Right, we don't transition to `TRANSFER_SRC_OPTIMAL` anymore.

HansKristian-Work: Using 0 dst mask seems like the correct option, since the access masks are resolved when application performance a transition away form COPY_SOURCE or continues to copy.

HansKristian-Work (approved):
LGTM, sneaky bugs :V

PR #368, opened 2020-11-06
vkd3d: Only care about DSV mismatch if the pipeline actually uses it.

doitsujin:
Avoids some unnecessary late compiles in Monster Hunter World, and probably others.


HansKristian-Work (approved):
LGTM.

PR #375, opened 2020-11-09
Add support for VK_VALVE_mutable_descriptor_type

HansKristian-Work:
RADV MR: https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/7967

### DIFF
--- a/libs/vkd3d/device.c
+++ b/libs/vkd3d/device.c
@@ -104,6 +104,8 @@ static const struct vkd3d_optional_extension_info optional_device_extensions[] =
     VK_EXTENSION(AMD_SHADER_CORE_PROPERTIES_2, AMD_shader_core_properties2),
     /* NV extensions */
     VK_EXTENSION(NV_SHADER_SM_BUILTINS, NV_shader_sm_builtins),
+    /* VALVE extensions */
+    VK_EXTENSION(VALVE_MUTABLE_DESCRIPTOR_TYPE, VALVE_mutable_descriptor_type),
 };
 
 static unsigned int get_spec_version(const VkExtensionProperties *extensions,
@@ -793,6 +795,7 @@ static void vkd3d_physical_device_info_init(struct vkd3d_physical_device_info *i
     VkPhysicalDeviceCustomBorderColorFeaturesEXT *custom_border_color_features;
     VkPhysicalDeviceTimelineSemaphoreFeaturesKHR *timeline_semaphore_features;
     VkPhysicalDevicePushDescriptorPropertiesKHR *push_descriptor_properties;
+    VkPhysicalDeviceMutableDescriptorTypeFeaturesVALVE *mutable_features;
     VkPhysicalDeviceShaderFloat16Int8FeaturesKHR *float16_int8_features;
     VkPhysicalDeviceShaderCoreProperties2AMD *shader_core_properties2;
     VkPhysicalDeviceDepthClipEnableFeaturesEXT *depth_clip_features;
@@ -839,6 +842,7 @@ static void vkd3d_physical_device_info_init(struct vkd3d_physical_device_info *i
     robustness2_features = &info->robustness2_features;
     extended_dynamic_state_features = &info->extended_dynamic_state_features;
     external_memory_host_properties = &info->external_memory_host_properties;
+    mutable_features = &info->mutable_descriptor_features;
 
     info->features2.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_FEATURES_2;
     info->properties2.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_PROPERTIES_2;
@@ -999,6 +1003,12 @@ static void vkd3d_physical_device_info_init(struct vkd3d_physical_device_info *i
     {
         shader_sm_builtins_properties->sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_SHADER_SM_BUILTINS_PROPERTIES_NV;
         vk_prepend_struct(&info->properties2, shader_sm_builtins_properties);
+    }
+
+    if (vulkan_info->VALVE_mutable_descriptor_type)
+    {
+        mutable_features->sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_MUTABLE_DESCRIPTOR_TYPE_FEATURES_VALVE;
+        vk_prepend_struct(&info->features2, mutable_features);
     }
 
     VK_CALL(vkGetPhysicalDeviceFeatures2(physical_device, &info->features2));
--- a/libs/vkd3d/resource.c
+++ b/libs/vkd3d/resource.c
@@ -5360,9 +5360,17 @@ static HRESULT d3d12_descriptor_heap_create_descriptor_pool(struct d3d12_descrip
     if (!pool_count)
         return S_OK;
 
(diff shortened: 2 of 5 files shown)


libs/vkd3d/device.c
```diff
@@ -754,6 +767,7 @@ static void vkd3d_physical_device_info_init(struct vkd3d_physical_device_info *i
     VkPhysicalDeviceCustomBorderColorFeaturesEXT *custom_border_color_features;
     VkPhysicalDeviceTimelineSemaphoreFeaturesKHR *timeline_semaphore_features;
     VkPhysicalDevicePushDescriptorPropertiesKHR *push_descriptor_properties;
+    VkPhysicalDeviceMutableDescriptorTypeFeaturesVALVE *mutable_features;
```
Joshua-Ashton, line 770: funny christmastree ordering

baryluk:
Hi Hans, I can test this Mesa and v3d3 changes on my AMD Fury X (FIJI, GFX8). Are there some specific game titles to test this with that you would recommend? For testing correctness and performance compared to without it.

libs/vkd3d/state.c
```diff
@@ -3625,6 +3676,57 @@ static uint32_t vkd3d_bindless_state_get_bindless_flags(struct d3d12_device *dev
...
+    set_layout_info.sType = VK_STRUCTURE_TYPE_DESCRIPTOR_SET_LAYOUT_CREATE_INFO;
+    set_layout_info.pNext = &binding_flags;
+    set_layout_info.flags = VK_DESCRIPTOR_SET_LAYOUT_CREATE_UPDATE_AFTER_BIND_POOL_BIT;
+    set_layout_info.bindingCount = 1;
+    set_layout_info.pBindings = &binding;
+
+    supported.sType = VK_STRUCTURE_TYPE_DESCRIPTOR_SET_LAYOUT_SUPPORT;
+    supported.pNext = NULL;
+    VK_CALL(vkGetDescriptorSetLayoutSupport(device->vk_device, &set_layout_info, &supported));
+    return supported.supported == VK_TRUE;
```
doitsujin, line 3727: do we need to check for the host local one as well or is that guaranteed to be supported?

libs/vkd3d/state.c
```diff
@@ -3652,33 +3754,52 @@ HRESULT vkd3d_bindless_state_init(struct vkd3d_bindless_state *bindless_state,
...
 
-    if (FAILED(hr = vkd3d_bindless_state_add_binding(bindless_state, device,
-            VKD3D_BINDLESS_SET_CBV | extra_bindings,
-            vkd3d_bindless_state_get_cbv_descriptor_type(bindless_state))))
-        goto fail;
+    if (vkd3d_bindless_supports_mutable_type(device))
+    {
+        INFO("Device supports VK_VALVE_mutable_descriptor_type.\n");
+        /* If we can, prefer to use one universal descriptor type which works for any descriptor. */
+        bindless_state->flags |= VKD3D_BINDLESS_MUTABLE_TYPE;
```
doitsujin, line 3761: I'd rather have this check + setting flags inside `vkd3d_bindless_state_get_bindless_flags`, and `if (flags & VKD3D_BINDLESS_MUTABLE_TYPE)` here instead.

doitsujin (changes requested):
few nits and questions.

libs/vkd3d/state.c
HansKristian-Work: Guaranteed by spec, but I'll check that as well for completeness' sake.

libs/vkd3d/resource.c
```diff
@@ -5463,7 +5478,8 @@ static HRESULT d3d12_descriptor_heap_create_descriptor_set(struct d3d12_descript
     }
 
     if (device->device_info.robustness2_features.nullDescriptor &&
-        binding->vk_descriptor_type != VK_DESCRIPTOR_TYPE_SAMPLER)
+        binding->vk_descriptor_type != VK_DESCRIPTOR_TYPE_SAMPLER &&
+        binding->vk_descriptor_type != VK_DESCRIPTOR_TYPE_MUTABLE_VALVE)
```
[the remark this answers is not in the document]
HansKristian-Work: Yes. There is no generic null descriptor we can use, but we could write NULL sampled images to clear out the memory at least. I don't think this code ever fixed any games in particular, it was a defensive measure that was trivial enough to do.

libs/vkd3d/state.c

libs/vkd3d/state.c

libs/vkd3d/resource.c
```diff
@@ -5360,9 +5360,24 @@ static HRESULT d3d12_descriptor_heap_create_descriptor_pool(struct d3d12_descrip
...
         return S_OK;
 
+    /* If using mutable type, we will allocate the most conservative size.
+     * This is fine since we're attempting to allocate a completely generic descriptor set. */
+
     vk_pool_info.sType = VK_STRUCTURE_TYPE_DESCRIPTOR_POOL_CREATE_INFO;
     vk_pool_info.pNext = NULL;
-    vk_pool_info.flags = VK_DESCRIPTOR_POOL_CREATE_UPDATE_AFTER_BIND_BIT_EXT;
+
+    if (descriptor_heap->desc.Flags & D3D12_DESCRIPTOR_HEAP_FLAG_SHADER_VISIBLE)
```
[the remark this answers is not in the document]

libs/vkd3d/state.c
HansKristian-Work: actually, derp, now I remember why I did the check there and not in get_bindless_flags ...

libs/vkd3d/resource.c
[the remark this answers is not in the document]
doitsujin: Yeh, something like that would potentially help out with the likes of AMDVLK/AMD proprietary on Linux.

libs/vkd3d/state.c

libs/vkd3d/resource.c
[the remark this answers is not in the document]

libs/vkd3d/state.c
```diff
@@ -3538,19 +3565,99 @@ static HRESULT vkd3d_bindless_state_add_binding(struct vkd3d_bindless_state *bin
...
+
+    supported.sType = VK_STRUCTURE_TYPE_DESCRIPTOR_SET_LAYOUT_SUPPORT;
+    supported.pNext = NULL;
+    VK_CALL(vkGetDescriptorSetLayoutSupport(device->vk_device, &set_layout_info, &supported));
+    if (!supported.supported)
+        return false;
+
+    set_layout_info.flags = VK_DESCRIPTOR_SET_LAYOUT_CREATE_HOST_ONLY_POOL_BIT_VALVE;
+    binding_flag = VK_DESCRIPTOR_BINDING_VARIABLE_DESCRIPTOR_COUNT_BIT_EXT;
+    VK_CALL(vkGetDescriptorSetLayoutSupport(device->vk_device, &set_layout_info, &supported));
```
doitsujin, line 3657: If it's guaranteed by spec then we should probably revert this since we now have some inconsistent checks (host-only checks for `if (device->device_info.mutable_descriptor_features.mutableDescriptorType)` but the new code suggests that this might in fact not be enough).

HansKristian-Work: Changed the checks to use the MUTABLE bindless flag.

doitsujin (approved):

PR #380, opened 2020-11-12
BDA root descriptors

doitsujin:
Changes root descriptors to have their GPU address passed in via push constants or the inline uniform block, rather than push descriptors or a regular descriptor set.

For the shader interface, root descriptor bindings will have `VKD3D_SHADER_BINDING_FLAG_RAW_VA | VKD3D_SHADER_BINDING_FLAG_BUFFER` set.

Passes all non-DXIL tests on RADV and was tested in Resident Evil 2, Control, SotTR.


libs/vkd3d/state.c
```diff
@@ -3541,6 +3541,10 @@ static uint32_t vkd3d_bindless_state_get_bindless_flags(struct d3d12_device *dev
     if (device_info->buffer_device_address_features.bufferDeviceAddress && (flags & VKD3D_BINDLESS_UAV))
         flags |= VKD3D_RAW_VA_UAV_COUNTER;
 
+    if (device_info->buffer_device_address_features.bufferDeviceAddress &&
+            device_info->features2.features.shaderInt64)
+        flags |= VKD3D_RAW_VA_ROOT_DESCRIPTOR;
```
HansKristian-Work, line 3547: We'll want to defer this commit until DXIL support lands.

libs/vkd3d/state.c
```diff
@@ -616,18 +633,23 @@ static HRESULT d3d12_root_signature_init_root_descriptors(struct d3d12_root_sign
...
 
-    if (!(vk_binding_info = vkd3d_malloc(sizeof(*vk_binding_info) * (info->push_descriptor_count + 1))))
-        return E_OUTOFMEMORY;
+    if (info->push_descriptor_count || (root_signature->flags & VKD3D_ROOT_SIGNATURE_USE_INLINE_UNIFORM_BLOCK))
+    {
+        if (!(vk_binding_info = vkd3d_malloc(sizeof(*vk_binding_info) * (info->push_descriptor_count + 1))))
+            return E_OUTOFMEMORY;
+    }
+    else if (!raw_va)
+        return S_OK;
```
HansKristian-Work, line 652: Is vk_set_layout initialized if we actually end up not creating a descriptor set layout here?

libs/vkd3d/command.c
```diff
@@ -3298,31 +3328,43 @@ static void d3d12_command_list_update_root_descriptors(struct d3d12_command_list
...
-                &inline_uniform_block_write, descriptor_set, root_signature, inline_uniform_block_data);
+                &inline_uniform_block_write, descriptor_set, root_signature, &root_parameter_data);
 
         descriptor_write_count += 1;
     }
+    else if (bindings->root_descriptor_dirty_mask)
+    {
+        va_count = d3d12_command_list_fetch_root_descriptor_vas(list, bind_point, &root_parameter_data);
+
+        VK_CALL(vkCmdPushConstants(list->vk_command_buffer,
```
HansKristian-Work, line 3363: Do push constants get pushed with an offset in this case? I didn't see any change for that.

HansKristian-Work (commented):
Comments.

libs/vkd3d/state.c
doitsujin: We zero-initialize the entire root signature struct on creation.

libs/vkd3d/command.c
doitsujin: `d3d12_root_signature_init_push_constants` takes care of the offset:
```c
        root_signature->parameters[i].constant.constant_index = push_constant_range->size / sizeof(uint32_t);
        root_signature->parameters[i].constant.constant_count = p->Constants.Num32BitValues;
```
where `push_constant_range->size` will be the total size of the VA portion for the first set of constants.

HansKristian-Work: Right, looked at the three relevant vkCmdPushConstants, and offsets were already applied.

HansKristian-Work (approved):
LGTM overall. I'll enable this when I get DXIL side implemented.

PR #383, opened 2020-11-13
Enable BDA descriptors in DXIL and globally.

HansKristian-Work:


HansKristian-Work:
Tested HZD, DS and RE2 locally on NV and AMD.

PR #384, opened 2020-11-13
include: Implement 64-bit and pointer atomics 

Joshua-Ashton:


doitsujin (approved):
Looks good to me (albeit a bit ugly with the MSVC hacks), but probably won't hurt if @HansKristian-Work could take a look next week as well.

doitsujin:
Ah, actually - we'll need compare and swap as well, which is currently completely missing.

Joshua-Ashton:
Added compare and swap.

HansKristian-Work (approved):
LGTM.

PR #390, opened 2020-11-15
WIP: Singleton devices

Joshua-Ashton:
D3D12CreateDevice is the same across an entire adapter for its lifespan, this implements that however there is a potential race condition to solve on Release.

We need a better solution for that race condition so I am opening this as a WIP for now.


HansKristian-Work:
Superseded PR.

PR #395, opened 2020-11-16
Singleton devices v2

doitsujin:
Supersedes #390 and should hopefully avoid any sort of race condition.

### DIFF
--- a/include/private/vkd3d_atomic.h
+++ b/include/private/vkd3d_atomic.h
@@ -108,11 +108,12 @@ FORCEINLINE uint32_t vkd3d_atomic_uint32_decrement(uint32_t *target, vkd3d_memor
     return result;
 }
 
-FORCEINLINE uint32_t vkd3d_atomic_uint32_compare_exchange(uint32_t* target, uint32_t expected, uint32_t desired, vkd3d_memory_order order)
+FORCEINLINE uint32_t vkd3d_atomic_uint32_compare_exchange(uint32_t* target, uint32_t expected, uint32_t desired,
+        vkd3d_memory_order success_order, vkd3d_memory_order fail_order)
 {
     uint32_t result;
     /* InterlockedCompareExchange has desired (ExChange) first, then expected (Comperand) */
-    vkd3d_atomic_choose_intrinsic(order, result, InterlockedCompareExchange, /* no suffix */, (LONG*)target, desired, expected);
+    vkd3d_atomic_choose_intrinsic(success_order, result, InterlockedCompareExchange, /* no suffix */, (LONG*)target, desired, expected);
     return result;
 }
 
@@ -156,11 +157,12 @@ FORCEINLINE uint64_t vkd3d_atomic_uint64_decrement(uint64_t *target, vkd3d_memor
     return result;
 }
 
-FORCEINLINE uint64_t vkd3d_atomic_uint64_compare_exchange(uint64_t* target, uint64_t expected, uint64_t desired, vkd3d_memory_order order)
+FORCEINLINE uint64_t vkd3d_atomic_uint64_compare_exchange(uint64_t* target, uint64_t expected, uint64_t desired,
+        vkd3d_memory_order success_order, vkd3d_memory_order fail_order)
 {
     uint64_t result;
     /* InterlockedCompareExchange has desired (ExChange) first, then expected (Comperand) */
-    vkd3d_atomic_choose_intrinsic(order, result, InterlockedCompareExchange, 64, (LONG64*)target, desired, expected);
+    vkd3d_atomic_choose_intrinsic(success_order, result, InterlockedCompareExchange, 64, (LONG64*)target, desired, expected);
     return result;
 }
 
@@ -187,10 +189,11 @@ typedef enum
 # define vkd3d_atomic_uint32_exchange_explicit(target, value, order) vkd3d_atomic_generic_exchange_explicit(target, value, order)
 # define vkd3d_atomic_uint32_increment(target, order)                vkd3d_atomic_generic_increment(target, order)
 # define vkd3d_atomic_uint32_decrement(target, order)                vkd3d_atomic_generic_decrement(target, order)
-FORCEINLINE uint32_t vkd3d_atomic_uint32_compare_exchange(uint32_t* target, uint32_t expected, uint32_t desired, vkd3d_memory_order order)
+static inline uint32_t vkd3d_atomic_uint32_compare_exchange(uint32_t* target, uint32_t expected, uint32_t desired,
+        vkd3d_memory_order success_order, vkd3d_memory_order fail_order)
 {
     /* Expected is written to with the old value in the case that *target != expected */
-    __atomic_compare_exchange_n(target, &expected, desired, 0, order, order);
+    __atomic_compare_exchange_n(target, &expected, desired, 0, success_order, fail_order);
     return expected;
 }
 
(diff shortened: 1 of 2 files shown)


include/private/vkd3d_atomic.h
```diff
@@ -108,7 +108,8 @@ FORCEINLINE uint32_t vkd3d_atomic_uint32_decrement(uint32_t *target, vkd3d_memor
     return result;
 }
 
-FORCEINLINE uint32_t vkd3d_atomic_uint32_compare_exchange(uint32_t* target, uint32_t expected, uint32_t desired, vkd3d_memory_order order)
+FORCEINLINE uint32_t vkd3d_atomic_uint32_compare_exchange(uint32_t* target, uint32_t expected, uint32_t desired,
```
HansKristian-Work, line 111: static?

include/private/vkd3d_atomic.h
```diff
@@ -156,7 +157,8 @@ FORCEINLINE uint64_t vkd3d_atomic_uint64_decrement(uint64_t *target, vkd3d_memor
     return result;
 }
 
-FORCEINLINE uint64_t vkd3d_atomic_uint64_compare_exchange(uint64_t* target, uint64_t expected, uint64_t desired, vkd3d_memory_order order)
+FORCEINLINE uint64_t vkd3d_atomic_uint64_compare_exchange(uint64_t* target, uint64_t expected, uint64_t desired,
```
HansKristian-Work, line 160: static?

libs/vkd3d/device.c
```diff
@@ -2165,6 +2165,65 @@ static void vkd3d_gpu_va_allocator_cleanup(struct vkd3d_gpu_va_allocator *alloca
...
+    struct d3d12_device *device;
+};
+
+static pthread_once_t d3d12_device_map_mutex_init_once = PTHREAD_ONCE_INIT;
+static pthread_mutex_t d3d12_device_map_mutex;
+static struct list d3d12_device_map;
+
+static void d3d12_init_device_map_once(void)
+{
+    pthread_mutex_init(&d3d12_device_map_mutex, NULL);
```
HansKristian-Work, line 2181: PTHREAD_MUTEX_INITIALIZER can be used instead.

HansKristian-Work (commented):
Nits. I wonder if there is a way to only do CAS on device creation rather than every release. ID3D12Device::Release feels like it would be semi-hot.

libs/vkd3d/device.c
```diff
@@ -2234,17 +2293,41 @@ static void d3d12_device_destroy(struct d3d12_device *device)
 static ULONG STDMETHODCALLTYPE d3d12_device_Release(d3d12_device_iface *iface)
 {
     struct d3d12_device *device = impl_from_ID3D12Device(iface);
-    ULONG refcount = InterlockedDecrement(&device->refcount);
+    ULONG cur_refcount = device->refcount;
```
HansKristian-Work, line 2296: This needs to be an (acquire?) atomic load.

include/private/vkd3d_atomic.h
Joshua-Ashton, line 111: This breaks MSVC build, you should pass success_order to where order is in chooseintrinsic

libs/vkd3d/device.c
Joshua-Ashton: I don't think we have something that maps to Win32 threads for that though

include/private/vkd3d_atomic.h
doitsujin: How?

include/private/vkd3d_atomic.h
doitsujin: None of the MSVC-specific functions are `static`. Not sure how useful it is to change it only for these now when it works without it anyway.

Joshua-Ashton:
You also need to update the ``InterlockedCompareExchange64``, etc defines for native builds.

include/private/vkd3d_atomic.h
Joshua-Ashton: FORCEINLINE on Windows is already static.

include/private/vkd3d_atomic.h
Joshua-Ashton: FORCEINLINE on Windows is already static.

libs/vkd3d/device.c
Joshua-Ashton: We don't have an equivalent of that for Win32 threads so it's not mapped in our thread wrapper iirc.

include/private/vkd3d_atomic.h
doitsujin: Ah, yeah, think I get what you mean. Should be fixed, but I can't really test MSVC builds right now.

libs/vkd3d/device.c
doitsujin: Done.

libs/vkd3d/device.c
HansKristian-Work: Ok.

libs/vkd3d/device.c
```diff
@@ -2234,17 +2293,41 @@ static void d3d12_device_destroy(struct d3d12_device *device)
...
+        cur_refcount = cas_refcount;
+
+        if (cur_refcount == 1 && !is_locked)
+        {
+            pthread_mutex_lock(&d3d12_device_map_mutex);
+            is_locked = true;
+        }
+
+        cas_refcount = vkd3d_atomic_uint32_compare_exchange((uint32_t*)&device->refcount, cur_refcount,
+                cur_refcount - 1, vkd3d_memory_order_release, vkd3d_memory_order_relaxed);
```
HansKristian-Work, line 2316: order needs to be acquire_release here, since if we observe refcount of 1, we must be able to see memory written by other threads before they released their writes.

HansKristian-Work (commented):
Tiny nit, otherwise, it seems to behave correctly.

HansKristian-Work (approved):
LGTM.

libs/vkd3d/device.c
DadSchoorse: @Joshua-Ashton There's `SRWLOCK_INIT`, so `#define PTHREAD_MUTEX_INITIALIZER {SRWLOCK_INIT}` should work afaik.

libs/vkd3d/device.c
```diff
@@ -2234,17 +2293,41 @@ static void d3d12_device_destroy(struct d3d12_device *device)
...
+        cur_refcount = cas_refcount;
+
+        if (cur_refcount == 1 && !is_locked)
+        {
+            pthread_mutex_lock(&d3d12_device_map_mutex);
+            is_locked = true;
+        }
+
+        cas_refcount = vkd3d_atomic_uint32_compare_exchange((uint32_t*)&device->refcount, cur_refcount,
+                cur_refcount - 1, vkd3d_memory_order_acquire_release, vkd3d_memory_order_relaxed);
```
Saancreed, line 2316: `vkd3d_memory_order_acquire_release` → `vkd3d_memory_order_acq_rel`, otherwise the build fails with

```
../vkd3d-proton/libs/vkd3d/device.c: In function ‘d3d12_device_Release’:
../vkd3d-proton/libs/vkd3d/device.c:2316:35: error: ‘vkd3d_memory_order_acquire_release’ undeclared (first use in this function); did you mean ‘vkd3d_memory_order_acq_rel’?
 2316 |                 cur_refcount - 1, vkd3d_memory_order_acquire_release, vkd3d_memory_order_relaxed);
      |                                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |                                   vkd3d_memory_order_acq_rel
```

Joshua-Ashton: ^

doitsujin: fixed.

HansKristian-Work:
Compiles fine on MSVC. Gonna go ahead and merge.

Issue #377, opened 2020-11-10
Investigate performance overhead of back to back timeline semaphores.


HansKristian-Work:
Consider emitting a pre-baked command buffer which does the full barrier instead.

PR #397, opened 2020-11-17
vkd3d: Use pipeline barrier command buffers for queue serialization.

HansKristian-Work:
We have observed a lot of large GPU bubbles when using back-to-back
timeline semaphores to synchronize GPU submissions. Use prebaked
pipeline barrier command buffers instead.

To resolve queue sparse serialization, use two binary semaphore pairs to
resolve this. There is no need to use timeline semaphores in this case.

Fix https://github.com/HansKristian-Work/vkd3d-proton/issues/377.


doitsujin (approved):
LGTM.

PR #399, opened 2020-11-17
vkd3d-shader: Correctly handle bit shifts greater than 31 bits.

doitsujin:
This is undefined behaviour in SPIR-V, but well-defined in DXBC, so we should explicitly 'and' the shift amount with 31.


HansKristian-Work (approved):
LGTM.

PR #400, opened 2020-11-17
vkd3d: Support offset buffers for raw/structured texel buffers.

doitsujin:
DXBC only, will need DXIL support.

PR #401, opened 2020-11-18
Add path for raw/structured TBV and fix game bug in Death Stranding.

### LINKED ISSUE #398: Death Stranding: icons turn black when you tilt the map
You can watch it [here](https://youtu.be/qaXvFhQoVo8) (YouTube link).
Log with VKD3D_DEBUG=warn: [DeathStranding-vkd3d.zip](https://github.com/HansKristian-Work/vkd3d-proton/files/5553231/DeathStranding-vkd3d.zip).

Screenshots:

Map - Normal:
map-normal

Map - Tilted:
map-tilted

HansKristian-Work:
- Enables offset buffers when using Raw/Structured buffers as well, this was missing since we normally always used SSBO path.
- Add a VKD3D_CONFIG option to disable bindless SSBO.
- Add a basic mechanism to add game-workarounds, which fiddle around with VKD3D_CONFIG flags. It is also possible to skip those workarounds with another flag.

Fix #398.
Supersedes #400.

Tested Death Stranding on NV, and test suite passes, except ByteAddressBuffer OOB tests, which are expected to fail since we do not deal with per-element robustness anymore. Normally 16-byte SSBO robustness would handle this case.


doitsujin (approved):
LGTM.

PR #405, opened 2020-11-18
Rework query initialization

doitsujin:
Extends the initial resource transition framework to allow for query heap initialization, which is necessary because D3D12 queries are immediately considered available. Fixes `test_resolve_query_data_in_reordered_command_list`, but as a compromise, returned values for uninitialized timestamp queries are non-zero, thus breaking another test. `todo`s have been changed accordingly.

This also greatly simplifies our impementation of `ResolveQueryData`.


HansKristian-Work (approved):
LGTM.

PR #413, opened 2020-11-19
Binary occlusion query fixup

doitsujin:
Replaces non-zero sample counts with 1 in order to match D3D12 behaviour.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -6296,6 +6296,75 @@ static size_t get_query_stride(D3D12_QUERY_TYPE type)
     return sizeof(uint64_t);
 }
 
+static void d3d12_command_list_resolve_binary_occlusion_queries(struct d3d12_command_list *list,
+        struct d3d12_resource *buffer, UINT64 offset, UINT query_count)
+{
+    const struct vkd3d_query_ops *query_ops = &list->device->meta_ops.query;
+    const struct vkd3d_vk_device_procs *vk_procs = &list->device->vk_procs;
+    VkDeviceSize alignment, aligned_offset, aligned_end;
+    VkDescriptorBufferInfo buffer_info;
+    struct vkd3d_query_op_args args;
+    VkWriteDescriptorSet vk_write;
+    unsigned int num_workgroups;
+    VkMemoryBarrier vk_barrier;
+    VkDescriptorSet vk_set;
+
+    d3d12_command_list_invalidate_current_pipeline(list, true);
+    d3d12_command_list_invalidate_root_parameters(list, VK_PIPELINE_BIND_POINT_COMPUTE, true);
+
+    alignment = list->device->device_info.properties2.properties.limits.minStorageBufferOffsetAlignment;
+    aligned_end = align(offset + query_count * sizeof(uint64_t), alignment);
+    aligned_offset = offset & ~(alignment - 1);
+
+    vk_set = d3d12_command_allocator_allocate_descriptor_set(list->allocator,
+            query_ops->vk_set_layout, VKD3D_DESCRIPTOR_POOL_TYPE_STATIC);
+
+    /* TODO suppprt some sort of scratch buffer so that
+     * we don't do write->read->write on host memory */
+    buffer_info.buffer = buffer->vk_buffer;
+    buffer_info.offset = buffer->heap_offset + aligned_offset;
+    buffer_info.range = aligned_end - aligned_offset;
+
+    vk_write.sType = VK_STRUCTURE_TYPE_WRITE_DESCRIPTOR_SET;
+    vk_write.pNext = NULL;
+    vk_write.dstSet = vk_set;
+    vk_write.dstBinding = 0;
+    vk_write.dstArrayElement = 0;
+    vk_write.descriptorCount = 1;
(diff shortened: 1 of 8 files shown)


libs/vkd3d/resource.c
```diff
@@ -1009,7 +1009,10 @@ HRESULT vkd3d_create_buffer(struct d3d12_device *device,
     if (heap_type == D3D12_HEAP_TYPE_UPLOAD)
         buffer_info.usage &= ~VK_BUFFER_USAGE_TRANSFER_DST_BIT;
     else if (heap_type == D3D12_HEAP_TYPE_READBACK)
-        buffer_info.usage = VK_BUFFER_USAGE_TRANSFER_DST_BIT;
+    {
+        buffer_info.usage = VK_BUFFER_USAGE_TRANSFER_DST_BIT
+                | VK_BUFFER_USAGE_STORAGE_BUFFER_BIT;
```
HansKristian-Work, line 1014: This formatting looks goofy, should probably just be on one line.

libs/vkd3d/command.c
```diff
@@ -6299,6 +6299,67 @@ static size_t get_query_stride(D3D12_QUERY_TYPE type)
...
+    VkDescriptorSet vk_set;
+
+    d3d12_command_list_invalidate_current_pipeline(list, true);
+    d3d12_command_list_invalidate_root_parameters(list, VK_PIPELINE_BIND_POINT_COMPUTE, true);
+
+    vk_set = d3d12_command_allocator_allocate_descriptor_set(list->allocator,
+            query_ops->vk_set_layout, VKD3D_DESCRIPTOR_POOL_TYPE_STATIC);
+
+    buffer_info.buffer = buffer->vk_buffer;
+    buffer_info.offset = buffer->heap_offset + offset;
```
HansKristian-Work, line 6321: Is alignment a concern here?

HansKristian-Work (commented):
Comments.

libs/vkd3d/command.c
```diff
@@ -6313,29 +6374,22 @@ static void STDMETHODCALLTYPE d3d12_command_list_ResolveQueryData(d3d12_command_
...
     }
 
+    d3d12_command_list_track_query_heap(list, query_heap);
     d3d12_command_list_end_current_render_pass(list, true);
 
     VK_CALL(vkCmdCopyQueryPoolResults(list->vk_command_buffer, query_heap->vk_query_pool,
             start_index, query_count, buffer->vk_buffer, buffer->heap_offset + aligned_dst_buffer_offset,
             get_query_stride(type), VK_QUERY_RESULT_64_BIT | VK_QUERY_RESULT_WAIT_BIT));
+
+    if (type == D3D12_QUERY_TYPE_BINARY_OCCLUSION)
```
HansKristian-Work, line 6390: Do we expect applications to not batch here? Resolving one query at a time sounds like it would get "fun".

libs/vkd3d/command.c
```diff
@@ -6313,29 +6374,22 @@ static void STDMETHODCALLTYPE d3d12_command_list_ResolveQueryData(d3d12_command_
...
     if (!d3d12_resource_is_buffer(buffer))
     {
         WARN("Destination resource is not a buffer.\n");
         return;
     }
 
+    d3d12_command_list_track_query_heap(list, query_heap);
     d3d12_command_list_end_current_render_pass(list, true);
 
     VK_CALL(vkCmdCopyQueryPoolResults(list->vk_command_buffer, query_heap->vk_query_pool,
```
HansKristian-Work, line 6386: Couldn't we resolve to a scratch buffer owned by the command allocator here to avoid ping-pong to host memory?

libs/vkd3d/command.c
doitsujin: Apps should batch if they aren't completely braindead. The D3D12 API kind of implies that this is implemented as some compute shader dispatch.

libs/vkd3d/command.c
doitsujin: We could, but I don't think we currently have the infastructure for that in place.

libs/vkd3d/command.c
doitsujin: Right; we can only guarantee 8 bytes here. Guess we need to pass in an offset via push constants as well.

libs/vkd3d/command.c
HansKristian-Work: Ok.

libs/vkd3d/command.c
HansKristian-Work: Could make it a TODO if some app happens to hammer this.

libs/vkd3d/resource.c
HansKristian-Work: Should mark this with a comment saying why it's used so we can remember to remove it in the future if we implement scratch buffer resolve.

HansKristian-Work: Nvm, we'll need SSBO usage anyways ...

libs/vkd3d/command.c
doitsujin: done.

libs/vkd3d/command.c

HansKristian-Work (approved):
LGTM.

PR #415, opened 2020-11-23
vkd3d-shader: Fix infinity handling in f32tof16.

doitsujin:
This came up in DXVK recently. The problem is that some Vulkan drivers (RADV+ACO, AMDVLK, Nvidia on *some* architectures) will return infinity when passing a number larger than `FLT16_MAX`, whereas DXBC requires us to return `FLT16_MAX` instead, so we need to manually fix that up in our shaders.


HansKristian-Work (approved):
LGTM.

PR #416, opened 2020-11-23
Be robust when faced with illegal VBO strides and offsets.

HansKristian-Work:


HansKristian-Work:
Tested on HZD, DS, RE2 and test suite.

libs/vkd3d/state.c
```diff
@@ -2802,6 +2806,23 @@ static HRESULT d3d12_pipeline_state_init_graphics(struct d3d12_pipeline_state *s
     graphics->vertex_buffer_mask = mask;
     vkd3d_shader_free_shader_signature(&input_signature);
 
+    for (i = 0; i < D3D12_IA_VERTEX_INPUT_RESOURCE_SLOT_COUNT; i++)
+    {
+        if (mask & (1u << i))
+        {
+            /* In D3D12, VBO strides must be aligned to min(4, max(sizeof(format))) for a given input slot.
```
doitsujin, line 2813: `max(sizeof(format))`?

libs/vkd3d/state.c
```diff
@@ -2802,6 +2806,23 @@ static HRESULT d3d12_pipeline_state_init_graphics(struct d3d12_pipeline_state *s
...
+             * Native drivers and Windows Vulkan drivers happen to be robust against buggy applications
+             * which use unaligned formats, but Vulkan drivers are not always robust here (they don't have to be).
+             * AMD Windows D3D12/Vulkan drivers return the wrong result for unaligned VBO strides,
+             * but it doesn't crash at least!
+             * RDNA will hang if we emit tbuffer_loads which don't conform to D3D12 rules. */
+
+            /* All VBO types in D3D12 are POT sized, so this works. */
```
doitsujin, line 2820: `R32G32B32_FLOAT` is not POT (although four bytes per channel, so I guess it doesn't matter here?)

libs/vkd3d/state.c
```diff
@@ -3287,6 +3308,11 @@ static bool d3d12_pipeline_state_can_use_dynamic_stride(struct d3d12_pipeline_st
                   graphics->minimum_vertex_buffer_dynamic_stride[slot]);
             return false;
         }
+        else if (dyn_state->vertex_strides[slot] & graphics->vertex_buffer_stride_align_mask[slot])
```
doitsujin, line 3311: Would we not erroneously hit this case if e.g. `vertex_buffer_stride_align_mask` is 4 and stride is 12?

doitsujin: nvm, we always hit `graphics->vertex_buffer_stride_align_mask[i]--;` for every vertex buffer.

libs/vkd3d/state.c
```diff
@@ -3287,6 +3308,11 @@ static bool d3d12_pipeline_state_can_use_dynamic_stride(struct d3d12_pipeline_st
...
+        {
+            TRACE("Stride alignment does not conform to D3D12. Falling back to static pipeline.\n");
+            return false;
```
doitsujin, line 3314: can't we just fix strides at bind time rather than compiling an entirely new pipeline?

HansKristian-Work: I guess we would modify the strides like we modify VBO offsets.

doitsujin: Yeh, I'd prefer doing that instead since there doesn't really seem to be any need to make the strides static.

libs/vkd3d/state.c
HansKristian-Work: Clarified comment.

libs/vkd3d/state.c
HansKristian-Work: Clarified in comment.

libs/vkd3d/state.c

doitsujin (approved):
If this at least remotely resembles what D3D12 drivers do in this case - LGTM.

HansKristian-Work:
Yes, AMD D3D12 and Vulkan drivers on Windows do not "correctly" render this case, but they do not crash the GPU, which is the significant factor here. It is still against D3D12 validation layers to misalign.

PR #417, opened 2020-11-24
Use scratch buffer for binary query fixup

doitsujin:
Paves the way for a potential predication rework, and potentially improves performance in case the app directly resolves queries into host memory.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -6367,46 +6367,16 @@ static size_t get_query_stride(D3D12_QUERY_TYPE type)
 }
 
 static void d3d12_command_list_resolve_binary_occlusion_queries(struct d3d12_command_list *list,
-        struct d3d12_resource *buffer, UINT64 offset, UINT query_count)
+        VkDeviceAddress src_va, VkDeviceAddress dst_va, UINT query_count)
 {
     const struct vkd3d_query_ops *query_ops = &list->device->meta_ops.query;
     const struct vkd3d_vk_device_procs *vk_procs = &list->device->vk_procs;
-    VkDeviceSize alignment, aligned_offset, aligned_end;
-    VkDescriptorBufferInfo buffer_info;
     struct vkd3d_query_op_args args;
-    VkWriteDescriptorSet vk_write;
     unsigned int num_workgroups;
     VkMemoryBarrier vk_barrier;
-    VkDescriptorSet vk_set;
 
     d3d12_command_list_invalidate_current_pipeline(list, true);
     d3d12_command_list_invalidate_root_parameters(list, VK_PIPELINE_BIND_POINT_COMPUTE, true);
-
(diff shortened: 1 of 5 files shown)


libs/vkd3d/shaders/cs_resolve_binary_queries.comp
```diff
@@ -1,26 +1,30 @@
...
+};
 
 layout(push_constant)
 uniform u_info_t {
-  uint query_offset;
+  query_buffer_t src_queries;
+  query_buffer_t dst_queries;
   uint query_count;
-} u_info;
+};
```
HansKristian-Work, line 17: Sure this compiles? push_constant must have an instance name.

libs/vkd3d/command.c
```diff
@@ -6455,16 +6427,28 @@ static void STDMETHODCALLTYPE d3d12_command_list_ResolveQueryData(d3d12_command_
...
-            get_query_stride(type), VK_QUERY_RESULT_64_BIT | VK_QUERY_RESULT_WAIT_BIT));
+            start_index, query_count, scratch.buffer, scratch.offset, stride,
+            VK_QUERY_RESULT_64_BIT | VK_QUERY_RESULT_WAIT_BIT));
 
-    if (type == D3D12_QUERY_TYPE_BINARY_OCCLUSION)
-        d3d12_command_list_resolve_binary_occlusion_queries(list,
-                buffer, aligned_dst_buffer_offset, query_count);
+    if (do_fixup)
+    {
+        d3d12_command_list_resolve_binary_occlusion_queries(list, scratch.va,
```
HansKristian-Work, line 6449: Probably a FIXME_ONCE here if device doesn't support BDA.

HansKristian-Work (commented):
Looks pretty clean, just some nits.

libs/vkd3d/shaders/cs_resolve_binary_queries.comp
doitsujin: glslang isn't complaining. Where does it say that push_constant needs a name (and what's the reason for that anyway, given that ubo doesn't need one)?

HansKristian-Work: hm ... ye, Vulkan GLSL spec says it's optional, I can swear it used to be required ... Anyways, guess this is a non-issue then if it works in glslang.

doitsujin: Ah, actually this requirement was removed three years ago, see https://github.com/KhronosGroup/Vulkan-Docs/issues/428.

HansKristian-Work: Ah, ok, so I'm not insane after all :D

libs/vkd3d/command.c
doitsujin: added one.

HansKristian-Work (approved):
LGTM.

PR #424, opened 2020-11-26
Rework predication

doitsujin:
This essentially solves three problems:
- D3D12 requires 64-bit predicates, whereas `VK_EXT_conditional_rendering` uses 32-bit conditions. This adds a resolve pass which emits a 32-bit predicate to a scratch buffer.
- Using the scratch buffer also means that changing the predicate value after enabling predication can no longer change the behaviour of predicated draws, as is required by D3D12. This seems to just work™ on Vulkan drivers, but should be undefined behaviour.
- Official AMD drivers do not support `VK_EXT_conditional_rendering`. To support predication on those drivers, we use indirect draws and dispatches (with indirect count for predicated indirect draws) instead.

Was a bit of a failed attempt to get AC:Valhalla to run on AMDVLK, which requires Predication support, unfortunately the game still hangs there.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -2536,7 +2536,8 @@ static VkAccessFlags vk_access_flags_all_possible_for_buffer(const struct d3d12_
             access |= VK_ACCESS_TRANSFORM_FEEDBACK_COUNTER_READ_BIT_EXT;
     }
 
-    if (consider_reads && (vk_queue_flags & VK_QUEUE_GRAPHICS_BIT) && device->vk_info.EXT_conditional_rendering)
+    if (consider_reads && (vk_queue_flags & VK_QUEUE_GRAPHICS_BIT) &&
+            device->device_info.conditional_rendering_features.conditionalRendering)
         access |= VK_ACCESS_CONDITIONAL_RENDERING_READ_BIT_EXT;
 
     if (consider_reads && (vk_queue_flags & VK_QUEUE_GRAPHICS_BIT))
@@ -2662,10 +2663,16 @@ static void vk_access_and_stage_flags_from_d3d12_resource_state(const struct d3d
                 *stages |= VK_PIPELINE_STAGE_DRAW_INDIRECT_BIT;
                 *access |= VK_ACCESS_INDIRECT_COMMAND_READ_BIT;
 
-                if (device->vk_info.EXT_conditional_rendering)
+                /* D3D12_RESOURCE_STATE_PREDICATION */
+                if (device->device_info.buffer_device_address_features.bufferDeviceAddress)
                 {
-                    *stages |= VK_PIPELINE_STAGE_CONDITIONAL_RENDERING_BIT_EXT;
-                    *access |= VK_ACCESS_CONDITIONAL_RENDERING_READ_BIT_EXT;
+                    *stages |= VK_PIPELINE_STAGE_COMPUTE_SHADER_BIT;
+                    *access |= VK_ACCESS_SHADER_READ_BIT;
+                }
+                else
+                {
+                    *stages |= VK_PIPELINE_STAGE_TRANSFER_BIT;
+                    *access |= VK_ACCESS_TRANSFER_READ_BIT;
                 }
                 break;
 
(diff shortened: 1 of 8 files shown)


libs/vkd3d/command.c
```diff
@@ -2662,10 +2663,16 @@ static void vk_access_and_stage_flags_from_d3d12_resource_state(const struct d3d
                 *stages |= VK_PIPELINE_STAGE_DRAW_INDIRECT_BIT;
                 *access |= VK_ACCESS_INDIRECT_COMMAND_READ_BIT;
 
-                if (device->vk_info.EXT_conditional_rendering)
+                /* D3D12_RESOURCE_STATE_PREDICATION */
```
HansKristian-Work, line 2666: Do we never use actual VK_PIPELINE_STAGE_CONDITIONAL_RENDERING_EXT?

libs/vkd3d/command.c
```diff
@@ -6557,69 +6653,113 @@ static void STDMETHODCALLTYPE d3d12_command_list_SetPredication(d3d12_command_li
...
     TRACE("iface %p, buffer %p, aligned_buffer_offset %#"PRIx64", operation %#x.\n",
             iface, buffer, aligned_buffer_offset, operation);
 
-    if (!vk_info->EXT_conditional_rendering)
+    d3d12_command_list_end_current_render_pass(list, true);
+
+    if (resource && (aligned_buffer_offset & 0x7))
+        return;
+
+    if (!list->device->device_info.buffer_device_address_features.bufferDeviceAddress &&
```
HansKristian-Work, line 6674: Shouldn't this be ||?

libs/vkd3d/command.c
```diff
@@ -6557,69 +6653,113 @@ static void STDMETHODCALLTYPE d3d12_command_list_SetPredication(d3d12_command_li
...
-        cond_info.sType = VK_STRUCTURE_TYPE_CONDITIONAL_RENDERING_BEGIN_INFO_EXT;
-        cond_info.pNext = NULL;
-        cond_info.buffer = resource->vk_buffer;
-        cond_info.offset = resource->heap_offset + aligned_buffer_offset;
-        switch (operation)
-        {
-            case D3D12_PREDICATION_OP_EQUAL_ZERO:
-                cond_info.flags = 0;
-                break;
+            VK_CALL(vkCmdCopyBuffer(list->vk_command_buffer,
```
HansKristian-Work, line 6725: Do we need a barrier before the copy?

libs/vkd3d/command.c
HansKristian-Work: Or hm, && is probably correct since there are multiple ways to deal with this.

libs/vkd3d/command.c
```diff
@@ -2662,10 +2663,16 @@ static void vk_access_and_stage_flags_from_d3d12_resource_state(const struct d3d
...
+                if (device->device_info.buffer_device_address_features.bufferDeviceAddress)
                 {
-                    *stages |= VK_PIPELINE_STAGE_CONDITIONAL_RENDERING_BIT_EXT;
-                    *access |= VK_ACCESS_CONDITIONAL_RENDERING_READ_BIT_EXT;
+                    *stages |= VK_PIPELINE_STAGE_COMPUTE_SHADER_BIT;
+                    *access |= VK_ACCESS_SHADER_READ_BIT;
+                }
+                else
+                {
+                    *stages |= VK_PIPELINE_STAGE_TRANSFER_BIT;
```
HansKristian-Work, line 2674: I see, this is the fallback when we copy the lower 32-bit predicate?

HansKristian-Work (approved):
LGTM. If none of my comments found a bug, please go ahead and merge.

libs/vkd3d/command.c
doitsujin: We do after we copy the predicate to its scratch memory; there's a separate barrier to take care of that.

libs/vkd3d/command.c
doitsujin: Yes, we only want to error out if *both* features are unsupported. We only require BDA for the fallback path, and for 64-bit predicates to work correctly.

HansKristian-Work: Well then.

libs/vkd3d/command.c
HansKristian-Work: That seems to be taken care of with the resource state barrier.

doitsujin: The app should be transitioning the buffer to the predication state beforehand, so we shouldn't need it.

libs/vkd3d/command.c
doitsujin: Yes.

Issue #419, opened 2020-11-24
Fix 64-bit modifiers test


hakzsam:
Invalid SPIR-V found for ps_dmodifier.

error: line 62: GLSL.std.450 NClamp: expected types of all operands to be equal to Result Type
  %42 = OpExtInst %double %1 NClamp %41 %float_0 %float_1

See test_shader_instructions().

Joshua-Ashton:
Thanks for catching -- will look into this after I sleep. :)

PR #425, opened 2020-11-27
vkd3d-shader: Fix saturates of fp64 types

Joshua-Ashton:
Closes: #419

### DIFF
--- a/libs/vkd3d-shader/spirv.c
+++ b/libs/vkd3d-shader/spirv.c
@@ -3715,15 +3715,21 @@ static uint32_t vkd3d_dxbc_compiler_emit_sat(struct vkd3d_dxbc_compiler *compile
 {
     enum vkd3d_component_type component_type = vkd3d_component_type_from_data_type(reg->data_type);
     unsigned int component_count = vkd3d_write_mask_component_count_typed(write_mask, component_type);
+    uint32_t type_id = vkd3d_dxbc_compiler_get_type_id_for_reg(compiler, reg, write_mask);
     struct vkd3d_spirv_builder *builder = &compiler->spirv_builder;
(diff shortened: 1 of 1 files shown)


libs/vkd3d-shader/spirv.c
```diff
@@ -3716,14 +3716,24 @@ static uint32_t vkd3d_dxbc_compiler_emit_sat(struct vkd3d_dxbc_compiler *compile
...
-    zero_id = vkd3d_dxbc_compiler_get_constant_float_vector(compiler, 0.0f, component_count);
-    one_id = vkd3d_dxbc_compiler_get_constant_float_vector(compiler, 1.0f, component_count);
+    uint32_t type_id;
 
     type_id = vkd3d_dxbc_compiler_get_type_id_for_reg(compiler, reg, write_mask);
-    if (reg->data_type == VKD3D_DATA_FLOAT || reg->data_type == VKD3D_DATA_DOUBLE)
+
+    if (component_type == VKD3D_TYPE_DOUBLE)
+    {
+        uint32_t zero_id, one_id;
```
doitsujin, line 3725: any reason why we redeclare this in both branches?

Joshua-Ashton: Just figured it'd be cleaner that way given it's only used within that scope.

doitsujin: just leave the original declaration in place tbh, we do this in *some* cases but usually only for constants and if there's a lot of stuff that really is only needed on one side.

Or just write it as `uint32_t zero_id = ...`.

Joshua-Ashton: Resolved with the latter.

PR #432, opened 2020-12-02
Batch vkCmdResetQueryPool calls in separate command buffer

doitsujin:
Improves performance in Ghostrunner a bit (~47 -> 50 FPS in the very beginning of the game on my Polaris).
The next step would be to begin/end occlusion queries without interrupting the current render pass.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -2844,6 +2844,7 @@ static ULONG STDMETHODCALLTYPE d3d12_command_list_Release(d3d12_command_list_ifa
             d3d12_command_allocator_free_command_buffer(list->allocator, list);
 
         vkd3d_free(list->init_transitions);
+        vkd3d_free(list->query_ranges);
         vkd3d_free(list);
 
         d3d12_device_release(device);
@@ -2955,6 +2956,114 @@ static HRESULT STDMETHODCALLTYPE d3d12_command_list_Close(d3d12_command_list_ifa
     return S_OK;
 }
 
+static bool d3d12_command_list_find_query(struct d3d12_command_list *list,
+        VkQueryPool vk_pool, uint32_t index, size_t *out_pos)
+{
+    const struct vkd3d_query_range *range;
+    size_t hi = list->query_ranges_count;
+    size_t lo = 0;
+
+    while (lo < hi)
+    {
+        size_t pos = lo + (hi - lo) / 2;
+        range = &list->query_ranges[pos];
+
+        if (vk_pool < range->vk_pool)
+            hi = pos;
+        else if (vk_pool > range->vk_pool)
+            lo = pos + 1;
+        else if (index < range->index)
+            hi = pos;
+        else if (index >= range->index + range->count)
+            lo = pos + 1;
+        else
+        {
+            if (out_pos)
+                *out_pos = pos;
+            return true;
+        }
+    }
+
+    if (out_pos)
+        *out_pos = lo;
+    return false;
+}
+
+static void d3d12_command_list_insert_query_range(struct d3d12_command_list *list, size_t pos,
+        VkQueryPool vk_pool, uint32_t index, uint32_t count, uint32_t flags)
+{
+    struct vkd3d_query_range *range;
+
+    vkd3d_array_reserve((void**)&list->query_ranges, &list->query_ranges_size,
+            list->query_ranges_count + 1, sizeof(*list->query_ranges));
+
+    for (size_t i = list->query_ranges_count; i > pos; i--)
+        list->query_ranges[i] = list->query_ranges[i - 1];
+
+    range = &list->query_ranges[pos];
+    range->vk_pool = vk_pool;
+    range->index = index;
+    range->count = count;
+    range->flags = flags;
+
+    list->query_ranges_count++;
+}
+
+static void d3d12_command_list_read_query_range(struct d3d12_command_list *list,
+        VkQueryPool vk_pool, uint32_t index, uint32_t count)
+{
+    const struct vkd3d_query_range *range;
+    size_t hi = index + count;
+    size_t lo = index;
+    size_t pos;
+
+    /* pos contains either the location of an existing range
+     * containing the first query of the new range, or the
+     * location where we need to insert it */
+    d3d12_command_list_find_query(list, vk_pool, index, &pos);
+
+    /* Do not attempt to merge adjacent ranges, but make sure
+     * that each query is only contained in one range */
+    while (lo < hi)
+    {
+        if (pos < list->query_ranges_count)
+        {
+            range = &list->query_ranges[pos];
+
+            if (lo >= range->index)
+                lo = max(lo, range->index + range->count);
+            else
+            {
+                size_t range_end = min(hi, range->index);
+                d3d12_command_list_insert_query_range(list, pos,
+                        vk_pool, lo, range_end - lo, 0);
+                lo = range_end;
+            }
+
+            pos += 1;
+        }
+        else
+        {
+            d3d12_command_list_insert_query_range(list, pos,
+                    vk_pool, lo, hi - lo, 0);
+            lo = hi;
+        }
+    }
+}
+
+static bool d3d12_command_list_reset_query(struct d3d12_command_list *list,
+        VkQueryPool vk_pool, uint32_t index)
+{
+    size_t pos;
+
+    if (d3d12_command_list_find_query(list, vk_pool, index, &pos))
+        return false;
+
+    d3d12_command_list_insert_query_range(list, pos,
+            vk_pool, index, 1, VKD3D_QUERY_RANGE_RESET);
+    return true;
+}
+
 static void d3d12_command_list_reset_state(struct d3d12_command_list *list,
         ID3D12PipelineState *initial_pipeline_state)
 {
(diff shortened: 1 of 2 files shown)


libs/vkd3d/command.c
```diff
@@ -2955,6 +2956,114 @@ static HRESULT STDMETHODCALLTYPE d3d12_command_list_Close(d3d12_command_list_ifa
...
+
+static void d3d12_command_list_insert_query_range(struct d3d12_command_list *list, size_t pos,
+        VkQueryPool vk_pool, uint32_t index, uint32_t count, uint32_t flags)
+{
+    struct vkd3d_query_range *range;
+
+    vkd3d_array_reserve((void**)&list->query_ranges, &list->query_ranges_size,
+            list->query_ranges_count + 1, sizeof(*list->query_ranges));
+
+    for (size_t i = list->query_ranges_count; i > pos; i--)
```
HansKristian-Work, line 3118: memmove?

libs/vkd3d/command.c
```diff
@@ -2955,6 +2956,114 @@ static HRESULT STDMETHODCALLTYPE d3d12_command_list_Close(d3d12_command_list_ifa
     return S_OK;
 }
 
+static bool d3d12_command_list_find_query(struct d3d12_command_list *list,
+        VkQueryPool vk_pool, uint32_t index, size_t *out_pos)
+{
+    const struct vkd3d_query_range *range;
```
HansKristian-Work, line 3080: What is the motivation to roll a custom binary search here? The fallback VA allocator uses bsearch(). Although, since this is kinda perf critical, I guess we can justify not using C stdlib (which cannot inline the comparator).

libs/vkd3d/vkd3d_private.h
```diff
@@ -1335,6 +1348,10 @@ struct d3d12_command_list
     size_t init_transitions_size;
     size_t init_transitions_count;
 
+    struct vkd3d_query_range *query_ranges;
```
HansKristian-Work, line 1352: Are these zero-initialized on creation?

libs/vkd3d/command.c
```diff
@@ -2955,6 +2956,114 @@ static HRESULT STDMETHODCALLTYPE d3d12_command_list_Close(d3d12_command_list_ifa
...
+    size_t pos;
+
+    /* pos contains either the location of an existing range
+     * containing the first query of the new range, or the
+     * location where we need to insert it */
+    d3d12_command_list_find_query(list, vk_pool, index, &pos);
+
+    /* Do not attempt to merge adjacent ranges, but make sure
+     * that each query is only contained in one range */
+    while (lo < hi)
```
HansKristian-Work, line 3145: I find this algorithm hard to understand by reading it. Would be helpful with an example here to explain in comment which cases it handles. From what I understand it attempts to split or merge ranges?

libs/vkd3d/command.c
```diff
@@ -2955,6 +2956,114 @@ static HRESULT STDMETHODCALLTYPE d3d12_command_list_Close(d3d12_command_list_ifa
...
+        {
+            range = &list->query_ranges[pos];
+
+            if (lo >= range->index)
+                lo = max(lo, range->index + range->count);
+            else
+            {
+                size_t range_end = min(hi, range->index);
+                d3d12_command_list_insert_query_range(list, pos,
+                        vk_pool, lo, range_end - lo, 0);
```
HansKristian-Work, line 3157: Why is flag 0 here and not RESET?

libs/vkd3d/command.c
```diff
@@ -1139,27 +1140,79 @@ static HRESULT d3d12_command_allocator_allocate_command_buffer(struct d3d12_comm
...
+    return S_OK;
+}
+
+static void d3d12_command_allocator_free_vk_command_buffer(struct d3d12_command_allocator *allocator,
+        VkCommandBuffer vk_command_buffer)
+{
+    struct d3d12_device *device = allocator->device;
+    const struct vkd3d_vk_device_procs *vk_procs = &device->vk_procs;
+
+    if (!(vk_command_buffer))
```
HansKristian-Work, line 1192: Weird parens.

libs/vkd3d/command.c
```diff
@@ -1139,27 +1140,79 @@ static HRESULT d3d12_command_allocator_allocate_command_buffer(struct d3d12_comm
...
+    if (list->vk_init_commands)
+        return S_OK;
+
+    command_buffer_info.sType = VK_STRUCTURE_TYPE_COMMAND_BUFFER_ALLOCATE_INFO;
+    command_buffer_info.pNext = NULL;
+    command_buffer_info.commandPool = allocator->vk_command_pool;
+    command_buffer_info.level = VK_COMMAND_BUFFER_LEVEL_PRIMARY;
+    command_buffer_info.commandBufferCount = 1;
+
+    if ((vr = VK_CALL(vkAllocateCommandBuffers(device->vk_device, &command_buffer_info,
```
HansKristian-Work, line 1163: Can't we just recycle a VkCommandBuffer from the allocator just like normal command lists?

libs/vkd3d/command.c
```diff
@@ -2955,6 +2956,114 @@ static HRESULT STDMETHODCALLTYPE d3d12_command_list_Close(d3d12_command_list_ifa
...
+
+static bool d3d12_command_list_reset_query(struct d3d12_command_list *list,
+        VkQueryPool vk_pool, uint32_t index)
+{
+    size_t pos;
+
+    if (d3d12_command_list_find_query(list, vk_pool, index, &pos))
+        return false;
+
+    d3d12_command_list_insert_query_range(list, pos,
```
HansKristian-Work, line 3180: What happens here if we read query index 2, and then later reset range [0, 4]. We would need to split this into three ranges. [0, 1] reset, 2 read, [3, 4] reset.

HansKristian-Work (changes requested):
Comments.

libs/vkd3d/command.c
doitsujin: We do not want to reset queries prior to `vkCmdCopyQueryPoolResults`.

libs/vkd3d/command.c
doitsujin: We can only ever reset one query at a time (since this happens in `{Begin|End}Query`, so we can never override read queries.

libs/vkd3d/vkd3d_private.h
doitsujin: Should be since we memset the entire struct at some point.

libs/vkd3d/command.c
doitsujin: If the query in question is not included in any range, bsearch would return `NULL`, but we need to know where to insert a new range in that case.

libs/vkd3d/command.c
doitsujin: I can add a comment. The goal here is to split the newly inserted range so that we never override any previously inserted ranges (important because the *first* use of a query decides whether or not we can reset it).

libs/vkd3d/command.c
doitsujin: Unless I misunderstand what you mean, we do not have a code path for that. `d3d12_command_allocator_allocate_command_buffer` also unconditionally calls `vkAllocateCommandBuffers`.

doitsujin: To be clear, there's no recycling going on at all. `d3d12_command_allocator_Reset` calls `vkFreeCommandBuffers`.

libs/vkd3d/command.c
HansKristian-Work: Right. I was wondering if it made sense to merge query ranges, say that the app is doing query 0, 1, 2, 3, ... in order. Ideally, we should be able to just bump the range instead of ending up with lots of different ranges, which doesn't seem ideal when we have a binary search in the hot path.

libs/vkd3d/command.c
HansKristian-Work: Ah, I was confused since there is allocator->command_buffers[] which is appended to. I see now it's purely there to vkFreeCommandBuffers on pool reset, which is a little silly. That is orthogonal to this PR, but seems useful to recycle command buffer handles if we can later at some point ...

libs/vkd3d/command.c
HansKristian-Work: Fair enough.

libs/vkd3d/command.c
HansKristian-Work: Right, I see now, because this only called when reading from a query, never reset ...

libs/vkd3d/command.c
doitsujin: I deliberately left that out since it would make the insertion code completely insane, since now we'd have to handle the case where inserting a range does not insert a new item, or even removes an item because it merges three adjacent ranges into one, and adjust our `pos` cursor in some way. I'm having a very hard time reasoning about all the special cases.

Binary search should never have more than 10-12 iterations per lookup anyway, and that is if the game uses an extreme number of queries.

I guess the bigger problem is that insertion is `O(n^2)`

HansKristian-Work: Ok, if it makes the code insane guess we can leave it on the table until it becomes a real problem.

HansKristian-Work (approved):
Ok, I grok the implementation now. LGTM, albeit complicated (but I don't think it can be done easier).

PR #435, opened 2020-12-03
vkd3d: Do not interrupt render pass for timestamp queries.

doitsujin:
Timestamp queries are not scoped, so ending the render pass is only necessary if we need to reset the query late.


HansKristian-Work (approved):
LGTM.

PR #438, opened 2020-12-04
vkd3d: Do not interrupt render pass for occlusion queries.

doitsujin:
Further increases performance in Ghostrunner. This works by deferring `vkCmdBeginQuery` to draw time, and ending these queries 

We should probably do a lot of real-world testing before merging this and see if we ever hit the `FIXME`, since the behaviour in that case is technically incorrect. We can't easily virtualize queries like DXVK does since D3D12 queries work quite similarly to Vulkan query pools; it's not completely impossible to do some local remapping for any command list that runs into this problem and allocate multiple queries per d3d12 query, but doing so would make resolving queries significantly harder.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -2488,9 +2488,95 @@ static void d3d12_command_list_emit_render_pass_transition(struct d3d12_command_
         j, vk_image_barriers));
 }
 
+static void d3d12_command_list_begin_active_query(struct d3d12_command_list *list,
+        struct vkd3d_active_query *query)
+{
+    const struct vkd3d_vk_device_procs *vk_procs = &list->device->vk_procs;
+    assert(query->state == VKD3D_ACTIVE_QUERY_RESET);
+
+    VK_CALL(vkCmdBeginQuery(list->vk_command_buffer, query->vk_pool, query->index, query->flags));
+    query->state = VKD3D_ACTIVE_QUERY_BEGUN;
+}
+
+static void d3d12_command_list_end_active_query(struct d3d12_command_list *list,
+        struct vkd3d_active_query *query)
+{
+    const struct vkd3d_vk_device_procs *vk_procs = &list->device->vk_procs;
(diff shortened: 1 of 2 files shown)


libs/vkd3d/command.c
```diff
@@ -2488,10 +2488,96 @@ static void d3d12_command_list_emit_render_pass_transition(struct d3d12_command_
...
+                d3d12_command_list_begin_active_query(list, query);
+            if (query->state == VKD3D_ACTIVE_QUERY_BEGUN)
+                d3d12_command_list_end_active_query(list, query);
+
+            *query = list->active_queries[--list->active_queries_count];
+            return;
+        }
+    }
+
+    WARN("Query (%p,%u) not active.\n", vk_query_pool, index);
```
HansKristian-Work, line 2551: Need to use PRIx64 with (uint64_t) for query since it's 64-bit on x86.

libs/vkd3d/command.c
```diff
@@ -2488,10 +2488,96 @@ static void d3d12_command_list_emit_render_pass_transition(struct d3d12_command_
...
+
+    for (i = 0; i < list->active_queries_count; i++)
+    {
+        struct vkd3d_active_query *query = &list->active_queries[i];
+
+        /* It is possible (although unexpected) that an app uses an occlusion
+         * query for multiple draw calls, which may require multiple render
+         * pass instances. In that case, we can only handle the first instance. */
+        if (query->state == VKD3D_ACTIVE_QUERY_ENDED && !end)
+            FIXME("Query (%p,%u) already ended.\n", query->vk_pool, query->index);
```
HansKristian-Work, line 2566: PRIx64 here as well.

HansKristian-Work (approved):
LGTM, just two minor formatting nits.

Tested DS, HZD, RE2 and Metro with no trips.

If there turns out to be issues I think we could add a hypothetical VKD3D_CONFIG flag which treats the query as scoped and not inline.

PR #441, opened 2020-12-07
tests: Remove old is_bug for conditional rendering.

HansKristian-Work:
Passes now.


doitsujin:
Actually - these aren't actually supposed to pass, and RADV does behave incorrectly here.

HansKristian-Work:
How does it behave incorrectly, and why do the test pass then?

doitsujin:
The test passes because it tests D3D12 behaviour, not Vulkan behaviour (which explicitly excludes clears from conditional rendering), and RADV implements D3D12 behaviour, not Vulkan, which is wrong.

PR #451, opened 2020-12-08
vkd3d: Add fast path for mutable descriptor copies

doitsujin:
Passes test suite. Needs testing w.r.t game stability and performance, thus marking as draft.

### DIFF
--- a/libs/vkd3d/resource.c
+++ b/libs/vkd3d/resource.c
@@ -3426,7 +3426,7 @@ void vkd3d_view_decref(struct vkd3d_view *view, struct d3d12_device *device)
         vkd3d_view_destroy(view, device);
 }
 
-void d3d12_desc_copy(struct d3d12_desc *dst, struct d3d12_desc *src,
+static void d3d12_desc_copy_single(struct d3d12_desc *dst, struct d3d12_desc *src,
         struct d3d12_device *device)
 {
     const struct vkd3d_vk_device_procs *vk_procs = &device->vk_procs;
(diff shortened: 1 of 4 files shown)


libs/vkd3d/resource.c
```diff
@@ -3513,6 +3513,88 @@ void d3d12_desc_copy(struct d3d12_desc *dst, struct d3d12_desc *src,
...
+    for (i = 0; i < count; i++)
+    {
+        dst[i].metadata = src[i].metadata;
+        dst[i].info = src[i].info;
+        dst[i].counter_address = src[i].counter_address;
+    }
+
+    if (copy_count)
+        VK_CALL(vkUpdateDescriptorSets(device->vk_device, 0, NULL, copy_count, vk_copies));
+
```
HansKristian-Work, line 3581: Tiny style nit.

libs/vkd3d/resource.c
```diff
@@ -3513,6 +3513,88 @@ void d3d12_desc_copy(struct d3d12_desc *dst, struct d3d12_desc *src,
...
+{
+    const struct vkd3d_vk_device_procs *vk_procs = &device->vk_procs;
+    VkCopyDescriptorSet vk_copies[2], *vk_copy;
+    struct vkd3d_descriptor_binding binding;
+    enum vkd3d_bindless_set_flag flag;
+    uint32_t copy_count = 0;
+    unsigned int i;
+
+    flag = heap_type == D3D12_DESCRIPTOR_HEAP_TYPE_SAMPLER ? VKD3D_BINDLESS_SET_SAMPLER : VKD3D_BINDLESS_SET_MUTABLE;
+    binding = vkd3d_bindless_state_find_set(&device->bindless_state, flag);
```
HansKristian-Work, line 3527: Is it possible to cache the binding for these sets? It feels a little suboptimal to scan through a binding list since this is in the ultra-hot path.

HansKristian-Work (commented):
Looks pretty straight forward to me.

libs/vkd3d/resource.c
doitsujin: what exactly about it?

HansKristian-Work: Empty line.

libs/vkd3d/resource.c
doitsujin: We're only going to have two bindings in total in this case though, and this path shouldn't be quite as hot if the app copies consecutive ranges.

We could in theory store this with the descriptor heap, but it's **only** going to work with the mutable type extension, and completely meaningless otherwise.

HansKristian-Work:
Tested with mutable_descriptor_type on HZD, DS, RE2, Metro without issues.

HansKristian-Work:
Tested without mutable_type on NV with the same titles. Test suite also works.

libs/vkd3d/resource.c
doitsujin: fixed.

HansKristian-Work (approved):

PR #455, opened 2020-12-10
Rework force_bindless_texel_buffer workaround

doitsujin:
Passes all DXBC tests, but will need DXIL support. Haven't come up with a new name yet.

### DIFF
--- a/libs/vkd3d-shader/spirv.c
+++ b/libs/vkd3d-shader/spirv.c
@@ -5685,13 +5685,20 @@ static void vkd3d_dxbc_compiler_emit_offset_buffer(struct vkd3d_dxbc_compiler *c
     const struct vkd3d_shader_interface_info *shader_interface = &compiler->shader_interface;
     struct vkd3d_spirv_builder *builder = &compiler->spirv_builder;
     uint32_t array_id, struct_id, pointer_id, var_id;
+    uint32_t member_ids[2];
 
     if (!(shader_interface->flags & (VKD3D_SHADER_INTERFACE_SSBO_OFFSET_BUFFER | VKD3D_SHADER_INTERFACE_TYPED_OFFSET_BUFFER)))
         return;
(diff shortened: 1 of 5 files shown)


libs/vkd3d-shader/spirv.c
```diff
@@ -5685,13 +5685,20 @@ static void vkd3d_dxbc_compiler_emit_offset_buffer(struct vkd3d_dxbc_compiler *c
...
-    vkd3d_spirv_build_op_decorate1(builder, array_id, SpvDecorationArrayStride, 8);
+    member_ids[0] = vkd3d_spirv_get_type_id(builder, VKD3D_TYPE_UINT, 2);
+    member_ids[1] = member_ids[0];
+
+    struct_id = vkd3d_spirv_build_op_type_struct(builder, member_ids, ARRAY_SIZE(member_ids));
+    vkd3d_spirv_build_op_member_decorate1(builder, struct_id, 0, SpvDecorationOffset, 0);
+    vkd3d_spirv_build_op_member_decorate1(builder, struct_id, 1, SpvDecorationOffset, 8);
+
+    array_id = vkd3d_spirv_build_op_type_runtime_array(builder, struct_id);
+    vkd3d_spirv_build_op_decorate1(builder, array_id, SpvDecorationArrayStride, 16);
```
HansKristian-Work, line 5701: This commit should be squashed I think so that we don't completely break bisects.

HansKristian-Work (commented):
Seems reasonable enough. Should be verified against the new test I made a PR for.

libs/vkd3d-shader/spirv.c
doitsujin: Will do once we get DXIL support, which then should also be squashed into this.

doitsujin:
Closing because this doesn't work.

PR #457, opened 2020-12-10
tests: Add test for reading typed R32 buffer as untyped.

HansKristian-Work:
Invokes undefined behavior that many games rely on by accident.


HansKristian-Work:
Passes on native AMD and NV drivers.

doitsujin (approved):

HansKristian-Work:
Just added a todo marker since we can't make this work easily right now since the mutable workaround didn't work.

Issue #373, opened 2020-11-08
Support for D3D12_HEAP_FLAG_ALLOW_WRITE_WATCH


Joshua-Ashton:
Needed for D3D12 apitrace.

This indicates that the memory of the heap is allocated with MEM_WRITE_WATCH which allows page-level granularity dirty tracking on the memory.

We should probably do this via a host pointer import allocated with MEM_WRITE_WATCH.

PR #463, opened 2020-12-12
vkd3d: Implement D3D12_HEAP_TYPE_WRITE_WATCH

Joshua-Ashton:
Needed for D3D12 APITrace.

Closes: #373

### DIFF
--- a/libs/vkd3d/resource.c
+++ b/libs/vkd3d/resource.c
@@ -462,6 +462,70 @@ static ULONG STDMETHODCALLTYPE d3d12_heap_AddRef(d3d12_heap_iface *iface)
 
 static ULONG d3d12_resource_decref(struct d3d12_resource *resource);
 
+static void *d3d12_allocate_write_watch_pointer(const D3D12_HEAP_PROPERTIES* properties, VkDeviceSize size)
+{
+#ifdef _WIN32
+    DWORD protect;
+    void *ptr;
+
+    switch (properties->Type)
+    {
+    case D3D12_HEAP_TYPE_DEFAULT:
+        return NULL;
+    case D3D12_HEAP_TYPE_UPLOAD:
+        protect = PAGE_READWRITE | PAGE_WRITECOMBINE;
+        break;
+    case D3D12_HEAP_TYPE_READBACK:
+        /* WRITE_WATCH fails for this type in native D3D12;
+         * otherwise it would be PAGE_READWRITE.
+         */
+        return NULL;
+    case D3D12_HEAP_TYPE_CUSTOM:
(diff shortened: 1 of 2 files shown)


libs/vkd3d/resource.c
```diff
@@ -433,6 +462,70 @@ static ULONG STDMETHODCALLTYPE d3d12_heap_AddRef(d3d12_heap_iface *iface)
...
+#endif
+}
+
+static void d3d12_free_write_watch_pointer(const D3D12_HEAP_DESC* desc, void *pointer)
+{
+#ifdef _WIN32
+    if (VirtualFree(pointer, 0, MEM_RELEASE) == 0)
+        ERR("Failed to free write watch pointer %#x.\n", GetLastError());
+#else
+    /* Not supported on other platforms. */
```
HansKristian-Work, line 525: Needs `(void)desc` and `(void)pointer` to avoid warnings.

libs/vkd3d/resource.c
```diff
@@ -433,6 +462,70 @@ static ULONG STDMETHODCALLTYPE d3d12_heap_AddRef(d3d12_heap_iface *iface)
...
+
+    if (!(ptr = VirtualAlloc(NULL, size, MEM_COMMIT | MEM_RESERVE | MEM_WRITE_WATCH, protect)))
+    {
+        ERR("Failed to allocate write watch pointer %#x.\n", GetLastError());
+        return NULL;
+    }
+
+    return ptr;
+#else
+    ERR("WRITE_WATCH not supported on this platform.\n");
```
HansKristian-Work, line 514: Needs `(void)properties` and `(void)size` to avoid warnings.

HansKristian-Work (changes requested):
I think I'm missing code which handles these cases:

https://github.com/HansKristian-Work/vkd3d-proton/blob/master/libs/vkd3d/resource.c#L982

When we create a resource on an external memory heap, we need to flag the resource as such.

libs/vkd3d/resource.c
```diff
@@ -372,9 +392,18 @@ static HRESULT vkd3d_allocate_image_memory(struct d3d12_device *device, VkImage
     if (FAILED(hr = vkd3d_select_memory_flags(device, heap_properties, &type_flags)))
         return hr;
 
-    if (FAILED(hr = vkd3d_allocate_memory(device, memory_requirements->size, type_flags,
-            memory_requirements->memoryTypeBits, pNext, vk_memory, vk_memory_type)))
-        return hr;
+    if (host_memory)
```
HansKristian-Work, line 395: Is WRITE_WATCH supported for images as well? Would be nice to have a test here.

PR #471, opened 2020-12-29
Github Actions and Docker support along with debug build support. 

dave-juicelabs:
- Adding support for Github Actions compilation.
- Adding support for a build environment in Docker.
- Adding debug build support through package-release.sh command line option --debug.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,15 +13,46 @@ jobs:
   # This workflow contains a single job called "build"
   build:
     # The type of runner that the job will run on
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
 
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
-      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
       - uses: actions/checkout@v2
         with:
           submodules: recursive
+          
+      - name: Setting up mingw
+        run: sudo apt install mingw-w64 mingw-w64
+
+      - name: Setting up wine
+        run: sudo apt install wine64 wine64-tools
+
+      - name: Setup python
+        uses: actions/setup-python@v1
+
+      - name: Setup meson
+        run: pip3 install meson
+
+      - name: Setup ninja
+        run: sudo apt install ninja-build
+
+      - name: Setup glslang
+        run: |
+          git clone --recursive https://github.com/KhronosGroup/glslang.git glslang
+          cd glslang
+          git checkout 8.13.3743
+          cmake .
+          sudo cmake --build . --config Release --target install
 
       # Runs a single command using the runners shell
       - name: Build
-        run: ./package-release.sh release build --no-package
+        run: |
+          ./package-release.sh ${GITHUB_REF##*/} build --no-package
+          echo "BRANCH_NAME=${GITHUB_REF##*/}" >> $GITHUB_ENV
+
+      - name: Uploading artifacts
+        uses: actions/upload-artifact@v2
+        with:
+          name: vkd3d-proton-${{ env.BRANCH_NAME }}
+          path: build/vkd3d-proton-${{ env.BRANCH_NAME }}
+          if-no-files-found: error
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,21 @@
+FROM ubuntu:20.04
+
+ENV TZ=US/Pacific
+RUN ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && \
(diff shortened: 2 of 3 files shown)


Joshua-Ashton:
Nice :)

I was actually planning on doing something like this soon but decided to hold off because I wanted to run the test suite on a self-hosted runner and currently GitHub has no way of not letting random people run arbitrary code on my machine.

Some things I am noticing:
 - Ubuntu's MinGW packages kinda suck.
    They're not only outdated, but the use sjlj instead of DWARF2 which causes some perf problems on x86 builds. (Not sure if that's as big of a problem for vkd3d-proton, but it was for DXVK)

    They also use a weird path for clock stuff in C++ but that only affects DXVK...
    I was planning on making an Arch-based docker for the builds myself because that's what we all use for development and it doesn't have weird problems with sjlj or gettimeofday.

 - Why not just use the Dockerfile for the build inside GitHub Actions also? We wouldn't need to keep setting up every time then as we could pull something from the Docker Cloud Registry or Steam Docker Registry which would make builds a lot faster; plus it consolidates 2 things that are doing the same thing.

 - I'd prefer to just use a packaged version of glslang -- I don't feel like we need to build that ourselves every time.

dave-juicelabs:
- I personally would like to see a switch to llvm and clang. I have not heard of any performance issues when compared to other compiler output and would also provide pdb generation which currently mingw does not. I am working from Windows so all Linux based stuff is somewhat more difficult to do.
- This honestly was an attempt to learn Github Actions so I haven't quite figured out the proper way to do Dockerfiles but yes it would be the best option.
- Again an issue with Github Actions. A packaged version would be preferred.

SveSop:
@Joshua-Ashton 
> Some things I am noticing:
> 
>     * Ubuntu's MinGW packages kinda suck.
>       They're not only outdated, but the use sjlj instead of DWARF2 which causes some perf problems on x86 builds. (Not sure if that's as big of a problem for vkd3d-proton, but it was for DXVK)
>       They also use a weird path for clock stuff in C++ but that only affects DXVK...
>       I was planning on making an Arch-based docker for the builds myself because that's what we all use for development and it doesn't have weird problems with sjlj or gettimeofday.
> 

I am indeed in agreement with you about the Ubuntu mingw-w64 package, however the new 21.04 seems to have this fixed for 8.0 ref: 
Changelog

gcc-mingw-w64 (24) unstable; urgency=medium

  * Following MSYS2 and Fedora’s example (among others), switch from SJLJ
    exception handling to Dwarf2; this will help the Rust toolchain in
    particular. Closes: #540782.

[https://launchpad.net/ubuntu/+source/gcc-mingw-w64](https://launchpad.net/ubuntu/+source/gcc-mingw-w64)

This is in "testing" atm, so not available outside of backport ppa for 20.04 or 20.10. (I got a PPA currently with packages for 20.04). DXVK requires mingw-w64-headers >=  8.0, and officially Ubuntu only have 7.0 for focal/groovy.

There IS however some build differences between native ubuntu build vs. running docker with arch and building there. This can be due to some ubuntu patches that changes stuff with AVX and such. [https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=939559](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=939559) (There is patches in the build-archive for Ubuntu package regarding this)

In short (after some testing from me), i have concluded with the following:
Building DXVK with my backport of mingw-w64-8.0 (gcc-10.2) builds an unstable binary when using the default -msse -msse2 flags, but seems to build "better" when using -mavx. However, recent commits to the DXVK tree has made both somewhat unstable (crashing SOMETIMES when starting apps).
The same source built with docker/arch works fine seemingly without those issues at all, but not able to produce running binaries when built with -mavx flag it seems.

This experience could be completely different for different processors (i have i7 8700K), so where a binary compiled with Ubuntu could just be doing something weird for me, but work fine for a difference cpu.

To sum up my hypothesis: I think the "avx patches" used by Ubuntu for gcc-mingw-w64 is not 100% stable when regarding avx/msse flag usage, atleast not for me when using gcc-mingw-w64(10.2) to build DXVK or VKD3D. The old gcc-mingw package built on gcc-9.3 seemed to work better. I had multiple fails with 32-bit apps, where 64-bit would work and vice versa. I cannot say the same for VKD3D, as the only thing i use regularly there is WoW (64-bit).

Sorry for a lengthy wall-of-text, but i felt it was somewhat related, and perhaps food for thought when considering making a "docker build system" or similar (kinda like Steam does).

Joshua-Ashton:
I manually merged in the debug stuff. I'm going to work on getting Github Actions on Arch today.

PR #477, opened 2021-01-11
tests: Mark sparse depth image test as TODO on RADV.

doitsujin:
Otherwise we'll have two test failures.

### DIFF
--- a/tests/d3d12.c
+++ b/tests/d3d12.c
@@ -42353,6 +42353,7 @@ static void test_get_resource_tiling(void)
         UINT tile_shape_h;
         UINT tile_shape_d;
         D3D12_TILED_RESOURCES_TIER min_tier;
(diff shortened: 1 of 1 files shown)


tests/d3d12.c
```diff
@@ -42353,6 +42353,7 @@ static void test_get_resource_tiling(void)
         UINT tile_shape_h;
         UINT tile_shape_d;
         D3D12_TILED_RESOURCES_TIER min_tier;
+        bool todo_amd;
```
HansKristian-Work, line 42356: Is this TODO on our end, or bug in RADV? Should probably be called todo_radv in that case.

HansKristian-Work (changes requested):
Nit.

HansKristian-Work (approved):

PR #479, opened 2021-01-12
Handle precise modifier

doitsujin:
This is a patch from upstream vkd3d that we should pull in as well to avoid potential precision issues.


HansKristian-Work (approved):
LGTM.

PR #481, opened 2021-01-13
Support multiple descriptors per heap entry

### LINKED ISSUE #460: Find a more robust solution to force_bindless_texel_buffer
This is a gross hack which does not really work if the application makes use of 16-bit storage. The solution will likely involve using multiple descriptors in some limited fashion :(

HansKristian-Work:
The goal here is to eliminate the ugly force_bindless_texel_buffer hack.

Tested on CP77 and Death Stranding which were the two games which required workaround to work.

Fix #460.


doitsujin (approved):
Annoying that we have to do this, but LGTM.

PR #482, opened 2021-01-13
vkd3d: Reset inline queries in BeginQuery.

doitsujin:
We currently never reset occlusion queries. For some reason, validation layers do not report this.


HansKristian-Work (approved):
LGTM.

PR #483, opened 2021-01-18
Virtualize occlusion and transform feedback queries

### LINKED ISSUE #459: Rewrite query implementation to use virtual queries
This is required to gracefully handle cases where queries cross the boundary of render passes.
Currently we either have two options:
- Just emit queries inside render passes and pray app is sensible (breaks stuff)
- Suspend render passes between every query (bad for performance)

doitsujin:
Allows us to allocate more than one Vulkan query per D3D12 query for certain query types, which makes the query optimization work even if a single query spreads across multiple render pass instances. This allows us to drop the workaround for AC:Valhalla.

Closes #459-


libs/vkd3d/shaders/cs_resolve_query.comp
```diff
@@ -0,0 +1,62 @@
...
+
+  if (thread_id >= query_count)
+    return;
+
+  // The query map is an array of linked lists, with the
+  // first query_count entries guaranteed to be list heads
+  query_map_entry_t entry = map.entries[thread_id + entry_offset];
+  uint64_t dst_data[c_field_count];
+
+  // By copyig the first query we get the reset for free
```
HansKristian-Work, line 47: Typo: copyig

libs/vkd3d/command.c
```diff
@@ -2572,31 +2603,44 @@ static void d3d12_command_list_end_active_query(struct d3d12_command_list *list,
...
+    query->heap = heap;
     query->index = index;
-    query->flags = flags;
+    query->type = type;
     query->state = VKD3D_ACTIVE_QUERY_RESET;
+    query->vk_pool = heap->vk_query_pool;
+    query->vk_index = index;
+    return true;
+
+/*    return d3d12_command_allocator_allocate_query(list->allocator,
```
HansKristian-Work: I guess this code should be removed?

HansKristian-Work: It's reinstated in a later commit so I guess it makes sense to do this.

libs/vkd3d/command.c
```diff
@@ -2692,6 +2696,339 @@ static void d3d12_command_list_handle_active_queries(struct d3d12_command_list *
...
+        dst_queries += d->virtual_query_count;
+    }
+
+    /* Upload query lists in chunks since vkCmdUpdateBuffer is limited to
+     * 64kiB per invocation. Normally, one single iteration should suffice. */
+    for (i = 0; i < list->pending_queries_count; i += 2048)
+    {
+        unsigned int count = min(2048, list->pending_queries_count - i);
+
+        VK_CALL(vkCmdUpdateBuffer(list->vk_command_buffer, entry_buffer.buffer,
```
HansKristian-Work, line 2943: Isn't the scratch buffer host visible?

tests/d3d12.c
```diff
@@ -47492,6 +47492,139 @@ static void test_undefined_read_typed_buffer_as_untyped_dxil(void)
...
+        0x43425844, 0x91af6cd0, 0x7e884502, 0xcede4f54, 0x6f2c9326, 0x00000001, 0x000000b0, 0x00000003,
+        0x0000002c, 0x0000003c, 0x00000070, 0x4e475349, 0x00000008, 0x00000000, 0x00000008, 0x4e47534f,
+        0x0000002c, 0x00000001, 0x00000008, 0x00000020, 0x00000000, 0x00000000, 0x00000003, 0xffffffff,
+        0x00000e01, 0x445f5653, 0x68747065, 0xababab00, 0x52444853, 0x00000038, 0x00000040, 0x0000000e,
+        0x04000059, 0x00208e46, 0x00000000, 0x00000001, 0x02000065, 0x0000c001, 0x05000036, 0x0000c001,
+        0x0020800a, 0x00000000, 0x00000000, 0x0100003e,
+    };
+    static const D3D12_SHADER_BYTECODE ps = {ps_code, sizeof(ps_code)};
+    static const uint32_t expected_results[] = {1,0,1,1,614400,0,307200,307200};
+    static const float depthOne = 1.0f;
```
HansKristian-Work, line 47530: weird camel case

libs/vkd3d/command.c
```diff
@@ -7433,13 +7513,25 @@ static void STDMETHODCALLTYPE d3d12_command_list_ResolveQueryData(d3d12_command_
...
-                1, &copy_region));
+            VK_CALL(vkCmdCopyBuffer(list->vk_command_buffer,
+                    query_heap->vk_buffer, buffer->vk_buffer,
+                    1, &copy_region));
+        }
+        else
+        {
+            uint32_t dst_index = aligned_dst_buffer_offset / sizeof(uint64_t);
+
+            d3d12_command_list_resolve_binary_occlusion_queries(list,
```
HansKristian-Work, line 7534: Shouldn't this be handled in the resolve shader with a spec const bool? I figure it would just write 1 or 0 depending on the final counter value.

libs/vkd3d/command.c
```diff
@@ -2692,6 +2696,339 @@ static void d3d12_command_list_handle_active_queries(struct d3d12_command_list *
...
+    struct dispatch_entry
+    {
+        struct d3d12_query_heap *heap;
+        uint32_t virtual_query_count;
+        uint32_t unique_query_count;
+        VkDeviceSize resolve_buffer_offset;
+        VkDeviceSize resolve_buffer_size;
+    };
+    
+    struct dispatch_entry *dispatches = NULL;
```
HansKristian-Work, line 2764: Any way to avoid having to allocate fresh arrays every time?

HansKristian-Work (commented):
Some comments. Overall the implementation makes sense, albeit fairly complicated resolve, but not much we can do about that.

libs/vkd3d/command.c
HansKristian-Work: Could technically store the arrays in command allocator, but I can see why it gets very hairy.

libs/vkd3d/command.c
```diff
@@ -2692,6 +2696,339 @@ static void d3d12_command_list_handle_active_queries(struct d3d12_command_list *
...
+    {
+        ERR("Failed to allocate query map.\n");
+        goto cleanup;
+    }
+
+    /* Active list for the current dispatch */
+    src_queries = list->pending_queries;
+    dst_queries = query_list;
+
+    for (i = 0; i < dispatch_count; i++)
```
HansKristian-Work, line 2897: Quite hard to follow this loop, but I understand the data structure it's trying to emit, so it's probably fine.

libs/vkd3d/command.c
doitsujin: We can't do that since you can allocate both binary and non-binary queries from an occlusion query heap.

There's the option to do this per query during the resolve by passing extra data to the shader, but there's a lot of room for the app to fuck up by doing this:
```
cmd->BeginQuery(..., D3D11_QUERY_TYPE_OCCLUSION);
...
cmd->EndQuery(..., D3D11_QUERY_TYPE_OCCLUSION);
cmd->ResolveQueryData(..., D3D11_QUERY_TYPE_BINARY_OCCLUSION);
```
in which case we should still do a binary resolve, even if it was specified incorrectly for the queries.

libs/vkd3d/command.c
doitsujin: Can do, it's just annoying, and we shouldn't hit this more than once per frame anyway.

libs/vkd3d/command.c
doitsujin: No, scratch buffers are device local.

tests/d3d12.c
doitsujin: fixed.

libs/vkd3d/shaders/cs_resolve_query.comp
doitsujin: fixed.

libs/vkd3d/command.c
doitsujin: Of course we can extend the scratch buffer system to allow host-visible buffers as well, but that's a larger change that would require a fair bit of work. Also we'd be doing random access on host memory in that case, which is probably bad.

HansKristian-Work: Right.

libs/vkd3d/command.c
HansKristian-Work: Ah, ok. I didn't realize app could resolve with BINARY separately.

libs/vkd3d/command.c
HansKristian-Work: Could just mark this as TODO at least so that we have a potential optimization avenue if games are dumb.

HansKristian-Work (approved):
LGTM, nice work!

libs/vkd3d/command.c
doitsujin: Done.

PR #484, opened 2021-01-23
vkd3d: Introduce VKD3D_FILTER_DEVICE_NAME.

DadSchoorse:
The same as `DXVK_FILTER_DEVICE_NAME`.

### DIFF
--- a/libs/vkd3d/device.c
+++ b/libs/vkd3d/device.c
@@ -1597,9 +1597,12 @@ static HRESULT vkd3d_select_physical_device(struct vkd3d_instance *instance,
     VkPhysicalDeviceProperties device_properties;
     VkPhysicalDevice device = VK_NULL_HANDLE;
     VkPhysicalDevice *physical_devices;
+    const char *filter;
     uint32_t count;
     unsigned int i;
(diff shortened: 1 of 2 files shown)


libs/vkd3d/device.c
```diff
@@ -1640,6 +1644,12 @@ static HRESULT vkd3d_select_physical_device(struct vkd3d_instance *instance,
         if (i == device_index)
             device = physical_devices[i];
 
+        if (!strstr(device_properties.deviceName, filter))
```
HansKristian-Work, line 1647: Is it well defined how strstr works with empty string? Seems cleaner to check `if (filter && strstr())`.

HansKristian-Work (changes requested):
Small nit.

libs/vkd3d/device.c
DadSchoorse: Yes, it's well defined (it will return the first argument). But yes, `if (filter && !strstr()).` is probably cleaner.

HansKristian-Work (approved):
LGTM, thanks.

PR #491, opened 2021-01-29
Implement first iteration of RTPSO

HansKristian-Work:
Implements the bare minimum API portion that is required to let the test pipeline compile without validation errors and is created without error on Windows / RX6800 driver.

- DXIL side should be pretty solid, everything is hooked up, barring bugs.
- StateObject is bare minimum for the test, still tons of work left, but this PR sets up the basic architecture and compilation flow.
- RTAS / CmdBuffer / Resource side has no code yet (depends on memory refactor work).


libs/vkd3d/raytracing_pipeline.c
```diff
@@ -0,0 +1,778 @@
...
+    *object = NULL;
+    return E_NOINTERFACE;
+}
+
+static HRESULT STDMETHODCALLTYPE d3d12_state_object_properties_QueryInterface(ID3D12StateObjectProperties *iface,
+        REFIID riid, void **object)
+{
+    TRACE("iface %p, riid %s, object %p.\n", iface, debugstr_guid(riid), object);
+
+    if (IsEqualGUID(riid, &IID_ID3D12StateObjectProperties)
```
doitsujin, line 67: Going by COM rules, this should call `d3d12_state_object_QueryInterface` with the parent state object.

May need some testing on Windows, but in general, if interface A allows querying interface B, then interface B must also allow querying interface A and pointers to IUnknown instances have to be consistent.

doitsujin:
Can't really comment on the DXIL bits or whether every little detail about RT pipeline creation is correct, but looks good overall besides QueryInterface stuff.

Do we always want to enable the VA aux buffer for SRVs though or should that only be done if RT support is actually present?

libs/vkd3d/state.c
```diff
@@ -399,11 +402,19 @@ static HRESULT d3d12_root_signature_info_from_desc(struct d3d12_root_signature_i
                     if (FAILED(hr = d3d12_root_signature_info_count_descriptors(info,
                             device, &p->DescriptorTable.pDescriptorRanges[j])))
                         return hr;
-                info->cost += 1;
+
+                /* Local root signature directly affects memory layout. */
+                if (local_root_signature)
+                    info->cost = (info->cost + 1u) & ~1u;
+                info->cost += local_root_signature ? 2 : 1;
```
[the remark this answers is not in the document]
HansKristian-Work: Descriptor heap VAs are 64-bit in the API, but the root cost is 32-bit for global root signatures. We get no chance to resolve the VA to index when placed in a buffer, so we'll need to convert it on GPU and the size is indeed 64-bit (sigh). I have a plan for how we can do it trivially, but that's for later.

libs/vkd3d/raytracing_pipeline.c
HansKristian-Work: Fixed.

HansKristian-Work:
I didn't touch the binding for SRV aux buffer vs UAV. It gets needlessly complicated and is harmless.

doitsujin (approved):

vityafx:
Guys, thanks much for your work! <3 Am I right to assume that these changes do not really bring full RT translation but only allow an internal test to pass with the feature being enabled? I just wonder why the branch has been deleted as I was tracking it for knowing the progress and also learning a bit :-)

Joshua-Ashton:
It got merged.

PR #497, opened 2021-02-04
Memory rework

doitsujin:
This is a large-scale rewrite of resource creation and memory allocation code.

The first half (up to ff9f7b5) of this commit dump is functional changes, the second half mostly removing old code and cleaning up.

This rework is supposed to solve the following problems:
- DXR requires us to pass actual GPU VAs to the application, rather than fake ones, which the current VA allocation scheme does not support.
- The resource creation and memory allocation code is a bit of a mess, which makes it very hard to e.g. implement a better workaround for image alignment issues on some AMD GPUs.

Quick overview:

#### Global memory allocator
Decouples memory allocation from `ID3D12Heap`, and provides a dedicated structure holding allocation info, `vkd3d_memory_allocation`. Like the old heap implementation, this will create an omnipotent buffer for allocations which we'll offset into when creating buffer resources.

Small buffers (<1MB) will be suballocated from larger chunks (16MB), both in order to reduce the number of BOs in some games like Control, and to allow the VA map to be efficient. These allocations need to be initialized to zero manually.

The chunk allocator is similar to that of DXVK in that it maintains a free list per chunk and uses a worst-fit algorithm to keep fragmentation low, which should work reasonably well since most allocations can be expected to be aligned to 64 KiB.

#### GPU VA to resource mapping
Based on #392 which supports O(1) lookup for native VAs, but with native support for small ranges for which we'll use a binary search over an array (requires locking). Since we suballocate small buffers, we will only hit that path when importing host memory. Alloc/Free are O(n).

The map returns a new structure, `vkd3d_unique_resource`, which is owned either by an allocation or a tiled resource. We need to support both since tiled resources do not own or reference an allocation, but can still be referenced via VAs.

In case a device does not support `bufferDeviceAddress`, fake VAs are still supported through a separate VA allocator, but we'll use the new code to perform the mapping. The old `vkd3d_gpu_va_allocator` was removed.

This change also allows us to pass VAs as root descriptors without remapping, which may reduce CPU overhead for those. On the other hand, VA lookup itself may be slightly more expensive than it is with the current scheme.

#### Resource creation rewrite
Decouples resource creation from memory allocation, and cleans up `d3d12_resource` in such a way that it no longer owns a heap (we still reference it though, not sure if we need to since the heap's ref count doesn't appear to change when allocating a placed resource on it?).

Committed buffers will no longer use dedicated allocations so that small committed buffers can be suballocated from larger chunks, just like small heaps.

#### TODO
- [x] Zero suballocated memory. This might turn out to be very hard to do with the current design, but we have no choice.
- [x] Add more debug output to track allocations in case something goes wrong
- [x] Test in more games.
- [x] Test on Nvidia.


libs/vkd3d/va_map.c
```diff
@@ -0,0 +1,382 @@
...
+static struct vkd3d_va_block *vkd3d_va_map_get_block(struct vkd3d_va_map *va_map, VkDeviceAddress va)
+{
+    VkDeviceAddress next_address = vkd3d_va_map_get_next_address(va);
+    struct vkd3d_va_tree *tree, **tree_ptr;
+    
+    tree = &va_map->va_tree;
+
+    while (next_address)
+    {
+        tree_ptr = &tree->next[next_address & VKD3D_VA_NEXT_MASK];
```
HansKristian-Work, line 59: Shouldn't this be an atomic load?

libs/vkd3d/va_map.c
```diff
@@ -0,0 +1,382 @@
...
+
+        if (!(tree = *tree_ptr))
+        {
+            void *orig;
+            tree = vkd3d_malloc(sizeof(*tree));
```
HansKristian-Work, line 64: calloc?

libs/vkd3d/va_map.c
```diff
@@ -0,0 +1,382 @@
...
+{
+    return (va >> VKD3D_VA_BLOCK_SIZE_BITS) & VKD3D_VA_BLOCK_MASK;
+}
+
+static const struct vkd3d_va_block *vkd3d_va_map_find_block(const struct vkd3d_va_map *va_map, VkDeviceAddress va)
+{
+    VkDeviceAddress next_address = vkd3d_va_map_get_next_address(va);
+    const struct vkd3d_va_tree *tree = &va_map->va_tree;
+
+    while (next_address && tree)
```
HansKristian-Work, line 38: Is the general idea here is that this will usually only take one iteration through the loop since this covers any VA up to 40 bits in the first iteration?

libs/vkd3d/va_map.c
```diff
@@ -0,0 +1,382 @@
...
+    {
+        min_va = resource->va;
+        max_va = resource->va + resource->size;
+        block_va = min_va & ~VKD3D_VA_LO_MASK;
+
+        while (block_va < max_va)
+        {
+            block = vkd3d_va_map_get_block(va_map, block_va);
+
+            if (block_va < min_va)
```
HansKristian-Work, line 143: Don't we somehow need atomic store here at least?

libs/vkd3d/va_map.c
```diff
@@ -0,0 +1,382 @@
...
+            }
+
+            block_va += VKD3D_VA_BLOCK_SIZE;
+        }
+    }
+    else
+    {
+        pthread_mutex_lock(&va_map->mutex);
+
+        if (vkd3d_va_map_find_entry(va_map, resource->va, &index) == resource)
```
HansKristian-Work, line 211: Should probably call this find_small_entry to make it more distinct?

libs/vkd3d/memory.c
```diff
@@ -411,18 +616,128 @@ void vkd3d_memory_allocator_cleanup(struct vkd3d_memory_allocator *allocator, st
...
+    type_mask &= device->memory_info.global_mask;
+    type_mask &= memory_requirements->memoryTypeBits;
+
+    for (i = 0; i < allocator->chunks_count; i++)
+    {
+        chunk = allocator->chunks[i];
+
+        /* Match flags since otherwise the backing buffer
+         * may not support our required usage flags */
+        if (chunk->allocation.d3d12_flags != heap_flags)
```
HansKristian-Work, line 679: Will this match CUSTOM properties as well? I.e. CACHED vs UNCACHED CUSTOM heap type.

libs/vkd3d/memory.c
```diff
@@ -411,18 +616,128 @@ void vkd3d_memory_allocator_cleanup(struct vkd3d_memory_allocator *allocator, st
...
+            continue;
+
+        /* Filter out unsupported memory types */
+        if (!(type_mask & (1u << chunk->allocation.vk_memory_type)))
```
HansKristian-Work, line 683: If we have stray host fallback chunks left, does this mean we can end up suballocating from those instead of device-only chunks?

HansKristian-Work (commented):
Had a look over this and I think it's a very solid start here. The layout of the patch set makes sense to me and I managed to follow things. I have some questions about the details in the low-level allocation. The latter half I didn't review in as much detail, but I like this general architecture at least.

libs/vkd3d/va_map.c
doitsujin: Yes.

libs/vkd3d/memory.c
doitsujin: Custom properties shouldn't matter at all since they only affect the memory type, not the usage flags.

I suppose we have to match the heap type though, since allocations on different heap types can target the same memory type and that does actually affect usage flags.

libs/vkd3d/memory.c
doitsujin: When looking only at this function, then yes, but `vkd3d_suballocate_memory` does two calls to this function with different memory type masks so that this doesn't happen.

libs/vkd3d/va_map.c
doitsujin: This is just address calculation, there's no load here.

doitsujin: I guess the load afterwards should be atomic, though.

libs/vkd3d/va_map.c
doitsujin: I guess every single access to the block resource pointers and va must be atomic.

libs/vkd3d/va_map.c
doitsujin: Fixed.

libs/vkd3d/va_map.c
doitsujin: Done.

libs/vkd3d/va_map.c
doitsujin: Should be fixed, needs a second pair of eyes though.

libs/vkd3d/va_map.c
doitsujin: Fixed.

libs/vkd3d/vkd3d_private.h
```diff
@@ -239,6 +239,75 @@ HRESULT vkd3d_fence_worker_start(struct vkd3d_fence_worker *worker,
...
+
+#define VKD3D_VA_NEXT_BITS (12)
+#define VKD3D_VA_NEXT_COUNT (1ull << VKD3D_VA_NEXT_BITS)
+#define VKD3D_VA_NEXT_MASK (VKD3D_VA_NEXT_COUNT - 1)
+
+struct vkd3d_unique_resource;
+
+struct vkd3d_va_entry
+{
+    VkDeviceAddress va;
```
HansKristian-Work, line 258: Since we're doing atomic load/store to this 64-bit, we need to ensure it's 8 byte aligned on 32-bit targets.

libs/vkd3d/va_map.c
HansKristian-Work: Ye, I think relaxed is fine here since other threads need a memory barrier to be able to even observe the newly allocated VA anyways.

libs/vkd3d/va_map.c
```diff
@@ -0,0 +1,381 @@
...
+    {
+        min_va = resource->va;
+        max_va = resource->va + resource->size;
+        block_va = min_va & ~VKD3D_VA_LO_MASK;
+
+        while (block_va < max_va)
+        {
+            block = vkd3d_va_map_get_block(va_map, block_va);
+
+            if (vkd3d_atomic_ptr_load_explicit(&block->l.resource, vkd3d_memory_order_relaxed) == resource)
```
HansKristian-Work, line 192: We release VA range *before* freeing VkDeviceMemory right? Just trying to think about weird ordering where other threads could potentially intervene.

libs/vkd3d/vkd3d_private.h
doitsujin: How do we do that? There's no alignas in C.

libs/vkd3d/va_map.c
doitsujin: There are possible race conditions if an app uses and destroys the same resource at the same time, but that's not a new issue and we can't really do much about that.

libs/vkd3d/vkd3d_private.h
doitsujin: Turns out we have DECLSPEC_ALIGN. Fixed.

HansKristian-Work:
Tested this branch on NV with many titles and no regressions as far as I can tell. Ran over:
- DS
- HZD
- CP77
- RE2
- DMC5
- Metro
- Test suite

libs/vkd3d/resource.c
```diff
@@ -2133,14 +2132,8 @@ static ULONG STDMETHODCALLTYPE d3d12_resource_Release(d3d12_resource_iface *ifac
     TRACE("%p decreasing refcount to %u.\n", resource, refcount);
 
     if (!refcount)
-    {
-        struct d3d12_device *device = resource->device;
-
         d3d12_resource_decref(resource);
 
-        d3d12_device_release(device);
```
HansKristian-Work, line 2141: Do resources not take device references?

doitsujin: We remove the device reference in `d3d12_resource_destroy` instead.

HansKristian-Work (approved):
LGTM. I've been testing this branch quite a lot over the last days.

PR #501, opened 2021-02-11
vkd3d: Fix compiler errors when using vkd3d_atomic_ptr_store_explicit.

doitsujin:


Joshua-Ashton (approved):
LGTM, not sure why it causes errors but it's definitely more correct.

PR #502, opened 2021-02-12
build: Add Github Actions workflows

Joshua-Ashton:
Supercedes: #471

Will add a MSVC build test down the line -- need to figure out the glslang logistics there.


HansKristian-Work (approved):
Can't prove apriori if it's correct, but this does not affect code at all, so LGTM. Thanks :D

PR #503, opened 2021-02-12
build: Add GCC problem matcher to build tests

Joshua-Ashton:
Uses my fork of gcc-problem-matcher to note down warnings and annotate them in files for pushes and PRs


HansKristian-Work (approved):
Seems useful. Can always be disabled later if it gets too annoying.

PR #507, opened 2021-02-12
tests: Fix warnings in tests

Joshua-Ashton:


tests/d3d12_crosstest.h
```diff
@@ -411,6 +411,8 @@ static inline bool is_depth_clip_enable_supported(ID3D12Device *device)
 static PFN_vkGetInstanceProcAddr pfn_vkGetInstanceProcAddr;
 static bool init_vulkan_loader(void)
 {
+    void *mod;
```
DadSchoorse, line 414: What if mod is `HMODULE`? and why does this even compile on mingw?

Joshua-Ashton: HMODULE == void*

DadSchoorse: you are still declaring mod two times tho in the `_WIN32` path

HansKristian-Work: I guess it's not as easy to trigger this case since it should only hit when using vkd3d-utils based Win32 test build.

HansKristian-Work (changes requested):
LGTM, just needs to fix the _WIN32 error as mentioned in discussion.

tests/d3d12_crosstest.h
Joshua-Ashton: Resolved.

HansKristian-Work (approved):
LGTM

PR #512, opened 2021-02-15
Add new memory allocator + VA map

doitsujin:


libs/vkd3d/vkd3d_private.h
```diff
@@ -406,6 +406,102 @@ struct d3d12_fence
...
+
+struct vkd3d_memory_chunk;
+
+struct vkd3d_allocate_memory_info
+{
+    VkMemoryRequirements memory_requirements;
+    D3D12_HEAP_PROPERTIES heap_properties;
+    D3D12_HEAP_FLAGS heap_flags;
+    VkBuffer vk_buffer;
+    void* host_ptr;
```
HansKristian-Work, line 494: `void *` many similar nits follow.

libs/vkd3d/vkd3d_private.h
```diff
@@ -406,6 +406,102 @@ struct d3d12_fence
 HRESULT d3d12_fence_create(struct d3d12_device *device,
         uint64_t initial_value, D3D12_FENCE_FLAGS flags, struct d3d12_fence **fence);
 
+enum vkd3d_allocation_flag
+{
+    VKD3D_ALLOCATION_FLAG_GLOBAL_BUFFER     = (1u << 0),
+    VKD3D_ALLOCATION_FLAG_DEDICATED_BUFFER  = (1u << 1),
```
HansKristian-Work, line 481: I'm a little confused about what the difference between DEDICATED and GLOBAL is here.

libs/vkd3d/memory.c
```diff
@@ -0,0 +1,482 @@
...
+    {
+        /* Respect existing memory requirements since there may not
+         * be any buffer resource to get memory requirements from. */
+        memory_requirements = info->memory_requirements;
+    }
+
+    /* For dedicated buffer allocations we should assign the existing
+     * buffer for address lookup purposes, but take care not to destroy
+     * it when freeing the allocation. */
+    if (allocation->flags & VKD3D_ALLOCATION_FLAG_DEDICATED_BUFFER)
```
HansKristian-Work, line 306: I guess DEDICATED and GLOBAL cannot be used together here, or we leak allocation->resource.vk_buffer?

libs/vkd3d/va_map.c
```diff
@@ -0,0 +1,381 @@
+/*
+ * Copyright 2020 Philip Rebohle for Valve Software
```
HansKristian-Work, line 2: 2021

doitsujin: It's technically code from 2020 but can change I guess.

libs/vkd3d/vkd3d_private.h
doitsujin: GLOBAL means that the allocation owns a buffer, whereas DEDICATED only references an existing buffer.

libs/vkd3d/memory.c
doitsujin: Yes, they are mutually exclusive, see above.

libs/vkd3d/va_map.c
```diff
@@ -0,0 +1,381 @@
...
+        }
+
+        pthread_mutex_unlock(&va_map->mutex);
+    }
+}
+
+const struct vkd3d_unique_resource *vkd3d_va_map_deref(struct vkd3d_va_map *va_map, VkDeviceAddress va)
+{
+    const struct vkd3d_va_block *block = vkd3d_va_map_find_block(va_map, va);
+    const struct vkd3d_unique_resource *resource = NULL;
```
HansKristian-Work, line 225: Looks like declare after statement to me.

libs/vkd3d/va_map.c
```diff
@@ -0,0 +1,381 @@
...
+
+        if (range.size)
+            allocator->free_ranges[i] = range;
+        else
+            allocator->free_range_count--;
+    }
+    else
+    {
+        va = allocator->next_va;
+        allocator->next_va += size;
```
HansKristian-Work, line 289: Should probably align the fake VAs to 64 kiB for sanity.

libs/vkd3d/va_map.c
doitsujin: Is it though? GCC isn't complaining and we do this sort of thing all the time (impl_from_ID3D12xxxx etc).

HansKristian-Work: Ok, no complaints then.

libs/vkd3d/va_map.c
```diff
@@ -0,0 +1,381 @@
...
+
+    /* Find and effectively delete any free range adjacent to new_range */
+    for (i = 0; i < allocator->free_range_count; i++)
+    {
+        const struct vkd3d_va_range *cur_range = &allocator->free_ranges[i];
+
+        if (range_shift)
+            allocator->free_ranges[i - range_shift] = *cur_range;
+
+        if (cur_range->base == va + size || cur_range->base + cur_range->size == va)
```
HansKristian-Work, line 323: Shouldn't this check against new_range.base/size instead of va/size? I can imagine a case where we have 3 adjacent ranges, A, B, C. If A and C are in the free list, and we free B, we should get two separate merge events I think?

doitsujin: Shouldn't really change anything since we can't have more than one adjacent range with a lower base address, for example, unless the merging has already gone wrong elsewhere. Will change anyway since aligning the fake VAs does get messy otherwise.

HansKristian-Work (changes requested):
Comments

HansKristian-Work (approved):
LGTM, to merge when CI passes.

libs/vkd3d/vkd3d_private.h
doitsujin: Fixed.

PR #513, opened 2021-02-15
Add suballocator for small allocations

doitsujin:


libs/vkd3d/memory.c
```diff
@@ -380,6 +394,197 @@ static HRESULT vkd3d_memory_allocation_init(struct vkd3d_memory_allocation *allo
...
+    r_length = pick_range->offset + pick_range->length
+            - allocation->offset - allocation->resource.size;
+
+    if (l_length)
+    {
+        pick_range->length = l_length;
+
+        if (r_length)
+        {
+            vkd3d_memory_chunk_insert_range(chunk, pick_index + 1,
```
HansKristian-Work, line 479: Does chunk order matter here?

HansKristian-Work: Ah, nvm, this is range within the chunk, not overall chunk order.

doitsujin: what do you mean with "chunk order"?

HansKristian-Work: I meant the order of elements in chunk->free_ranges.

libs/vkd3d/memory.c
```diff
@@ -380,6 +394,197 @@ static HRESULT vkd3d_memory_allocation_init(struct vkd3d_memory_allocation *allo
...
+        else
+            lo = index + 1;
+    }
+
+    return lo;
+}
+
+static void vkd3d_memory_chunk_free_range(struct vkd3d_memory_chunk *chunk, const struct vkd3d_memory_allocation *allocation)
+{
+    struct vkd3d_memory_free_range *range;
```
HansKristian-Work, line 520: Do we never release chunks back to the system with vkFreeMemory? I can see a potential problem if we have fully freed chunks hanging around in a memory type which isn't usually allocated from.

doitsujin: Can add that, but the reasoning (at least for DXVK where this was largely take from) is that it's very unlikely to have a fully freed chunk in the first place. I guess it might make more sense here though.

libs/vkd3d/memory.c
```diff
@@ -380,6 +394,197 @@ static HRESULT vkd3d_memory_allocation_init(struct vkd3d_memory_allocation *allo
...
+        if (range->length == memory_requirements->size)
+        {
+            pick_index = i;
+            pick_range = range;
+            break;
+        }
+
+        /* Alignment is almost always going to be 64 KiB, so
+         * don't worry too much about misalignment gaps here */
+        if (!pick_range || range->length > pick_range->length)
```
HansKristian-Work, line 453: I guess the intent here is to sub-allocate from the current largest range?

doitsujin: yes.

libs/vkd3d/memory.c
HansKristian-Work: I can see a problem with RADV since the "freed" chunks will still contribute to global BO list and potentially cause other performance issues.

HansKristian-Work: A middle ground would be to keep a maximum number of chunks alive per memory type perhaps. Could be a separate "garbage collection" after freeing a range and the chunk is now fully available.

HansKristian-Work (changes requested):
Largely makes sense. Given that this is taken from DXVK I'll trust that the algorithmic details work. My only real concern is how we deal with fully freed memory chunks which will claim residence on RADV, but they won't on DXVK.

libs/vkd3d/memory.c
doitsujin: I went with just unconditionally destroying the chunk. Reason being that there are quite a large number of conditions to meet for a chunk to be considered for suballocation, and we probably wouldn't want to keep chunks alive that were allocated on system memory when we really wanted device memory etc.

HansKristian-Work (approved):
LGTM, will merge when CI passes.

PR #516, opened 2021-02-16
Zero suballocated memory

doitsujin:
Executes `vkCmdFillBuffer` on a compute queue for small allocations before each queue submission.

Couple of things to note:
- We obviously have to stall all (other) device queues that may access the cleared memory. We do this by submitting a timeline semaphore wait to those queues immediately after submitting the clear command buffer. There doesn't seem to be a nicer way to do this since all the submission thread stuff etc works on D3D12CommandQueue objects, which we have no knowledge of in the memory allocator.
- We have to ensure that no Vulkan resources get freed while still being cleared. The same timeline semaphore is used for that purpose.
- However, we only synchronize with the GPU if the underlying Vulkan allocation gets freed, not when freeing a suballocation. This is safe because clears are executed in order, so even if the app ends up freeing and then immediately reallocating memory that is currently being cleared, the worst thing that will happen is that we clear some memory regions twice, but subsequent submissions will have to wait for both clears to finish regardless.

I've tested this in Control with some printf-debugging to verify that we're indeed submitting the clear commands. We don't really have a good test for this otherwise.


doitsujin:
looks like clang not seem to support 64-bit CAS on x86 and fails to build.

Joshua-Ashton:
@doitsujin see: https://github.com/HansKristian-Work/vkd3d-proton/pull/517

libs/vkd3d/memory.c
```diff
@@ -637,15 +637,112 @@ static void vkd3d_memory_allocator_remove_chunk(struct vkd3d_memory_allocator *a
...
+    command_pool_info.pNext = NULL;
+    command_pool_info.flags = VK_COMMAND_POOL_CREATE_TRANSIENT_BIT;
+    command_pool_info.queueFamilyIndex = device->queues[VKD3D_QUEUE_FAMILY_INTERNAL_COMPUTE]->vk_family_index;
+
+    command_buffer_info.sType = VK_STRUCTURE_TYPE_COMMAND_BUFFER_ALLOCATE_INFO;
+    command_buffer_info.pNext = NULL;
+    command_buffer_info.level = VK_COMMAND_BUFFER_LEVEL_PRIMARY;
+    command_buffer_info.commandBufferCount = 1;
+
+    for (i = 0; i < VKD3D_MEMORY_CLEAR_COMMAND_BUFFER_COUNT; i++)
```
HansKristian-Work, line 691: There isn't much of a gain to use N command pools here. We can just as easily use 1 pool and just reset the command buffers individually.

HansKristian-Work: The queue code which does initial transitions uses that pattern as well fwiw, so might be good for consistency's sake.

libs/vkd3d/memory.c
```diff
@@ -659,9 +756,303 @@ void vkd3d_memory_allocator_cleanup(struct vkd3d_memory_allocator *allocator, st
...
+    /* Buffers are CONCURRENT due to D3D12 API restrictions,
+     * so we don't need to bother with ownership transfers */
+    memory_barrier.sType = VK_STRUCTURE_TYPE_MEMORY_BARRIER;
+    memory_barrier.pNext = NULL;
+    memory_barrier.srcAccessMask = VK_ACCESS_TRANSFER_WRITE_BIT;
+    memory_barrier.dstAccessMask = 0;
+
+    VK_CALL(vkCmdPipelineBarrier(vk_cmd_buffer,
+            VK_PIPELINE_STAGE_TRANSFER_BIT,
+            VK_PIPELINE_STAGE_BOTTOM_OF_PIPE_BIT,
```
HansKristian-Work, line 831: A memory barrier here is moot since we're using semaphores to synchronize the access with other queues.

libs/vkd3d/memory.c
```diff
@@ -659,9 +756,303 @@ void vkd3d_memory_allocator_cleanup(struct vkd3d_memory_allocator *allocator, st
...
+        return S_OK;
+
+    /* Record commands late so that we can simply remove allocations from
+     * the queue if they got freed before the clear commands got dispatched,
+     * rather than rewriting the command buffer or dispatching the clear */
+    internal_queue = device->queues[VKD3D_QUEUE_FAMILY_INTERNAL_COMPUTE];
+    vk_cmd_buffer = clear_queue->vk_command_buffers[clear_queue->command_buffer_index];
+    vk_cmd_pool = clear_queue->vk_command_pools[clear_queue->command_buffer_index];
+
+    if ((vr = VK_CALL(vkResetCommandPool(device->vk_device, vk_cmd_pool, 0))))
```
HansKristian-Work, line 797: Need to wait for timeline to complete before we can reset a command pool.

libs/vkd3d/device.c
```diff
@@ -1788,6 +1789,7 @@ static HRESULT d3d12_device_create_vkd3d_queues(struct d3d12_device *device,
                 &queue_info->vk_properties[i], &device->queues[i]))))
             goto out_destroy_queues;
 
+        device->unique_queue_mask |= 1u << i;
```
HansKristian-Work, line 1792: Should the internal compute queue be included here? From what I understand, this mask should only be used for user-visible queues since the memory clear code will do internal compute -> all user submissions.

libs/vkd3d/memory.c
```diff
@@ -659,9 +756,303 @@ void vkd3d_memory_allocator_cleanup(struct vkd3d_memory_allocator *allocator, st
...
+            allocation->chunk->allocation.clear_semaphore_value = clear_queue->next_signal_value;
+
+        clear_queue->allocations[clear_queue->allocations_count++] = allocation;
+        clear_queue->num_bytes_pending += allocation->resource.size;
+
+        if (clear_queue->num_bytes_pending >= VKD3D_MEMORY_CLEAR_QUEUE_MAX_PENDING_BYTES)
+            vkd3d_memory_allocator_flush_clears_locked(allocator, device);
+
+        pthread_mutex_unlock(&clear_queue->mutex);
+    }
```
HansKristian-Work, line 958: What if we suballocate from a DENY_BUFFERS heap? I guess there won't be any vk_buffer we can call vkCmdFillBuffer on?

libs/vkd3d/memory.c
```diff
@@ -659,9 +756,303 @@ void vkd3d_memory_allocator_cleanup(struct vkd3d_memory_allocator *allocator, st
...
+            return;
+        }
+    }
+
+    /* If this is a chunk and a suballocation from it had been immediately
+     * freed, it is possible that the suballocation got removed from the
+     * clear queue so that the chunk's wait value never gets signaled. Wait
+     * for the last signaled value in that case. */
+    if (wait_value == clear_queue->next_signal_value)
+        wait_value = clear_queue->next_signal_value - 1;
```
HansKristian-Work, line 1046: Where is wait_value used? Below, it's still allocation->clear_semaphore_value

libs/vkd3d/memory.c
```diff
@@ -659,9 +762,303 @@ void vkd3d_memory_allocator_cleanup(struct vkd3d_memory_allocator *allocator, st
...
+        {
+            ERR("Failed to insert free range.\n");
+            pthread_mutex_unlock(&clear_queue->mutex);
+            return;
+        }
+
+        allocation->clear_semaphore_value = clear_queue->next_signal_value;
+
+        if (allocation->chunk)
+            allocation->chunk->allocation.clear_semaphore_value = clear_queue->next_signal_value;
```
HansKristian-Work, line 949: I guess we don't need max here because next_signal_value is monotonically increasing?

HansKristian-Work (changes requested):
Comments. There is a -Watomic-align warning that seems to be generated by this PR somehow.

libs/vkd3d/memory.c
doitsujin: Don't we at least need to make sure that the writes from vkCmdFillBuffer actually becomes available?

libs/vkd3d/device.c
doitsujin: I've not excluded it here since we have to filter out the internal queue anyway. The internal queue *may* still point to a user queue so it doesn't make much of a difference anyway.

libs/vkd3d/memory.c
doitsujin: `    else if (allocation->resource.vk_buffer)` should take care of that. We don't handle image-only heaps, and I don't think we can unless we disable suballocation entirely for those.

libs/vkd3d/memory.c
doitsujin: Yes, and this happens while the mutex is locked, so there can't be any concurrency issues either.

libs/vkd3d/memory.c
HansKristian-Work: Signaling a semaphore makes all device writes available. Waiting for semaphore makes all device memory available.

doitsujin: Actually we do still need the barrier if this maps to a user queue since we will not wait on the timeline semaphore on that particular queue.

doitsujin: Changed the logic here so that we wait on all queues except the internal compute one if it is unique.

libs/vkd3d/memory.c
doitsujin: Fixed.

libs/vkd3d/device.c
doitsujin: Changed this regardless.

libs/vkd3d/memory.c
doitsujin: Fixed.

libs/vkd3d/memory.c
doitsujin: Fixed.

libs/vkd3d/device.c
```diff
@@ -1843,6 +1847,9 @@ static HRESULT vkd3d_select_queues(const struct vkd3d_instance *vkd3d_instance,
     info->family_index[VKD3D_QUEUE_FAMILY_SPARSE_BINDING] = vkd3d_find_queue(count, queue_properties,
             VK_QUEUE_SPARSE_BINDING_BIT, VK_QUEUE_SPARSE_BINDING_BIT);
 
+    /* TODO make this a low-priority compute queue that does not conflict with the actual compute queue */
+    info->family_index[VKD3D_QUEUE_FAMILY_INTERNAL_COMPUTE] = info->family_index[VKD3D_QUEUE_FAMILY_COMPUTE];
+
```
HansKristian-Work, line 1852: Do we intend to add
```

    if (info->family_index[VKD3D_QUEUE_FAMILY_COMPUTE] == VK_QUEUE_FAMILY_IGNORED || VKD3D_FORCE_SINGLE_QUEUE)
        info->family_index[VKD3D_QUEUE_FAMILY_COMPUTE] = info->family_index[VKD3D_QUEUE_FAMILY_GRAPHICS];
```
below?

HansKristian-Work (commented):
Just have one question.

libs/vkd3d/device.c
HansKristian-Work: I guess we *always* want to use async compute queue here, but we might have to handle the VK_QUEUE_FAMILY_IGNORED case at least?

doitsujin: We actually do need to handle the case where we don't have a compute queue, but not sure if we want to enforce single-queue mode here. Do you have a test case to check whether having a compute queue still leads to bad perf on RADV?

I've tested this in Control, but since the game doesn't constantly allocate and deallocate memory, it wouldn't trigger the issue.

HansKristian-Work: I think we want to always use async compute if we can here. I have a test case in Granite still. Can test it next time I swap out to AMD GPU.

doitsujin: I could test this on my end if you can share the Granite test.

HansKristian-Work (approved):
LGTM.

PR #517, opened 2021-02-16
build: Link against libatomic on x86 when using Clang

Joshua-Ashton:
Needed for 64-bit atomics on 32-bit architectures on Clang.


HansKristian-Work (approved):
LGTM

PR #518, opened 2021-02-16
Add heap suballocation stress test

HansKristian-Work:
Stress tests internal implementation details which we will hit internally with CreateCommittedResource and CreateHeap.

Creates many allocations, aims to either clear all of them to a fixed value, or does nothing, which should return 0.

Also verifies safety when attempting to free an idle ID3D12Heap, since it might be attempting to zero memory on a GPU queue.

### DIFF
--- a/tests/d3d12.c
+++ b/tests/d3d12.c
@@ -47850,6 +47850,227 @@ static void test_vrs(void)
     destroy_test_context(&context);
 }
 
+static void test_stress_suballocation(void)
+{
+#define SUBALLOC_TEST_NUM_BUFFERS 128
+#define SUBALLOC_TEST_NUM_ITERATIONS 64
+    ID3D12Resource *readback_buffers[SUBALLOC_TEST_NUM_BUFFERS] = { NULL };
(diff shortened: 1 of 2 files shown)


HansKristian-Work:
This test exposes existing bugs in our implementation since we're relying on driver zerovram, but recycled allocations don't seem to get zeroed on NV atm, which is perfectly reasonable driver behavior:

```
d3d12:48052: Test failed: Expected all words to be 0, but got 10.
d3d12:48052: Test failed: Expected all words to be 0, but got 23.
d3d12:48052: Test failed: Expected all words to be 0, but got 20.
d3d12:48052: Test failed: Expected all words to be 0, but got 33.
d3d12:48052: Test failed: Expected all words to be 0, but got 48.
d3d12:48052: Test failed: Expected all words to be 0, but got 50.
d3d12:48052: Test failed: Expected all words to be 0, but got 44.
```
The test passes on native D3D12 drivers. I expect it should pass on RADV as well.

doitsujin (approved):
LGTM.

PR #520, opened 2021-02-16
vkd3d: Use UINT64* instead of uint64_t* in 64-bit CAS.

HansKristian-Work:
Avoids alignment warnings on 32-bit.


Joshua-Ashton:
Shouldn't this need to be the actual type going into it from a structure that's aligned?

HansKristian-Work:
> 
> 
> Shouldn't this need to be the actual type going into it from a structure that's aligned?

Yes, the problem is that user code is creating a UINT64, this is cast to uint64_t*, and then the CAS loop uses uint64_t* instead, which confuses it since uint64_t doesn't guarantee alignment.

Joshua-Ashton:
Makes sense.

PR #523, opened 2021-02-18
vkd3d: Fix SetName for inline query types

Joshua-Ashton:

### DIFF
--- a/libs/vkd3d/resource.c
+++ b/libs/vkd3d/resource.c
@@ -6474,8 +6474,16 @@ static HRESULT STDMETHODCALLTYPE d3d12_query_heap_SetName(ID3D12QueryHeap *iface
 
     TRACE("iface %p, name %s.\n", iface, debugstr_w(name));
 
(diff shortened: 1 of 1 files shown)


libs/vkd3d/resource.c
```diff
@@ -6474,8 +6474,16 @@ static HRESULT STDMETHODCALLTYPE d3d12_query_heap_SetName(ID3D12QueryHeap *iface
 
     TRACE("iface %p, name %s.\n", iface, debugstr_w(name));
 
-    return vkd3d_set_vk_object_name(heap->device, (uint64_t)heap->vk_query_pool,
-            VK_OBJECT_TYPE_QUERY_POOL, name);
+    if (!d3d12_query_heap_type_is_inline(heap->desc.Type))
```
doitsujin, line 6477: Can we check whether vk_query_pool is non-null instead? That would remain correct even if we end up changing things again.

Joshua-Ashton: Done.

PR #524, opened 2021-02-18
Add more tests to stress allocation with multithreading

HansKristian-Work:
Also adds some minor test fixes while I was testing Windows behavior.


doitsujin:
The RT PSO thing looks weird, test should be good.

doitsujin (approved):

PR #526, opened 2021-02-19
vkd3d: Account for front buffer in swapchain image count

Joshua-Ashton:

### DIFF
--- a/libs/vkd3d/swapchain.c
+++ b/libs/vkd3d/swapchain.c
@@ -1358,12 +1358,13 @@ static HRESULT d3d12_swapchain_create_vulkan_swapchain(struct d3d12_swapchain *s
         return hresult_from_vk_result(vr);
     }
 
-    image_count = swapchain->desc.BufferCount;
+    /* Need to account for the front buffer, so + 1 */
+    image_count = swapchain->desc.BufferCount + 1;
     image_count = max(image_count, surface_caps.minImageCount);
     if (surface_caps.maxImageCount)
         image_count = min(image_count, surface_caps.maxImageCount);
 
-    if (image_count != swapchain->desc.BufferCount)
+    if (image_count != (swapchain->desc.BufferCount + 1))
     {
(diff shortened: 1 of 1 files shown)


libs/vkd3d/swapchain.c
```diff
@@ -1358,12 +1358,13 @@ static HRESULT d3d12_swapchain_create_vulkan_swapchain(struct d3d12_swapchain *s
...
 
-    image_count = swapchain->desc.BufferCount;
+    /* Need to account for the front buffer, so + 1 */
+    image_count = swapchain->desc.BufferCount + 1;
     image_count = max(image_count, surface_caps.minImageCount);
     if (surface_caps.maxImageCount)
         image_count = min(image_count, surface_caps.maxImageCount);
 
-    if (image_count != swapchain->desc.BufferCount)
+    if (image_count != (swapchain->desc.BufferCount + 1))
```
doitsujin, line 1367: Do we need this WARN at all?

IMO we should just do our thing here and stay quiet since the app never sees the effect of having too many Vulkan back buffers anyway.

doitsujin (commented):
Looks good, although I think we can just delete one rather useless log message.

libs/vkd3d/swapchain.c
HansKristian-Work: We don't now, since we always use userbuffers.

HansKristian-Work:
Should this depend on whether or not we're doing fullscreen? From https://docs.microsoft.com/en-us/windows/win32/api/dxgi/ns-dxgi-dxgi_swap_chain_desc#remarks:

> In full-screen mode, there is a dedicated front buffer; in windowed mode, the desktop is the front buffer.
> If you create a swap chain with one buffer, specifying DXGI_SWAP_EFFECT_SEQUENTIAL does not cause the contents of the single buffer to be swapped with the front buffer.

HansKristian-Work (approved):

PR #530, opened 2021-02-19
vkd3d: Fix integer underflow when checking for suitable free ranges.

doitsujin:
The difference between a range's offset and the aligned offset may be greater than the size of that range.


HansKristian-Work (approved):
LGTM.

PR #531, opened 2021-02-19
Suballocate committed textures if possible.

doitsujin:
Drastically reduces the number of allocations in some games, which helps RADV performance.

We still use dedicated allocations if the driver sets the corresponding hint, or if we cannot create a buffer on the same memory type to clear the allocation.

### DIFF
--- a/libs/vkd3d/resource.c
+++ b/libs/vkd3d/resource.c
@@ -2306,6 +2306,7 @@ HRESULT d3d12_resource_create_committed(struct d3d12_device *device, const D3D12
         const D3D12_HEAP_PROPERTIES *heap_properties, D3D12_HEAP_FLAGS heap_flags, D3D12_RESOURCE_STATES initial_state,
         const D3D12_CLEAR_VALUE *optimized_clear_value, struct d3d12_resource **resource)
 {
+    const struct vkd3d_vk_device_procs *vk_procs = &device->vk_procs;
     struct d3d12_resource *object;
     HRESULT hr;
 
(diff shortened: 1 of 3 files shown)


libs/vkd3d/resource.c
```diff
@@ -2329,11 +2355,32 @@ HRESULT d3d12_resource_create_committed(struct d3d12_device *device, const D3D12
...
+        }
+        else
+        {
+            /* We want to allow suballocations and we need the allocation to
+             * be cleared to zero, which only works if we allow buffers */
+            allocate_info.heap_flags &= ~D3D12_HEAP_FLAG_DENY_BUFFERS;
+            allocate_info.flags = VKD3D_ALLOCATION_FLAG_GLOBAL_BUFFER;
+        }
 
-        if (FAILED(hr = vkd3d_allocate_resource_memory(device,
```
HansKristian-Work, line 2334: I think vkd3d_allocate_resource_memory is unused after this.

doitsujin: Yes, that's why the second commit removes it.

HansKristian-Work: nvm, it's removed in next commit.

libs/vkd3d/resource.c
```diff
@@ -2329,11 +2355,32 @@ HRESULT d3d12_resource_create_committed(struct d3d12_device *device, const D3D12
...
+            dedicated_info.pNext = NULL;
+            dedicated_info.image = object->res.vk_image;
+            dedicated_info.buffer = VK_NULL_HANDLE;
+            allocate_info.pNext = &dedicated_info;
+        }
+        else
+        {
+            /* We want to allow suballocations and we need the allocation to
+             * be cleared to zero, which only works if we allow buffers */
+            allocate_info.heap_flags &= ~D3D12_HEAP_FLAG_DENY_BUFFERS;
```
HansKristian-Work, line 2370: For NON_ZERO_HEAP I think we don't need buffer memory at all for suballocation.

libs/vkd3d/resource.c
```diff
@@ -2329,11 +2355,32 @@ HRESULT d3d12_resource_create_committed(struct d3d12_device *device, const D3D12
...
+            allocate_info.flags = VKD3D_ALLOCATION_FLAG_GLOBAL_BUFFER;
```
HansKristian-Work, line 2371: Why FLAG_GLOBAL_BUFFER?

doitsujin: `vkd3d_allocate_memory` won't create a buffer otherwise.

HansKristian-Work: I guess it's there to request that we want the allocation to have a VkBuffer.

HansKristian-Work (approved):
LGTM

libs/vkd3d/resource.c
doitsujin: This requires some more thinking, probably best to leave it as-is for now.

PR #532, opened 2021-02-19
Mask some heap flags for chunk lookup

doitsujin:


HansKristian-Work (approved):
LGTM, to merge when CI passes

PR #533, opened 2021-02-19
vkd3d: Do not attempt to enable anything RT related.

HansKristian-Work:
Temporary workaround for release build to avoid causing vkCreateDevice failures on Ampere
with Steam Linux runtime.


doitsujin (approved):
Kind of agree with Josh that commenting out would be enough, but I guess this'll do as well for a release branch.

PR #537, opened 2021-02-22
Use virtual descriptor heap GPU VAs

HansKristian-Work:
This is required for DXR since we will need to be able to place table VAs directly in shader record buffers.
This is also a decent CPU optimization since we no longer need to make potentially cold derefs to resolve heap_offset when recording command buffers.

### DIFF
--- a/libs/vkd3d/device.c
+++ b/libs/vkd3d/device.c
@@ -4757,8 +4757,6 @@ static HRESULT d3d12_device_init(struct d3d12_device *device,
     HRESULT hr;
     int rc;
 
-    memset(device, 0, sizeof(*device));
-
 #ifdef VKD3D_ENABLE_PROFILING
     if (vkd3d_uses_profiling())
         device->ID3D12Device_iface.lpVtbl = &d3d12_device_vtbl_profiled;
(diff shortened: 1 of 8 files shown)


libs/vkd3d/device.c
```diff
@@ -2138,6 +2138,33 @@ void d3d12_device_return_scratch_buffer(struct d3d12_device *device, const struc
     }
 }
 
+uint64_t d3d12_device_get_descriptor_heap_gpu_va(struct d3d12_device *device)
+{
+    uint64_t va;
+
+    /* The virtual GPU descriptor VAs are of form (unique << 32) | (desc index * sizeof(d3d12_desc)),
```
doitsujin, line 2145: Can anything bad happen if these overlap with resource VAs?

HansKristian-Work: There is no way to use these VAs in any other context than as a table, so I don't see how.

HansKristian-Work: Technically we could always just return 0 for the top bits, but tools like apitrace will probably get very confused.

doitsujin (approved):

PR #544, opened 2021-02-25
Don't use VK_SHADER_STAGE_ALL for push constants

doitsujin:
Instead, gather the necessary bits from the root signature. May be more efficient on some drivers.


HansKristian-Work:
This conflicts a little with https://github.com/HansKristian-Work/vkd3d-proton/pull/546, but I think we can use restricted stages for graphics pipeline layout with this approach.

HansKristian-Work (approved):
LGTM, it won't be enough to deal with compute since that takes ALL visibility, but I'll rebase my PR to fix that.

PR #546, opened 2021-02-25
vkd3d: Redirect push constants to their bind point stages.

HansKristian-Work:
Gives a massive boost on NVIDIA for some reason.
RADV defers push constant update, so ALL_STAGES doesn't have
that much of a perf hit.

~20% uplift in RE2, ~5% uplift in CP77 from some quick and dirty testing.
Seems to be heavily content dependent either way.

Also a bug fix, since we would clobber graphics push constants from
compute and vice versa if both graphics and compute used the same root
signature.

### DIFF
--- a/libs/vkd3d/state.c
+++ b/libs/vkd3d/state.c
@@ -67,7 +67,9 @@ static void d3d12_root_signature_cleanup(struct d3d12_root_signature *root_signa
     vkd3d_sampler_state_free_descriptor_set(&device->sampler_state, device,
             root_signature->vk_sampler_set, root_signature->vk_sampler_pool);
 
-    VK_CALL(vkDestroyPipelineLayout(device->vk_device, root_signature->vk_pipeline_layout, NULL));
+    VK_CALL(vkDestroyPipelineLayout(device->vk_device, root_signature->graphics.vk_pipeline_layout, NULL));
+    VK_CALL(vkDestroyPipelineLayout(device->vk_device, root_signature->compute.vk_pipeline_layout, NULL));
+    VK_CALL(vkDestroyPipelineLayout(device->vk_device, root_signature->raygen.vk_pipeline_layout, NULL));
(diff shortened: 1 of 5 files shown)


libs/vkd3d/state.c
```diff
@@ -67,7 +67,9 @@ static void d3d12_root_signature_cleanup(struct d3d12_root_signature *root_signa
     vkd3d_sampler_state_free_descriptor_set(&device->sampler_state, device,
             root_signature->vk_sampler_set, root_signature->vk_sampler_pool);
 
-    VK_CALL(vkDestroyPipelineLayout(device->vk_device, root_signature->vk_pipeline_layout, NULL));
+    VK_CALL(vkDestroyPipelineLayout(device->vk_device, root_signature->graphics.vk_pipeline_layout, NULL));
+	VK_CALL(vkDestroyPipelineLayout(device->vk_device, root_signature->compute.vk_pipeline_layout, NULL));
+	VK_CALL(vkDestroyPipelineLayout(device->vk_device, root_signature->raygen.vk_pipeline_layout, NULL));
```
doitsujin, line 72: Weird indentation?

libs/vkd3d/state.c
```diff
@@ -972,7 +993,7 @@ static HRESULT d3d12_root_signature_init_global(struct d3d12_root_signature *roo
     const struct vkd3d_bindless_state *bindless_state = &device->bindless_state;
     VkDescriptorSetLayout set_layouts[VKD3D_MAX_DESCRIPTOR_SETS];
     struct vkd3d_descriptor_set_context context;
-    unsigned int i, push_constant_range_count;
+    unsigned int i;
```
doitsujin, line 996: :christmas_tree:

doitsujin:
LGTM besides one last formatting nit. Not a huge fan of having to do this, but.... what can we do.

libs/vkd3d/state.c

doitsujin (approved):

PR #548, opened 2021-03-01
Enable debug_utils conditionally, disable tracing for Release

Joshua-Ashton:
## Enable VK_EXT_debug_utils conditionally

Enabling VK_EXT_debug_utils comes at some overhead in Wine due to the object tracking required. There is also likely a non-zero overhead in some native implementations also.

By enabling this conditionally, we can also avoid additional overhead from apps that set debug labels on both the Vulkan and front-end side.

The default condition is to enable it when building with Renderdoc integration or in debug builds.

## Disable TRACE calls in release builds

It's needless overhead to have control flow at the start of every entrypoint. Branch predictors are good but there's no reason to keep what is essentially dead code behind an if sitting around (especially given we aren't marking it as unlikely).

## Other minor cleanup stuff to get compiling without warnings...

### DIFF
--- a/libs/vkd3d/device.c
+++ b/libs/vkd3d/device.c
@@ -75,7 +75,7 @@ struct vkd3d_optional_extension_info
 static const struct vkd3d_optional_extension_info optional_instance_extensions[] =
 {
     /* EXT extensions */
-    VK_EXTENSION(EXT_DEBUG_UTILS, EXT_debug_utils),
+    VK_EXTENSION_COND(EXT_DEBUG_UTILS, EXT_debug_utils, VKD3D_CONFIG_FLAG_DEBUG_UTILS),
 };
 
 static const struct vkd3d_optional_extension_info optional_device_extensions[] =
@@ -475,6 +475,7 @@ static const struct vkd3d_debug_option vkd3d_config_options[] =
     /* Enable Vulkan debug extensions. */
     {"vk_debug", VKD3D_CONFIG_FLAG_VULKAN_DEBUG},
     {"skip_application_workarounds", VKD3D_CONFIG_FLAG_SKIP_APPLICATION_WORKAROUNDS},
+    {"debug_utils", VKD3D_CONFIG_FLAG_DEBUG_UTILS},
 };
 
 static void vkd3d_config_flags_init_once(void)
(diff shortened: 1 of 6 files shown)


libs/vkd3d/device.c
```diff
@@ -70,8 +70,10 @@ struct vkd3d_optional_extension_info
 
 static const struct vkd3d_optional_extension_info optional_instance_extensions[] =
 {
+#ifdef VKD3D_ENABLE_DEBUG_UTILS
```
Themaister, line 73: Empty array fails MSVC.

HansKristian-Work, line 73: Empty array will fail MSVC.

Joshua-Ashton: Thanks for the heads up. I'll work on getting proper CI for MSVC soon so this can be avoided in the future.

HansKristian-Work (commented):
Makes sense in general, see MSVC issue. And there's lots of -Wunused warnings when ifdeffing out stuff, probably best to macro away the entire function call to set_debug_blah instead?

DadSchoorse:
Why disable `VK_EXT_debug_utils` completely at build time? Most of the benefits still apply if you disable it by default and enable it at runtime with `VKD3D_CONFIG=vk_debug`.

Joshua-Ashton:
@HansKristian-Work Those warnings should be solved by the commit after in `Simplify proeprties/feature tracing`

Joshua-Ashton:
@DadSchoorse My idea was also to eliminate the need to check the UUID of stuff in SetPrivateData this way and eliminate all the name code.

Joshua-Ashton:
I think it's fine to probably do that on run-time on second thought -- I will just do that.

HansKristian-Work:
Please don't tie it to vk_debug. vk_debug adds validation layers as well. If we want to capture on Windows for example, we should only enable debug_utils, but not validation.

doitsujin:
I think we should just have a simple environment variable to enable debug_utils stuff. Unlike the `TRACE` changes there isn't much of a reason to skip over the code entirely.

HansKristian-Work:
> I think we should just have a simple environment variable to enable debug_utils stuff. Unlike the `TRACE` changes there isn't much of a reason to skip over the code entirely.

VKD3D_CONFIG seems up for the task for that kind of boolean flag.

doitsujin:
Right, I didn't see your comment yet as I was typing this. But yeah, shouldn't be a compile-time option.

HansKristian-Work:
> @HansKristian-Work Those warnings should be solved by the commit after in `Simplify proeprties/feature tracing`, I will re-order these.

Ah, right. I expected -Wunused-parameter to trigger, but it doesn't seem like it, only unused stack variables trigger the warning.

HansKristian-Work (approved):
LGTM.

PR #550, opened 2021-03-03
Implement DispatchRays

HansKristian-Work:
Can successfully run the (albeit simple) RT test on NV now. :tada:

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -4072,6 +4072,7 @@ static bool d3d12_command_list_update_compute_pipeline(struct d3d12_command_list
 static bool d3d12_command_list_update_raygen_pipeline(struct d3d12_command_list *list)
 {
     const struct vkd3d_vk_device_procs *vk_procs = &list->device->vk_procs;
+    bool stack_size_dirty = false;
 
     if (list->current_pipeline != VK_NULL_HANDLE)
         return true;
@@ -4088,12 +4089,20 @@ static bool d3d12_command_list_update_raygen_pipeline(struct d3d12_command_list 
                 VK_PIPELINE_BIND_POINT_RAY_TRACING_KHR,
                 list->rt_state->pipeline));
         list->command_buffer_pipeline = list->rt_state->pipeline;
+        stack_size_dirty = true;
+    }
+    else
+    {
+        stack_size_dirty = list->dynamic_state.pipeline_stack_size != list->rt_state->pipeline_stack_size;
     }
 
-    /* Pipeline stack size is part of the PSO, not any command buffer state
-     * for some reason ... */
-    VK_CALL(vkCmdSetRayTracingPipelineStackSizeKHR(list->vk_command_buffer,
-            list->rt_state->pipeline_stack_size));
+    if (stack_size_dirty)
+    {
+        /* Pipeline stack size is part of the PSO, not any command buffer state for some reason ... */
+        VK_CALL(vkCmdSetRayTracingPipelineStackSizeKHR(list->vk_command_buffer,
+                list->rt_state->pipeline_stack_size));
+        list->dynamic_state.pipeline_stack_size = list->rt_state->pipeline_stack_size;
+    }
 
     return true;
 }
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -6319,11 +6319,10 @@ static void d3d12_command_list_set_root_signature(struct d3d12_command_list *lis
 
         case VK_PIPELINE_BIND_POINT_COMPUTE:
             bindings->layout = root_signature->compute;
+            bindings->rt_layout = root_signature->raygen;
             break;
 
         default:
-            /* TODO, RT will be relevant here later somehow.
-             * It will get awkward since RayGen happens in compute on DXR. */
             break;
     }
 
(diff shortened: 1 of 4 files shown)


libs/vkd3d/command.c
```diff
@@ -4069,6 +4069,42 @@ static bool d3d12_command_list_update_compute_pipeline(struct d3d12_command_list
...
+
+    if (list->command_buffer_pipeline != list->rt_state->pipeline)
+    {
+        VK_CALL(vkCmdBindPipeline(list->vk_command_buffer,
+                VK_PIPELINE_BIND_POINT_RAY_TRACING_KHR,
+                list->rt_state->pipeline));
+        list->command_buffer_pipeline = list->rt_state->pipeline;
+        /* Ensures we update the pipeline stack size.
+         * It is somewhat overkill to introduce this to the dynamic state bitmask system. */
+        list->dynamic_state.pipeline_stack_size = UINT32_MAX;
```
Joshua-Ashton, line 4093: I am not a fan of this, why not just move the `vkCmdSetRayTracingPipelineStackSizeKHR` into this block and scrap the `dynamic_state` tracking of `pipeline_stack_size`entirely?

D3D12 only has this in PSO state so, idk why it's being tracked in dynamic here when we can resolve this at the same time we bind the pipeline.

Joshua-Ashton (changes requested):

libs/vkd3d/command.c
HansKristian-Work: The stack size can change on the fly after binding the PSO, so we have to call SetPipelineStackSize every time. I guess it can't hurt all that much.

Joshua-Ashton: Oh it can? I wasn't aware of that. Makes sense then.

libs/vkd3d/resource.c
```diff
@@ -3502,6 +3502,22 @@ static void vkd3d_create_buffer_srv(struct d3d12_desc *descriptor,
...
+        /* We implement this as a raw VA in the aux buffer. */
+        if (device->bindless_state.flags & VKD3D_RAW_VA_AUX_BUFFER)
+        {
+            VkDeviceAddress *counter_addresses = descriptor->heap->raw_va_aux_buffer.host_ptr;
+            uint32_t descriptor_index = d3d12_desc_heap_offset(descriptor);
+            counter_addresses[descriptor_index] = desc->RaytracingAccelerationStructure.Location;
+            descriptor->metadata.flags = VKD3D_DESCRIPTOR_FLAG_RAW_VA_AUX_BUFFER;
+        }
+        else
+            FIXME("Using CreateSRV for RTAS without RAW_VA_AUX_BUFFER support?\n");
```
doitsujin, line 3516: We're only enabling RT support if the aux buffer thing is supported, right? This should be `WARN` then since it'll only trigger with broken apps.

libs/vkd3d/resource.c
```diff
@@ -3502,6 +3502,22 @@ static void vkd3d_create_buffer_srv(struct d3d12_desc *descriptor,
...
 
+    if (desc->ViewDimension == D3D12_SRV_DIMENSION_RAYTRACING_ACCELERATION_STRUCTURE)
+    {
+        /* We implement this as a raw VA in the aux buffer. */
+        if (device->bindless_state.flags & VKD3D_RAW_VA_AUX_BUFFER)
+        {
+            VkDeviceAddress *counter_addresses = descriptor->heap->raw_va_aux_buffer.host_ptr;
+            uint32_t descriptor_index = d3d12_desc_heap_offset(descriptor);
+            counter_addresses[descriptor_index] = desc->RaytracingAccelerationStructure.Location;
+            descriptor->metadata.flags = VKD3D_DESCRIPTOR_FLAG_RAW_VA_AUX_BUFFER;
```
doitsujin, line 3513: We should reset `cookie` and `set_info_mask` as well or otherwise descriptor copies will end up being weird.

libs/vkd3d/command.c
```diff
@@ -4183,11 +4182,11 @@ static void d3d12_command_list_update_descriptor_table_offsets(struct d3d12_comm
...
 
     /* Set descriptor offsets */
-    if (bindings->layout.vk_push_stages)
+    if (push_stages)
     {
         VK_CALL(vkCmdPushConstants(list->vk_command_buffer,
-                bindings->layout.vk_pipeline_layout,
-                bindings->layout.vk_push_stages,
+                layout,
+                push_stages,
```
doitsujin, line 4218: weird newline

libs/vkd3d/command.c
```diff
@@ -8311,10 +8354,51 @@ static void STDMETHODCALLTYPE d3d12_command_list_SetPipelineState1(d3d12_command
...
+    VkStridedDeviceAddressRegionKHR callable_table;
+    VkStridedDeviceAddressRegionKHR raygen_table;
+    VkStridedDeviceAddressRegionKHR miss_table;
+    VkStridedDeviceAddressRegionKHR hit_table;
+
+    TRACE("iface %p, desc %p\n", iface, desc);
+
+    if (!d3d12_device_supports_ray_tracing_tier_1_0(list->device))
+    {
+        ERR("Ray tracing is not supported. Calling this is invalid.\n");
```
doitsujin, line 8381: `WARN` because app bug

doitsujin (changes requested):
See comments, the descriptor update changes make sense and overall this looks less annoying than anticipated.

libs/vkd3d/resource.c
HansKristian-Work: Fixed.

libs/vkd3d/command.c

libs/vkd3d/command.c

Joshua-Ashton (approved):
LGTM after revision

libs/vkd3d/resource.c

HansKristian-Work: Yes, drivers cannot enable RT without BDA support.

doitsujin (approved):
LGTM.

PR #551, opened 2021-03-03
build: Use --file-alignment=4096 with MinGW

Joshua-Ashton:
Avoids a copy in the Wine loader as well as enables debug symbols to work in perf.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -125,6 +125,14 @@ if cpu_family == 'x86'
   endif
 endif
 
+if not vkd3d_is_msvc
+  # We need to set the section alignment for debug symbols to
+  # work properly as well as avoiding a memcpy from the Wine loader.
+  if vkd3d_compiler.has_link_argument('-Wl,--file-alignment=4096')
+    add_global_link_arguments('-Wl,--file-alignment=4096', language : [ 'c', 'cpp' ])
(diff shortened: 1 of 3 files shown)


meson.build
```diff
@@ -119,12 +119,20 @@ if cpu_family == 'x86'
...
     lib_atomic = vkd3d_compiler.find_library('atomic')
     vkd3d_extra_libs += lib_atomic
   endif
 endif
 
+if not vkd3d_is_msvc
+  # We need to set the section alignment for debug symbols to
+  # work properly as well as avoiding a memcpy from the Wine loader.
+  if vkd3d_compiler.has_link_argument('-Wl,--file-alignment=4096')
+    add_global_link_arguments('-Wl,--file-alignment=4096', language: 'cpp')
```
DadSchoorse, line 132: vkd3d-proton uses C as well

Joshua-Ashton: Resolved.

HansKristian-Work: The linker has to use C++ due to dxil-spirv though.

Joshua-Ashton: I set it for both C and C++

HansKristian-Work (approved):
LGTM

PR #553, opened 2021-03-04
vkd3d: Silence log spam around Min LOD Clamp.

doitsujin:
This seriously hurts performance in AC:Valhalla.


HansKristian-Work (approved):
LGTM

PR #554, opened 2021-03-04
Implement more DXR features and tests

HansKristian-Work:
- Implement RTAS copy operations for CLONE and COMPACT. Serialization is FIXME, but according to docs, it's only supposed to work in developer mode OS and PIX, so ... :shrug:
- Verifies that all required VBO formats and IBO formats work as expected.

Cleans up the ray tracing test a bit to make it a little less insanely hardcoded.

### DIFF
--- a/tests/d3d12.c
+++ b/tests/d3d12.c
@@ -41695,15 +41695,22 @@ static void test_stencil_export_dxil(void)
 
 static void test_raytracing(void)
 {
+#define NUM_GEOM_DESC 6
+#define NUM_UNMASKED_INSTANCES 4
+#define INSTANCE_OFFSET_Y (100.0f)
+#define GEOM_OFFSET_X (10.0f)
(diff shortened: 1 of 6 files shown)


tests/d3d12.c
```diff
@@ -41695,15 +41695,22 @@ static void test_stencil_export_dxil(void)
 
 static void test_raytracing(void)
 {
+#define NUM_GEOM_DESC 6
```
Joshua-Ashton, line 41698: const

Joshua-Ashton (changes requested):
Minor nit, otherwise LGTM -- although I didn't review the tests in-depth.

doitsujin (approved):

tests/d3d12.c
Joshua-Ashton: Okay, nevermind, can't do that in C :v

Joshua-Ashton (approved):

PR #557, opened 2021-03-05
Implement descriptor table hoisting through RS 1.1

HansKristian-Work:
Implement RS 1.1 STATIC semantics, and add config flags to override this if the application happens to obey these rules.
We only add hoisting for a descriptor range if the descriptor range count is 1, since we cannot deal correctly with arrays.
We only hoist if descriptor range count is 1, so we should avoid all scenarios where the application truly wants update-after-bind behavior.

When hoisting, we pretend that we have extra root parameters in the root signature. When we update descriptor table offsets, we poke out the VkBufferInfo from the descriptor heap and turn that into a root descriptor. We always need to use descriptors for robustness, so no BDA is possible sadly. It would also greatly complicate the implementation.

From trivial testing I got +10% FPS on DS with CBV hoist on Ampere ... I expect Pascal would fare even better here.


doitsujin:
> We only add hoisting for a descriptor range if the descriptor range count is 1, since we cannot deal correctly with arrays.

Unfortunately this isn't quite enough to know whether we're dealing with descriptor arrays or not, since the descriptor ranges in the root signature don't have to match the descriptor ranges used in the shader. It seems to be legal to specify two consecutive CBV descriptors in the root signature but access them as a single array in the shader, and vice versa. There are ways to get this right by just using the bindless path in the shader if we have to, since both are available.

Anyway, I think we should really sort out the root signature v1.1 stuff first so that we *can* actually check for STATIC flags, and then enable any technically illegal perf hacks afterwards.

HansKristian-Work:
> Unfortunately this isn't quite enough to know whether we're dealing with descriptor arrays or not, since the descriptor ranges in the root signature don't have to match the descriptor ranges used in the shader. It seems to be legal to specify two consecutive CBV descriptors in the root signature but access them as a single array in the shader, and vice versa.

Wat ... I guess we'll need tests for this.

doitsujin (changes requested):
As discussed on Discord, I'd kind of like the SRV/UAV-specific code on the DXIL/DBXC side to be dropped since we don't plan to support it, and the whole `can_hoist` logic there is somewhat complicated.

HansKristian-Work:
Dropped the can_hoist cruft.

libs/vkd3d/command.c
```diff
@@ -4527,6 +4576,10 @@ static void d3d12_command_list_update_descriptors(struct d3d12_command_list *lis
     if (bindings->dirty_flags & VKD3D_PIPELINE_DIRTY_STATIC_SAMPLER_SET)
         d3d12_command_list_update_static_samplers(list, bindings, vk_bind_point, layout);
 
+    /* If we can, hoist descriptors from the descriptor heap into fake root parameters. */
+    if (bindings->dirty_flags & VKD3D_PIPELINE_DIRTY_DESCRIPTOR_TABLE_OFFSETS)
```
doitsujin, line 4580: I think we need a separate flag here. `VKD3D_PIPELINE_DIRTY_DESCRIPTOR_TABLE_OFFSETS` will not be set when the app calls SetDescriptorHeaps, but we want to update hoisted descriptors in that case.

Something like `VKD3D_PIPELINE_DIRTY_STATIC_DESCRIPTORS` should do, then set that when changing the CBV_SRV_UAV heap or when setting descriptor tables.

libs/vkd3d/vkd3d_private.h
```diff
@@ -1142,6 +1143,20 @@ struct d3d12_bind_point_layout
     VkShaderStageFlags vk_push_stages;
 };
 
+#define VKD3D_MAX_HOISTED_DESCRIPTORS 4
```
doitsujin, line 1146: Do we know how many CBVs apps typically use in a pipeline? 4 seems very low; at least in D3D11 it's not uncommon to have 8-12 CBVs used in the vertex shader and then another bunch in the fragment shader.

Even if we end up triggering the descriptor set fallback (rather than push descriptors) when exceeding driver limits, this should still provide an uplift on NV.

doitsujin (commented):
Looks mostly good except the flag thing + 1 question.

libs/vkd3d/vkd3d_private.h
HansKristian-Work: I guess we can bump this. I am a bit worried about push descriptor spam if root signature has a lot of descriptors and shaders only use a few anyways. 16 sounds fair?

libs/vkd3d/command.c

doitsujin (approved):

PR #566, opened 2021-03-11
Allow enabling RT tier 1.0.

HansKristian-Work:
Control seems to work.

The implementation is not complete enough to unconditionally enable it apparently, since we crash UE4 games when enabling RT 1.0.


libs/vkd3d/state.c
```diff
@@ -1050,6 +1052,9 @@ static HRESULT d3d12_root_signature_init_local(struct d3d12_root_signature *root
         goto fail;
     }
 
+    root_signature->binding_count = info.binding_count;
```
Joshua-Ashton, line 1055: Even though static samplers are unsupported right now on this path for us, I think they are meant to be.

So maybe just set that count from the desc here anyway so we don't forget in the future? :)

Joshua-Ashton (changes requested):
Other than minor nit, LGTM.

libs/vkd3d/state.c
HansKristian-Work: Oh right, I completely forgot about static samplers in local RS. :V Can just add a FIXME ...

vityafx:
Sorry for chiming in, but are all games expected to understand that the video card is capable of ray tracing now? In my case, cyberpunk 2077 didn't pick up the ray-tracing support, but Control did and works quite well.

HansKristian-Work:
> 
> 
> Sorry for chiming in as it is not really related to this pull request, but are all games expected to understand that the video card is capable of ray tracing now? In my case, cyberpunk 2077 didn't pick up the ray-tracing support, but Control did and works quite well.
> I could also test a few other games should you like. Given my hardware, I can also test the performance quite well, can export the mangohud's stats.

The implementation is the bare minimum that happened to get Control to run, don't expect other content to work yet.

libs/vkd3d/state.c
HansKristian-Work: Fixed.

doitsujin:
> In my case, cyberpunk 2077 didn't pick up the ray-tracing support

It's likely that Cyberpunk requires access to Nvidia's Windows driver to expose Raytracing. Reminder that RT in that game does not work on AMD GPUs either, unlike in Control.

libs/vkd3d/device.c
```diff
@@ -4432,21 +4432,20 @@ static D3D12_RAYTRACING_TIER d3d12_device_determine_ray_tracing_tier(struct d3d1
...
         VK_FORMAT_R16G16_SNORM,
-        /* DXR specifies RGBA16 here, but it also completely ignores A,
-         * so it's *actually* DXGI_FORMAT_R16G16B16_FLOAT/SNORM,
-         * but those formats don't exist in D3D,
-         * and they couldn't be bothered to add those apparently <_<. */
-        VK_FORMAT_R16G16B16_SFLOAT,
-        VK_FORMAT_R16G16B16_SNORM,
+        /* DXR specifies RGBA16 here, but it also completely ignores A.
+         * So does Vulkan, and you can technically load out of bounds with this,
+         * but behavior in DXR and Vulkan is actually supposed to match here, so ... */
```
doitsujin, line 4447: Do we need this comment anymore? It's a 1:1 mapping with no real quirks attached.

doitsujin:
LGTM otherwise.

Joshua-Ashton (approved):
LGTM

libs/vkd3d/device.c
HansKristian-Work: Fixed.

PR #572, opened 2021-03-13
vkd3d: Implement ClearState.

HansKristian-Work:

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -3770,7 +3770,7 @@ bool d3d12_command_list_reset_query(struct d3d12_command_list *list,
     return true;
 }
 
-static void d3d12_command_list_reset_state(struct d3d12_command_list *list,
+static void d3d12_command_list_reset_api_state(struct d3d12_command_list *list,
         ID3D12PipelineState *initial_pipeline_state)
 {
     d3d12_command_list_iface *iface = &list->ID3D12GraphicsCommandList_iface;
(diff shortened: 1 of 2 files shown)


libs/vkd3d/command.c
```diff
@@ -3916,12 +3916,12 @@ static HRESULT STDMETHODCALLTYPE d3d12_command_list_Reset(d3d12_command_list_ifa
...
-static HRESULT STDMETHODCALLTYPE d3d12_command_list_ClearState(d3d12_command_list_iface *iface,
+static void STDMETHODCALLTYPE d3d12_command_list_ClearState(d3d12_command_list_iface *iface,
         ID3D12PipelineState *pipeline_state)
 {
-    FIXME("iface %p, pipline_state %p stub!\n", iface, pipeline_state);
-
-    return E_NOTIMPL;
+    struct d3d12_command_list *list = impl_from_ID3D12GraphicsCommandList(iface);
+    TRACE("iface %p, pipline_state %p!\n", iface, pipeline_state);
+    d3d12_command_list_reset_state(list, pipeline_state);
```
doitsujin, line 3924: This doesn't quite work, as `d3d12_command_list_reset_state` affects state that isn't D3D state but rather some internal tracking that needs to remain consistent (such as whether we are inside a render pass). I think we need to split that function into two.

doitsujin (changes requested):

libs/vkd3d/command.c

doitsujin (approved):

PR #574, opened 2021-03-15
Re-enable multi-queue mode

doitsujin:
Locked behind a config flag for now because there are several caveats:
- Initial resource transitions are still not handled correctly in a simultaneous-use scenario.
- The fence worker currently does not handle out-of-order signals very well, which introduces GPU bubbles in some games.
- Using `SHARING_MODE_CONCURRENT` for all images may cause performance issues on some GPUs.


libs/vkd3d/device.c
```diff
@@ -478,6 +478,7 @@ static const struct vkd3d_debug_option vkd3d_config_options[] =
     {"debug_utils", VKD3D_CONFIG_FLAG_DEBUG_UTILS},
     {"force_static_cbv", VKD3D_CONFIG_FLAG_FORCE_STATIC_CBV},
     {"dxr", VKD3D_CONFIG_FLAG_DXR},
+    {"multi_queue", VKD3D_CONFIG_FLAG_MULTI_QUEUE},
```
Saancreed, line 481: Might be a good idea to document that in the README.

libs/vkd3d/memory.c
```diff
@@ -743,6 +743,9 @@ HRESULT vkd3d_memory_allocator_init(struct vkd3d_memory_allocator *allocator, st
     }
 
     vkd3d_va_map_init(&allocator->va_map);
+
+    allocator->vkd3d_queue = d3d12_device_allocate_vkd3d_queue(device,
```
HansKristian-Work, line 747: Is this queue ever released?

libs/vkd3d/memory.c
```diff
@@ -914,18 +916,21 @@ static HRESULT vkd3d_memory_allocator_flush_clears_locked(struct vkd3d_memory_al
     while (queue_mask)
     {
         queue_index = vkd3d_bitmask_iter32(&queue_mask);
-        queue = device->queues[queue_index];
+        queue_family = device->queue_families[queue_index];
 
-        if (!(vk_queue = vkd3d_queue_acquire(queue)))
-            return E_FAIL;
+        for (i = 0; i < queue_family->queue_count; i++)
```
HansKristian-Work, line 921: I think we need to move to a model where submission threads wait on the allocator timeline on-demand rather than broadcasting to ~10+ queues every time we zero memory. Could make this a TODO.

libs/vkd3d/device.c
HansKristian-Work: Yes, but can wait until it's more complete.

HansKristian-Work (changes requested):
Small nits.

libs/vkd3d/memory.c
doitsujin: No, because the memory allocator gets destroyed after the queues.

libs/vkd3d/memory.c
doitsujin: Foxed.

libs/vkd3d/memory.c
HansKristian-Work: I suppose allocate_vkd3d_queue just bumps the virtual queue count, and doesn't really *own* anything anyways?

libs/vkd3d/memory.c
HansKristian-Work: 🦊

HansKristian-Work (approved):
LGTM

PR #578, opened 2021-03-15
Create one fence worker per command queue

doitsujin:
Fixes GPU bubbles in some games with multi-queue mode enabled.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -8844,6 +8844,7 @@ static ULONG STDMETHODCALLTYPE d3d12_command_queue_Release(ID3D12CommandQueue *i
         vkd3d_private_store_destroy(&command_queue->private_store);
 
         d3d12_command_queue_submit_stop(command_queue);
+        vkd3d_fence_worker_stop(&command_queue->fence_worker, device);
         d3d12_device_unmap_vkd3d_queue(device, command_queue->vkd3d_queue);
         pthread_join(command_queue->submission_thread, NULL);
         pthread_mutex_destroy(&command_queue->queue_lock);
@@ -9557,7 +9558,7 @@ static void d3d12_command_queue_signal(struct d3d12_command_queue *command_queue
         return;
     }
 
-    if (FAILED(hr = vkd3d_enqueue_timeline_semaphore(&device->fence_worker, fence, physical_value, vkd3d_queue)))
+    if (FAILED(hr = vkd3d_enqueue_timeline_semaphore(&command_queue->fence_worker, fence, physical_value, vkd3d_queue)))
     {
         /* In case of an unexpected failure, try to safely destroy Vulkan objects. */
         vkd3d_queue_wait_idle(vkd3d_queue, vk_procs);
@@ -10394,6 +10395,9 @@ static HRESULT d3d12_command_queue_init(struct d3d12_command_queue *queue,
 
     d3d12_device_add_ref(queue->device = device);
 
+    if (FAILED(hr = vkd3d_fence_worker_start(&queue->fence_worker, device)))
+        goto fail_fence_worker_start;
+
     if ((rc = pthread_create(&queue->submission_thread, NULL, d3d12_command_queue_submission_worker_main, queue)) < 0)
     {
         d3d12_device_release(queue->device);
@@ -10403,7 +10407,9 @@ static HRESULT d3d12_command_queue_init(struct d3d12_command_queue *queue,
 
     return S_OK;
 
-fail_pthread_create:;
+fail_pthread_create:
+    vkd3d_fence_worker_stop(&queue->fence_worker, device);
+fail_fence_worker_start:;
 #ifdef VKD3D_BUILD_STANDALONE_D3D12
 fail_swapchain_factory:
     vkd3d_private_store_destroy(&queue->private_store);
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -288,79 +288,29 @@ static HRESULT vkd3d_enqueue_timeline_semaphore(struct vkd3d_fence_worker *worke
     waiting_fence->queue = queue;
     ++worker->enqueued_fence_count;
 
-    InterlockedIncrement(&fence->pending_worker_operation_count);
-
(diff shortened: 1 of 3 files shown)


libs/vkd3d/command.c
```diff
@@ -408,6 +380,7 @@ static void vkd3d_wait_for_gpu_timeline_semaphores(struct vkd3d_fence_worker *wo
             if (FAILED(hr = d3d12_fence_signal(current->fence, counter_value)))
                 ERR("Failed to signal D3D12 fence, hr %#x.\n", hr);
 
+            d3d12_fence_dec_ref(current->fence);
```
HansKristian-Work: Should dec_ref *after* the InterlockedDecrement below perhaps?

libs/vkd3d/command.c
```diff
@@ -408,6 +380,7 @@ static void vkd3d_wait_for_gpu_timeline_semaphores(struct vkd3d_fence_worker *wo
...
             InterlockedDecrement(&current->fence->pending_worker_operation_count);
```
HansKristian-Work: Is pending_worker_operation_count meaningful anymore?

libs/vkd3d/command.c
```diff
@@ -772,6 +772,9 @@ static HRESULT d3d12_fence_signal(struct d3d12_fence *fence, uint64_t physical_v
         return hresult_from_errno(rc);
     }
 
+    if (fence->physical_value > physical_value)
+        return S_OK;
```
HansKristian-Work, line 688: I guess this can theoretically happen if we submit Signal(1) -> Wait(1) -> Signal(2) on different queue threads, and we observe the signal 2 before signal 1?

HansKristian-Work: Probably a good idea to comment how this can happen.

libs/vkd3d/command.c
HansKristian-Work: Seems like it's removed in the following commits.

libs/vkd3d/command.c
```diff
@@ -288,164 +288,101 @@ static HRESULT vkd3d_enqueue_timeline_semaphore(struct vkd3d_fence_worker *worke
...
+    wait_info.flags = 0;
+    wait_info.semaphoreCount = 1;
+    wait_info.pSemaphores = &fence->vk_semaphore;
+    wait_info.pValues = &fence->waiting_fence.value;
+
+    do
+    {
+        vr = VK_CALL(vkWaitSemaphoresKHR(device->vk_device, &wait_info, ~(uint64_t)0));
+    }
+    while (vr == VK_TIMEOUT);
```
HansKristian-Work, line 316: TIMEOUT cannot happen with UINT64_MAX.

libs/vkd3d/command.c
```diff
@@ -288,164 +288,101 @@ static HRESULT vkd3d_enqueue_timeline_semaphore(struct vkd3d_fence_worker *worke
...
+        ERR("Failed to wait for Vulkan timeline semaphore, vr %d.\n", vr);
         return;
     }
 
     /* This is a good time to kick the debug ring thread into action. */
     if (device->debug_ring.active)
         pthread_cond_signal(&device->debug_ring.ring_cond);
 
-    for (i = 0, j = 0; i < worker->fence_count; ++i)
+    if ((vr = VK_CALL(vkGetSemaphoreCounterValueKHR(device->vk_device, fence->vk_semaphore, &counter_value))) < 0)
```
HansKristian-Work, line 328: Is it even meaningful to get the counter value? We should just signal d3d12_fence_signal with the physical value we waited on.

libs/vkd3d/command.c
```diff
@@ -288,164 +288,101 @@ static HRESULT vkd3d_enqueue_timeline_semaphore(struct vkd3d_fence_worker *worke
...
-                }
-            }
+        pthread_mutex_unlock(&worker->mutex);
 
-            pthread_mutex_unlock(&worker->mutex);
-        }
+        for (i = 0; i < fence_count; i++)
+            vkd3d_wait_for_gpu_timeline_semaphore(worker, &fences[i]);
+
+        vkd3d_free(fences);
```
HansKristian-Work, line 382: We should probably just pop one fence per iteration to avoid forcing allocation on every single wait.

HansKristian-Work (changes requested):
Comments.

libs/vkd3d/command.c
doitsujin: Should we still fix this given that the entire function gets reworked in the next commit anyway?

libs/vkd3d/command.c
doitsujin: Yeah, can remove. For some reason the old code was also handling timeouts.

libs/vkd3d/command.c
doitsujin: Yeah, will fix. I guess the old code did it to check which of the fences should be signalled?

libs/vkd3d/command.c
doitsujin: Done.

libs/vkd3d/command.c
doitsujin: Fixed, but this gets reworked in the following commits anyway.

libs/vkd3d/command.c
doitsujin: Went with the swap thing.

libs/vkd3d/command.c
```diff
@@ -772,9 +772,6 @@ static HRESULT d3d12_fence_signal(struct d3d12_fence *fence, uint64_t physical_v
         return hresult_from_errno(rc);
     }
 
-    /* With multiple fence workers, it is possible that signal calls are
```
HansKristian-Work: Was it the intention to remove this comment here?

HansKristian-Work (changes requested):
Just a tiny nit with comment.

libs/vkd3d/command.c
doitsujin: Ah, no, must have screwed up while rebasing.

HansKristian-Work (approved):
LGTM!

Issue #568, opened 2021-03-12
Issue running simple Microsoft DXR SDK samples using https://github.com/HansKristian-Work/vkd3d-proton/pull/566..


oscarbg:
Hi,
first congrats for success in running Control with DXR enabled!
sadly, I get no success, just trying some DXR samples from Microsoft SDK: https://github.com/microsoft/DirectX-Graphics-Samples/tree/master/Samples/Desktop/D3D12Raytracing
I attach prebuilt binaries for fast testing:
[dxrsamples.zip](https://github.com/HansKristian-Work/vkd3d-proton/files/6127686/dxrsamples.zip)
using a build from: https://github.com/HansKristian-Work/vkd3d-proton/pull/566
I'm getting in all samples same error:
```
15152:err:vkd3d_dxil_log_callback: dxil-spirv: Failed to convert function.
15152:err:d3d12_state_object_compile_pipeline: Failed to convert DXIL export: MyRaygenShader
```
hope is a minor issue and getting it solved helps running more DXR apps..

full log:
```
warn:  DXGIDeclareAdapterRemovalSupport: Stub
info:  Game: D3D12RaytracingHelloWorld.exe
info:  DXVK: v1.8
info:  Built-in extension providers:
info:    Win32 WSI
info:    OpenVR
info:    OpenXR
warn:  OpenVR: Failed to locate module
info:  Enabled instance extensions:
info:    VK_KHR_get_surface_capabilities2
info:    VK_KHR_surface
info:    VK_KHR_win32_surface
info:  TITAN V:
info:    Driver: 461.81.0
info:    Vulkan: 1.2.155
info:    Memory Heap[0]: 
info:      Size: 12079 MiB
info:      Flags: 0x1
info:      Memory Type[7]: Property Flags = 0x1
info:    Memory Heap[1]: 
info:      Size: 16362 MiB
info:      Flags: 0x0
info:      Memory Type[0]: Property Flags = 0x0
info:      Memory Type[1]: Property Flags = 0x0
info:      Memory Type[2]: Property Flags = 0x0
info:      Memory Type[3]: Property Flags = 0x0
info:      Memory Type[4]: Property Flags = 0x0
info:      Memory Type[5]: Property Flags = 0x0
info:      Memory Type[6]: Property Flags = 0x0
info:      Memory Type[8]: Property Flags = 0x6
info:      Memory Type[9]: Property Flags = 0xe
info:    Memory Heap[2]: 
[...]

PR #580, opened 2021-03-16
Implement DXR object collections

HansKristian-Work:
Fix https://github.com/HansKristian-Work/vkd3d-proton/issues/568
Fix https://github.com/HansKristian-Work/vkd3d-proton/issues/592

### DIFF
--- a/libs/vkd3d/device.c
+++ b/libs/vkd3d/device.c
@@ -91,6 +91,7 @@ static const struct vkd3d_optional_extension_info optional_device_extensions[] =
     VK_EXTENSION_COND(KHR_RAY_TRACING_PIPELINE, KHR_ray_tracing_pipeline, VKD3D_CONFIG_FLAG_DXR),
     VK_EXTENSION_COND(KHR_ACCELERATION_STRUCTURE, KHR_acceleration_structure, VKD3D_CONFIG_FLAG_DXR),
     VK_EXTENSION_COND(KHR_DEFERRED_HOST_OPERATIONS, KHR_deferred_host_operations, VKD3D_CONFIG_FLAG_DXR),
+    VK_EXTENSION_COND(KHR_PIPELINE_LIBRARY, KHR_pipeline_library, VKD3D_CONFIG_FLAG_DXR),
     VK_EXTENSION(KHR_SPIRV_1_4, KHR_spirv_1_4),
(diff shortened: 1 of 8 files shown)


libs/vkd3d/raytracing_pipeline.c
```diff
@@ -110,14 +137,7 @@ static ULONG d3d12_state_object_release(struct d3d12_state_object *state_object)
     TRACE("%p decreasing refcount to %u.\n", state_object, refcount);
 
     if (!refcount)
-    {
```
doitsujin, line 112: Can there ever be a situation where the app can re-claim a reference to the object after the (public) ref count has reached zero? In that case, `AddRef` would have to handle that case and add a private ref again.

libs/vkd3d/device.c
```diff
@@ -4750,9 +4754,12 @@ static void d3d12_device_caps_init_shader_model(struct d3d12_device *device)
          * FP16, Denorm modes (float controls extension)
          */
         if (device->device_info.float16_int8_features.shaderFloat16 &&
-            device->device_info.subgroup_extended_types_features.shaderSubgroupExtendedTypes)
+                device->device_info.storage_16bit_features.storageBuffer16BitAccess &&
+                device->device_info.subgroup_extended_types_features.shaderSubgroupExtendedTypes)
```
doitsujin, line 4758: Do we have enough 16-bit support in dxil-spirv to enable `Native16BitShaderOpsSupported`?

doitsujin (commented):
LGTM, just two questions.

libs/vkd3d/device.c
HansKristian-Work: Yes, I think so.

libs/vkd3d/raytracing_pipeline.c
HansKristian-Work: I doubt it. The DXR wording says: "Collection to include in state object. Enclosing state object holds a ref on the existing collection."
I proved it's not a real reference, but I don't think this means being able to make those objects live again.

doitsujin (approved):

libs/vkd3d/device.c
doitsujin: Guess we should write a test and enable that at some point then, but can be done separately.

PR #582, opened 2021-03-16
Collection of adapter fixes

Joshua-Ashton:


DadSchoorse:
Calling `vkGetPhysicalDeviceProperties2` is not allowed on 1.0 devices. You should check with `vkGetPhysicalDeviceProperties` if a device supports `VK_API_VERSION_1_1`.

Joshua-Ashton:
We don't ever make a 1.0 instance or device.

DadSchoorse:
That doesn't matter, if you create a 1.1 device and call `vkEnumeratePhysicalDevices`, you will get devices which only support 1.0 as well.

doitsujin:
> We don't ever make a 1.0 instance or logical device, can the physical device still expose that? That's annoying.

Yes, and this is exactly why things like the device select layer and old DXVK versions break when there's a 1.0 driver present.

Joshua-Ashton:
Okay, will fix that. Mildly annoying.

Joshua-Ashton:
Hopefully resolved.

libs/d3d12/main.c
```diff
@@ -233,39 +197,46 @@ static VkPhysicalDevice d3d12_get_vk_physical_device(struct vkd3d_instance *inst
...
+        properties2.pNext = &id_properties;
 
-            if (!memcmp(id_properties.deviceLUID, &adapter_desc->AdapterLuid, VK_LUID_SIZE))
-            {
-                vk_physical_device = vk_physical_devices[i];
-                break;
-            }
+        pfn_vkGetPhysicalDeviceProperties2(vk_physical_devices[i], &properties2);
+
+        if (!memcmp(id_properties.deviceLUID, &adapter_desc->AdapterLuid, VK_LUID_SIZE))
```
DadSchoorse, line 219: You should check `id_properties.deviceLUIDValid` first.

libs/d3d12/main.c
```diff
@@ -233,39 +197,46 @@ static VkPhysicalDevice d3d12_get_vk_physical_device(struct vkd3d_instance *inst
...
-            memset(&id_properties, 0, sizeof(id_properties));
-            id_properties.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_ID_PROPERTIES;
+        /* Skip over < 1.1 physical devices because... 1.1 instances
+         * still expose those apparently... */
+        if (properties2.properties.apiVersion < VK_API_VERSION_1_1)
+            continue;
 
-            properties2.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_PROPERTIES_2;
-            properties2.pNext = &id_properties;
+        memset(&id_properties, 0, sizeof(id_properties));
```
DadSchoorse, line 211: Why `memset` the whole struct instead of just setting `pNext` to `NULL`?

DadSchoorse (changes requested):
comments

libs/d3d12/main.c
Joshua-Ashton: The existing code did this, but, fixed.

libs/d3d12/main.c
Joshua-Ashton: Resolved.

DadSchoorse (approved):

libs/d3d12/main.c
```diff
@@ -233,39 +197,46 @@ static VkPhysicalDevice d3d12_get_vk_physical_device(struct vkd3d_instance *inst
...
-        TRACE("Matching adapters by LUIDs.\n");
+        pfn_vkGetPhysicalDeviceProperties(vk_physical_devices[i], &properties2.properties);
 
-        for (i = 0; i < count; ++i)
-        {
-            memset(&id_properties, 0, sizeof(id_properties));
-            id_properties.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_ID_PROPERTIES;
+        /* Skip over < 1.1 physical devices because... 1.1 instances
+         * still expose those apparently... */
+        if (properties2.properties.apiVersion < VK_API_VERSION_1_1)
```
doitsujin, line 208: Should probably check `VKD3D_MIN_API_VERSION` (which is currently 1_1 but may be bumped to 1_2 at some point) and remove the comment above, instead WARN that we skipped an adapter.

doitsujin:
I think there's more that can be done here, since the `vkd3d_select_physical_device` cruft still exists. That should probably be removed and we should make sure that we always find a valid non-null physical device that supports all required extensions right away, and if we can't, error out.

doitsujin (changes requested):

Joshua-Ashton:
`vkd3d_select_physical_device` is used for native builds. We can clean that up later, I want to remove the whole vkd3d-utils thing and consolidate it into d3d12 eventually.

HansKristian-Work (commented):
Don't have further concerns than existing comments.

libs/d3d12/main.c
Joshua-Ashton: Resolved.

doitsujin (approved):

PR #595, opened 2021-03-23
Use VK_KHR_create_renderpass2 + Renderpass Key Flags Refactor

Joshua-Ashton:
We need this before implementing TIER_2 variable rate shading.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -2074,12 +2074,14 @@ static void d3d12_command_list_clear_attachment_pass(struct d3d12_command_list *
         const D3D12_RECT *rects)
 {
     const struct vkd3d_vk_device_procs *vk_procs = &list->device->vk_procs;
-    VkAttachmentDescription attachment_desc;
-    VkAttachmentReference attachment_ref;
-    VkSubpassDependency dependencies[2];
-    VkSubpassDescription subpass_desc;
+    VkAttachmentDescription2KHR attachment_desc;
+    VkAttachmentReference2KHR attachment_ref;
+    VkSubpassBeginInfoKHR subpass_begin_info;
+    VkSubpassDependency2KHR dependencies[2];
+    VkSubpassDescription2KHR subpass_desc;
+    VkSubpassEndInfoKHR subpass_end_info;
+    VkRenderPassCreateInfo2KHR pass_info;
     VkRenderPassBeginInfo begin_info;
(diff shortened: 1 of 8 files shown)


libs/vkd3d/command.c
```diff
@@ -2209,16 +2231,19 @@ static void d3d12_command_list_clear_attachment_pass(struct d3d12_command_list *
     begin_info.clearValueCount = clear_op ? 1 : 0;
     begin_info.pClearValues = clear_op ? clear_value : NULL;
 
-    VK_CALL(vkCmdBeginRenderPass(list->vk_command_buffer,
-            &begin_info, VK_SUBPASS_CONTENTS_INLINE));
+    VK_CALL(vkCmdBeginRenderPass2KHR(list->vk_command_buffer,
```
HansKristian-Work, line 2234: Do we actually need to use BeginRenderPass2 as well?

Joshua-Ashton: I guess not but I don't see why we'd mix and match stuff here. Any Mesa driver will just wrap it for us.

HansKristian-Work: Fair enough I guess.

libs/vkd3d/command.c
```diff
@@ -5399,7 +5445,7 @@ static void d3d12_command_list_copy_image(struct d3d12_command_list *list,
 
         VK_CALL(vkUpdateDescriptorSets(list->device->vk_device, 1, &vk_descriptor_write, 0, NULL));
 
-        VK_CALL(vkCmdBeginRenderPass(list->vk_command_buffer, &begin_info, VK_SUBPASS_CONTENTS_INLINE));
+        VK_CALL(vkCmdBeginRenderPass2KHR(list->vk_command_buffer, &begin_info, &subpass_begin_info));
```
doitsujin, line 5448: I really don't think we'll ever need this for image copies.

doitsujin (commented):
Nit, looks good otherwise.

libs/vkd3d/command.c
HansKristian-Work: I guess it's more for consistency.

Joshua-Ashton: Yeah, mixing and matching renderpass and renderpass2 stuff is weird. Mesa and stuff must wrap this to that anyway so it avoids another  step.

HansKristian-Work (approved):
LGTM.

PR #598, opened 2021-03-25
Implement Conservative Rasterization Tier 3

doitsujin:
Draft for now because this will need DXIL support as well. On AMD we can only expose Tier 1 due to driver limitations, but Nvidia has full Tier 3 support.


HansKristian-Work (approved):
LGTM!

PR #615, opened 2021-04-12
Force TGSM Barriers Config + Enable for F1 2020

Joshua-Ashton:
Mirrors the DXVK app profile for emitting a barrier after TGSM stores because F1 still haven't fixed this bug in their game for three years running. :D


libs/vkd3d/state.c
```diff
@@ -2888,6 +2889,7 @@ static HRESULT d3d12_pipeline_state_init_graphics(struct d3d12_pipeline_state *s
     compile_args.target = VKD3D_SHADER_TARGET_SPIRV_VULKAN_1_0;
     compile_args.target_extension_count = vk_info->shader_extension_count;
     compile_args.target_extensions = vk_info->shader_extensions;
+    compile_args.config_flags = vkd3d_config_flags;
```
HansKristian-Work, line 2892: Need it for RTPSO as well.

HansKristian-Work (changes requested):
Small nit, otherwise fine.

libs/vkd3d/state.c
Joshua-Ashton: Done!

PR #621, opened 2021-04-15
Various PSO robustness fixes

### LINKED ISSUE #618: DXBC: Correctly fail pipeline compilation if binding is missing
If we miss a binding in the root signature, DXBC backend will not fail compilation and we end up hitting driver with broken SPIR-V.

HansKristian-Work:
Works around a lot of broken code in FH4.

The D3D12 runtime is verified to be robust against these bugs, so we also need to be.

Fix #618.

### DIFF
--- a/libs/vkd3d-shader/vkd3d_shader_main.c
+++ b/libs/vkd3d-shader/vkd3d_shader_main.c
@@ -300,6 +300,29 @@ static void vkd3d_shader_scan_destroy(struct vkd3d_shader_scan_info *scan_info)
     hash_map_clear(&scan_info->register_map);
 }
 
+static int vkd3d_shader_validate_shader_type(enum vkd3d_shader_type type, VkShaderStageFlagBits stages)
+{
+    static const VkShaderStageFlagBits table[VKD3D_SHADER_TYPE_COUNT] = {
+        VK_SHADER_STAGE_FRAGMENT_BIT,
+        VK_SHADER_STAGE_VERTEX_BIT,
+        VK_SHADER_STAGE_GEOMETRY_BIT,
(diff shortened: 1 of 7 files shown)


libs/vkd3d-shader/vkd3d_shader_main.c
```diff
@@ -300,6 +302,29 @@ static void vkd3d_shader_scan_destroy(struct vkd3d_shader_scan_info *scan_info)
...
+        VK_SHADER_STAGE_GEOMETRY_BIT,
+        VK_SHADER_STAGE_TESSELLATION_CONTROL_BIT,
+        VK_SHADER_STAGE_TESSELLATION_EVALUATION_BIT,
+        VK_SHADER_STAGE_COMPUTE_BIT,
+    };
+
+    if (type >= VKD3D_SHADER_TYPE_COUNT)
+        return VKD3D_ERROR_INVALID_ARGUMENT;
+
+    if (!(table[type] & stages))
```
doitsujin, line 319: Is the intention here that `stages` can contain multiple bits or just one? Should either fix the type (to `VkShaderStageFlags`) or make this `!=` instead.

HansKristian-Work: Can be STAGE_ALL if every stage is acceptable (mostly for RT)

HansKristian-Work: I guess it's not that relevant since I don't let RT paths go through this check anyways, so should probably just be !=

HansKristian-Work: Fixed.

doitsujin (approved):

PR #623, opened 2021-04-16
vkd3d: Unify _mm_pause detection.

DadSchoorse:
I was unreasonably annoyed by the detection duplication when I randomly opened one of the headers. :frog:

### DIFF
--- a/include/private/vkd3d_spinlock.h
+++ b/include/private/vkd3d_spinlock.h
@@ -28,6 +28,13 @@
 #include <emmintrin.h>
 #endif
 
+static inline void vkd3d_pause(void)
+{
+#ifdef __SSE2__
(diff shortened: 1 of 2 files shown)


include/private/vkd3d_spinlock.h
```diff
@@ -26,6 +26,9 @@
 
 #ifdef __SSE2__
 #include <emmintrin.h>
+#define VKD3D_PAUSE() _mm_pause()
+#else
+#define VKD3D_PAUSE() do {} while(false)
```
doitsujin, line 31: this should be an inline function rather than a macro.

doitsujin (changes requested):

HansKristian-Work (approved):

PR #646, opened 2021-04-27
build: Install dlls inside prefix without d3d12 dlls (only with Windows 10)

IroAlexis:
As I had reported on VKx (a while ago), if we use a Wine build without vkd3d, the prefixes generated with this build doesn't contain the d3d12 dlls and the script return an error (it is logical).

I have fixed this behavior but it contains some workarounds and it may be that the script is not very clear now. I would understand if you don't want this fix.

### DIFF
--- a/setup_vkd3d_proton.sh
+++ b/setup_vkd3d_proton.sh
@@ -70,8 +70,8 @@ fi
 
 # ensure wine placeholder dlls are recreated
 # if they are missing
-$wineboot -u && $wine wineserver --wait
-winver=$($wine winecfg -v)
+$wineboot -u && wineserver --wait
+winver=$($wine winecfg -v 2> /dev/null)
 
 win64_sys_path=$($wine64 winepath -u 'C:\windows\system32' 2> /dev/null)
 win64_sys_path="${win64_sys_path/$'\r'/}"
--- a/setup_vkd3d_proton.sh
+++ b/setup_vkd3d_proton.sh
@@ -44,6 +44,7 @@ fi
 
 # find wine executable
 export WINEDEBUG=-all
+export WINEDLLOVERRIDES="mscoree,mshtml="
 
 wine="wine"
 wine64="wine64"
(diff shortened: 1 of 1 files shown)


Joshua-Ashton:
Please split the part adding `export WINEDLLOVERRIDES="mscoree,mshtml="`

I don't really see the purpose for the Win10 checks. There is 12on7 stuff so it could work in a prefix there too.

I don't understand the last two workarounds also, I haven't encountered these issues at all on Wine-TKG with fsync.

Joshua-Ashton:
Ah, I see, the last two checks are just for the Windows version checks... I'd just get rid of that entirely and just always install in the pfx.

IroAlexis:
> Please split the part adding export `WINEDLLOVERRIDES="mscoree,mshtml="` and add a comment about what it does.

Ok no problem

> I don't really see the purpose for the Win10 checks. There is 12on7 stuff so it could work in a prefix there too.

> Ah, I see, the last two checks are just for the Windows version checks... I'd just get rid of that entirely and just always install in the pfx.

If we follow the Windows recommandations, d3d12 is compatible only win10. There is too win7 with 12on7 as you say him but I don't know how Wine will behave and maybe there will be bugs that will be reported here and that will probably be annoying.
I can get rid this checks but it is dangerous in my opinion.
 
> I don't understand the last two workarounds also, I haven't encountered these issues at all on Wine-TKG with fsync.

Wrong test on my end....

Joshua-Ashton:
Yeah, I don't think we need to be checking the Wine prefix version, that can change at any time anyway.

IroAlexis:
> > I don't understand the last two workarounds also, I haven't encountered these issues at all on Wine-TKG with fsync.
> 
> Wrong test on my end....

I got mixed up with my different builds, here is an example :
```
$ WINEDEBUG=-all /opt/lunion-play-git/bin/wine winecfg -v 1> /tmp/winver && cat /tmp/winver && rm /tmp/winver
wineserver: using server-side synchronization.
win10
```
But it doesn't matter if you maintain that you don't want to check the Windows version.

Joshua-Ashton:
Superceded by: https://github.com/HansKristian-Work/vkd3d-proton/pull/648

PR #651, opened 2021-04-28
vkd3d: Enable and require shaderDrawParameters.

DadSchoorse:
Unless I miss something in the spec, `shaderDrawParameters` is an optional feature in 1.1 and should be enabled with `VkPhysicalDeviceShaderDrawParametersFeatures` before use. It seems like this was missed while porting from 1.0 to 1.1 because the original extension doesn't have a feature struct.


Joshua-Ashton (approved):
LGTM

PR #653, opened 2021-05-04
vkd3d: Handle copies of depth-stencil image -> color

HansKristian-Work:


Joshua-Ashton:
Is there any benefit to using separate depth stencil here instead of transitioning all of depth stencil?

HansKristian-Work:
> Is there any benefit to using separate depth stencil here instead of transitioning all of depth stencil?

Not really when looking closer at it. Fixed it.

Joshua-Ashton (approved):
LGTM but it'd be nice to avoid requiring COLOR_ATTACHMENT on all R8 in the future...

PR #659, opened 2021-05-06
Type error fix for clang-cl

dave-juicelabs:


libs/vkd3d/device.c
```diff
@@ -2309,7 +2309,7 @@ static void d3d12_device_destroy_query_pool(struct d3d12_device *device, const s
     VK_CALL(vkDestroyQueryPool(device->vk_device, pool->vk_query_pool, NULL));
 }
 
-HRESULT d3d12_device_get_query_pool(struct d3d12_device *device, uint32_t type_index, struct vkd3d_query_pool *pool)
+HRESULT d3d12_device_get_query_pool(struct d3d12_device *device, D3D12_QUERY_HEAP_TYPE type_index, struct vkd3d_query_pool *pool)
```
Joshua-Ashton, line 2312: This is not correct, every function that calls this passes in a `uint32_t` which is a `VKD3D_QUERY_TYPE_INDEX_*` value, this function also passes it's `type_index` to `uint32_t`s on other functions only also.

What is the error you are seeing, and why does this fix it? All this commit does is break type stuff.

Joshua-Ashton (changes requested):

libs/vkd3d/device.c
dave-juicelabs:
```
../libs/vkd3d/device.c(2312,9): error: conflicting types for 'd3d12_device_get_query_pool'
HRESULT d3d12_device_get_query_pool(struct d3d12_device *device, uint32_t type_index, struct vkd3d_query_pool *pool)
        ^
..\libs\vkd3d/vkd3d_private.h(2634,9): note: previous declaration is here
HRESULT d3d12_device_get_query_pool(struct d3d12_device *device, D3D12_QUERY_HEAP_TYPE heap_type, struct vkd3d_query_pool *pool);
        ^
```

This is the error that I get from clang-cl. I assume then the better fix would be to update the header file.

Joshua-Ashton: Yes

Joshua-Ashton (approved):
LGTM!

PR #660, opened 2021-05-06
clang-cl support for native Windows builds.

dave-juicelabs:


meson.build
```diff
@@ -64,7 +64,7 @@ if not enable_trace
   add_project_arguments('-DVKD3D_NO_TRACE_MESSAGES', language : 'c')
 endif
 
-vkd3d_external_includes = [ './subprojects/Vulkan-Headers/include', './subprojects/SPIRV-Headers/include' ]
+vkd3d_external_includes = [ './subprojects/SPIRV-Headers/include' ]
```
Joshua-Ashton, line 67: Looks like an unintentional change here.

If not, why were the header includes removed? We really don't want a dependency on whatever outdated crap the SDK is shipping.

Joshua-Ashton (changes requested):

meson.build
dave-juicelabs: It was unintentional to include that. I will fix this and also update the commit message as requested in the other PR.

Joshua-Ashton (changes requested):
Miiinor nit, commit prefix should be `build:`

Joshua-Ashton (approved):
LGTM!

PR #663, opened 2021-05-12
Add runtime descriptor QA checks

HansKristian-Work:
Supports instrumenting DXIL shaders and DXBC shaders for descriptor heap access. The intention here is to tightly integrate this in vkd3d-proton for greatly improved report quality and runtime efficiency.


doitsujin (commented):
looks like `2d398f6e5687f843f9edb5a504025dba6d0c195b` accidentally added a whole bunch of weird files?

HansKristian-Work:
> looks like `2d398f6e5687f843f9edb5a504025dba6d0c195b` accidentally added a whole bunch of weird files?

weird. Fixed.

libs/vkd3d/resource.c
```diff
@@ -5076,6 +5078,9 @@ static HRESULT d3d12_descriptor_heap_create_descriptor_set(struct d3d12_descript
     VkDescriptorSetAllocateInfo vk_set_info;
     VkResult vr;
 
+    if (vkd3d_descriptor_debug_active_qa_checks() && descriptor_heap->desc.Type == D3D12_DESCRIPTOR_HEAP_TYPE_CBV_SRV_UAV)
```
Joshua-Ashton, line 5081: Need account for this in the pool also otherwise it fails to allocate on AMD.

Joshua-Ashton (changes requested):

libs/vkd3d/resource.c
HansKristian-Work: Fixed.

libs/vkd3d-shader/spirv.c
```diff
@@ -8385,6 +8735,52 @@ static uint32_t vkd3d_dxbc_compiler_get_resource_index(struct vkd3d_dxbc_compile
...
+                ERR("Invalid binding type %u.\n", binding->type);
+                type_flags = 0;
+                break;
+        }
+
+        descriptor_qa_args[0] = index_id;
+        descriptor_qa_args[1] = vkd3d_dxbc_compiler_get_constant_uint(compiler, type_flags);
+        descriptor_qa_args[2] = vkd3d_dxbc_compiler_get_constant_uint(compiler,
+                ++compiler->descriptor_qa_instruction_count);
+        index_id = vkd3d_spirv_build_op_function_call(builder,
```
doitsujin, line 8786: is this valid? spirv-val already complains about our hull shader phase functions taking raw IDs rather than variables.

HansKristian-Work: It is. It's when passing pointers that the pointer must bind directly to an OpVariable declaration.

HansKristian-Work: Verified to be sure, and it passes spirv-val just fine.

doitsujin (approved):

PR #671, opened 2021-05-26
Require robustness2

HansKristian-Work:
This is widely supported on desktop, and the non-robustness path is essentially completely untested at this point. For upcoming workarounds, having non-robustness2 in the picture is going to get more and more annoying, so it's time to flip it to required.

The non-robustness2 path is also broken for any robust UAV either way, since stores are not masked.


Joshua-Ashton (approved):
LGTM.

I also saw your minor fix for one of the null descriptors not being WHOLE_SIZE. :)

doitsujin (approved):

PR #673, opened 2021-05-27
Adding the dxil-spirv shader conversion to the d3d12 pipeline creation cached blob

dave-juicelabs:
Massive speed up in AAA game creation of thousands of pipelines that utilize caching of the blobs but not ID3D12PipelineLibrary.

Putting this here for feedback, NOT direct integration as there is surely some work still to do.

### DIFF
--- a/include/vkd3d_shader.h
+++ b/include/vkd3d_shader.h
@@ -722,6 +722,8 @@ int vkd3d_shader_compile_dxil_export(const struct vkd3d_shader_code *dxil,
         const struct vkd3d_shader_interface_local_info *shader_interface_local_info,
         const struct vkd3d_shader_compile_arguments *compiler_args);
 
+vkd3d_shader_hash_t vkd3d_shader_hash(const struct vkd3d_shader_code* shader);
+
 #endif  /* VKD3D_SHADER_NO_PROTOTYPES */
 
 /*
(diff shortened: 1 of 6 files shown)


doitsujin:
There's a number of reasons why we haven't done this:
- Our generated shaders are *huge*. Games like Horizon Zero Dawn that generate a ~700MB disk cache with the current solution as well as on Windows would easily end up creating cache files in excess of 10GB without compression. This alone is a dealbreaker.
- We'd have to implement full validation of the pipeline state vector against the cached pipeline, since the app may keep using the same pipeline identifiers while changing the content (e.g. shaders due to a game update, or render target formats when changing in-game settings, etc).
- Wiring this up is just annoying, and the whole serializer thing proposed here seems kind of overengineerd.

dave-juicelabs:
The reason I started down this path was due to the 20 minute start time as the game compiled 17,000 pipelines. After adding this, which I agree is over engineered in places, the compilation was negligible.

As for your explanation of why this hasn't been added before, makes sense and I will look at the dxil conversion for potential speedups. The game I am looking at, which hasn't been released yet but is from a AAA game studio, has up to 16 threads compiling pipelines. With performance analysis, the threads are spending the most time in dxil spirv conversion doing cfg related work. (Off the top of my head, don't have the right words in mind).

Last, thanks for the feedback. The first two bullets were exactly what I was looking for to help point me in the right direction.

HansKristian-Work:
> dxil spirv conversion doing cfg related work

Not surprised, as this is the most gnarly part of DXIL -> SPIR-V and efficient traversal algorithms are quite difficult ... 20 minutes sounds a bit much though, probably hits some pathological case.

PR #675, opened 2021-05-27
vkd3d: Fix TODO in ClearUAV.

HansKristian-Work:


libs/vkd3d/command.c
```diff
@@ -7429,6 +7429,32 @@ static void vkd3d_clear_uav_info_from_desc(struct vkd3d_clear_uav_info *args, co
     }
 }
 
+static void vkd3d_mask_uint_clear_color(uint32_t color[4], VkFormat vk_format)
+{
+    unsigned int i;
+
+    /* Need to mask the clear value, since apparently driver can saturate the clear value instead. */
+    switch (vk_format)
```
doitsujin, line 7437: Need to handle R10G10B10A2 as well here, other than that, LGTM.

doitsujin (changes requested):

libs/vkd3d/command.c
HansKristian-Work: Fixed.

doitsujin (approved):

PR #676, opened 2021-05-27
tests: Fix Clear UAV test constant.

HansKristian-Work:
Was using 0x8000 / 0xffff instead of 0x200 / 0x3ff, rounded differently
on NV.


HansKristian-Work:
It's a bit awkward to express tolerance for something like RGB10A2. The reference is a packed 32-bit word, and if we can express something that doesn't require +/- ULP delta, it's probably best to do that.

doitsujin (approved):

PR #682, opened 2021-06-01
Low latency presentation and frame latency fixes

Joshua-Ashton:
## Low latency presentation

In cases where acquire image is blocking, we should call that after
presentation to avoid latency when the app calls present.

This avoids weird inverse frame cadences with Mesa WSI right now,
as acquiring an image is always a blocking call until it is complete.

In cases when we aren't blocking, this kicks off the acquisition so
it can be waited upon by the next frame -- and in the future by
the submissions themselves as semaphores.

## Handle frame latency without WAITABLE_OBJECT

Documentation says that this should always be 3 without WAITABLE_OBJECT
unlike in D3D11 where it will use the DXGI device's frame latency.

This stops runaway presentations in the non-blocking acquire image case
with the new semaphore setup.

## Wait for swapchain image acquisition on GPU

Use another set of semaphores to wait for the image acquisition on the
GPU.

In the non-blocking vkAcquireNextImageKHR case, this means that a
potential bubble of time between waiting on the fence and submitting
the blit + presentation is eliminated.

Runaway presentation in this setup is avoided by frame latency objects
and normal frame latency which is always 3 according to documentation.

### DIFF
--- a/libs/vkd3d/swapchain.c
+++ b/libs/vkd3d/swapchain.c
@@ -190,6 +190,7 @@ struct d3d12_swapchain
     unsigned int vk_swapchain_width;
     unsigned int vk_swapchain_height;
     VkPresentModeKHR present_mode;
+    bool is_suboptimal;
 
     struct
     {
@@ -1499,6 +1500,7 @@ static HRESULT d3d12_swapchain_create_vulkan_swapchain(struct d3d12_swapchain *s
     swapchain->vk_swapchain_height = height;
 
     swapchain->vk_image_index = INVALID_VK_IMAGE_INDEX;
+    swapchain->is_suboptimal = false;
 
     if (swapchain->vk_swapchain != VK_NULL_HANDLE)
     {
@@ -1745,7 +1747,13 @@ static VkResult d3d12_swapchain_queue_present(struct d3d12_swapchain *swapchain,
                 VK_NULL_HANDLE, &swapchain->vk_image_index);
 
         if (vr >= 0)
+        {
             swapchain->vk_acquire_semaphores_signaled[swapchain->frame_id] = true;
+            /* If we have observed suboptimal once, guarantees that we keep observing it
+             * until we have recreated the swapchain. */
+            if (swapchain->is_suboptimal)
+                vr = VK_SUBOPTIMAL_KHR;
+        }
 
         if (vr == VK_SUBOPTIMAL_KHR)
         {
@@ -1810,6 +1818,14 @@ static VkResult d3d12_swapchain_queue_present(struct d3d12_swapchain *swapchain,
         swapchain->frame_id = (swapchain->frame_id + 1) % swapchain->buffer_count;
         swapchain->vk_image_index = INVALID_VK_IMAGE_INDEX;
 
+        if (vr == VK_SUBOPTIMAL_KHR)
+            swapchain->is_suboptimal = true;
+
+        /* If we have observed suboptimal once, guarantees that we keep observing it
+         * until we have recreated the swapchain. */
+        if (swapchain->is_suboptimal)
+            vr = VK_SUBOPTIMAL_KHR;
+
         /* Could get SUBOPTIMAL here. Defer acquiring if we hit that path.
          * On next present, we can recreate the swapchain. */
         if (vr == VK_SUCCESS)
@@ -1823,6 +1839,12 @@ static VkResult d3d12_swapchain_queue_present(struct d3d12_swapchain *swapchain,
             if (vr >= 0)
             {
                 swapchain->vk_acquire_semaphores_signaled[swapchain->frame_id] = true;
+                /* If we observe suboptimal here, we cannot set INVALID_VK_IMAGE_INDEX yet,
+                 * since the next present will acquire again, before we have had a chance to present.
+                 * This can potentially deadlock, or cause other weirdness that we're not ready to handle.
+                 * Mark the swapchain is suboptimal, so that we're guaranteed to recreate the swapchain in due time. */
+                if (vr == VK_SUBOPTIMAL_KHR)
+                    swapchain->is_suboptimal = true;
             }
             else
             {
--- a/libs/vkd3d/swapchain.c
+++ b/libs/vkd3d/swapchain.c
@@ -1764,7 +1764,6 @@ static VkResult d3d12_swapchain_queue_present(struct d3d12_swapchain *swapchain,
 static HRESULT d3d12_swapchain_present(struct d3d12_swapchain *swapchain,
         unsigned int sync_interval, unsigned int flags)
 {
-    HANDLE frame_latency_event;
     VkQueue vk_queue;
     VkResult vr;
     HRESULT hr;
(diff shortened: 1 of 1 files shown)


libs/vkd3d/swapchain.c
```diff
@@ -2658,39 +2666,28 @@ static HRESULT d3d12_swapchain_init(struct d3d12_swapchain *swapchain, IDXGIFact
...
     swapchain->current_buffer_index = 0;
 
+    swapchain->frame_number = DXGI_MAX_SWAP_CHAIN_BUFFERS;
+    swapchain->frame_latency = 3;
+
     if (swapchain_desc->Flags & DXGI_SWAP_CHAIN_FLAG_FRAME_LATENCY_WAITABLE_OBJECT)
-    {
-        swapchain->frame_number = DXGI_MAX_SWAP_CHAIN_BUFFERS;
-        swapchain->frame_latency = 1;
+        swapchain->frame_latency = 3;
```
DadSchoorse, line 2675: You already set it to three, why are you doing it again?

Joshua-Ashton: I made a typo.

libs/vkd3d/swapchain.c
```diff
@@ -1742,13 +1721,24 @@ static VkResult d3d12_swapchain_queue_present(struct d3d12_swapchain *swapchain,
...
         /* Could get SUBOPTIMAL here. Deal with it later. */
         vr = VK_SUCCESS;
+
+        /* Try to acquire our next image to avoid blocking next frame. */
+        if ((vr = vk_procs->vkAcquireNextImageKHR(vk_device, swapchain->vk_swapchain, UINT64_MAX,
+                    swapchain->vk_acquire_semaphores[swapchain->frame_id], VK_NULL_HANDLE, &swapchain->vk_image_index)))
+        {
+            /* Didn't manage to get a non-suboptimal image or any at all.
+             * The last present we did was a success so don't remake the
+             * swapchain now. Retry again at the next presentation. */
```
DadSchoorse, line 1739: Aren't you leaving the semaphore signaled here in case of `VK_SUBOPTIMAL_KHR`, without waiting for it? And if you succeed next time, eventually you will reuse your signaled semaphore, which is invalid.

Joshua-Ashton: Yes. I guess I should just recreate in this case as well to avoid that.

libs/vkd3d/swapchain.c
```diff
@@ -1687,6 +1654,8 @@ static HRESULT d3d12_swapchain_set_sync_interval(struct d3d12_swapchain *swapcha
 static VkResult d3d12_swapchain_queue_present(struct d3d12_swapchain *swapchain, VkQueue vk_queue)
 {
     const struct vkd3d_vk_device_procs *vk_procs = d3d12_swapchain_procs(swapchain);
+    const uint32_t acquire_wait_mask = VK_PIPELINE_STAGE_ALL_COMMANDS_BIT;
```
HansKristian-Work, line 1657: Should be COLOR_ATTACHMENT_OUTPUT_BIT

libs/vkd3d/swapchain.c
```diff
@@ -1698,8 +1667,18 @@ static VkResult d3d12_swapchain_queue_present(struct d3d12_swapchain *swapchain,
...
     {
-        if ((vr = d3d12_swapchain_acquire_next_vulkan_image(swapchain)) < 0)
+        if ((vr = vk_procs->vkAcquireNextImageKHR(vk_device, swapchain->vk_swapchain, UINT64_MAX,
+                swapchain->vk_acquire_semaphores[swapchain->frame_id], VK_NULL_HANDLE, &swapchain->vk_image_index)))
+        {
+            if (vr == VK_SUBOPTIMAL_KHR)
+            {
+                /* Suboptimal is still considered success, but we always want
+                 * to recreate swapchains in this case. */
+                return VK_ERROR_OUT_OF_DATE_KHR;
```
HansKristian-Work, line 1677: Semaphore is left signalled here. Should submit a dummy submit to wait for the semaphore to get signaled, wait for a fence, and then destroy fence, and recreate semaphore.

libs/vkd3d/swapchain.c
HansKristian-Work: Same comment as above.

libs/vkd3d/swapchain.c
```diff
@@ -2658,39 +2666,28 @@ static HRESULT d3d12_swapchain_init(struct d3d12_swapchain *swapchain, IDXGIFact
...
-        WARN("Failed to create Vulkan fence, vr %d.\n", vr);
-        d3d12_swapchain_destroy(swapchain);
-        return hresult_from_vk_result(vr);
-    }
-    swapchain->vk_fence = vk_fence;
-
     swapchain->current_buffer_index = 0;
 
+    swapchain->frame_number = DXGI_MAX_SWAP_CHAIN_BUFFERS;
+    swapchain->frame_latency = 3;
```
HansKristian-Work, line 2672: Can we make 3 a define?

libs/vkd3d/swapchain.c
```diff
@@ -1687,6 +1654,8 @@ static HRESULT d3d12_swapchain_set_sync_interval(struct d3d12_swapchain *swapcha
 static VkResult d3d12_swapchain_queue_present(struct d3d12_swapchain *swapchain, VkQueue vk_queue)
 {
     const struct vkd3d_vk_device_procs *vk_procs = d3d12_swapchain_procs(swapchain);
+    const uint32_t acquire_wait_mask = VK_PIPELINE_STAGE_ALL_COMMANDS_BIT;
+    VkDevice vk_device = d3d12_swapchain_device(swapchain)->vk_device;
```
HansKristian-Work, line 1658: Code before declaration?

HansKristian-Work (changes requested):
Overall design seems fine, but some bugs with SUBOPTIMAL.

libs/vkd3d/swapchain.c
```diff
@@ -1311,8 +1278,11 @@ static void d3d12_swapchain_destroy_buffers(struct d3d12_swapchain *swapchain, B
     {
         for (i = 0; i < swapchain->buffer_count; ++i)
         {
-            vk_procs->vkDestroySemaphore(swapchain->command_queue->device->vk_device, swapchain->vk_semaphores[i], NULL);
-            swapchain->vk_semaphores[i] = VK_NULL_HANDLE;
+            vk_procs->vkDestroySemaphore(swapchain->command_queue->device->vk_device, swapchain->vk_acquire_semaphores[i], NULL);
```
HansKristian-Work, line 1281: Do we risk destroying a semaphore signal that is in flight here if user calls ResizeBuffers or something?

libs/vkd3d/swapchain.c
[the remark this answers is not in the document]
HansKristian-Work: GCC doesn't complain, so I guess not.

HansKristian-Work (commented):
Pushed some fixes.

libs/vkd3d/swapchain.c
```diff
@@ -1727,27 +1778,53 @@ static VkResult d3d12_swapchain_queue_present(struct d3d12_swapchain *swapchain,
...
+        {
+            /* Try to acquire our next image to avoid blocking next frame. */
+            assert(!swapchain->vk_acquire_semaphores_signaled[swapchain->frame_id]);
+            vr = vk_procs->vkAcquireNextImageKHR(vk_device, swapchain->vk_swapchain, UINT64_MAX,
+                    swapchain->vk_acquire_semaphores[swapchain->frame_id], VK_NULL_HANDLE,
+                    &swapchain->vk_image_index);
+
+            if (vr >= 0)
+            {
+                swapchain->vk_acquire_semaphores_signaled[swapchain->frame_id] = true;
```
Joshua-Ashton, line 1818: We still want to `swapchain->vk_image_index = INVALID_VK_IMAGE_INDEX;` here

HansKristian-Work: Do we though? The next present will call Acquire again, while we have an image acquired.

Joshua-Ashton:
LGTM

PR #690, opened 2021-06-09
vkd3d: Rewrite image layout transition handling

HansKristian-Work:
Yet another iteration of this.

- Honor resource barriers for resource states which cannot automatically
  decay or promote. This includes COLOR_ATTACHMENT, UNORDERED_ACCESS and
  VRS image. If SIMULTANEOUS_ACCESS is used, we can still promote, and
  we handle that by setting common layout to GENERAL for these resources.

- Avoid redundant barriers in render passes since normal resource
  barriers will always make sure we are already in
  COLOR_ATTACHMENT_OPTIMAL.

- Do not force GENERAL layout if resource has UNORDERED_ACCESS flag set.
  As this is not a promotable state, we have to explicitly transition
  into it. I tested this on validation layers, where even COMMON state
  refuses to promote to UAV state. The exception here of course is
  SIMULTANOUS_ACCESS, but we handle that properly now.

- Verify that UAV or SIMULTANEOUS access is not used together with DSV
  state. This is explicitly banned in the API docs.

- Actually emit image barriers. Batch the image transitions as that's
  what D3D12 docs encourage app developers to do, and it also expects
  that drivers can optimize this. Ensure that we respect the in-order
  resource barrier rules by splitting batches if there are overlaps in
  the transitions.

- Ensure that correct image layout is used when clearing a suspended
  render pass attachment.

I also cleaned up a ton of validation errors. There are some benign ones left, but it runs 100% clean w.r.t. image layout validation.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -2469,6 +2469,18 @@ enum vkd3d_render_pass_transition_mode
     VKD3D_RENDER_PASS_TRANSITION_MODE_END,
 };
 
+static bool d3d12_resource_requires_shader_visibility_after_transition(
+        const struct d3d12_resource *resource,
+        VkImageLayout old_layout, VkImageLayout new_layout)
+{
+    return !(resource->desc.Flags & D3D12_RESOURCE_FLAG_DENY_SHADER_RESOURCE) &&
+            old_layout != VK_IMAGE_LAYOUT_UNDEFINED &&
+            old_layout != new_layout &&
+            (new_layout == VK_IMAGE_LAYOUT_DEPTH_READ_ONLY_OPTIMAL ||
+                    new_layout == VK_IMAGE_LAYOUT_DEPTH_ATTACHMENT_STENCIL_READ_ONLY_OPTIMAL ||
+                    new_layout == VK_IMAGE_LAYOUT_DEPTH_READ_ONLY_STENCIL_ATTACHMENT_OPTIMAL);
+}
+
 static VkPipelineStageFlags vk_render_pass_barrier_from_view(const struct vkd3d_view *view, const struct d3d12_resource *resource,
         enum vkd3d_render_pass_transition_mode mode, VkImageLayout layout, bool clear, VkImageMemoryBarrier *vk_barrier)
 {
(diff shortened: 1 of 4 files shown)


HansKristian-Work:
On RX 6800, I got some gains:
HZD: ~2.5%
DS: ~2.5%
RE2: ~1.5%
CP77: ~2.5%

I verified with a HZD trace that there are a lot of internal barriers which vanished as a result. Being smarter about depth-stencil will be a challenge and should yield further gains, but we should be able to eliminate a ton of further redundant barriers this way.

libs/vkd3d/state.c
```diff
@@ -2626,12 +2626,35 @@ static unsigned int vkd3d_get_rt_format_swizzle(const struct vkd3d_format *forma
 
 STATIC_ASSERT(sizeof(struct vkd3d_shader_transform_feedback_element) == sizeof(D3D12_SO_DECLARATION_ENTRY));
 
+static VkImageAspectFlags vk_dsv_format_to_aspects(VkFormat vk_format)
```
doitsujin, line 2629: Do we know if it's legal to specify something like `R32_UINT` or `R32_FLOAT_X8X24_TYPELESS` as pipeline DSV formats?

If not, `vkd3d_format` already knows the exact aspect mask, so this could potentially be simplified (we'd need to fix the `DXGI_FORMAT_R24_UNORM_X8_TYPELESS` mapping + fallback logic, though).

HansKristian-Work: I suppose we could just look at the dsv->resource's aspect mask instead.

HansKristian-Work: Should be fixed now.

libs/vkd3d/vkd3d_private.h
```diff
@@ -838,6 +839,19 @@ static inline VkImageLayout d3d12_resource_pick_layout(const struct d3d12_resour
...
 }
 
+static inline VkImageLayout d3d12_resource_get_outside_render_pass_color_layout(const struct d3d12_resource *resource)
+{
+    if (resource->desc.Flags & D3D12_RESOURCE_FLAG_ALLOW_SIMULTANEOUS_ACCESS)
+        return resource->common_layout;
+    else
+        return d3d12_resource_pick_layout(resource, VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL);
+}
+
```
doitsujin, line 849: wouldn't it be more useful to fold this into `d3d12_resource_pick_layout` instead?

i.e. adding a new resource flag (`VKD3D_RESOURCE_SIMULTANEOUS_ACCESS`) and then changing that to something like
```c
    return resource->flags & (VKD3D_RESOURCE_LINEAR_TILING | VKD3D_RESOURCE_SIMULTANEOUS_ACCESS)
            ? resource->common_layout : layout;
```
This was the intended purpose at least.

libs/vkd3d/vkd3d_private.h
```diff
@@ -838,6 +839,19 @@ static inline VkImageLayout d3d12_resource_pick_layout(const struct d3d12_resour
...
+static inline VkImageLayout d3d12_resource_get_outside_render_pass_depth_stencil_layout(const struct d3d12_resource *resource)
+{
+    return resource->common_layout;
+}
```
doitsujin, line 853: do we really need this function?

libs/vkd3d/command.c
```diff
@@ -3324,6 +3333,14 @@ static void vk_access_and_stage_flags_from_d3d12_resource_state(const struct d3d
             case D3D12_RESOURCE_STATE_DEPTH_WRITE:
             case D3D12_RESOURCE_STATE_DEPTH_READ:
                 *stages |= VK_PIPELINE_STAGE_EARLY_FRAGMENT_TESTS_BIT | VK_PIPELINE_STAGE_LATE_FRAGMENT_TESTS_BIT;
+                *access |= VK_ACCESS_DEPTH_STENCIL_ATTACHMENT_READ_BIT;
+                /* If our default DS layout is attachment optimal, we will not perform implicit
+                 * memory barriers as part of a render pass. */
+                if (d3d12_resource_get_outside_render_pass_depth_stencil_layout(resource) ==
+                        VK_IMAGE_LAYOUT_DEPTH_STENCIL_ATTACHMENT_OPTIMAL)
+                {
+                    *access |= VK_ACCESS_DEPTH_STENCIL_ATTACHMENT_WRITE_BIT;
```
doitsujin, line 3389: is this necessary even in the `D3D12_RESOURCE_STATE_DEPTH_READ` case?

libs/vkd3d/vkd3d_private.h
HansKristian-Work: I think that approach is fine.

libs/vkd3d/command.c
HansKristian-Work: Given that it's not valid to write to depth in that layout, it's probably fine to omit the access flag in DEPTH_READ.

libs/vkd3d/vkd3d_private.h
```diff
@@ -734,6 +734,7 @@ enum vkd3d_resource_flag
     VKD3D_RESOURCE_LINEAR_TILING          = (1u << 4),
     VKD3D_RESOURCE_EXTERNAL               = (1u << 5),
     VKD3D_RESOURCE_ACCELERATION_STRUCTURE = (1u << 6),
+    VKD3D_RESOURCE_SIMULTANEOUS_ACCESS    = (1u << 7)
```
doitsujin, line 737: `,` at the end, unless we want to change two lines next time we add a flag.

libs/vkd3d/command.c
```diff
@@ -3330,6 +3335,14 @@ static void vk_access_and_stage_flags_from_d3d12_resource_state(const struct d3d
             case D3D12_RESOURCE_STATE_DEPTH_WRITE:
             case D3D12_RESOURCE_STATE_DEPTH_READ:
                 *stages |= VK_PIPELINE_STAGE_EARLY_FRAGMENT_TESTS_BIT | VK_PIPELINE_STAGE_LATE_FRAGMENT_TESTS_BIT;
+                *access |= VK_ACCESS_DEPTH_STENCIL_ATTACHMENT_READ_BIT;
+                /* If our default DS layout is attachment optimal, we will not perform implicit
+                 * memory barriers as part of a render pass. */
+                if (state == D3D12_RESOURCE_STATE_DEPTH_WRITE &&
```
doitsujin, line 3388: nit: why not just move this above the `D3D12_RESOURCE_STATE_DEPTH_READ` thing and do a good old `/* fall through */` at the end?

doitsujin:
Small nits now, looks good otherwise.

libs/vkd3d/command.c

libs/vkd3d/vkd3d_private.h
HansKristian-Work: Fixed. pretty sure trailing `,` is not really compliant invalid, but w/e, we do that everywhere.

doitsujin: It's actually explicitly allowed in C99.

doitsujin (approved):

libs/vkd3d/vkd3d_private.h
HansKristian-Work: I must be confusing this with C++

PR #691, opened 2021-06-10
Fix a shader bug in constant double vector handling

conor42:
The code path for operations using a constant double2 is currently not tested. These patches modify one of the tests and fix a bug.

Constant double vectors are also bugged in Windows, at least in the RX 580 driver, but the effect is to skip the operation for the second double entirely (or treat it as zero).


Joshua-Ashton:
Good catch! LGTM :frog:

Joshua-Ashton (approved):

PR #692, opened 2021-06-10
 vkd3d: Enable binary import and image view handle extensions

rgc183:
This also updates subprojects/Vulkan-Headers to 1.2.180

This is needed for VK_NVX_binary_import and VK_NVX_image_view_handle.

Signed-off-by: Roshan Chaudhari <rochaudhari@nvidia.com>

Reviewed-by: Liam Middlebrook <lmiddlebrook@nvidia.com>

### DIFF
--- a/include/vkd3d_win32.h
+++ b/include/vkd3d_win32.h
@@ -42,8 +42,19 @@
 #define WIDL_C_INLINE_WRAPPERS
 #include <vkd3d_windows.h>
 
+/* Vulkan headers include static const declarations. Enable static keyword for
+ * them.
+ */
+#ifdef __MINGW32__
+# undef static
+#endif
+
 #define VK_USE_PLATFORM_WIN32_KHR
 #include <vulkan/vulkan.h>
+
+#ifdef __MINGW32__
+# define static
+#endif
 
 #include <dxgi1_6.h>
 
--- a/libs/vkd3d/device.c
+++ b/libs/vkd3d/device.c
@@ -99,6 +99,8 @@ static const struct vkd3d_optional_extension_info optional_device_extensions[] =
     VK_EXTENSION(AMD_SHADER_CORE_PROPERTIES_2, AMD_shader_core_properties2),
     /* NV extensions */
     VK_EXTENSION(NV_SHADER_SM_BUILTINS, NV_shader_sm_builtins),
+    VK_EXTENSION(NVX_BINARY_IMPORT, NVX_binary_import),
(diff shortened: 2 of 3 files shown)


include/vkd3d_win32.h
```diff
@@ -42,9 +42,20 @@
 #define WIDL_C_INLINE_WRAPPERS
 #include <vkd3d_windows.h>
 
+/* Vulkan headers include static const declarations. Enable static keyword for
```
liam-middlebrook, line 45: Could you explain more why this is necessary? It's not really clear to me. Why is is just ifdef'd for MINGW32? What is specific about MINGW32 that causes issues here?

rgc183: line 32, makes hack for MINGW32 and causes `static` to meaningless. So we undo this hack for vulkan headers. If we do not, header files will have declaration of variables without static keyword and then linking with multiple libraries will throw "multiple definition"

liam-middlebrook: Could we move the inclusion of the Vulkan headers before this hack? Or does that cause issues with the WIN32 platform definition?

rgc183: vulkan header is calling windows.h which eventually calls to winbase.h in mingw, and it fails to find declaration for `IUnknown_**` .

HansKristian-Work: Right, we get silly stuff like this if we don't do this hack:
```
/usr/lib/gcc/x86_64-w64-mingw32/11.1.0/../../../../x86_64-w64-mingw32/bin/ld: libs/vkd3d/libvkd3d-proton.a(swapchain.c.obj):swapchain.c:(.rdata+0x29a8): multiple definition of `VK_PIPELINE_STAGE_2_TOP_OF_PIPE_BIT_KHR'; libs/d3d12/d3d12.dll.p/main.c.obj:main.c:(.rdata+0x8d0): first defined here
/usr/lib/gcc/x86_64-w64-mingw32/11.1.0/../../../../x86_64-w64-mingw32/bin/ld: libs/vkd3d/libvkd3d-proton.a(swapchain.c.obj):swapchain.c:(.rdata+0x29b0): multiple definition of `VK_PIPELINE_STAGE_2_NONE_KHR'; libs/d3d12/d3d12.dll.p/main.c.obj:main.c:(.rdata+0x8d8): first defined here
```
We already do stupid hacks for MinGW here, so I guess I'll let this slide.

HansKristian-Work (approved):
LGTM

PR #698, opened 2021-06-11
Promote and keep depth-stencil resources in OPTIMAL as long as possible

HansKristian-Work:
Goal here is to avoid unnecessary image layout transitions when render
passes toggle depth-stencil PSO states. Since we cannot know which
states a resource is in, we have to be conservative, and assume that
shader reads *could* happen.

The best effort we can do is to detect when writes happen to a DSV
resource. In this scenario, we can deduce that the aspect cannot be
read, since DEPTH_WRITE | RESOURCE state is not allowed.

To make the tracking somewhat sane, we only promote to OPTIMAL if an
entire image's worth of subresources for a given aspect is transitioned.
The common case for depth-stencil images is 1 mip / 1 layer anyways.

Some other changes are required here:
- Instead of common_layout for the depth image, we need to consult the
  command list, which might promote the layout to optimal.
- We make use of render pass compatibility rules which state that we can
  change attachment reference layouts as well as initial/finalLayout.
  To make this change, a pipeline will fill in a
  vkd3d_render_pass_compat struct.
- A command list has a dsv_plane_optimal_mask which keeps track
  of the plane aspects we have promoted to OPTIMAL, and we know cannot
  be read by shaders.
  The desired optimal mask is (existing optimal | PSO write).
  The initial existing optimal is inherited from the command list's
  tracker.
- RTV/DSV/views no longer keep track of VkImageLayout. This is
  unnecessary since we always deduce image layout based on context.

Overall, this shows a massive gain in HZD benchmark (RADV, 1440p ultimate, ~16% FPS on RX 6800).

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -2495,7 +2495,7 @@ static void d3d12_command_list_clear_attachment_pass(struct d3d12_command_list *
     dependencies[0].dstSubpass = 0;
     dependencies[0].srcStageMask = stages;
     dependencies[0].dstStageMask = stages;
-    dependencies[0].srcAccessMask = 0;
+    dependencies[0].srcAccessMask = clear_op ? access : 0;
     dependencies[0].dstAccessMask = access;
     dependencies[0].dependencyFlags = 0;
     dependencies[0].viewOffset = 0;
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -56,6 +56,10 @@ static void d3d12_command_list_notify_decay_dsv_resource(struct d3d12_command_li
 static uint32_t d3d12_command_list_notify_dsv_writes(struct d3d12_command_list *list,
         struct d3d12_resource *resource, const struct vkd3d_view *view,
         uint32_t plane_write_mask);
+static void d3d12_command_list_notify_dsv_discard(struct d3d12_command_list *list,
+        struct d3d12_resource *resource,
+        uint32_t first_subresource, uint32_t subresource_count,
+        uint32_t resource_subresource_count);
 static uint32_t d3d12_command_list_get_depth_stencil_resource_layout(const struct d3d12_command_list *list,
         const struct d3d12_resource *resource, uint32_t *plane_optimal_mask);
 static void d3d12_command_list_decay_optimal_dsv_resource(struct d3d12_command_list *list,
@@ -2246,6 +2250,36 @@ static uint32_t d3d12_command_list_notify_dsv_writes(struct d3d12_command_list *
     }
 }
 
+static void d3d12_command_list_notify_dsv_discard(struct d3d12_command_list *list,
+        struct d3d12_resource *resource,
+        uint32_t first_subresource, uint32_t subresource_count,
+        uint32_t resource_subresource_count)
+{
+    if (subresource_count == resource_subresource_count)
+    {
+        d3d12_command_list_promote_dsv_resource(list, resource,
+                VKD3D_DEPTH_PLANE_OPTIMAL | VKD3D_STENCIL_PLANE_OPTIMAL);
+    }
+    else if (resource->format->vk_aspect_mask == (VK_IMAGE_ASPECT_DEPTH_BIT | VK_IMAGE_ASPECT_STENCIL_BIT))
+    {
+        /* Can we at least discard a plane fully? */
+
+        if (first_subresource == 0 && subresource_count >= resource_subresource_count / 2)
+        {
+            if (resource->format->vk_aspect_mask & VK_IMAGE_ASPECT_DEPTH_BIT)
+                d3d12_command_list_promote_dsv_resource(list, resource, VKD3D_DEPTH_PLANE_OPTIMAL);
+            else if (resource->format->vk_aspect_mask & VK_IMAGE_ASPECT_STENCIL_BIT)
+                d3d12_command_list_promote_dsv_resource(list, resource, VKD3D_STENCIL_PLANE_OPTIMAL);
+        }
+        else if (first_subresource <= resource_subresource_count / 2 &&
+                first_subresource + subresource_count == resource_subresource_count)
+        {
+            if (resource->format->vk_aspect_mask & VK_IMAGE_ASPECT_STENCIL_BIT)
+                d3d12_command_list_promote_dsv_resource(list, resource, VKD3D_STENCIL_PLANE_OPTIMAL);
+        }
+    }
+}
+
 static void d3d12_command_list_notify_dsv_state(struct d3d12_command_list *list,
         struct d3d12_resource *resource, D3D12_RESOURCE_STATES state, UINT subresource)
 {
@@ -2363,6 +2397,7 @@ static void d3d12_command_list_clear_attachment_pass(struct d3d12_command_list *
     VkFramebuffer vk_framebuffer;
     VkRenderPass vk_render_pass;
     VkPipelineStageFlags stages;
+    uint32_t plane_write_mask;
     bool separate_ds_layouts;
     VkAccessFlags access;
     VkExtent3D extent;
@@ -2388,7 +2423,7 @@ static void d3d12_command_list_clear_attachment_pass(struct d3d12_command_list *
         if (is_bound)
             attachment_desc.initialLayout = list->dsv_layout;
         else
-            attachment_desc.initialLayout = resource->common_layout;
+            attachment_desc.initialLayout = d3d12_command_list_get_depth_stencil_resource_layout(list, resource, NULL);
 
         if (separate_ds_layouts)
         {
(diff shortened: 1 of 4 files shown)


libs/vkd3d/command.c
```diff
@@ -2079,6 +2109,202 @@ static void d3d12_command_list_clear_attachment_inline(struct d3d12_command_list
...
+{
+    const VkImageAspectFlags ds_aspects = VK_IMAGE_ASPECT_DEPTH_BIT | VK_IMAGE_ASPECT_STENCIL_BIT;
+    static const VkImageLayout layouts[] = {
+            VK_IMAGE_LAYOUT_DEPTH_STENCIL_READ_ONLY_OPTIMAL,
+            VK_IMAGE_LAYOUT_DEPTH_ATTACHMENT_STENCIL_READ_ONLY_OPTIMAL,
+            VK_IMAGE_LAYOUT_DEPTH_READ_ONLY_STENCIL_ATTACHMENT_OPTIMAL,
+            VK_IMAGE_LAYOUT_DEPTH_STENCIL_ATTACHMENT_OPTIMAL,
+    };
+
+    if ((aspects & ds_aspects) != ds_aspects)
```
doitsujin, line 2126: Is `aspects` supposed to be the aspect mask supported by the image format? This isn't super clear from the naming.

libs/vkd3d/command.c
```diff
@@ -2079,6 +2109,202 @@ static void d3d12_command_list_clear_attachment_inline(struct d3d12_command_list
...
+    assert(!(plane_optimal_mask & ~(VKD3D_DEPTH_PLANE_OPTIMAL | VKD3D_STENCIL_PLANE_OPTIMAL)));
+
+    /* No point in adding these since they are always deduced to be optimal. */
+    if (resource->desc.Flags & D3D12_RESOURCE_FLAG_DENY_SHADER_RESOURCE)
+        return VKD3D_DEPTH_PLANE_OPTIMAL | VKD3D_STENCIL_PLANE_OPTIMAL;
+
+    if (!(resource->format->vk_aspect_mask & VK_IMAGE_ASPECT_STENCIL_BIT))
+        plane_optimal_mask |= (plane_optimal_mask & VKD3D_DEPTH_PLANE_OPTIMAL) ? VKD3D_STENCIL_PLANE_OPTIMAL : 0;
+    if (!(resource->format->vk_aspect_mask & VK_IMAGE_ASPECT_DEPTH_BIT))
+        plane_optimal_mask |= (plane_optimal_mask & VKD3D_STENCIL_PLANE_OPTIMAL) ? VKD3D_DEPTH_PLANE_OPTIMAL : 0;
```
doitsujin, line 2195: I guess this means that unavailable aspects are expected to be `OPTIMAL`?

libs/vkd3d/command.c
```diff
@@ -2079,6 +2109,202 @@ static void d3d12_command_list_clear_attachment_inline(struct d3d12_command_list
...
+    }
+}
+
+static void d3d12_command_list_notify_dsv_state(struct d3d12_command_list *list,
+        struct d3d12_resource *resource, D3D12_RESOURCE_STATES state, UINT subresource)
+{
+    /* Need to decide if we should promote or decay or promote DSV optimal state.
+     * We can promote if we know for sure that all subresources are optimal.
+     * If we observe any barrier which leaves this state, we must decay. */
+    bool dsv_optimal = state == D3D12_RESOURCE_STATE_DEPTH_READ ||
```
doitsujin, line 2268: Do we know of any broken apps trying to read a `D3D12_RESOURCE_STATE_DEPTH_READ` resource inside shaders without also specifying a shader resource state?

I hope I'm wrong but this smells like it'll need app workarounds in the future.

libs/vkd3d/command.c
```diff
@@ -2079,6 +2109,202 @@ static void d3d12_command_list_clear_attachment_inline(struct d3d12_command_list
...
+        if (subresource == 0)
+        {
+            if (resource->format->vk_aspect_mask & VK_IMAGE_ASPECT_DEPTH_BIT)
+                d3d12_command_list_promote_dsv_resource(list, resource, VKD3D_DEPTH_PLANE_OPTIMAL);
+            else if (resource->format->vk_aspect_mask & VK_IMAGE_ASPECT_STENCIL_BIT)
+                d3d12_command_list_promote_dsv_resource(list, resource, VKD3D_STENCIL_PLANE_OPTIMAL);
+        }
+        else if (subresource == 1)
+        {
+            if (resource->format->vk_aspect_mask == (VK_IMAGE_ASPECT_STENCIL_BIT | VK_IMAGE_ASPECT_STENCIL_BIT))
```
doitsujin, line 2293: 2x `VK_IMAGE_ASPECT_STENCIL_BIT`, one of these neds to be `VK_IMAGE_ASPECT_DEPTH_BIT`

libs/vkd3d/command.c
```diff
@@ -2079,6 +2109,202 @@ static void d3d12_command_list_clear_attachment_inline(struct d3d12_command_list
...
+    barrier.dstQueueFamilyIndex = VK_QUEUE_FAMILY_IGNORED;
+    barrier.srcAccessMask = VK_ACCESS_DEPTH_STENCIL_ATTACHMENT_WRITE_BIT;
+    barrier.dstAccessMask = VK_ACCESS_DEPTH_STENCIL_ATTACHMENT_READ_BIT | VK_ACCESS_SHADER_READ_BIT;
+    barrier.subresourceRange.aspectMask = resource->format->vk_aspect_mask;
+    barrier.subresourceRange.baseMipLevel = 0;
+    barrier.subresourceRange.baseArrayLayer = 0;
+    barrier.subresourceRange.layerCount = VK_REMAINING_ARRAY_LAYERS;
+    barrier.subresourceRange.levelCount = VK_REMAINING_MIP_LEVELS;
+    barrier.image = resource->res.vk_image;
+    batch->src_stage_mask |= VK_PIPELINE_STAGE_LATE_FRAGMENT_TESTS_BIT;
```
doitsujin, line 2158: Don't we need `EARLY | LATE` here? My understanding is that `LATE` may not be performed and I'm not sure what drivers are or aren't allowed to optmize away here.

libs/vkd3d/command.c
```diff
@@ -2079,6 +2109,202 @@ static void d3d12_command_list_clear_attachment_inline(struct d3d12_command_list
...
+    batch->dst_stage_mask |= VK_PIPELINE_STAGE_ALL_COMMANDS_BIT;
+    d3d12_command_list_barrier_batch_add_layout_transition(list, batch, &barrier);
```
doitsujin, line 2160: I guess another question is whether we want to switch to using the separable depth and stencil layout thing at some point, assuming we are allowed to specify partial aspect masks for barriers if we do. Haven't looked at it in detail though.

doitsujin (commented):
Looked at the first commit for now, couple of questions + one bug.

In general, this stuff is getting complicated enough that key functions should probably have a comment about what their purpose is, it's not always immediately obvious by just looking at the code.

libs/vkd3d/command.c
HansKristian-Work: Derp

libs/vkd3d/command.c
HansKristian-Work: I don't know of any case. Worst case, we can pretend DEPTH_READ is not an optimal-type layout.

libs/vkd3d/command.c
HansKristian-Work: Goal here is to make sure depth/stencil states are both OPTIMAL or neither OPTIMAL so that we don't get separate layouts for no reason.

libs/vkd3d/command.c
HansKristian-Work: Need to check and comment a bit more.

libs/vkd3d/command.c
HansKristian-Work: src_stage can always be LATE as and store op is defined to happen in LATE stage.

libs/vkd3d/command.c
HansKristian-Work: The main point of separate DS layouts is that we can indeed use barriers on one aspect at a time. It won't buy us too much. At least we can enter OPTIMAL layout only for one aspect.

libs/vkd3d/command.c
HansKristian-Work: Added comments.

libs/vkd3d/command.c
HansKristian-Work: Commented.

libs/vkd3d/command.c
HansKristian-Work: Make code clearer and added comment.

libs/vkd3d/command.c
```diff
@@ -2246,6 +2250,36 @@ static uint32_t d3d12_command_list_notify_dsv_writes(struct d3d12_command_list *
...
+    if (subresource_count == resource_subresource_count)
+    {
+        d3d12_command_list_promote_dsv_resource(list, resource,
+                VKD3D_DEPTH_PLANE_OPTIMAL | VKD3D_STENCIL_PLANE_OPTIMAL);
+    }
+    else if (resource->format->vk_aspect_mask == (VK_IMAGE_ASPECT_DEPTH_BIT | VK_IMAGE_ASPECT_STENCIL_BIT))
+    {
+        /* Can we at least discard a plane fully? */
+
+        if (first_subresource == 0 && subresource_count >= resource_subresource_count / 2)
```
doitsujin, line 2267: why `/ 2`? The subresource count that's being passed into this function does not take planes into account.

This whole thing is weird anyway since we know the d3d12_resource, so this function could just infer the subresource count from that directly.

HansKristian-Work: It looked like it did when I wrote it, but will have another look.

HansKristian-Work: Ye, it doesn't. But that's a bug I think. Gonna separate that fix out to a different PR.

doitsujin (approved):

PR #700, opened 2021-06-16
Necromunda workaround

### LINKED ISSUE #681: Necromunda: Hired Gun: black/rainbow glow around some models
Mesa 21.1.1 on a AMD 5700 XT.

I've not noticed it anywhere else besides around NPCs, but they get a weird rainbow/black glow around them.

This does not happen with DXVK (i.e. when the game is set to Dx11 instead of 12).

Tracelog: [vkd3d.log](https://github.com/HansKristian-Work/vkd3d-proton/files/6576855/vkd3d.log)

Screenshot (not the best, but shows the issue on the female at the front):

HansKristian-Work:
Not the prettiest way to solve it, but any other solution will be extremely painful.

Fix #681.


DadSchoorse:
Why are you adding another app config system instead of doing what `VKD3D_CONFIG_FLAG_FORCE_TGSM_BARRIERS` does?

HansKristian-Work:
> Why are you adding another app config system instead of doing what `VKD3D_CONFIG_FLAG_FORCE_TGSM_BARRIERS` does?

I plan to move TGSM barriers to the quirk system to avoid the interface leak. For F1 I can add a default quirk.

Joshua-Ashton:
Does the global leak really matter given vkd3d-shader is statically linked?

HansKristian-Work:
> Does the global leak really matter given vkd3d-shader is statically linked?

Not a big deal, but it's not particularly elegant either.

libs/vkd3d-shader/spirv.c
```diff
@@ -9109,6 +9136,12 @@ static void vkd3d_dxbc_compiler_emit_sample(struct vkd3d_dxbc_compiler *compiler
             return;
     }
 
+    if (force_explicit_lod)
+    {
+        operands_mask |= SpvImageOperandsLodMask;
+        image_operands[image_operand_count++] = vkd3d_dxbc_compiler_get_constant_float(compiler, 0.0f);
```
doitsujin, line 9142: is image_operands large enough?

HansKristian-Work: Think so, we block Bias and LOD clamp this way.

HansKristian-Work: At most we get explicit LOD + texel offset in this path.

doitsujin (approved):

PR #702, opened 2021-06-17
vkd3d: reject creating a resource that is placed if the heap is too s…

hakzsam:
Fixes a Cyberpunk 2077 game bug, again...

### DIFF
--- a/libs/vkd3d/resource.c
+++ b/libs/vkd3d/resource.c
@@ -2552,6 +2552,24 @@ HRESULT d3d12_resource_create_placed(struct d3d12_device *device, const D3D12_RE
         /* Align manually. This works because we padded the required allocation size reported to the app. */
         VK_CALL(vkGetImageMemoryRequirements(device->vk_device, object->res.vk_image, &memory_requirements));
         heap_offset = align(heap_offset, memory_requirements.alignment);
+
+        if (heap_offset + memory_requirements.size > heap->allocation.resource.size)
+        {
+            ERR("Heap too small for the texture (heap=%"PRIu64", res=%"PRIu64".\n",
+                heap->allocation.resource.size, heap_offset + memory_requirements.size);
+            hr = E_INVALIDARG;
+            goto fail;
+        }
+    }
+    else
+    {
+        if (heap_offset + desc->Width > heap->allocation.resource.size)
+        {
+            ERR("Heap too small for the buffer (heap=%"PRIu64", res=%"PRIu64".\n",
+                heap->allocation.resource.size, heap_offset + desc->Width);
+            hr = E_INVALIDARG;
+            goto fail;
+        }
     }
 
     vkd3d_memory_allocation_slice(&object->mem, &heap->allocation, heap_offset, 0);


libs/vkd3d/resource.c
```diff
@@ -2552,6 +2552,12 @@ HRESULT d3d12_resource_create_placed(struct d3d12_device *device, const D3D12_RE
         /* Align manually. This works because we padded the required allocation size reported to the app. */
         VK_CALL(vkGetImageMemoryRequirements(device->vk_device, object->res.vk_image, &memory_requirements));
         heap_offset = align(heap_offset, memory_requirements.alignment);
+
+        if (heap_offset + memory_requirements.size > heap->allocation.resource.size) {
```
HansKristian-Work, line 2556: Brace is on next line

HansKristian-Work: Need to consider buffer case as well.

libs/vkd3d/resource.c
```diff
@@ -2552,6 +2552,12 @@ HRESULT d3d12_resource_create_placed(struct d3d12_device *device, const D3D12_RE
...
+            ERR("Heap too small for the resource.\n");
```
HansKristian-Work, line 2557: Probably log the values as well for good measure.

hakzsam: Done.

libs/vkd3d/resource.c
hakzsam: Sounds good.

libs/vkd3d/resource.c
```diff
@@ -2552,6 +2552,16 @@ HRESULT d3d12_resource_create_placed(struct d3d12_device *device, const D3D12_RE
         /* Align manually. This works because we padded the required allocation size reported to the app. */
         VK_CALL(vkGetImageMemoryRequirements(device->vk_device, object->res.vk_image, &memory_requirements));
         heap_offset = align(heap_offset, memory_requirements.alignment);
+    } else {
+        VK_CALL(vkGetBufferMemoryRequirements(device->vk_device, object->res.vk_buffer, &memory_requirements));
```
HansKristian-Work, line 2556: Braces, also as mentioned cannot use vkGetBufferMemoryRequirements here, since we always map to the underlying VkBuffer in the heap allocation. A buffer resource is just offset + range into a heap. So I think it's enough to check `heap_offset + desc->Width <= heap->allocation.resource.size`

HansKristian-Work (changes requested):

libs/vkd3d/resource.c
hakzsam: Oh right, misread your suggestion about vkGetBufferMemoryReqs.

libs/vkd3d/resource.c
```diff
@@ -2552,6 +2552,24 @@ HRESULT d3d12_resource_create_placed(struct d3d12_device *device, const D3D12_RE
...
+            hr = E_INVALIDARG;
+            goto fail;
+        }
+    }
+    else
+    {
+        if (heap_offset + desc->Width > heap->allocation.resource.size)
+        {
+            ERR("Heap too small for the buffer (heap=%"PRIu64", res=%"PRIu64".\n",
+                heap->allocation.resource.size, heap_offset + memory_requirements.size);
```
HansKristian-Work, line 2569: Width not memory_requirement.size

HansKristian-Work (approved):

PR #708, opened 2021-06-18
Use indexing by DXGI format to get Vulkan format.

DadSchoorse:


libs/vkd3d/utils.c
```diff
@@ -22,15 +22,43 @@
...
+#define EMPTY2 EMPTY1 EMPTY
+#define EMPTY3 EMPTY2 EMPTY
+#define EMPTY4 EMPTY3 EMPTY
+#define EMPTY5 EMPTY4 EMPTY
+#define EMPTY6 EMPTY5 EMPTY
+#define EMPTY7 EMPTY6 EMPTY
+#define EMPTY8 EMPTY7 EMPTY
+#define EMPTY9 EMPTY8 EMPTY
+#define EMPTY00
+#define EMPTY01 EMPTY9 EMPTY
```
Joshua-Ashton, line 47: Why is this 01 and not 10?

DadSchoorse: I didn't know that it as that easy to insert something in a macro :frog:

libs/vkd3d/utils.c
```diff
@@ -22,15 +22,43 @@
...
+#define EMPTY20 EMPTY10 EMPTY10
+#define EMPTY30 EMPTY20 EMPTY10
+#define EMPTY40 EMPTY30 EMPTY10
+#define EMPTY50 EMPTY40 EMPTY10
+#define EMPTY60 EMPTY50 EMPTY10
+#define EMPTY70 EMPTY60 EMPTY10
+#define EMPTY80 EMPTY70 EMPTY10
+#define EMPTY90 EMPTY80 EMPTY10
+/* creates 10 * tens + zero initialized empty array elements */
+#define UNUSED(tens, ones) EMPTY##tens##0 EMPTY##ones
```
doitsujin, line 57: why do we need this mess?

doitsujin: I'd rather just have a slow lookup for depth/stencil formats or store a pointer to a depth-stencil entry in the regular format lookup thingy or something than this.

HansKristian-Work:
Wouldn't it make more sense to build an indexable array once on device creation instead? We have to fixup formats for DS anyways on AMD, so might as well do it there. I'm not a fan of the macro weirdness either.

DadSchoorse:
> Wouldn't it make more sense to build an indexable array once on device creation instead? We have to fixup formats for DS anyways on AMD, so might as well do it there. I'm not a fan of the macro weirdness either.

Seems reasonable, I'll take that approach.

libs/vkd3d/vkd3d_private.h
```diff
@@ -2608,7 +2632,7 @@ struct d3d12_device
 
     HRESULT removed_reason;
 
-    const struct vkd3d_format *depth_stencil_formats;
+    const struct vkd3d_format depth_stencil_formats[VKD3D_MAX_DXGI_FORMAT + 1];
```
HansKristian-Work, line 2636: This should just be `struct vkd3d_format` to avoid the weird cast to `void *`.

libs/vkd3d/utils.c
```diff
@@ -385,10 +385,27 @@ static void vkd3d_init_depth_stencil_formats(struct d3d12_device *device)
     }
 }
 
+static void vkd3d_init_non_depth_stencil_formats(struct d3d12_device *device)
+{
+    struct vkd3d_format *formats = (void *) device->formats;
```
HansKristian-Work, line 390: Again, weird cast to `void*`

HansKristian-Work (changes requested):
Seems sensible to me. The main nit is the const-correctness.

libs/vkd3d/utils.c
DadSchoorse: Would  `(struct vkd3d_format *)` be okay for you? Imo it's better to ignore const once then to not have it all. This way we can at least make sure it's only modified here.

Joshua-Ashton: If we're throwing away const correctness anywhere, we shouldn't have it.

doitsujin: if we write to an array then it shouldn't be const.

doitsujin: ofc the big brain strategy would be to have a function return a `const struct vkd3d_format *` which both allocates and initializes that array, using non-const internally.

DadSchoorse: In that case you still have cast away const to free the memory.

Joshua-Ashton: That's a pretty common trope in C though. I don't have a problem with that.

libs/vkd3d/utils.c
```diff
@@ -353,60 +355,89 @@ static void vkd3d_cleanup_format_compatibility_lists(struct d3d12_device *device
...
-    struct vkd3d_format *formats;
     unsigned int i;
 
+    if (!(formats = vkd3d_calloc(VKD3D_MAX_DXGI_FORMAT + 1, sizeof(*formats))))
+        return E_OUTOFMEMORY;
+
     VK_CALL(vkGetPhysicalDeviceFormatProperties(device->vk_physical_device,
             VK_FORMAT_D24_UNORM_S8_UINT, &properties));
 
     if (properties.optimalTilingFeatures & VK_FORMAT_FEATURE_DEPTH_STENCIL_ATTACHMENT_BIT)
```
HansKristian-Work, line 369: This test needs to be inverted.

HansKristian-Work (changes requested):
Found a bug when I ran test suite

libs/vkd3d/utils.c
```diff
@@ -353,60 +355,89 @@ static void vkd3d_cleanup_format_compatibility_lists(struct d3d12_device *device
...
-
-        memcpy(formats, vkd3d_depth_stencil_formats, sizeof(vkd3d_depth_stencil_formats));
-        for (i = 0; i < count; ++i)
+    for (i = 0; i < ARRAY_SIZE(vkd3d_depth_stencil_formats); ++i)
+    {
+        assert(vkd3d_depth_stencil_formats[i].dxgi_format <= VKD3D_MAX_DXGI_FORMAT);
+        format = &formats[vkd3d_depth_stencil_formats[i].dxgi_format];
+        *format = vkd3d_depth_stencil_formats[i];
+        if (format->vk_format == VK_FORMAT_D24_UNORM_S8_UINT &&
+            properties.optimalTilingFeatures & VK_FORMAT_FEATURE_DEPTH_STENCIL_ATTACHMENT_BIT)
```
HansKristian-Work, line 381: And here

HansKristian-Work (approved):
LGTM

PR #709, opened 2021-06-21
vkd3d: Handle plane index when creating image descriptors.

doitsujin:
This will be necessary to handle multi-plane images correctly, where the view format will be a color format (with `VK_IMAGE_ASPECT_COLOR_BIT`) while the image itself will have `VK_IMAGE_ASPECT_PLANE_{0,1}_BIT` set. Also removes some unnecessary warnings.

For depth-stencil images, this also behaves correctly: In order to sample the stencil aspect of a depth-stencil image, the plane index must be set to 1 in D3D12, and failing to do so crashes the AMD native driver.


HansKristian-Work (approved):
LGTM

PR #716, opened 2021-06-25
Recover window position when leaving fullscreen properly

### LINKED ISSUE #715: Horizon Zero Dawn stops rendering when focusing out in fullscreen mode or leaving it with Alt+Enter
Alt+Enter does its job fine when going from windowed mode to fullscreen. When going from fullscreen to windowed, the game window disappears (the game still works as a process).

Alt-tabbing out of the fullscreen behaves leaves a black window stays on screen without rendering anything in there.

An immediate reason for this behaviour is the game initially (before creating the swapchain) calls SetWindowPos setting the window origin to 0,-10000 (out of the screen).
vkd3d-proton gets the window size just once at swapchain initialization, and sets this size back when going out of fullscreen in d3d12_swapchain_SetFullscreenState(). Meanwhile, application is changing window position and size after the swapchain is created. Also probably worth mentioning that the game calls MakeWindowAssociation(DXGI_MWA_NO_ALT_ENTER | DXGI_MWA_NO_WINDOW_CHANGES) but this did not affect anything in my testing.
It looks like the game aims to fully manage its window itself, including Alt+Enter processing (which, as I understand, would not currently work with vkd3d-proton otherwise).

I am attaching the patch which roughly simulates the game behaviour in triangle demo, and also has a potential fix which queries  window size in d3d12_swapchain_SetFullscreenState() when going from windowed mode. 
To see the effect, run the triangle with the changes and go to fullscreen with Alt+Enter and then back to windowed mode. 
Being run on Windows with native d3d12, the window ends up being 800x600.
On Windows with vkd3d-proton d3d12 (without the potential fix in the patch), the window disappears after going out of fullscreen just like in game.

[0001-vkd3d-Update-original_window_rect-in-d3d12_swapchain.txt](https://github.com/HansKristian-Work/vkd3d-proton/files/6712382/0001-vkd3d-Update-original_window_rect-in-d3d12_swapchain.txt)

HansKristian-Work:
This matches DXVK implementation behavior and fixes HZD alt-tabbing and alt+enter jank.

Fix #715.

### DIFF
--- a/libs/vkd3d/swapchain.c
+++ b/libs/vkd3d/swapchain.c
@@ -2069,6 +2069,9 @@ static HRESULT STDMETHODCALLTYPE d3d12_swapchain_SetFullscreenState(dxgi_swapcha
     }
 
     original_state = swapchain->fullscreen_desc.Windowed;
+    if (original_state)
+        GetWindowRect(swapchain->window, &swapchain->state.original_window_rect);
+
     swapchain->fullscreen_desc.Windowed = !fullscreen;
     hr = d3d12_swapchain_set_fullscreen(swapchain, target, original_state);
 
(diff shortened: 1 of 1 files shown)


Joshua-Ashton (approved):
LGTM

PR #719, opened 2021-06-25
vkd3d: Fix top-most handling when restoring from fullscreen

Joshua-Ashton:


HansKristian-Work (approved):
Inherited from DXVK, so LGTM based on that.

PR #723, opened 2021-06-29
tests: Test edge cases for VBO robustness with weird strides

HansKristian-Work:
- Passes WARP.
- Expected failures on AMD native windows driver are marked. The range check behaves as expected if we were to ignore dynamic stride rules in Vulkan. Elements which should be OOB are not marked as such.
- Fails quite a lot of RADV. Will file bugs.
- (Need to test NV)

### DIFF
--- a/tests/d3d12.c
+++ b/tests/d3d12.c
@@ -19737,6 +19737,164 @@ static void test_null_rtv(void)
     destroy_test_context(&context);
 }
 
+static void test_vbv_stride_edge_cases(void)
+{
+    D3D12_GRAPHICS_PIPELINE_STATE_DESC pso_desc;
+    D3D12_STREAM_OUTPUT_BUFFER_VIEW so_view;
+    D3D12_INPUT_LAYOUT_DESC input_layout;
+    D3D12_SO_DECLARATION_ENTRY so_entry;
+    struct test_context_desc desc;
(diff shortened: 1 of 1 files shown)


libs/vkd3d/command.c
```diff
@@ -4764,27 +4764,32 @@ static void d3d12_command_list_update_dynamic_state(struct d3d12_command_list *l
...
+                    dyn_state->viewport_count, dyn_state->scissors));
+        }
+        else if (dyn_state->dirty_flags & VKD3D_DYNAMIC_STATE_SCISSOR)
+        {
+            VK_CALL(vkCmdSetScissor(list->vk_command_buffer,
+                    0, dyn_state->viewport_count, dyn_state->scissors));
+        }
     }
+    else
+        FIXME_ONCE("Viewport count is zero. Is this allowed?\n");
```
doitsujin, line 4792: D3D11 allows this; it behaves as if rasterization is disabled so that geometry-only pipelines can still run without viewports. DXVK's not-so-sophisticated way of dealing with this is to bind a dummy viewport with an empty scissor rect; we should probably do the same.

HansKristian-Work: I'll do that then, and split it out to a separate PR.

HansKristian-Work:
Tested on NV as well. Passes fine.

doitsujin (approved):

PR #725, opened 2021-07-02
vkd3d: Return format2 information for d3d12_device_CheckFeatureSupport

rochaudhari:
Currently only format1 information is being returned for D3D12_FORMAT_SUPPORT.

Signed-off-by: Roshan Chaudhari <rochaudhari@nvidia.com>

### DIFF
--- a/libs/vkd3d/device.c
+++ b/libs/vkd3d/device.c
@@ -2933,7 +2933,13 @@ static HRESULT STDMETHODCALLTYPE d3d12_device_CheckFeatureSupport(d3d12_device_i
             if (image_features & VK_FORMAT_FEATURE_BLIT_SRC_BIT)
                 data->Support1 |= D3D12_FORMAT_SUPPORT1_MULTISAMPLE_RESOLVE;
             if (image_features & VK_FORMAT_FEATURE_STORAGE_IMAGE_BIT)
+            {
                 data->Support1 |= D3D12_FORMAT_SUPPORT1_TYPED_UNORDERED_ACCESS_VIEW;
+                if (device->device_info.features2.features.shaderStorageImageReadWithoutFormat)
+                    data->Support2 |= D3D12_FORMAT_SUPPORT2_UAV_TYPED_LOAD;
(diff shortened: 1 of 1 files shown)


libs/vkd3d/device.c
```diff
@@ -2933,7 +2933,10 @@ static HRESULT STDMETHODCALLTYPE d3d12_device_CheckFeatureSupport(d3d12_device_i
             if (image_features & VK_FORMAT_FEATURE_BLIT_SRC_BIT)
                 data->Support1 |= D3D12_FORMAT_SUPPORT1_MULTISAMPLE_RESOLVE;
             if (image_features & VK_FORMAT_FEATURE_STORAGE_IMAGE_BIT)
+            {
                 data->Support1 |= D3D12_FORMAT_SUPPORT1_TYPED_UNORDERED_ACCESS_VIEW;
+                data->Support2 |= D3D12_FORMAT_SUPPORT2_UAV_TYPED_LOAD | D3D12_FORMAT_SUPPORT2_UAV_TYPED_STORE;
```
HansKristian-Work, line 2938: This needs to check if ReadWithoutFormat and WriteWithoutFormat is supported I think.

HansKristian-Work (changes requested):
Nit.

libs/vkd3d/device.c
rochaudhari: Is it because` Load/Store `in HLSL does not have equivalent `OpTypeImage` ?

HansKristian-Work: Yes, all typed UAV loads and stores in DX are without format in the shader. The exception to this rule is FL 11.0 requirements where typed loads only work on R32_UINT/SINT/FLOAT, so we can infer the format in certain scenarios. Typed stores are without format always.

rochaudhari: Thanks for explaining.!

HansKristian-Work (approved):
LGTM

PR #729, opened 2021-07-02
Handle separate DS layouts in ClearDSView and CopyTexture

HansKristian-Work:
Leaving it as draft since there are some dubious validation warnings which don't make much sense.


HansKristian-Work:
Opening for review, as it was validation layer bugs getting in the way. Got a validation layer fix in flight.

doitsujin (approved):

PR #730, opened 2021-07-06
vkd3d: Add workaround for MinLODClamp.

HansKristian-Work:
Not correct, will need spec additions to handle it properly.
Fixes ground rendering in DIRT 5.

### DIFF
--- a/libs/vkd3d/resource.c
+++ b/libs/vkd3d/resource.c
@@ -671,6 +671,7 @@ static uint32_t vkd3d_view_entry_hash(const void *key)
             hash = hash_combine(hash, (uintptr_t)k->u.texture.format);
             hash = hash_combine(hash, k->u.texture.miplevel_idx);
             hash = hash_combine(hash, k->u.texture.miplevel_count);
+            hash = hash_combine(hash, float_bits_to_uint32(k->u.texture.miplevel_clamp));
             hash = hash_combine(hash, k->u.texture.layer_idx);
             hash = hash_combine(hash, k->u.texture.layer_count);
             hash = hash_combine(hash, k->u.texture.components.r);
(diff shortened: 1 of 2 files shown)


libs/vkd3d/resource.c
```diff
@@ -3415,6 +3418,25 @@ bool vkd3d_create_texture_view(struct d3d12_device *device, const struct vkd3d_t
...
+     * It should at least avoid a scenario where implicit LOD fetch from invalid levels.
+     * TODO: We will need an extension with vkCreateImageView pNext specificing minLODClamp.
+     * It will be trivial to add in RADV at least ... */
+    if (desc->miplevel_clamp >= 1.0f)
+    {
+        uint32_t new_base_level;
+        uint32_t level_offset;
+        uint32_t end_level;
+
+        assert(view_desc.subresourceRange.levelCount != VK_REMAINING_MIP_LEVELS);
```
doitsujin, line 3432: Is there a reason why we have an assert here? This shouldn't be hard at all to handle properly since we could just not change levelCount in that case.

libs/vkd3d/resource.c
```diff
@@ -3881,20 +3903,28 @@ static void vkd3d_create_texture_srv(struct d3d12_desc *descriptor,
                 key.u.texture.view_type = VK_IMAGE_VIEW_TYPE_2D;
                 key.u.texture.miplevel_idx = desc->Texture2D.MostDetailedMip;
                 key.u.texture.miplevel_count = desc->Texture2D.MipLevels;
+                key.u.texture.miplevel_clamp = desc->Texture2D.ResourceMinLODClamp;
                 key.u.texture.layer_count = 1;
                 key.u.texture.aspect_mask = vk_image_aspect_flags_from_d3d12(resource->format, desc->Texture2D.PlaneSlice);
                 if (desc->Texture2D.ResourceMinLODClamp)
-                    FIXME_ONCE("Unhandled min LOD clamp %.8e.\n", desc->Texture2D.ResourceMinLODClamp);
+                {
+                    FIXME_ONCE("Wrong handling of min LOD clamp %.8e. Falling back to view clamp.\n",
```
doitsujin, line 3911: If we want to print a warning, we should do so inside `vkd3d_create_texture_view` to avoid duplication.

Joshua-Ashton:
As mentioned on Discord we should do this for 1D too.

libs/vkd3d/resource.c

libs/vkd3d/resource.c

doitsujin (approved):

Joshua-Ashton (approved):

PR #734, opened 2021-07-07
Resolve subresource region

HansKristian-Work:


libs/vkd3d/command.c
```diff
@@ -9056,15 +9075,107 @@ static void STDMETHODCALLTYPE d3d12_command_list_SetSamplePositions(d3d12_comman
...
+        image_copy.extent = extent;
+
+        d3d12_command_list_copy_image(list, dst_resource, dst_resource->format,
+                src_resource, src_resource->format, &image_copy,
+                writes_full_subresource, overlapping_subresource);
+    }
+    else
+    {
+        /* TODO: MIN/MAX resolve for depth/stencil. */
+        FIXME("Unsupported resolve mode: %u.\n", mode);
```
doitsujin, line 9177: I feel like we should probably just pass the resolve mode to `d3d12_command_list_resolve_subresource` and handle it there, there are other situations where we'd want to use render pass resolves even with `AVERAGE` and I'd like to avoid unnecessary code duplication

doitsujin (approved):

PR #735, opened 2021-07-07
Reflect HS patch count in DXIL

HansKristian-Work:
Avoids late pipeline compile in DIRT5. Use the meta struct instead of random scan entry point.

### DIFF
--- a/include/vkd3d_shader.h
+++ b/include/vkd3d_shader.h
@@ -57,7 +57,7 @@ typedef uint64_t vkd3d_shader_hash_t;
 struct vkd3d_shader_meta
 {
     vkd3d_shader_hash_t hash;
-    uint32_t cs_workgroup_size[3]; /* Only contains valid data if uses_subgroup_size is true. */
+    unsigned int cs_workgroup_size[3]; /* Only contains valid data if uses_subgroup_size is true. */
     bool replaced;
     bool uses_subgroup_size;
 };
--- a/include/vkd3d_shader.h
+++ b/include/vkd3d_shader.h
@@ -58,6 +58,7 @@ struct vkd3d_shader_meta
 {
     vkd3d_shader_hash_t hash;
     unsigned int cs_workgroup_size[3]; /* Only contains valid data if uses_subgroup_size is true. */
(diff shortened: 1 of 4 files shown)


doitsujin (approved):

include/vkd3d_shader.h
```diff
@@ -57,8 +57,8 @@ typedef uint64_t vkd3d_shader_hash_t;
 struct vkd3d_shader_meta
 {
     vkd3d_shader_hash_t hash;
-    uint32_t cs_workgroup_size[3]; /* Only contains valid data if uses_subgroup_size is true. */
-    uint32_t patch_vertex_count; /* Relevant for HS. May be 0, in which case the patch vertex count is not known. */
+    unsigned int cs_workgroup_size[3]; /* Only contains valid data if uses_subgroup_size is true. */
```
Joshua-Ashton, line 60: Why did this change? These are equivalent.

include/vkd3d_shader.h
```diff
@@ -58,6 +58,7 @@ struct vkd3d_shader_meta
 {
     vkd3d_shader_hash_t hash;
     uint32_t cs_workgroup_size[3]; /* Only contains valid data if uses_subgroup_size is true. */
+    uint32_t patch_vertex_count; /* Relevant for HS. May be 0, in which case the patch vertex count is not known. */
```
Joshua-Ashton: Then this is uint32_t

include/vkd3d_shader.h
```diff
@@ -624,6 +625,7 @@ struct vkd3d_shader_scan_info
     bool has_side_effects;
     bool needs_late_zs;
     bool discards;
+    unsigned int patch_vertex_count;
```
Joshua-Ashton, line 628: Then the same thing here is unsigned int

Joshua-Ashton (changes requested):

include/vkd3d_shader.h
HansKristian-Work: The DXIL C API takes these by pointers, and uint32_t might not be == unsigned int.

HansKristian-Work: I'll reorder the changes so they make more sense.

Joshua-Ashton (approved):

PR #737, opened 2021-07-08
meta: Update version to 2.4.

HansKristian-Work:


HansKristian-Work:
Did a sweep over AMD and NV with various games and test suite. All checks out. Just one regression in Ghostrunner DXR that will be fixed first.

PR #741, opened 2021-07-09
vkd3d: Add config to use host-visible vram for UPLOAD heap

Joshua-Ashton:
Adds the "upload_hvv" config flag, which will make D3D12_HEAP_TYPE_UPLOAD attempt to use host-visible VRAM for allocations.

This takes advantage of large or resizable BAR if available.

I see a perf delta of 83-84 -> 92-94 (~12%) when using this in Horizon Zero Dawn.


Joshua-Ashton:
No change to `GetCustomHeapProperties` as this hasn't changed on AMD/NV drivers despite them using HVV for some of the UPLOAD heap.

DadSchoorse:
With that option there no memory types should get removed anymore. It even kinda works with the smol bar, at least CME saw some gains in Scarlet Nexus.
```diff
diff --git a/libs/vkd3d/resource.c b/libs/vkd3d/resource.c
index eeb722fd..19bc1676 100644
--- a/libs/vkd3d/resource.c
+++ b/libs/vkd3d/resource.c
@@ -5822,6 +5822,9 @@ static uint32_t vkd3d_memory_info_find_global_mask(struct d3d12_device *device)
     uint32_t heap_index;
     uint32_t i, mask;
 
+    if (vkd3d_config_flags & VKD3D_CONFIG_FLAG_UPLOAD_HVV)
+        return UINT32_MAX;
+
     for (i = 0; i < device->memory_properties.memoryHeapCount; i++)
     {
         heap_size = device->memory_properties.memoryHeaps[i].size;
```

HansKristian-Work:
I think this looks fine, and I added some extra commits. We forgot to use BAR for descriptor heap helper buffers.

To make this more robust, I think we need to consider a few things:
- Add AMD_overallocation_behavior
- Use memory priority on BAR allocations
- Add a budgeting system based on VRAM so we don't allocate endlessly on BAR and trigger VRAM trashing because of our allocations. Perhaps 4GB gets 256MB, 8GB gets 512MB, 16GB gets 2GB, etc. Once we reach the budget, we can just fail any internal allocation which exceeds the limit and fallback to host allocations.

Since this is hidden behind a config option, it's probably okay to leave it like this for now and then we can start working towards enabling this path by default.

HansKristian-Work:
Another thing to consider is that if DEVICE_LOCAL only allocation fails, we should not try to fallback to a DEVICE_LOCAL | HOST_VISIBLE allocation.

Joshua-Ashton:
LGTM

PR #745, opened 2021-07-15
Support ID3D12Fence::SetEvent with NULL event

HansKristian-Work:


doitsujin:
This seems overly complicated; can't we just create a new event and add then wait on it in SetEvent?

And just return immediately if the value to wait for has already been reached.

HansKristian-Work:
We'd need a new callback in the interface to be able to create and destroy events, and that seems more messy than this solution.

doitsujin:
Ah, right, `.so` builds screwing us over yet again.

libs/vkd3d/command.c
```diff
@@ -907,8 +924,20 @@ static HRESULT STDMETHODCALLTYPE d3d12_fence_SetEventOnCompletion(d3d12_fence_if
...
     ++fence->event_count;
 
+    /* If event is NULL, we need to block until the fence value completes.
+     * Implement this in a uniform way where we pretend we have a dummy event.
+     * A NULL fence->events[].event means that we should set latch to true
+     * and signal a condition variable instead of calling external signal_event callback. */
+    if (!event)
+    {
+        latch = false;
+        while (!latch)
```
doitsujin, line 937: do we not need atomic load/store for this, given that it happens on another thread?

HansKristian-Work: It happens while the fence object is locked, so not needed. It's a typical cond.wait([&]()); kind of thing.

HansKristian-Work: Signalling events/latches also happens while fence is locked.

doitsujin: ah yeah, derp.

doitsujin (approved):

PR #756, opened 2021-07-27
Fixing CopyTextureRegion going out of bounds when src_box is null.

dave-juicelabs:


dave-juicelabs:
This issue came about by having a src image with size (1840, 1840, 1) and a destination image with size (1024, 1024, 1). src_box was null and the src image size was causing an error during vkCmdCopyImage.

libs/vkd3d/command.c
```diff
@@ -5362,8 +5362,22 @@ static void vk_image_copy_from_d3d12(VkImageCopy *image_copy,
...
     else
     {
+        VkExtent3D dstExtent;
+
+        // Calculate the extent based on the source first
         unsigned int miplevel = image_copy->srcSubresource.mipLevel;
         vk_extent_3d_from_d3d12_miplevel(&image_copy->extent, src_desc, miplevel);
+
+        // Now make sure it does not extend past the destination when there is no src_box
+        miplevel = image_copy->dstSubresource.mipLevel;
```
Joshua-Ashton, line 5372: Just pass this directly to vk_extent_3d_from_d3d12_miplevel... Same for the above srcSubresource.

libs/vkd3d/command.c
```diff
@@ -5362,8 +5362,22 @@ static void vk_image_copy_from_d3d12(VkImageCopy *image_copy,
...
+        vk_extent_3d_from_d3d12_miplevel(&dstExtent, dst_desc, miplevel);
+
+        if((dst_x + image_copy->extent.width) > dstExtent.width)
```
Joshua-Ashton, line 5375: Use min here.

Joshua-Ashton (changes requested):
Please add a test case for this also.

Joshua-Ashton (changes requested):
Please split the test change into another commit before the original one.

HansKristian-Work (approved):
LGTM

Issue #669, opened 2021-05-24
Metro Exodus Enhanced Edition (1449560) - Title requires DXR 1.1 and FP16 support.


ZereoX:
### Edit:
See comment below: https://github.com/HansKristian-Work/vkd3d-proton/issues/669#issuecomment-899723484

Update logs with 7ff3ef2:
|VKD3D_LOG|VKD3D_DESCRIPTOR_QA|
| ---- | ---- |
|[VKD3D.zip](https://github.com/HansKristian-Work/vkd3d-proton/files/7039870/VKD3D.zip)|[VKD3D_DESCRIPTOR_QA.zip](https://github.com/HansKristian-Work/vkd3d-proton/files/7039871/VKD3D_DESCRIPTOR_QA.zip)|

### Description:
**_Metro Exodus Enhanced Edition_** (1449560) title reports that the minimum system requirements are not met when launched and closes to a crash handler if **_Run Anyway_** is used. Title indicates it requires SM6.5, but VKD3D-Proton only advertises / supports 6.2 (or 6.3 with `VKD3D_CONFIG=dxr`)

| VKD3D-Proton (2.3.1) | Crash Handler |
| ---- | ---- |
|![Screenshot from 2021-05-23 10-21-25](https://user-images.githubusercontent.com/761853/119345688-3ca48380-bc67-11eb-86d1-50f19a28a6a2.png)|![Crash Handler](https://user-images.githubusercontent.com/761853/119345862-72496c80-bc67-11eb-9019-7281e5659b42.png)|

Please let me know if any additional information is required or if additional testing is needed.

### Notes
- To run this application on Windows, you must be on version **20H2 or newer** to support SM 6.5.
- On Proton 6.3-4, the game remains open as long as the crash handler is not dismissed and plays through the opening cinematic before remaining stuck on a black screen .

### Software information
- Metro Exodus Enhanced Edition (1449560) is available with the purchase of the base game, [Metro Exodus](https://store.steampowered.com/app/412020/Metro_Exodus/).

### System information
OS: Windows 10 Pro 21H1 / Ubuntu 20.04
GPU: AMD Radeon RX 5700 XT
Drivers: 21.5.2 / 21.1.1
Wine version: Proton 6.3-4
VKD3D-Proton version: 2.3.1 / Master (925a930)

### Log files
[...]

PR #758, opened 2021-07-28
Expose shader model 6.4

HansKristian-Work:
- Improve checking for when we can expose Native16BitOperations.
- Validate use of native FP16 when creating pipelines. Apps which use min16float will fall back to FP32 on GPUs which cannot support it. For now, the fallback always takes place since there is no known scenario where native FP16 actually helps, but dxil-spirv has an option we can use for later if desired.
- Add test for edge cases in min16float behavior.
- Rewrite how shader models are exposed. We only need to consider required features, not optional ones. This allows us to expose SM 6.4 widely with latest dxil-spirv changes on master.
- Add tests for the esoteric SM 6.2 and 6.4 features exposed.


jqadev:
@HansKristian-Work what about SM 6.5 https://github.com/HansKristian-Work/vkd3d-proton/issues/669 ? Is it possible to add support for it in this PR?

HansKristian-Work:
> @HansKristian-Work what about SM 6.5 #669 ? Is it possible to add support for it in this PR?

SM 6.5 is for later. Requires some very difficult wave intrinsics.

Joshua-Ashton (approved):
Lgtm

PR #771, opened 2021-08-24
Implement DXR local static samplers

HansKristian-Work:
WIP


HansKristian-Work:
RT PSOs in Metro Exodus (the original build) all pass now. Ran into unrelated bugs though, but not RT related.

PR #772, opened 2021-08-24
vkd3d: Optimize repeated null descriptor updates

rlocatti-nv:
There are titles clearing the same descriptors constantly.
This leads to unnecessary updates that can become costly.

This commit introduces a new flag to track when D3D12 descriptors are not null, and skips clearing them if they are already null.
Descriptors are assumed to be null by default.

This fixes a performance regression introduced by 9983a1720f683b6366a9cf7a864a6efc94978e1d

Signed-off-by: Rodrigo Locatti <rlocatti@nvidia.com>

### DIFF
--- a/libs/vkd3d/resource.c
+++ b/libs/vkd3d/resource.c
@@ -3494,6 +3494,18 @@ static void d3d12_descriptor_heap_write_null_descriptor_template(struct d3d12_de
     VkDeviceAddress *va;
 
     heap = desc->heap;
+
+    /* When mutable descriptors are not supported, set a dummy type.
+       This will make those drivers not care about the null type being different between
+       null writes. */
+    if (!heap->null_descriptor_template.has_mutable_descriptors)
+        vk_mutable_descriptor_type = 0;
+
+    /* Skip writes with the same null type that are already null. */
+    if (!(desc->metadata.flags & VKD3D_DESCRIPTOR_FLAG_NON_NULL)
+            && desc->current_null_type == vk_mutable_descriptor_type)
+        return;
+
     num_writes = heap->null_descriptor_template.num_writes;
     vk_procs = &heap->device->vk_procs;
     offset = desc->heap_offset;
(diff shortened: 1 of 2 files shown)


Joshua-Ashton (approved):

libs/vkd3d/resource.c
```diff
@@ -3493,6 +3493,9 @@ static void d3d12_descriptor_heap_write_null_descriptor_template(struct d3d12_de
     unsigned int offset;
     VkDeviceAddress *va;
 
+    if (!(desc->metadata.flags & VKD3D_DESCRIPTOR_FLAG_NON_NULL))
```
HansKristian-Work, line 3496: I don't think this is correct for mutable_descriptor_type. When writing consecutive null descriptors, they can change type between two NULL descriptors. This seems correct for non-mutable however, so adding a check for that should be fine. Since this is pretty hot code, could be fine to add a bool to `heap->null_descriptor_template` to check if duplicate null descriptors should be elided.

rlocatti-nv:
Added a bool in `d3d12_descriptor_heap` to read whether the device supports mutable descriptors. There's a new VkDescriptorType in `d3d12_desc` on the space previously wasted on padding (on 64-bit builds) to track the last null descriptor type.

When a driver doesn't have mutable descriptors, the null write type is set to zero to avoid unnecessary clears (all Vulkan descriptors are already null). I'm not a big fan of the bool inside `d3d12_descriptor_heap` though.

libs/vkd3d/resource.c
```diff
@@ -3494,6 +3494,18 @@ static void d3d12_descriptor_heap_write_null_descriptor_template(struct d3d12_de
     VkDeviceAddress *va;
 
     heap = desc->heap;
+
+    /* When mutable descriptors are not supported, set a dummy type.
+       This will make those drivers not care about the null type being different between
+       null writes. */
+    if (!heap->has_mutable_descriptors)
```
HansKristian-Work, line 3501: I was thinking of placing this in heap->null_descriptor_template to keep the pointer chasing down to a bare minimum since we already load the set_mask in same cache line. The alternative is just to chase desc->heap->device->blah every time, but that seems a bit inoptimal since this is a performance PR after all.

libs/vkd3d/resource.c
```diff
@@ -5387,6 +5403,7 @@ static HRESULT d3d12_descriptor_heap_init(struct d3d12_descriptor_heap *descript
     descriptor_heap->refcount = 1;
     descriptor_heap->device = device;
     descriptor_heap->desc = *desc;
+    descriptor_heap->null_descriptor_template.has_mutable_descriptors = device->vk_info.VALVE_mutable_descriptor_type;
```
HansKristian-Work, line 5406: Should probably be moved to add_null_descriptor_template path where stuff is set up once.

HansKristian-Work (approved):
LGTM, thanks

PR #783, opened 2021-09-01
Planar copyable footprints

HansKristian-Work:
Implement GetCopyableFootprints for depth-stencil, which fixes crashes when entering benchmark in F1 2021.

### DIFF
--- a/libs/vkd3d/device.c
+++ b/libs/vkd3d/device.c
@@ -3846,10 +3846,12 @@ static void STDMETHODCALLTYPE d3d12_device_GetCopyableFootprints(d3d12_device_if
 {
     struct d3d12_device *device = impl_from_ID3D12Device(iface);
     static const struct vkd3d_format vkd3d_format_unknown
-            = {DXGI_FORMAT_UNKNOWN, VK_FORMAT_UNDEFINED, 1, 1, 1, 1, 0};
+            = {DXGI_FORMAT_UNKNOWN, VK_FORMAT_UNDEFINED, 1, 1, 1, 1, 0, 1};
 
     unsigned int i, sub_resource_idx, miplevel_idx, row_count, row_size, row_pitch;
-    unsigned int width, height, depth, array_size;
+    unsigned int width, height, depth, array_size, num_planes, num_subresources;
+    unsigned int num_subresources_per_plane, plane_idx;
+    const struct vkd3d_format *plane_format;
     const struct vkd3d_format *format;
     uint64_t offset, size, total;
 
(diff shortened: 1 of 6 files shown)


libs/vkd3d/device.c
```diff
@@ -3883,10 +3885,13 @@ static void STDMETHODCALLTYPE d3d12_device_GetCopyableFootprints(d3d12_device_if
         return;
     }
 
+    num_planes = max(vkd3d_popcount(format->vk_aspect_mask), 1);
```
doitsujin, line 3888: `format->plane_count`?

libs/vkd3d/device.c
```diff
@@ -3883,10 +3885,13 @@ static void STDMETHODCALLTYPE d3d12_device_GetCopyableFootprints(d3d12_device_if
...
     array_size = d3d12_resource_desc_get_layer_count(desc);
+    num_subresources_per_plane = desc->MipLevels * array_size;
+    num_subresources = num_subresources_per_plane * num_planes;
```
doitsujin, line 3891: We actually have `d3d12_resource_desc_get_sub_resource_count_per_plane` (and we arguably *should* have `d3d12_resource_desc_get_sub_resource_count` as well, but currently that one takes a resource rather than a resource desc).

libs/vkd3d/utils.c
```diff
@@ -483,6 +483,25 @@ const struct vkd3d_format *vkd3d_get_format(const struct d3d12_device *device,
     return format->dxgi_format ? format : NULL;
 }
 
+const struct vkd3d_format *vkd3d_format_footprint_for_plane(const struct d3d12_device *device,
```
doitsujin, line 486: Not sure I like this design; I'd rather have some sort of
```
struct {
  uint32_t bytes_per_pixel;
  VkExtent2D block_size;
};
```
that we can query from a vkd3d_format for any given plane (maybe omitting `block_size` for now since we don't support subsampled formats yet, but we'll probably need to at some point).

libs/vkd3d/device.c
HansKristian-Work: Fixed.

libs/vkd3d/utils.c
HansKristian-Work: Fixed.

libs/vkd3d/device.c
HansKristian-Work: Fixed.

doitsujin (approved):

PR #786, opened 2021-09-02
vkd3d: Remove deferred clears.

doitsujin:
Emitting render pass clears while we're in the process of starting a render pass overrides dsv layout tracking info.


HansKristian-Work (approved):
LGTM

PR #787, opened 2021-09-03
vkd3d: Fix regressions when introducing null-copy elision.

HansKristian-Work:
Need to initialize the set mask so that copies happen properly
on default-initialized descriptors. Also, move the current_null_type to
metadata so that it's properly copied on descriptor copy.


Joshua-Ashton (approved):
LGTM

PR #795, opened 2021-09-08
Narrow scope of pipeline cache workaround

HansKristian-Work:
470.62.02 driver fixes the bug, but need to keep workaround for older 470 releases.


Joshua-Ashton (approved):
lgtm, hopefully we can drop this entirely at somepoint though

HansKristian-Work:
Ye, when 470 series is no longer relevant I suppose we can nuke this.

PR #801, opened 2021-09-17
vkd3d: Copy all aspects in CopyResource.

HansKristian-Work:
Just like we're promoting layer count, also promote aspect mask.


HansKristian-Work:
Superseded by #802.

PR #803, opened 2021-09-18
Return E_INVALIDARG for SampleDesc.Count == 0

conor42:


doitsujin:
didn't we run into an app that required this to succeed?

conor42:
> 
> 
> didn't we run into an app that required this to succeed?

The reverted patch was for Psychonaughts 2, but it turns out to run better if the call fails, at least on an RX580/RADV. Was there something else?

Joshua-Ashton:
Windows doesnt return E_INVALIDARG though, not in this case. I have an apitrace that shows a call with it succeeding on native Windows. Maybe there is something else at play?

Joshua-Ashton:
inb4 this is an amd vs nv difference or something... The trace was recorded on NV

conor42:
> 
> 
> inb4 this is an amd vs nv difference or something... The trace was recorded on NV

Looks like it. I tested again in Windows + RX580, definitely E_INVALIDARG.

The game in DX12 mode has issues with excessive vram usage and needing to be run at low graphics settings. It wouldn't run at all on my system at default settings with upstream Wine + vkd3d-proton until I made this change. After that it did great.

Joshua-Ashton:
If its causing issues then I have no problem reverting this.

conor42:
Maybe the game should be tested on nvidia to make sure there isn't any hardware dependency weirdness going on

conor42:
Also if the test produces a different result on nvidia it will need a bug_if, or we could just omit testing entirely. Btw I stumbled upon this problem when adding a test to upstream vkd3d, where they're generally needed pass review.

Joshua-Ashton:
I will test this on NV today.

Joshua-Ashton:
Hm, test seems to pass fine here on NV...

PR #805, opened 2021-09-19
Backport "vkd3d-shader: Read the SM5 resource type instruction modifier." from vkd3d

sulix:
This is a hopefully-straightforward backport of @zfigura's commit from upstream vkd3d:
https://source.winehq.org/git/vkd3d.git/commit/c3a5df7375ba5d172fe6dd4d7987ed34777bc198

This doesn't actually fix the problem I was having (with Age of Empires IV), but it does get rid of a couple of the ``Unhandled modifier`` warnings which were cluttering up the log. Other than that, I've not tested it much, and know next-to-nothing about d3d12 (and only a little about Vulkan), so maybe it's broken…

Feel free to just close this if it's going to complicate a future bulk merge from vkd3d too much, but it's here if you want it. :-)

### DIFF
--- a/libs/vkd3d-shader/dxbc.c
+++ b/libs/vkd3d-shader/dxbc.c
@@ -23,7 +23,11 @@
 
 #define VKD3D_SM4_INSTRUCTION_MODIFIER        (0x1u << 31)
 
-#define VKD3D_SM4_MODIFIER_AOFFIMMI           0x1
+#define VKD3D_SM4_MODIFIER_MASK               0x3fu
+
+#define VKD3D_SM5_MODIFIER_RESOURCE_TYPE_SHIFT 6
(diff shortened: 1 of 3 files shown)


HansKristian-Work:
Is this a functional change? I see some changes to logging, but that's about it.

sulix:
I think so: it does plumb the resource_type through into the ``vkd3d_shader_instruction``, but there's nothing in the SPIRV emitter which seems to actually use it yet.

I admit to not knowing either DXBC or SPIRV well enough to know what (if any) the sensible translation for these per-instruction modifiers would be, so was wildly making changes trying to get AoE4 working, and this was the closest existing change I found. (My access to the game has expired, so I probably won't be looking at this again until it's properly released at the end of October, but I left a few notes in [the Proton bug](https://github.com/ValveSoftware/Proton/issues/5048) in case you're curious.)

No worries if just changing the logging doesn't seem worthwhile and you don't want to merge this in: but since I'd backported it while experimenting, I thought I'd at least give you the option. :-)

PR #807, opened 2021-09-21
impl_from overhead refactor

Joshua-Ashton:
- Move `impl_from`s into the header.
   - Previously we were doing function calls for every single object cast from another compilation unit which is... no.
   - Initial thoughts are that it's gross to require vtable exposed everywhere for the asserts, but C++ requires that for dynamic_cast so, meh.
 - Remove `unsafe_impl_from`s. These can just be the normal one and the checks get optimized away now. (yay)
 - Optimization for calling ResolveQueryData with zero queries.
 - Moved `d3d12_query_heap_type_get_data_size` to header.


HansKristian-Work (approved):
LGTM

PR #809, opened 2021-09-23
Update setup_vkd3d_proton.sh

Ahmed-E-86:
Add an optional WINE variable that can be used with custom wine builds.

### DIFF
--- a/setup_vkd3d_proton.sh
+++ b/setup_vkd3d_proton.sh
@@ -48,18 +48,9 @@ export WINEDEBUG=-all
 # wine gecko and mono
 export WINEDLLOVERRIDES="mscoree,mshtml="
 
-if [ -n "$WINE" ]; then
-   wine="$WINE"
-   wine64="${WINE}64"
-   wineboot="${WINE} wineboot"
-fi
-
-# avoid empty WINE variable
-if [ -z "$WINE" ]; then
-   wine="wine"
-   wine64="wine64"
-   wineboot="wineboot"
-fi
+wine="wine"
+wine64="wine64"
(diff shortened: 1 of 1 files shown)


HansKristian-Work:
Needs Signed-off-by line. `git commit --amend -s` to do this automatically.

doitsujin:
As mentioned on the DXVK PR, I don't see a good reason to do this when `PATH` and `WINELOADER` exist.

Ahmed-E-86:
I tried to use `WINELOADER` the same way I use `WINE`, and I end up with an error message. Could you explain how it can be used?

IroAlexis:
You can use $PATH variable like this `PATH=/path/to/wine-dir/bin:$PATH ./setup_vkd3d_proton.sh install`

Ahmed-E-86:
> You can use $PATH variable like this `PATH=/path/to/wine-dir/bin:$PATH ./setup_vkd3d_proton.sh install`

No, this does not work as it should be, because PATH does not define the location of wine64, and wineboot. Also, Proton and Proton-GE do not have wineboot binary, so in order to execute wineboot, you need to execute it as `wine wineboot` instead of just `wineboot`.

Ahmed-E-86:
> You can use $PATH variable like this `PATH=/path/to/wine-dir/bin:$PATH ./setup_vkd3d_proton.sh install`

Thanks for your comment. $PATH will work fine after `wineboot="wineboot"` changed to `wineboot="wine wineboot"`, otherwise wineboot of wine installed on the OS will be used instead, especially with Proton and Proton-GE, so a small change needs to be made.

PR #819, opened 2021-09-28
Fix unbound RTV rendering scenarios

HansKristian-Work:
In DIRT5, the game attempts to rendering to RTV1, but RTV 1 is unbound. Based on testing on native drivers, this is supposed to work with no validation errors. The fix is to introduce fallback pipeline checks somewhat similar to DSV unknown tests, except this is in reverse. Rendering to NULL DSV is actually disallowed, but NULL RTV is allowed ... *shrug*

### DIFF
--- a/libs/vkd3d/state.c
+++ b/libs/vkd3d/state.c
@@ -1346,7 +1346,8 @@ struct vkd3d_render_pass_entry
     VkRenderPass vk_render_pass;
 };
 
-STATIC_ASSERT(sizeof(struct vkd3d_render_pass_key) == 48);
+/* Ensure that key is packed, and can be memcmp'd. */
+STATIC_ASSERT(sizeof(struct vkd3d_render_pass_key) == 52);
 
 static VkImageLayout vkd3d_render_pass_get_depth_stencil_layout(const struct vkd3d_render_pass_key *key)
 {
@@ -1399,7 +1400,7 @@ static HRESULT vkd3d_render_pass_cache_create_pass_locked(struct vkd3d_render_pa
 
     for (index = 0, attachment_index = 0; index < rt_count; ++index)
     {
-        if (!key->vk_formats[index])
+        if (!(key->rtv_active_mask & (1u << index)))
         {
             attachment_references[index].sType = VK_STRUCTURE_TYPE_ATTACHMENT_REFERENCE_2_KHR;
             attachment_references[index].pNext = NULL;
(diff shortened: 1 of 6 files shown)


libs/vkd3d/state.c
```diff
@@ -1346,7 +1346,7 @@ struct vkd3d_render_pass_entry
     VkRenderPass vk_render_pass;
 };
 
-STATIC_ASSERT(sizeof(struct vkd3d_render_pass_key) == 48);
+STATIC_ASSERT(sizeof(struct vkd3d_render_pass_key) == 52);
```
doitsujin, line 1349: is there any particular reason we're insisting on a specific size here anyway?

libs/vkd3d/state.c
```diff
@@ -2657,7 +2657,7 @@ STATIC_ASSERT(sizeof(struct vkd3d_shader_transform_feedback_element) == sizeof(D
 
 static HRESULT d3d12_graphics_pipeline_state_create_render_pass_for_plane_mask(
         struct d3d12_graphics_pipeline_state *graphics, struct d3d12_device *device,
-        const struct vkd3d_format *dynamic_dsv_format,
+        uint32_t rtv_active_mask, const struct vkd3d_format *dynamic_dsv_format,
```
doitsujin, line 2660: we might wan to refactor this to use a struct at some point, rather than adding more arguments

libs/vkd3d/state.c
```diff
@@ -2779,7 +2780,8 @@ static HRESULT d3d12_graphics_pipeline_state_create_render_pass(
          plane_optimal_mask++)
     {
         if (FAILED(hr = d3d12_graphics_pipeline_state_create_render_pass_for_plane_mask(
-                graphics, device, dynamic_dsv_format,
+                graphics, device,
+                rtv_active_mask, dynamic_dsv_format,
```
doitsujin, line 2784: weird newline

doitsujin (commented):
Looks good overall, just one formatting nit and one question

libs/vkd3d/state.c
HansKristian-Work: It's to ensure there is no packing padding, which would screw up with the key compare which is memcmp.

libs/vkd3d/state.c

libs/vkd3d/state.c
HansKristian-Work: ye, it's getting a bit crowded. Save that for later.

libs/vkd3d/state.c
HansKristian-Work: added a comment

doitsujin (approved):

PR #826, opened 2021-10-03
Add a generic set of UE4 quirks

Joshua-Ashton:
I encountered another game, Bus Simulator 22, which has the same broken shader as Necromunda, including the same hash.

Looks like a bunch of UE4 games are affected, so check if the exe name ends in `-Shipping.exe` and apply these quirks for these shader hashes going forward.

I reeeeally would have liked to have done a regex here, but it looks like there's nothing nice and portable in the std and pulling in some external dependency is kinda gross, there is this which was licensed nice and is small however https://github.com/kokke/tiny-regex-c if we want to consider it.

Fixes Bus Simulator 22

### DIFF
--- a/include/private/vkd3d_string.h
+++ b/include/private/vkd3d_string.h
@@ -47,4 +47,35 @@ static inline bool vkd3d_string_ends_with(const char *str, const char *ending)
     return vkd3d_string_ends_with_n(str, strlen(str), ending, strlen(ending));
 }
 
+enum vkd3d_string_compare_mode
+{
+    VKD3D_STRING_COMPARE_NEVER,
+    VKD3D_STRING_COMPARE_ALWAYS,
+    VKD3D_STRING_COMPARE_EXACT,
+    VKD3D_STRING_COMPARE_STARTS_WITH,
+    VKD3D_STRING_COMPARE_ENDS_WITH,
+    VKD3D_STRING_COMPARE_CONTAINS,
+};
+
+static inline bool vkd3d_string_compare(enum vkd3d_string_compare_mode mode, const char *string, const char *comparator)
+{
+    switch (mode)
+    {
+        default:
+        case VKD3D_STRING_COMPARE_NEVER:
+            return false;
+        case VKD3D_STRING_COMPARE_ALWAYS:
+            return true;
+        case VKD3D_STRING_COMPARE_EXACT:
+            return !strcmp(string, comparator);
+        case VKD3D_STRING_COMPARE_STARTS_WITH:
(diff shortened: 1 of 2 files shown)


include/private/vkd3d_string.h
```diff
@@ -37,4 +37,14 @@ char *vkd3d_strdup_n(const char *str, size_t n);
 WCHAR *vkd3d_wstrdup(const WCHAR *str);
 WCHAR *vkd3d_wstrdup_n(const WCHAR *str, size_t n);
 
+static inline bool vkd3d_string_ends_with_n(const char *str, size_t str_len, const char *ending, size_t ending_len)
+{
+    return str_len >= ending_len && !strcmp(str + (str_len - ending_len), ending);
```
HansKristian-Work, line 42: _n naming here is somewhat misleading, since I would expect that to only look at len number of characters (i.e. strncmp).

libs/vkd3d/device.c
```diff
@@ -449,6 +449,7 @@ static const struct vkd3d_shader_quirk_info f1_2020_quirks = {
     NULL, 0, VKD3D_SHADER_QUIRK_FORCE_TGSM_BARRIERS,
 };
 
+/* Checks if the end of the application's name matches. */
```
HansKristian-Work, line 445: I would instead add a compare mode field to the vkd3d_shader_quirk_meta, where we either go for exact or "ends with" matching. Would also allows us to add "begins with" and "contains" later if necessary.

HansKristian-Work (changes requested):

include/private/vkd3d_string.h
```diff
@@ -37,4 +37,45 @@ char *vkd3d_strdup_n(const char *str, size_t n);
...
+            return true;
+        case VKD3D_STRING_COMPARE_EXACT:
+            return !strcmp(string, comparator);
+        case VKD3D_STRING_COMPARE_STARTS_WITH:
+            return !strncmp(string, comparator, strlen(comparator));
+        case VKD3D_STRING_COMPARE_ENDS_WITH:
+            return vkd3d_string_ends_with(string, comparator);
+        case VKD3D_STRING_COMPARE_CONTAINS:
+            return strstr(string, comparator) != NULL;
+    };
```
HansKristian-Work, line 77: weird semicolon

HansKristian-Work (changes requested):
Trivial nit

include/private/vkd3d_string.h
Joshua-Ashton: Dunno how he got there, fixed.

PR #846, opened 2021-10-08
Experimentally expose DXR 1.1

HansKristian-Work:
Improves test coverage to add:
- Intersection shaders
- Anyhit shaders
- AABB primitives
- Test behavior with opacity control flags
- TraceRay flags
- DXR 1.1 TraceRay flags
- DXR 1.1 RTPSO cull flags

Fix some minor bugs I found a long the way.


doitsujin (approved):

sfjuocekr:
Shadow of the Tomb Raider won't enable ray tracing either with VKD3D_CONFIG set to either/both dxr and dxr11.

Galcian79:
Never mind. I just discovered RADV does not support VK_KHR_ray_query extension yet. I was said only AMD beta drivers support that and it is broken anyway.

PR #852, opened 2021-10-12
Implement workarounds for Invariant position

HansKristian-Work:
Introduces a scheme where we can enable Invariant vertex positions, either:

- Globally
- Per-game
- Per-shader (ye, no)

After many trials and tribulations, make the choice to enable Invariant by default.


HansKristian-Work:
> You might want to consider "Monster Hunter World" too.

We should add all games which are known to be broken. Is it just MHW?

hakzsam:
AFAIK, yes. At least, RADV has a similar workaround for MHW DX12.

Oschowa:
RADV also has the same workaround applied for Shadow of the Tomb Raider dx11/12.

CME42:
Life is Strange True Colors should also be added: Siren-Win64-Shipping.exe

PR #854, opened 2021-10-13
vkd3d: Virtual query optimizations

misyltoad:

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -3123,6 +3123,7 @@ static bool d3d12_command_list_gather_pending_queries(struct d3d12_command_list 
         struct d3d12_query_heap *heap;
         uint32_t virtual_query_count;
         uint32_t unique_query_count;
+        uint32_t map_size;
         VkDeviceSize resolve_buffer_offset;
         VkDeviceSize resolve_buffer_size;
     };
@@ -3194,6 +3195,7 @@ static bool d3d12_command_list_gather_pending_queries(struct d3d12_command_list 
             d->heap = q->heap;
             d->virtual_query_count = 1;
             d->unique_query_count = 0;
+            d->map_size = 0;
             d->resolve_buffer_offset = resolve_buffer_size;
 
             r = NULL;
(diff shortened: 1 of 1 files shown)


libs/vkd3d/command.c
```diff
@@ -3263,7 +3263,8 @@ static bool d3d12_command_list_gather_pending_queries(struct d3d12_command_list
     for (i = 0; i < dispatch_count; i++)
     {
         struct dispatch_entry *d = &dispatches[i];
-        memset(query_map, 0, sizeof(*query_map) * query_map_size);
+        if (i != 0)
+            memset(query_map, 0, sizeof(*query_map) * query_map_size);
```
doitsujin: why is this better than the current code? *Something* needs to zero-initialize the memory in question anyway so this doesn't make a huge amount of sense to me.

And if we want to overcomplicate this for no reason, might as well clear the query map at the end rather than the beginning of the loop.

Joshua-Ashton: calloc doesn't necessarily memset. If it calls mmap to allocate, it can assume the page is already pre zero'ed and do nothing. So it's always better to use calloc instead of malloc + memset.

libs/vkd3d/command.c
```diff
@@ -3263,7 +3266,8 @@ static bool d3d12_command_list_gather_pending_queries(struct d3d12_command_list
     for (i = 0; i < dispatch_count; i++)
     {
         struct dispatch_entry *d = &dispatches[i];
-        memset(query_map, 0, sizeof(*query_map) * query_map_size);
```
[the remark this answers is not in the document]
Joshua-Ashton: We could split by start index too and offset I guess, that way query_map_size is the max size of the set of dispatches rather than dispatch + dispatch + dispatch

libs/vkd3d/command.c
doitsujin: Yeah that would be ideal, in which case this would have to be reverted anyway.

HansKristian-Work: From Discord, it seems like this wasn't flagged as a significant perf issue, just an observation (I think I misunderstood the discussion), should we table this for later? It seems like the correct solution is a rethink of the algorithm at a higher level.

PR #859, opened 2021-10-15
vkd3d: Pad reserved resources to 64k alignment.

HansKristian-Work:
Fix GPU crashes when attempting to bind non-aligned reserved resource.

### DIFF
--- a/libs/vkd3d/resource.c
+++ b/libs/vkd3d/resource.c
@@ -119,6 +119,11 @@ HRESULT vkd3d_create_buffer(struct d3d12_device *device,
         buffer_info.flags |= VK_BUFFER_CREATE_SPARSE_BINDING_BIT |
                 VK_BUFFER_CREATE_SPARSE_RESIDENCY_BIT |
                 VK_BUFFER_CREATE_SPARSE_ALIASED_BIT;
+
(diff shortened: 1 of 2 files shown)


libs/vkd3d/resource.c
```diff
@@ -2389,15 +2391,25 @@ static void d3d12_resource_destroy(struct d3d12_resource *resource, struct d3d12
...
     if (resource->desc.Dimension == D3D12_RESOURCE_DIMENSION_BUFFER)
     {
+        padded_desc = resource->desc;
+
+        /* If we attempt to bind sparse buffer with non-64k pages, we crash drivers.
+         * Specs seems a bit unclear how non-aligned VkBuffer sizes are supposed to work,
+         * so be safe. Pad out sparse buffers to their natural page size. */
+        if (resource->desc.Dimension == D3D12_RESOURCE_DIMENSION_BUFFER &&
+                (resource->flags & VKD3D_RESOURCE_RESERVED))
+            padded_desc.Width = align64(padded_desc.Width, VKD3D_TILE_SIZE);
```
doitsujin, line 2409: Can we handle this in vkd3d_create_buffer instead? We do some stuff regarding sparse resources there already, I think that would be the better place.

PR #861, opened 2021-10-18
vkd3d-shader: Merge i/o variables using the same location.

### LINKED ISSUE #609: Terrain misrendering in Anno 1800 (and poor performance)
When using the D3D12 renderer of Anno 1800, the terrain seems to be composed of multiples blocky textures.
All of this was tested on a RX 570 4GB VRAM.

This bug is present with RADV as well as AMDVLK.

Also note that the game will crash in a few seconds after starting a game with both AMDVLK and Windows 
driver. RADV looks stable.

There is also an issue with performance often being bad (~15fps instead of 40 with DXVK on a specific place). I suspect is has something to do VRAM limitation as it doesn't happen when using low quality texture instead of high. MangoHud shows that VRAM is used at near 4GB, even with low quality textures. I don't have this issue is both DXVK and windows D3D12. 
Can't tell for sure as the game crash fast on Windows, but I haven't encountered this performance issue on windows VKD3D.

RADV is using mesa 21.0 from kisak ppa.
VKD3D log at "warn" level: 
[vkd3d_anno1800_radv.txt](https://github.com/HansKristian-Work/vkd3d-proton/files/6275310/vkd3d_anno1800_radv.txt)
Tested on both VKD3D 2.2 (from Proton 6.3-2) and VKD3D git.
Tested on Proton 6.3-2

doitsujin:
Fixes #609.

### DIFF
--- a/libs/vkd3d-shader/spirv.c
+++ b/libs/vkd3d-shader/spirv.c
@@ -1244,17 +1244,6 @@ static void vkd3d_spirv_build_op_storev(struct vkd3d_spirv_builder *builder,
                 pointer_id, object_id, memory_access, operands, operand_count);
 }
 
-static void vkd3d_spirv_build_op_copy_memory(struct vkd3d_spirv_builder *builder,
-        uint32_t target_id, uint32_t source_id, uint32_t memory_access)
-{
-    if (!memory_access)
-        vkd3d_spirv_build_op2(&builder->function_stream, SpvOpCopyMemory,
-            target_id, source_id);
-    else
-        vkd3d_spirv_build_op3(&builder->function_stream, SpvOpCopyMemory,
-            target_id, source_id, memory_access);
-}
-
 static uint32_t vkd3d_spirv_build_op_select(struct vkd3d_spirv_builder *builder,
         uint32_t result_type, uint32_t condition_id, uint32_t object0_id, uint32_t object1_id)
 {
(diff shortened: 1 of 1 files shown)


libs/vkd3d-shader/spirv.c
```diff
@@ -4420,23 +4441,116 @@ static bool needs_private_io_variable(const struct vkd3d_shader_signature *signa
...
+    {
+        vkd3d_spirv_build_op_decorate1(builder, var_id, SpvDecorationLocation, 0);
+        vkd3d_spirv_build_op_decorate1(builder, var_id, SpvDecorationIndex, location);
+    }
+    else
+    {
+        vkd3d_spirv_build_op_decorate1(builder, var_id, SpvDecorationLocation, location);
+    }
+
+    if (compiler->shader_type == VKD3D_SHADER_TYPE_PIXEL && storage_class == SpvStorageClassInput)
```
HansKristian-Work, line 4541: Might make more sense to decorate Patch here, considering there's an early out for `var_ids`?

libs/vkd3d-shader/spirv.c
```diff
@@ -5085,6 +5148,10 @@ static void vkd3d_dxbc_compiler_emit_output(struct vkd3d_dxbc_compiler *compiler
 
     compiler->output_info[signature_idx].id = id;
     compiler->output_info[signature_idx].component_type = component_type;
+    compiler->output_info[signature_idx].dst_write_mask = (VKD3DSP_WRITEMASK_0 << output_component_count) - 1;
+
+    if (!builtin)
```
HansKristian-Work, line 5153: Not sure I understand the rationale behind this. Are there cases where write_mask would be smaller than num_components? Would that happen if multiple signature outputs share the same physical IO variable?

libs/vkd3d-shader/spirv.c
doitsujin: Yeah, oversight.

Issue #880, opened 2021-10-28
Age of Empires 4 (1466860) most units fail to render


aeikum:
Most of the in-game unit graphics fail to render in Age of Empires 4. Screenshot and log attached, notice the floating swords in the bottom-left.

The game is playable currently on the bleeding-edge branch of Proton Experimental. (Just needed 06104aec963 picked from upstream Wine to become playable.)

[steam-1466860.log](https://github.com/HansKristian-Work/vkd3d-proton/files/7435729/steam-1466860.log)

![Screenshot_2021-10-28_11-07-00](https://user-images.githubusercontent.com/34554541/139294151-1b4b10b8-99cd-4e6c-b5ad-e6fb4988e32b.png)

techy2493:
Trying to generate some logs for this, but haven't had a lot of luck so far getting logging turned on, am going to keep trying though. In the meantime wanted to indicate that the linked thread is now closed and a new one was created which continues to track this error on the Proton side.

https://github.com/ValveSoftware/Proton/issues/5258

I posted this there, but I'll include it  here as well for ease of access. Here's an image of rendering, I'll include settings as configured in a moment as an edit.

![image](https://media.giphy.com/media/CCtYDO5FOTFzLtiHmf/giphy.gif)

sulix:
I've re-run a few of my experiments on the final game, version 5.0.5544.0 (appid 1466860, build id 7595436), and my observations are:

There are two "types" of issues here (which are likely at least related, if not the same issue):
- Some objects don't display at all.
- Some display, but have incorrect transforms (the "wiggling" seen in the rocks, spears, some trees, etc).
It seems possible the former is simply the latter, with a big enough transform flaw that the rendered item is not visible at all (offscreen, culled, etc.)

Looking at the game under renderdoc, there are two possible codepaths which can occur, both of which have this issue. I think they probably depend on the game settings or something.
1. A depth pre-pass renders the objects (typically in Depth-only pass 3 or 4, with shader hash a2675ea7efda0520 or 063ac6fa716bddb6). See [this renderdoc capture](https://davidgow.net/stuff/aoe4-berrybush-depthpass.rdc.bz2).
2. The colour and depth are rendered at the same time (typically Color Pass 3, with shader hash 2a62ba8f721daf38 or fbda694dd47ef82e). See [this renderdoc capture](https://davidgow.net/stuff/aoe4-berrybush-colour.rdc.bz2)

[...]

PR #884, opened 2021-11-01
vkd3d: Handle D3D12_APPEND_ALIGNED_ELEMENT for <4 byte wide elements

sulix:
In d3d12, input element alignment needs to be the _minimum_ of 4 and the size of the type. See the D3D11 spec, section 4.4.6, which behaves similarly:
https://microsoft.github.io/DirectX-Specs/d3d/archive/D3D11_3_FunctionalSpec.htm#4.4.6%20Element%20Alignment

This is correctly taken into account when generating, e.g., the ``vertex_buffer_stride_align_mask`` used for validation, but is not taken into account when ``D3D12_APPEND_ALIGNED_ELEMENT`` is used to automatically place input elements. Currently, vkd3d always assumes the alignment is 4.

This means that, for example, bytes or shorts should be packed tightly together when ``D3D12_APPEND_ALIGNED_ELEMENT`` is used, but are instead padded to 4 bytes.

Fixing this makes units appear in Age of Empires IV (see vkd3d-proton issue #880 for examples.)

### DIFF
--- a/libs/vkd3d/state.c
+++ b/libs/vkd3d/state.c
@@ -2807,9 +2807,9 @@ static HRESULT compute_input_layout_offsets(const struct d3d12_device *device,
         if (e->AlignedByteOffset != D3D12_APPEND_ALIGNED_ELEMENT)
             offsets[i] = e->AlignedByteOffset;
         else
-            offsets[i] = input_slot_offsets[e->InputSlot];
+            offsets[i] = align(input_slot_offsets[e->InputSlot], min(4, format->byte_count));
 
-        input_slot_offsets[e->InputSlot] = align(offsets[i] + format->byte_count, 4);
+        input_slot_offsets[e->InputSlot] = offsets[i] + format->byte_count;
     }
 
     return S_OK;


libs/vkd3d/state.c
```diff
@@ -2809,7 +2809,7 @@ static HRESULT compute_input_layout_offsets(const struct d3d12_device *device,
         else
             offsets[i] = input_slot_offsets[e->InputSlot];
 
-        input_slot_offsets[e->InputSlot] = align(offsets[i] + format->byte_count, 4);
+        input_slot_offsets[e->InputSlot] = align(offsets[i] + format->byte_count, min(4, format->byte_count));
```
HansKristian-Work, line 2812: Nice find! Not sure this is a correct fix though. This aligns the next element based on the current one. I think a more correct formulation would be:

```
offsets[i] = align(offsets[i], min(4, format->byte_count));
input_slot_offsets[e->InputSlot] = offsets[i] + format->byte_count;
```

HansKristian-Work (changes requested):
Comment.

libs/vkd3d/state.c
sulix: Yeah, that makes more sense.

I guess ``input_slot_offsets`` needs to contain the unaligned offset, as there's no way to know how to align it in advance.

I've updated the patch with this fix.

HansKristian-Work (approved):
LGTM, thanks!

PR #887, opened 2021-11-02
Disable primitive restart for list topologies

HansKristian-Work:
IBStripCut only applies to strip primitive types. Use extended_dynamic_state2 when we can and fallback pipeline when we can't.

Builds on #1038.

### DIFF
--- a/libs/vkd3d/vkd3d_private.h
+++ b/libs/vkd3d/vkd3d_private.h
@@ -149,6 +149,7 @@ struct vkd3d_vulkan_info
     bool EXT_transform_feedback;
     bool EXT_vertex_attribute_divisor;
     bool EXT_extended_dynamic_state;
+    bool EXT_extended_dynamic_state2;
     bool EXT_external_memory_host;
     bool EXT_4444_formats;
     bool EXT_shader_image_atomic_int64;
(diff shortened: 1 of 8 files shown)


libs/vkd3d/vkd3d_private.h
```diff
@@ -1415,7 +1416,7 @@ HRESULT vkd3d_create_descriptor_set_layout(struct d3d12_device *device,
         VkDescriptorSetLayoutCreateFlags flags, unsigned int binding_count,
         const VkDescriptorSetLayoutBinding *bindings, VkDescriptorSetLayout *set_layout);
 
-#define VKD3D_MAX_DYNAMIC_STATE_COUNT (7)
+#define VKD3D_MAX_DYNAMIC_STATE_COUNT (8)
```
Joshua-Ashton, line 1419: Unrelated to your change but isn't this already wrong and now still wrong?

Before `dynamic_state_list` had 8 elements that could add a VkDynamicState, now it has 9...

https://github.com/HansKristian-Work/vkd3d-proton/blob/621c34a4f3d76849d267619ab6d271db4fa95dce/libs/vkd3d/state.c#L2815

doitsujin (approved):

PR #899, opened 2021-11-11
vkd3d: Handle 64KB_UNDEFINED_SWIZZLE.

HansKristian-Work:

Triggers in FH5.


Joshua-Ashton (approved):

DadSchoorse:
Shouldn't there be a warning if these images are host visible? Because writing to the mapped memory doesn't work as expected.

Joshua-Ashton:
@DadSchoorse this is undefined not standard swizzle.

DadSchoorse:
yeah sorry misread the docs, not only the mip packing is unknown but also the layout within the tiles

PR #905, opened 2021-11-16
Use VK_KHR_dynamic_rendering

doitsujin:
Keeping this as a draft for now since merging without support from stable drivers is not an option.

Tests pass on Nvidia, requires Vulkan beta driver 470.62.12 or later.

Needs more testing on RADV as its implementation of the extension comes along, and testing in actual games, although for that we'll need some sort of working Proton build with winevulkan support for this extension. I haven't checked if this is provided by proton-experimental bleeding edge yet.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -4270,7 +4270,6 @@ static void d3d12_command_list_reset_api_state(struct d3d12_command_list *list,
 
     list->current_pipeline = VK_NULL_HANDLE;
     list->command_buffer_pipeline = VK_NULL_HANDLE;
-    list->pso_render_pass = VK_NULL_HANDLE;
 
     memset(&list->dynamic_state, 0, sizeof(list->dynamic_state));
     list->dynamic_state.blend_constants[0] = D3D12_DEFAULT_BLEND_FACTOR_RED;
@@ -4422,6 +4421,7 @@ static bool d3d12_command_list_update_rendering_info(struct d3d12_command_list *
 
     graphics = &list->state->graphics;
     rtv_mask = graphics->rtv_active_mask & list->rtv_nonnull_mask;
+    rendering_info->rtv_mask = rtv_mask;
 
     rendering_info->info.colorAttachmentCount = vkd3d_get_color_attachment_count(graphics->rtv_active_mask);
     rendering_info->rtv_mask = rtv_mask;
@@ -4565,10 +4565,10 @@ static bool d3d12_command_list_update_graphics_pipeline(struct d3d12_command_lis
 {
     const struct vkd3d_vk_device_procs *vk_procs = &list->device->vk_procs;
     const struct vkd3d_render_pass_compatibility *render_pass_compat;
-    VkRenderPass vk_render_pass;
+    uint32_t dsv_plane_optimal_mask;
     uint32_t new_active_flags;
+    VkImageLayout dsv_layout;
     VkPipeline vk_pipeline;
-    uint32_t i;
 
     if (list->current_pipeline != VK_NULL_HANDLE)
         return true;
@@ -4590,42 +4590,28 @@ static bool d3d12_command_list_update_graphics_pipeline(struct d3d12_command_lis
             return false;
     }
 
-    /* Try to match render passes so we can stay compatible. */
-    for (i = 1, vk_render_pass = render_pass_compat->dsv_layouts[0];
-         vk_render_pass != list->pso_render_pass && i < ARRAY_SIZE(render_pass_compat->dsv_layouts);
-         i++)
+    if (d3d12_command_list_has_depth_stencil_view(list))
     {
-        vk_render_pass = render_pass_compat->dsv_layouts[i];
+        /* Select new dsv_layout. Any new PSO write we didn't observe yet must be updated here. */
+        dsv_plane_optimal_mask = list->dsv_plane_optimal_mask | list->state->graphics.dsv_plane_optimal_mask;
+        dsv_layout = dsv_plane_optimal_mask_to_layout(dsv_plane_optimal_mask, list->dsv.format->vk_aspect_mask);
+    }
+    else
+    {
+        dsv_plane_optimal_mask = 0;
+        dsv_layout = VK_IMAGE_LAYOUT_UNDEFINED;
     }
 
-    /* The render pass cache ensures that we use the same Vulkan render pass
-     * object for compatible render passes.
-     * If vk_render_pass == list->pso_render_pass, we know for certain
-     * that the PSO does not add any write masks we didn't already account for. */
-    if (list->pso_render_pass != vk_render_pass)
+    /* If we need to bind or unbind certain render targets or if the DSV layout changed, interrupt rendering */
+    if (((list->state->graphics.rtv_active_mask & list->rtv_nonnull_mask) != list->rendering_info.rtv_mask) ||
+            (dsv_layout != list->rendering_info.dsv.imageLayout))
     {
         d3d12_command_list_invalidate_rendering_info(list);
-        /* Don't end render pass if none is active, or otherwise
-         * deferred clears are not going to work as intended. */
-        if (list->rendering_info.state_flags & (VKD3D_RENDERING_ACTIVE | VKD3D_RENDERING_SUSPENDED))
-            d3d12_command_list_end_current_render_pass(list, false);
-
-        if (d3d12_command_list_has_depth_stencil_view(list))
-        {
-            /* Select new dsv_layout. Any new PSO write we didn't observe yet must be updated here. */
-            list->dsv_plane_optimal_mask |= list->state->graphics.dsv_plane_optimal_mask;
-            list->dsv_layout = dsv_plane_optimal_mask_to_layout(list->dsv_plane_optimal_mask,
-                    list->dsv.format->vk_aspect_mask);
-            /* Pick render pass based on new plane optimal mask. */
-            list->pso_render_pass = render_pass_compat->dsv_layouts[list->dsv_plane_optimal_mask];
-        }
-        else
-        {
-            list->pso_render_pass = render_pass_compat->dsv_layouts[0];
-            list->dsv_plane_optimal_mask = 0;
-            list->dsv_layout = VK_IMAGE_LAYOUT_UNDEFINED;
-        }
+        d3d12_command_list_end_current_render_pass(list, false);
     }
+
+    list->dsv_plane_optimal_mask = dsv_plane_optimal_mask;
+    list->dsv_layout = dsv_layout;
 
     if (list->command_buffer_pipeline != vk_pipeline)
     {
@@ -5300,7 +5286,6 @@ static bool d3d12_command_list_begin_render_pass(struct d3d12_command_list *list
 {
     const struct vkd3d_vk_device_procs *vk_procs = &list->device->vk_procs;
     struct d3d12_graphics_pipeline_state *graphics;
-    VkRenderPass vk_render_pass;
 
     d3d12_command_list_promote_dsv_layout(list);
     if (!d3d12_command_list_update_graphics_pipeline(list))
@@ -5318,9 +5303,6 @@ static bool d3d12_command_list_begin_render_pass(struct d3d12_command_list *list
         d3d12_command_list_handle_active_queries(list, false);
         return true;
     }
-
-    vk_render_pass = list->pso_render_pass;
-    assert(vk_render_pass);
 
     if (!(list->rendering_info.state_flags & VKD3D_RENDERING_SUSPENDED))
         d3d12_command_list_emit_render_pass_transition(list, VKD3D_RENDER_PASS_TRANSITION_MODE_BEGIN);
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -2045,9 +2045,9 @@ static inline struct d3d12_command_list *impl_from_ID3D12GraphicsCommandList(d3d
     return CONTAINING_RECORD(iface, struct d3d12_command_list, ID3D12GraphicsCommandList_iface);
 }
 
-static void d3d12_command_list_invalidate_current_framebuffer(struct d3d12_command_list *list)
+static void d3d12_command_list_invalidate_rendering_info(struct d3d12_command_list *list)
 {
-    list->current_framebuffer = VK_NULL_HANDLE;
+    list->rendering_info.state_flags &= ~VKD3D_RENDERING_CURRENT;
 }
 
 static void d3d12_command_list_invalidate_current_pipeline(struct d3d12_command_list *list, bool meta_shader)
@@ -3497,21 +3497,17 @@ static void d3d12_command_list_end_current_render_pass(struct d3d12_command_list
                 list->so_counter_buffers, list->so_counter_buffer_offsets));
     }
 
-    if (list->current_render_pass)
-    {
-        VkSubpassEndInfoKHR subpass_end_info;
-        subpass_end_info.sType = VK_STRUCTURE_TYPE_SUBPASS_END_INFO_KHR;
-        subpass_end_info.pNext = NULL;
-
-        VK_CALL(vkCmdEndRenderPass2KHR(list->vk_command_buffer, &subpass_end_info));
-    }
+    if (list->rendering_info.state_flags & VKD3D_RENDERING_ACTIVE)
+        VK_CALL(vkCmdEndRenderingKHR(list->vk_command_buffer));
 
     /* Don't emit barriers for temporary suspension of the render pass */
-    if (!suspend && (list->current_render_pass || list->render_pass_suspended))
+    if (!suspend && (list->rendering_info.state_flags & (VKD3D_RENDERING_ACTIVE | VKD3D_RENDERING_SUSPENDED)))
         d3d12_command_list_emit_render_pass_transition(list, VKD3D_RENDER_PASS_TRANSITION_MODE_END);
 
-    list->render_pass_suspended = suspend && (list->current_render_pass || list->render_pass_suspended);
-    list->current_render_pass = VK_NULL_HANDLE;
+    if (suspend && (list->rendering_info.state_flags & (VKD3D_RENDERING_ACTIVE)))
+        list->rendering_info.state_flags |= VKD3D_RENDERING_SUSPENDED;
+
+    list->rendering_info.state_flags &= ~VKD3D_RENDERING_ACTIVE;
 
     if (list->xfb_enabled)
     {
@@ -3528,11 +3524,6 @@ static void d3d12_command_list_end_current_render_pass(struct d3d12_command_list
 
         list->xfb_enabled = false;
     }
-}
-
-static void d3d12_command_list_invalidate_current_render_pass(struct d3d12_command_list *list)
-{
-    d3d12_command_list_end_current_render_pass(list, false);
 }
 
 static void d3d12_command_list_invalidate_push_constants(struct vkd3d_pipeline_bindings *bindings)
@@ -4304,11 +4295,9 @@ static void d3d12_command_list_reset_api_state(struct d3d12_command_list *list,
 
     list->has_valid_index_buffer = false;
 
-    list->current_framebuffer = VK_NULL_HANDLE;
     list->current_pipeline = VK_NULL_HANDLE;
     list->command_buffer_pipeline = VK_NULL_HANDLE;
     list->pso_render_pass = VK_NULL_HANDLE;
-    list->current_render_pass = VK_NULL_HANDLE;
 
     memset(&list->dynamic_state, 0, sizeof(list->dynamic_state));
     list->dynamic_state.blend_constants[0] = D3D12_DEFAULT_BLEND_FACTOR_RED;
@@ -4361,7 +4350,7 @@ static void d3d12_command_list_reset_internal_state(struct d3d12_command_list *l
     list->dsv_resource_tracking_count = 0;
     list->tracked_copy_buffer_count = 0;
 
-    list->render_pass_suspended = false;
+    list->rendering_info.state_flags = 0;
 }
 
 static void d3d12_command_list_reset_state(struct d3d12_command_list *list,
@@ -4482,29 +4471,37 @@ static bool d3d12_command_list_create_framebuffer(struct d3d12_command_list *lis
     return true;
 }
 
-static bool d3d12_command_list_update_current_framebuffer(struct d3d12_command_list *list)
+static bool d3d12_command_list_update_rendering_info(struct d3d12_command_list *list)
 {
-    VkImageView views[D3D12_SIMULTANEOUS_RENDER_TARGET_COUNT + 2];
+    struct vkd3d_rendering_info *rendering_info = &list->rendering_info;
     struct d3d12_graphics_pipeline_state *graphics;
-    VkFramebuffer vk_framebuffer;
-    unsigned int view_count;
     uint32_t rtv_mask;
-    VkExtent3D extent;
     unsigned int i;
 
-    if (list->current_framebuffer != VK_NULL_HANDLE)
+    if (rendering_info->state_flags & VKD3D_RENDERING_CURRENT)
         return true;
 
     graphics = &list->state->graphics;
     rtv_mask = graphics->rtv_active_mask & list->rtv_nonnull_mask;
-    view_count = 0;
 
-    /* The pipeline has fallback render passes / PSO in case we're
-     * attempting to render to unbound RTV. */
-    while (rtv_mask)
+    rendering_info->info.colorAttachmentCount = vkd3d_get_color_attachment_count(graphics->rtv_active_mask);
+    rendering_info->rtv_mask = rtv_mask;
+
+    /* The pipeline has fallback PSO in case we're attempting to render to unbound RTV. */
+    for (i = 0; i < D3D12_SIMULTANEOUS_RENDER_TARGET_COUNT; i++)
     {
-        i = vkd3d_bitmask_iter32(&rtv_mask);
-        views[view_count++] = list->rtvs[i].view->vk_image_view;
+        VkRenderingAttachmentInfoKHR *attachment = &rendering_info->rtv[i];
+
+        if (rtv_mask & (1u << i))
+        {
+            attachment->imageView = list->rtvs[i].view->vk_image_view;
+            attachment->imageLayout = VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL;
+        }
+        else
+        {
+            attachment->imageView = VK_NULL_HANDLE;
+            attachment->imageLayout = VK_IMAGE_LAYOUT_UNDEFINED;
+        }
     }
 
     if (d3d12_command_list_has_depth_stencil_view(list))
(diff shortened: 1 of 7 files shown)


libs/vkd3d/command.c
```diff
@@ -2671,96 +2645,76 @@ static void d3d12_command_list_clear_attachment_pass(struct d3d12_command_list *
...
     }
 
-    if (!d3d12_command_allocator_add_render_pass(list->allocator, vk_render_pass))
+    memset(&image_barriers[0], 0, sizeof(image_barriers[0]));
+    image_barriers[0].sType = VK_STRUCTURE_TYPE_IMAGE_MEMORY_BARRIER;
+    image_barriers[0].srcQueueFamilyIndex = VK_QUEUE_FAMILY_IGNORED;
+    image_barriers[0].dstQueueFamilyIndex = VK_QUEUE_FAMILY_IGNORED;
+    image_barriers[0].image = resource->res.vk_image;
+    image_barriers[0].newLayout = attachment_info.imageLayout;
+    image_barriers[0].subresourceRange.aspectMask = view->format->vk_aspect_mask;
```
HansKristian-Work, line 2611: This seems wrong if we only want to do single-aspect clear.

libs/vkd3d/command.c
```diff
@@ -2578,65 +2578,41 @@ static void d3d12_command_list_clear_attachment_pass(struct d3d12_command_list *
...
+        final_layouts[0] = dsv_plane_optimal_mask_to_layout(
                 d3d12_command_list_notify_dsv_writes(list, resource, view, plane_write_mask),
                 resource->format->vk_aspect_mask);
 
         if (separate_ds_layouts)
         {
-            stencil_attachment_desc.stencilFinalLayout = vk_separate_stencil_layout(attachment_desc.finalLayout);
-            attachment_desc.finalLayout = vk_separate_depth_layout(attachment_desc.finalLayout);
+            /* Do not transition aspects that we are not supposed to clear */
+            attachment_info.imageLayout = (clear_aspects & VK_IMAGE_ASPECT_DEPTH_BIT)
```
HansKristian-Work, line 2543: Wouldn't it be more correct to use attachment desc layout = final_layout here? The clear aspects we pass in to notify_dsv_writes will promote anything to OPTIMAL if we have writes.

libs/vkd3d/command.c
```diff
@@ -2645,12 +2621,12 @@ static void d3d12_command_list_clear_attachment_pass(struct d3d12_command_list *
             if (separate_ds_layouts)
             {
                 if (clear_aspects & VK_IMAGE_ASPECT_DEPTH_BIT)
-                    attachment_desc.initialLayout = VK_IMAGE_LAYOUT_UNDEFINED;
+                    initial_layouts[0] = VK_IMAGE_LAYOUT_UNDEFINED;
                 if (clear_aspects & VK_IMAGE_ASPECT_STENCIL_BIT)
-                    stencil_attachment_desc.stencilInitialLayout = VK_IMAGE_LAYOUT_UNDEFINED;
+                    initial_layouts[0] = VK_IMAGE_LAYOUT_UNDEFINED;
```
HansKristian-Work: Typo? It seems like this gets addressed in a following commit, but should probably be part of this commit instead.

libs/vkd3d/command.c
```diff
@@ -2671,96 +2645,76 @@ static void d3d12_command_list_clear_attachment_pass(struct d3d12_command_list *
...
+        image_barriers[1] = image_barriers[0];
+        image_barriers[1].newLayout = stencil_attachment_info.imageLayout;
+        image_barriers[1].subresourceRange.aspectMask = VK_IMAGE_ASPECT_STENCIL_BIT;
+        image_barrier_count = 2;
+    }
+    else
+        image_barrier_count = 1;
 
-    if (!d3d12_command_list_create_framebuffer(list, vk_render_pass, 1, &view->vk_image_view, extent, &vk_framebuffer))
+    for (i = 0; i < image_barrier_count; i++)
```
HansKristian-Work, line 2635: This seems like it would pessimize attachments which don't need layout transition. I would imagine that if oldLayout == newLayout, we can elide barrier for that case. That is what we did implicitly with the older method with external subpass barrier.

libs/vkd3d/command.c
```diff
@@ -2671,96 +2645,76 @@ static void d3d12_command_list_clear_attachment_pass(struct d3d12_command_list *
...
         d3d12_command_list_clear_attachment_inline(list, resource, view, 0,
                 clear_aspects, clear_value, rect_count, rects);
     }
 
-    subpass_end_info.sType = VK_STRUCTURE_TYPE_SUBPASS_END_INFO_KHR;
-    subpass_end_info.pNext = NULL;
+    VK_CALL(vkCmdEndRenderingKHR(list->vk_command_buffer));
 
-    VK_CALL(vkCmdEndRenderPass2KHR(list->vk_command_buffer, &subpass_end_info));
+    for (i = 0; i < image_barrier_count; i++)
```
HansKristian-Work, line 2656: final_layouts should always be equal to attachment layout I think, so there should be no need to do barriers here.

Technically, I suppose we would need access -> access flush for COLOR_WRITE | DEPTH_WRITE, etc, but we don't do that for normal render passes due to the massive slowdown.

libs/vkd3d/command.c
```diff
@@ -2543,8 +2521,13 @@ static void d3d12_command_list_clear_attachment_pass(struct d3d12_command_list *
 
     stencil_attachment_info = attachment_info;
 
-    memset(&rendering_info, 0, sizeof(rendering_info));    
+    memset(&rendering_info, 0, sizeof(rendering_info));
     rendering_info.sType = VK_STRUCTURE_TYPE_RENDERING_INFO_KHR;
+    rendering_info.renderArea.offset.x = 0;
```
HansKristian-Work, line 2498: This part seems like it should be folded into the previous commit.

libs/vkd3d/command.c
```diff
@@ -5979,33 +5975,31 @@ static void d3d12_command_list_copy_image(struct d3d12_command_list *list,
...
         }
 
-        extent.width = d3d12_resource_desc_get_width(&dst_resource->desc, dst_view_desc.miplevel_idx);
-        extent.height = d3d12_resource_desc_get_height(&dst_resource->desc, dst_view_desc.miplevel_idx);
-        extent.depth = dst_view_desc.layer_count;
+        memset(&attachment_info, 0, sizeof(attachment_info));
+        attachment_info.sType = VK_STRUCTURE_TYPE_RENDERING_ATTACHMENT_INFO_KHR;
+        attachment_info.imageView = dst_view->vk_image_view;
+        attachment_info.imageLayout = dst_layout;
+        attachment_info.loadOp = VK_ATTACHMENT_LOAD_OP_LOAD;
```
HansKristian-Work, line 5875: Ideally we'd use DONT_CARE and restrict renderArea to copy dest region to be friendlier to hypothetical TBDR.

libs/vkd3d/vkd3d_private.h
```diff
@@ -2471,15 +2470,13 @@ struct vkd3d_copy_image_pipeline_key
     const struct vkd3d_format *format;
     VkImageViewType view_type;
     VkSampleCountFlagBits sample_count;
-    VkImageLayout layout;
```
HansKristian-Work, line 2477: Could there be packing shenanigans here? (memcmp checks for hashmap)?

libs/vkd3d/state.c
```diff
@@ -4082,8 +4084,25 @@ VkPipeline d3d12_pipeline_state_create_pipeline_variant(struct d3d12_pipeline_st
...
+        if (rtv_active_mask & (1u << i))
+            rtv_formats[i] = graphics->rtv_formats[i];
+        else
+            rtv_formats[i] = VK_FORMAT_UNDEFINED;
+    }
+
+    rendering_info.sType = VK_STRUCTURE_TYPE_PIPELINE_RENDERING_CREATE_INFO_KHR;
+    rendering_info.pNext = NULL;
+    rendering_info.viewMask = 0;
+    rendering_info.colorAttachmentCount = D3D12_SIMULTANEOUS_RENDER_TARGET_COUNT;
```
HansKristian-Work, line 3666: Shouldn't colorAttachmentCount be restricted to findMSB(rtv_active_mask) + 1?

libs/vkd3d/state.c
```diff
@@ -4082,8 +4084,25 @@ VkPipeline d3d12_pipeline_state_create_pipeline_variant(struct d3d12_pipeline_st
     vp_desc.scissorCount = key && !key->dynamic_viewport ? max(key->viewport_count, 1) : 0;
     vp_desc.pScissors = NULL;
 
+    rtv_active_mask = key ? key->rtv_active_mask : graphics->rtv_active_mask;
```
HansKristian-Work, line 3654: We should be able to dynamically render to NULL RTV with dynamic rendering, so not sure if we even need a key for rtv_active_mask here.

libs/vkd3d/command.c
```diff
@@ -1990,9 +1990,9 @@ static inline struct d3d12_command_list *impl_from_ID3D12GraphicsCommandList(d3d
     return CONTAINING_RECORD(iface, struct d3d12_command_list, ID3D12GraphicsCommandList_iface);
 }
 
-static void d3d12_command_list_invalidate_current_framebuffer(struct d3d12_command_list *list)
+static void d3d12_command_list_invalidate_rendering_info(struct d3d12_command_list *list)
 {
-    list->current_framebuffer = VK_NULL_HANDLE;
+    list->rendering_info.state_flags &= ~VKD3D_RENDERING_CURRENT;
```
HansKristian-Work, line 1970: Does this mean we risk dropping calls to vkCmdEndRenderingKHR?

HansKristian-Work: Nah. CURRENT vs ACTIVE I suppose.

libs/vkd3d/vkd3d_private.h
```diff
@@ -1876,6 +1876,7 @@ struct vkd3d_rendering_info
     VkRenderingAttachmentInfoKHR dsv;
     VkRenderingFragmentShadingRateAttachmentInfoKHR vrs;
     uint32_t state_flags;
+    uint32_t rtv_mask;
```
HansKristian-Work, line 1841: I think we should be able to not consider RTV null fallbacks at all. Binding a NULL attachment in vkCmdBeginRenderingKHR is legal I think. The only fallback we need to keep around is the "NULL dsv + depth enable" stuff, but other than that, we should be okay just binding NULL attachments.

HansKristian-Work: > If the imageView member of any element of pColorAttachments is [VK_NULL_HANDLE](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#VK_NULL_HANDLE), writes to the corresponding location by a fragment are discarded.

HansKristian-Work: For depth-stencil, we can also bind NULL:

> Values in imageView are loaded and stored according to the values of loadOp and storeOp, within the render area for each device specified in [VkRenderingInfo](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#VkRenderingInfo). If imageView is [VK_NULL_HANDLE](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#VK_NULL_HANDLE), other members of this structure are ignored; writes to this attachment will be discarded, and no load, store, or resolve operations will be performed.

doitsujin: I'm aware of this, but it's such a complex rework that I'd like to revisit this at some other point.

libs/vkd3d/state.c
doitsujin: See below, I'd rather land dynamic_rendering first and then do the optimization.

libs/vkd3d/vkd3d_private.h
doitsujin: Not a new problem but I suppose we should memset the key first.

libs/vkd3d/state.c
doitsujin: do we have a way to find the msb?

libs/vkd3d/command.c
doitsujin: done.

libs/vkd3d/command.c
doitsujin: We shouldn't ever drop any, removing `CURRENT` just notifies that we need to restart the render pass.

libs/vkd3d/command.c
doitsujin: as in:
```
            /* Do not transition aspects that we are not supposed to clear */
            final_layouts[1] = vk_separate_stencil_layout(final_layouts[0]);
            final_layouts[0] = vk_separate_depth_layout(final_layouts[0]);

            attachment_info.imageLayout = final_layouts[0];
            stencil_attachment_info.imageLayout = final_layouts[1];
```
?

libs/vkd3d/vkd3d_private.h
doitsujin: Fixed.

libs/vkd3d/command.c
HansKristian-Work: Yes.

libs/vkd3d/state.c
doitsujin: Shouldn't make any difference, but I guess we can do that.

libs/vkd3d/vkd3d_private.h
HansKristian-Work: k, we can do this afterwards.

libs/vkd3d/command.c
doitsujin: Done.

libs/vkd3d/command.c
doitsujin: fixed, sort of. In practice we'll still be doing `UNDEFINED`->`OPTIMAL` for now.

libs/vkd3d/command.c

libs/vkd3d/command.c
```diff
@@ -2671,96 +2648,64 @@ static void d3d12_command_list_clear_attachment_pass(struct d3d12_command_list *
...
-        return;
+        VK_CALL(vkCmdPipelineBarrier(list->vk_command_buffer,
+            stages, stages, 0, 0, NULL, 0, NULL,
+            image_barrier_count, image_barriers));
     }
 
-    subpass_begin_info.sType = VK_STRUCTURE_TYPE_SUBPASS_BEGIN_INFO_KHR;
-    subpass_begin_info.pNext = NULL;
-    subpass_begin_info.contents = VK_SUBPASS_CONTENTS_INLINE;
+    rendering_info.renderArea.offset.x = 0;
```
HansKristian-Work: Looks like code dupe with renderArea set above?

libs/vkd3d/command.c
```diff
@@ -2691,12 +2669,6 @@ static void d3d12_command_list_clear_attachment_pass(struct d3d12_command_list *
             image_barrier_count, image_barriers));
     }
 
-    rendering_info.renderArea.offset.x = 0;
```
HansKristian-Work: Ah ... ye. splitting commits in rebases is not fun :(

libs/vkd3d/command.c
```diff
@@ -5965,33 +5961,31 @@ static void d3d12_command_list_copy_image(struct d3d12_command_list *list,
...
+        attachment_info.imageView = dst_view->vk_image_view;
+        attachment_info.imageLayout = dst_layout;
+        attachment_info.loadOp = VK_ATTACHMENT_LOAD_OP_DONT_CARE;
+        attachment_info.storeOp = VK_ATTACHMENT_STORE_OP_STORE;
+
+        memset(&rendering_info, 0, sizeof(rendering_info));
+        rendering_info.sType = VK_STRUCTURE_TYPE_RENDERING_INFO_KHR;
+        rendering_info.renderArea.offset.x = region->dstOffset.x;
+        rendering_info.renderArea.offset.y = region->dstOffset.y;
+        rendering_info.renderArea.extent.width = region->extent.width;
```
HansKristian-Work, line 5868: Do we know if this is clamped?

libs/vkd3d/swapchain.c
```diff
@@ -1052,23 +1035,56 @@ static VkResult d3d12_swapchain_record_swapchain_blit(struct d3d12_swapchain *sw
...
+    image_barrier.dstQueueFamilyIndex = VK_QUEUE_FAMILY_IGNORED;
+    image_barrier.image = swapchain->vk_swapchain_images[dst_index];
+    image_barrier.subresourceRange.aspectMask = VK_IMAGE_ASPECT_COLOR_BIT;
+    image_barrier.subresourceRange.baseMipLevel = 0;
+    image_barrier.subresourceRange.levelCount = 1;
+    image_barrier.subresourceRange.baseArrayLayer = 0;
+    image_barrier.subresourceRange.layerCount = 1;
+
+    if (attachment_info.loadOp != VK_ATTACHMENT_LOAD_OP_DONT_CARE)
+        image_barrier.dstAccessMask = VK_ACCESS_COLOR_ATTACHMENT_READ_BIT;
```
HansKristian-Work, line 1054: Shouldn't this if loadOp == LOAD_OP_LOAD? Also, |= instead of = probably?

libs/vkd3d/command.c
```diff
@@ -3442,21 +3442,19 @@ static void d3d12_command_list_end_current_render_pass(struct d3d12_command_list
...
-        subpass_end_info.pNext = NULL;
+        list->rendering_info.state_flags &= ~VKD3D_RENDERING_ACTIVE;
 
-        VK_CALL(vkCmdEndRenderPass2KHR(list->vk_command_buffer, &subpass_end_info));
+        VK_CALL(vkCmdEndRenderingKHR(list->vk_command_buffer));
     }
 
     /* Don't emit barriers for temporary suspension of the render pass */
-    if (!suspend && (list->current_render_pass || list->render_pass_suspended))
+    if (!suspend && (list->rendering_info.state_flags & (VKD3D_RENDERING_ACTIVE | VKD3D_RENDERING_SUSPENDED)))
```
HansKristian-Work, line 3425: I'm a little bit confused about this check. VKD3D_RENDERING_ACTIVE is always cleared above, and we still check for it here. In the old path, list->current_render_pass was not set to NULL_HANDLE.

libs/vkd3d/command.c
```diff
@@ -3442,21 +3442,19 @@ static void d3d12_command_list_end_current_render_pass(struct d3d12_command_list
...
         d3d12_command_list_emit_render_pass_transition(list, VKD3D_RENDER_PASS_TRANSITION_MODE_END);
 
-    list->render_pass_suspended = suspend && (list->current_render_pass || list->render_pass_suspended);
-    list->current_render_pass = VK_NULL_HANDLE;
+    if (suspend && (list->rendering_info.state_flags & (VKD3D_RENDERING_ACTIVE)))
```
HansKristian-Work, line 3428: This branch should never hit I think?

HansKristian-Work: Maybe it's supposed to check CURRENT flag?

libs/vkd3d/vkd3d_private.h
```diff
@@ -1876,6 +1876,22 @@ struct vkd3d_query_range
     uint32_t flags;
 };
 
+enum vkd3d_rendering_flags
+{
+    VKD3D_RENDERING_ACTIVE    = (1u << 0),
```
HansKristian-Work, line 1829: I'm struggling a bit reasoning about these flags. From what I understand from earlier:
- ACTIVE: We're inside CmdBeginRenderingKHR.
- SUSPENDED: We had to do non-rendering, so we called EndRenderingKHR. However, we did not perform any kind of "end pass barriers".
- CURRENT: Not sure what this represents anymore.

libs/vkd3d/state.c
```diff
@@ -2933,9 +2893,6 @@ static HRESULT d3d12_graphics_pipeline_state_create_render_pass_for_plane_mask(
         key.vk_formats[key.attachment_count++] = dsv_format;
     }
 
-    if (variant_flags & VKD3D_GRAPHICS_PIPELINE_STATIC_VARIANT_VRS_ATTACHMENT)
```
HansKristian-Work, line 2936: I think we still need to cover for:
> If the current render pass instance was begun with [vkCmdBeginRenderingKHR](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#vkCmdBeginRenderingKHR) and [VkRenderingFragmentShadingRateAttachmentInfoKHR](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#VkRenderingFragmentShadingRateAttachmentInfoKHR)::imageView was not [VK_NULL_HANDLE](https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/vkspec.html#VK_NULL_HANDLE), the currently bound graphics pipeline must have been created with VK_PIPELINE_CREATE_RENDERING_FRAGMENT_SHADING_RATE_ATTACHMENT_BIT_KHR

If we set this unconditionally if VRS is supported, we can still use null VRS attachment.

libs/vkd3d/command.c
HansKristian-Work: Sure. We're kinda force to do UNDEFINED -> OPTIMAL in many clear scenarios due to dumb alias rules + initial resource layout shenanigans. I guess there's a theoretical improvement for later if we deduce:
- Initial layout transition has already happened
- Committed resource

libs/vkd3d/state.c
HansKristian-Work: Alternatively, NumRenderTargets could be used, might be more instructive.

libs/vkd3d/command.c
doitsujin: I don't know how to fix this at this point.

libs/vkd3d/command.c
doitsujin: It's not if the app is beyond broken.

libs/vkd3d/swapchain.c
```diff
@@ -1052,23 +1035,56 @@ static VkResult d3d12_swapchain_record_swapchain_blit(struct d3d12_swapchain *sw
...
+
+    VK_CALL(vkCmdPipelineBarrier(vk_cmd_buffer,
+            VK_PIPELINE_STAGE_TOP_OF_PIPE_BIT,
```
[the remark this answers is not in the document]
doitsujin: why is the src stage important?

libs/vkd3d/command.c
doitsujin: Yeah the reset should happen later.

libs/vkd3d/swapchain.c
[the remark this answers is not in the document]
doitsujin: fixed

libs/vkd3d/swapchain.c
doitsujin: There was a reason for the DONT_CARE check but I can't quite remember what it was.

missing |= is a typo.

libs/vkd3d/command.c
doitsujin: *should* be gone now.

libs/vkd3d/command.c
doitsujin: fixed.

libs/vkd3d/command.c
doitsujin: See above.

libs/vkd3d/state.c
doitsujin: fixed, wasn't aware that this exists.

libs/vkd3d/vkd3d_private.h
doitsujin: `CURRENT` is a dirty flag for the rendering info.

libs/vkd3d/state.c
doitsujin: fixed.

libs/vkd3d/command.c
```diff
@@ -9882,6 +9869,43 @@ static struct d3d12_command_list *unsafe_impl_from_ID3D12CommandList(ID3D12Comma
...
+    attachment_info->sType = VK_STRUCTURE_TYPE_RENDERING_ATTACHMENT_INFO_KHR;
+    attachment_info->loadOp = VK_ATTACHMENT_LOAD_OP_LOAD;
+    attachment_info->storeOp = VK_ATTACHMENT_STORE_OP_STORE;
+}
+
+static void d3d12_command_list_init_rendering_info(struct d3d12_device *device, struct vkd3d_rendering_info *rendering_info)
+{
+    unsigned int i;
+
+    rendering_info->info.sType = VK_STRUCTURE_TYPE_RENDERING_INFO_KHR;
```
HansKristian-Work, line 9787: I don't see state flags being cleared out here? Is that taken care of with command list creation memset?

libs/vkd3d/command.c
```diff
@@ -4426,29 +4415,35 @@ static bool d3d12_command_list_create_framebuffer(struct d3d12_command_list *lis
...
-    view_count = 0;
 
     /* The pipeline has fallback render passes / PSO in case we're
      * attempting to render to unbound RTV. */
-    while (rtv_mask)
+    for (i = 0; i < D3D12_SIMULTANEOUS_RENDER_TARGET_COUNT; i++)
     {
-        i = vkd3d_bitmask_iter32(&rtv_mask);
-        views[view_count++] = list->rtvs[i].view->vk_image_view;
+        VkRenderingAttachmentInfoKHR *attachment = &rendering_info->rtv[i];
```
HansKristian-Work, line 4373: colorAttachmentCount should be updated here to reflect the colorAttachment count used in PSO I think.

libs/vkd3d/command.c
```diff
@@ -9745,12 +9732,12 @@ static void STDMETHODCALLTYPE d3d12_command_list_RSSetShadingRateImage(d3d12_com
...
 
-    d3d12_command_list_invalidate_current_framebuffer(list);
+    d3d12_command_list_invalidate_rendering_info(list);
 
     /* Need to end the renderpass if we have one to make
      * way for the new VRS attachment */
-    if (list->current_render_pass || list->render_pass_suspended)
-        d3d12_command_list_invalidate_current_render_pass(list);
+    if (list->rendering_info.state_flags & (VKD3D_RENDERING_ACTIVE | VKD3D_RENDERING_SUSPENDED))
+        d3d12_command_list_end_current_render_pass(list, false);
```
HansKristian-Work, line 9644: Below, the pipeline is invalidated on `if (!list->vrs_image != !vrs_image)`. This is now obsolete.

libs/vkd3d/command.c
doitsujin: Yes, we're assuming that everything is zero-initialized.

libs/vkd3d/command.c
doitsujin: fixed, and changed the way we compute the pso rt count.

libs/vkd3d/command.c
doitsujin: fixed.

hurricane-dorian:
Since branch update 8a7bfd6 all games just black-screen then crash.
Mesa-git on rx5700 amdgpu 5.16.9

doitsujin:
Yes, this is a known issue with the current rebase.

HansKristian-Work:
Found one bug for reference, caught by https://github.com/HansKristian-Work/vkd3d-proton/pull/1027:

```
    if (d3d12_command_list_has_depth_stencil_view(list))
    {
        /* Select new dsv_layout. Any new PSO write we didn't observe yet must be updated here. */
        dsv_plane_optimal_mask = list->dsv_plane_optimal_mask | list->state->graphics.dsv_plane_optimal_mask;
        dsv_layout = dsv_plane_optimal_mask_to_layout(list->dsv_plane_optimal_mask, // Should be dsv_plane_optimal_mask
                list->dsv.format->vk_aspect_mask);
    }
    else
    {
        dsv_plane_optimal_mask = 0;
        dsv_layout = VK_IMAGE_LAYOUT_UNDEFINED;
    }
```

PR #912, opened 2021-11-18
Bump interfaces to ID3D12Device9 and ID3D12GraphicsCommandList6

doitsujin:
Lots of boilerplate here since D3D12_RESOURCE_DESC1 was introduced for sampler feedback.

### DIFF
--- a/include/vkd3d_d3d12.idl
+++ b/include/vkd3d_d3d12.idl
@@ -2158,6 +2158,15 @@ interface ID3D12Resource1 : ID3D12Resource
 {
     HRESULT GetProtectedResourceSession(REFIID riid, void **protected_session);
 }
+[
+    uuid(be36ec3b-ea85-4aeb-a45a-e9d76404a495),
+    object,
+    local,
+    pointer_default(unique)
+]
+interface ID3D12Resource2 : ID3D12Resource1 {
+    D3D12_RESOURCE_DESC1 GetDesc1();
+}
 
 [
     uuid(7116d91c-e7e4-47ce-b8c6-ec8168f437e5),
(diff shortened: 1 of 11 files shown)


include/vkd3d_d3d12.idl
```diff
@@ -3313,6 +3322,44 @@ typedef struct D3D12_DISPATCH_MESH_ARGUMENTS
     UINT ThreadGroupCountZ;
 } D3D12_DISPATCH_MESH_ARGUMENTS;
 
+typedef enum D3D12_SHADER_CACHE_MODE {
```
HansKristian-Work, line 3325: Brace

libs/vkd3d/resource.c
```diff
@@ -1543,17 +1544,26 @@ static void STDMETHODCALLTYPE d3d12_resource_Unmap(d3d12_resource_iface *iface,
 static D3D12_RESOURCE_DESC * STDMETHODCALLTYPE d3d12_resource_GetDesc(d3d12_resource_iface *iface,
         D3D12_RESOURCE_DESC *resource_desc)
 {
-    struct d3d12_resource *resource = impl_from_ID3D12Resource1(iface);
+    struct d3d12_resource *resource = impl_from_ID3D12Resource2(iface);
 
     TRACE("iface %p, resource_desc %p.\n", iface, resource_desc);
 
-    *resource_desc = resource->desc;
+    resource_desc->Dimension = resource->desc.Dimension;
```
HansKristian-Work, line 1551: Not related to the patch, but I noticed that this function signature looks completely different from https://docs.microsoft.com/en-us/windows/win32/api/d3d12/nf-d3d12-id3d12resource-getdesc. Shouldn't we return the struct by value?

libs/vkd3d/device.c
```diff
@@ -4849,7 +4725,242 @@ static HRESULT STDMETHODCALLTYPE d3d12_device_CreateProtectedResourceSession1(d3
     return E_NOTIMPL;
 }
 
-CONST_VTBL struct ID3D12Device7Vtbl d3d12_device_vtbl =
+static D3D12_RESOURCE_ALLOCATION_INFO* STDMETHODCALLTYPE d3d12_device_GetResourceAllocationInfo2(d3d12_device_iface *iface,
```
HansKristian-Work, line 4729: Odd that we're returning a pointer. Is this some COM ABI shenanigans?

HansKristian-Work (commented):
Some nits/questions. CI build has a bunch of warnings as well with pointer types that were missed.

libs/vkd3d/resource.c
doitsujin: This signature is correct, any function that returns a struct is turned into a function that takes a pointer to an out-struct and returns the same pointer. Don't ask me why, but this is done for all COM functions of this kind.

libs/vkd3d/device.c
doitsujin: See above.

include/vkd3d_d3d12.idl
doitsujin: Fixed.

libs/vkd3d/resource.c
HansKristian-Work: OK.

HansKristian-Work:
Still a bunch of warnings: https://github.com/HansKristian-Work/vkd3d-proton/actions/runs/1481011923

HansKristian-Work (approved):
LGTM, thanks.

PR #924, opened 2021-11-26
Fix format compatibility stuff

doitsujin:
Allows the `MUTABLE | EXTENDED` bits to be set with non-typeless formats, and always uses a (hopefully exact) format list even for UAVs.


tests/d3d12_clear.c
```diff
@@ -716,6 +716,9 @@ void test_clear_unordered_access_view_image(void)
         {DXGI_FORMAT_R11G11B10_FLOAT, 1, 1, 0, 0, 1, 0, {{0}},
                 {0x3f000000 /* 1.0f */, 0 /* 0.0f */, 0xbf800000 /* -1.0f */, 0x3f000000 /* 1.0f */},
                 0x00000380, true},
+        {DXGI_FORMAT_B8G8R8A8_UNORM,  1, 1, 0, 0, 1, 0, {{0}},
+                {0, 0, 0x3f000080 /* 0.5f + epsilon */, 0x3f800000 /* 1.0f */}, 0xff000080, true},
+        {DXGI_FORMAT_B8G8R8A8_UNORM,  1, 1, 0, 0, 1, 0, {{0}}, {1, 2, 3, 4}, 0x04010203},
```
DadSchoorse, line 721: Doesn't look like the code currently checks if uavs are supported for the format, which is an issues since bgra8 uavs are optional.

doitsujin: Is there a single driver that doesn't support them?

DadSchoorse: Old anv, intel on windows, turnip I think.

libs/vkd3d/utils.c
```diff
@@ -153,6 +153,239 @@ static const struct vkd3d_format vkd3d_depth_stencil_formats[] =
 #undef SINT
 #undef UINT
 
+static const struct vkd3d_format_compatibility_info_2
+{
+    DXGI_FORMAT image_format;
+    DXGI_FORMAT view_formats[VKD3D_MAX_COMPATIBLE_FORMAT_COUNT];
+    DXGI_FORMAT uint_format;
```
HansKristian-Work, line 160: Reading later commits, I understood what this was used for, but would be good to add a comment saying it's to handle ClearUAVUint.

libs/vkd3d/utils.c
```diff
@@ -153,6 +153,239 @@ static const struct vkd3d_format vkd3d_depth_stencil_formats[] =
...
+    {DXGI_FORMAT_BC3_TYPELESS,
+            {DXGI_FORMAT_BC3_UNORM, DXGI_FORMAT_BC3_UNORM_SRGB}},
+    {DXGI_FORMAT_BC3_UNORM,
+            {DXGI_FORMAT_BC3_UNORM_SRGB}},
+    {DXGI_FORMAT_BC3_UNORM_SRGB,
+            {DXGI_FORMAT_BC3_UNORM}},
+
+    {DXGI_FORMAT_BC4_TYPELESS,
+            {DXGI_FORMAT_BC4_UNORM, DXGI_FORMAT_BC4_SNORM}},
+    {DXGI_FORMAT_BC5_TYPELESS,
```
HansKristian-Work, line 376: Missed BC4_UNORM, BC4_SNORM? Same with BC5.

libs/vkd3d/resource.c
```diff
@@ -209,15 +209,40 @@ static unsigned int max_miplevel_count(const D3D12_RESOURCE_DESC1 *desc)
...
+    if (desc->Flags & D3D12_RESOURCE_FLAG_ALLOW_UNORDERED_ACCESS)
+    {
+        const struct vkd3d_format *uint_format = vkd3d_get_format(device, list.uint_format, false);
+
+        /* Format used for ClearUnorderedAccessViewUint */
+        if (uint_format)
+            vkd3d_format_compatibility_list_add_format(&list, uint_format->vk_format);
+
+        /* Legacy D3D11 compatibility rule that allows typed UAV loads on FL11.0 hardware */
+        if (format->type == VKD3D_FORMAT_TYPE_TYPELESS && format->byte_count == 4)
```
HansKristian-Work, line 234: I'm not sure this is correct. In the test I observed that it was possible to cast RGBA8_UNORM to R32UI UAV, but not R32_FLOAT. Is the D3D11 legacy rule only for TYPELESS? Might be a bug in the validation layer fwiw.

tests/d3d12_clear.c
HansKristian-Work: I guess we should just add a format feature query before running the test.

HansKristian-Work (changes requested):
Looks good overall, just some nits I think.

libs/vkd3d/utils.c
doitsujin: Is this allowed under the no SNORM <-> UNORM rule?

HansKristian-Work: Oh, I see. It's not I think.

libs/vkd3d/resource.c
doitsujin: I'm not sure what the exact rules are here. Although I suppose it makes sense to allow non-typeless since there's no R11G11B10 typeless format but you can still use it for UAVs.

libs/vkd3d/utils.c
HansKristian-Work: I figure if a format is not present in this list, it's considered non-mutable?

libs/vkd3d/resource.c
HansKristian-Work: I'll add some tests for that case.

tests/d3d12_clear.c
doitsujin: done.

libs/vkd3d/utils.c
doitsujin: Yes, however we might still make images mutable regardless if UAV usage is enabled (under the uint/R32UI rule).

HansKristian-Work (approved):
LGTM, thanks!

PR #932, opened 2021-11-29
vkd3d: Remove misleading FIXME.

HansKristian-Work:
We can bind texel buffers at scalar alignment now.
The warning is misleading for placed resources, since 64k never aligns
with a float3.

### DIFF
--- a/libs/vkd3d/resource.c
+++ b/libs/vkd3d/resource.c
@@ -3153,33 +3153,6 @@ void d3d12_desc_copy(struct d3d12_desc *dst, struct d3d12_desc *src,
     }
 }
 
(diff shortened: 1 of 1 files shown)


libs/vkd3d/resource.c
```diff
@@ -3218,8 +3218,6 @@ static bool vkd3d_create_vk_buffer_view(struct d3d12_device *device,
     }
 
     alignment = vkd3d_get_required_texel_buffer_alignment(device, format);
```
DadSchoorse, line 3220: This is dead code now, and it's the only user of that function.

doitsujin (approved):

PR #934, opened 2021-11-30
vkd3d: Retry to create Vk device without NVX extensions.

ivyl:
The creation with those extensions fails in a few cases:
 * older 32 bit drivers
 * missing or inaccessible /dev/nvidia-uvm

I believe that DXVK has a workaround for the first problem.

There's also a mysterious crash that some Debian users experience with 64bit titles and a correct /dev/nvidia-uvm, see https://github.com/ValveSoftware/Proton/issues/4450#issuecomment-982356716 and preceding comments. I believe we've checked everything else that could possibly cause the issue and the build with the NVX extensions disabled worked for the user. The bad  performance observed by @fezie was due to force enabled trace logging in the build.

### DIFF
--- a/libs/vkd3d/device.c
+++ b/libs/vkd3d/device.c
@@ -273,6 +273,55 @@ static unsigned int vkd3d_enable_extensions(const char *extensions[],
     return extension_count;
 }
 
+static bool vkd3d_remove_extension(const char *to_remove, const char *extensions[], uint32_t *extension_count)
+{
+    uint32_t i;
+
+    for (i = 0; i < *extension_count; i++)
+    {
+        if (strcmp(to_remove, extensions[i]) != 0)
+            continue;
+
+        WARN("Removing %s extension from the array.\n", to_remove);
+
+        if (i < (*extension_count - 1))
+            extensions[i] = extensions[*extension_count - 1];
+
+        (*extension_count)--;
+
+        return true;
+    }
+
+    return false;
+}
+
+static bool vkd3d_disable_nvx_extensions(struct d3d12_device *device, const char *extensions[], uint32_t *enabled_extension_count)
+{
+    struct vkd3d_vulkan_info *vk_info = &device->vk_info;
+    bool disabled = false;
+    unsigned int i;
+
+    static const struct vkd3d_optional_extension_info nvx_extensions[] =
+    {
+        VK_EXTENSION(NVX_BINARY_IMPORT, NVX_binary_import),
+        VK_EXTENSION(NVX_IMAGE_VIEW_HANDLE, NVX_image_view_handle),
+    };
+
+    for (i = 0; i < ARRAY_SIZE(nvx_extensions); ++i)
+    {
+        const char *extension_name = nvx_extensions[i].extension_name;
+        bool *supported = (void *)((uintptr_t)vk_info + nvx_extensions[i].vulkan_info_offset);
+
+        if (vkd3d_remove_extension(extension_name, extensions, enabled_extension_count))
+        {
+            *supported = false;
+            disabled = true;
+        }
+    }
+
+    return disabled;
+}
+
 static HRESULT vkd3d_init_instance_caps(struct vkd3d_instance *instance,
         const struct vkd3d_instance_create_info *create_info,
         uint32_t *instance_extension_count, bool *user_extension_supported)
@@ -2230,6 +2279,12 @@ static HRESULT vkd3d_create_vk_device(struct d3d12_device *device,
     vkd3d_free(user_extension_supported);
 
     vr = VK_CALL(vkCreateDevice(physical_device, &device_info, NULL, &vk_device));
+    if (vr == VK_ERROR_INITIALIZATION_FAILED &&
+            vkd3d_disable_nvx_extensions(device, extensions, &device_info.enabledExtensionCount))
+    {
+        WARN("Disabled extensions that can cause Vulkan device creation to fail, retrying.\n");
+        vr = VK_CALL(vkCreateDevice(physical_device, &device_info, NULL, &vk_device));
+    }
     vkd3d_free((void *)extensions);
     if (vr < 0)
     {


libs/vkd3d/device.c
```diff
@@ -273,6 +273,49 @@ static unsigned int vkd3d_enable_extensions(const char *extensions[],
...
+
+    return found;
+}
+
+static bool vkd3d_disable_nvx_extensions(struct d3d12_device *device, const char *extensions[], uint32_t *enabled_extension_count)
+{
+    struct vkd3d_vulkan_info *vk_info = &device->vk_info;
+    bool disabled = false;
+
+    if (vkd3d_remove_extension(VK_NVX_BINARY_IMPORT_EXTENSION_NAME, extensions, enabled_extension_count))
```
HansKristian-Work, line 304: I would expect ray tracing to complicate things as well, since that requires the UVM module.

libs/vkd3d/device.c
```diff
@@ -273,6 +273,49 @@ static unsigned int vkd3d_enable_extensions(const char *extensions[],
...
+    bool found = false;
+    uint32_t i;
+
+    for (i = 0; i < *extension_count; i++)
+    {
+        if (!found && strcmp(to_remove, extensions[i]) == 0)
+            found = true;
+
+        if (found && i < (*extension_count - 1))
+            extensions[i] = extensions[i+1];
```
HansKristian-Work, line 287: `i + 1` formatting.

HansKristian-Work: This seems a bit inefficient, I'd rather do `if (found) { extensions[i] = extensions[last_index]; extension_count--; }` or something. Ordering is not important.

libs/vkd3d/device.c
HansKristian-Work: Would make more sense here to have an array of "problematic extensions", remove them one by one, and attempt to create a device until it works.

ivyl: Those are hidden behind `VKD3D_CONFIG` option so I thought it was okay to leave them alone.

I can add handling for them ones that require uvm (tested with `while true; do rm -f /dev/nvidia-uvm; done` on RTX 3060):

```c
static bool vkd3d_disable_problematic_extensions(struct d3d12_device *device, const char *extensions[], uint32_t *enabled_extension_count)
{
    struct vkd3d_vulkan_info *vk_info = &device->vk_info;
    bool disabled = false;
    unsigned int i;

    static const struct vkd3d_optional_extension_info problematic_extensions[] =
    {
        VK_EXTENSION(KHR_RAY_TRACING_PIPELINE, KHR_ray_tracing_pipeline),
        VK_EXTENSION(KHR_ACCELERATION_STRUCTURE, KHR_acceleration_structure),
        VK_EXTENSION(KHR_RAY_QUERY, KHR_ray_query),
        VK_EXTENSION(NVX_BINARY_IMPORT, NVX_binary_import),
        VK_EXTENSION(NVX_IMAGE_VIEW_HANDLE, NVX_image_view_handle),
    };

    for (i = 0; i < ARRAY_SIZE(problematic_extensions); ++i)
    {
        const char *extension_name = problematic_extensions[i].extension_name;
        ptrdiff_t offset = problematic_extensions[i].vulkan_info_offset;
        bool *supported = (void *)((uintptr_t)vk_info + offset);

        if (vkd3d_remove_extension(extension_name, extensions, enabled_extension_count))
        {
            *supported = false;
            disabled = true;
        }
    }

    return disabled;
}
```

But we would need to also add something like:

```diff
diff --git a/libs/vkd3d/device.c b/libs/vkd3d/device.c
index a62b6fb7..44e0bd29 100644
--- a/libs/vkd3d/device.c
+++ b/libs/vkd3d/device.c
@@ -5198,6 +5208,7 @@ static bool d3d12_device_supports_rtas_formats(struct d3d12_device *device, cons

 static D3D12_RAYTRACING_TIER d3d12_device_determine_ray_tracing_tier(struct d3d12_device *device)
 {
+    const struct vkd3d_vulkan_info *vk_info = &device->vk_info;
     const struct vkd3d_physical_device_info *info = &device->device_info;
     D3D12_RAYTRACING_TIER tier = D3D12_RAYTRACING_TIER_NOT_SUPPORTED;
     bool supports_vbo_formats;
@@ -5226,7 +5237,9 @@ static D3D12_RAYTRACING_TIER d3d12_device_determine_ray_tracing_tier(struct d3d1
         VK_FORMAT_R8G8_SNORM,
     };

-    if (info->ray_tracing_pipeline_features.rayTracingPipeline &&
+    if (vk_info->KHR_acceleration_structure &&
+        vk_info->KHR_ray_tracing_pipeline &&
+        info->ray_tracing_pipeline_features.rayTracingPipeline &&
         info->acceleration_structure_features.accelerationStructure &&
         info->ray_tracing_pipeline_properties.maxRayHitAttributeSize >= D3D12_RAYTRACING_MAX_ATTRIBUTE_SIZE_IN_BYTES &&
         /* Group handle size must match exactly or ShaderRecord layout will not match.
@@ -5253,7 +5266,8 @@ static D3D12_RAYTRACING_TIER d3d12_device_determine_ray_tracing_tier(struct d3d1
         }
     }

-    if (tier == D3D12_RAYTRACING_TIER_1_0 && info->ray_query_features.rayQuery &&
+    if (tier == D3D12_RAYTRACING_TIER_1_0 && vk_info->KHR_ray_query &&
+            info->ray_query_features.rayQuery &&
             info->ray_tracing_pipeline_features.rayTraversalPrimitiveCulling &&
        
```

or overwrite the values in `info`.

libs/vkd3d/device.c
ivyl: Ah, yes. Since the order does not matter something like this should do:

```c
static bool vkd3d_remove_extension(const char *to_remove, const char *extensions[], uint32_t *extension_count)
{
    uint32_t i;

    for (i = 0; i < *extension_count; i++)
    {
        if (strcmp(to_remove, extensions[i]) != 0)
            continue;

        WARN("Removing %s extension from the array.\n", to_remove);

        if (i < (*extension_count - 1))
            extensions[i] = extensions[*extension_count - 1];

        (*extension_count)--;

        return true;
    }

    return false;
}
```

libs/vkd3d/device.c
HansKristian-Work: Ugh ... right, we'd have to disable the feature bits as well, otherwise vkCreateDevice must fail since we're trying to enable unsupported features without extension. True, it's hidden behind a VKD3D_CONFIG, so I guess it's fine to do with just the two NVX extensions.

HansKristian-Work (approved):
LGTM. Annoying that we have to do this, but should hopefully avoid lots of the bug reports.

PR #936, opened 2021-12-02
vkd3d: Attempt to create linear image without EXTENDED_USAGE.

HansKristian-Work:
NVIDIA drivers apparently cannot support EXTENDED_USAGE linear
images for whatever reason, so attempt to create these images without
the creation flag.

Fixes crash in DEATHLOOP on NV.


doitsujin:
Do we know which bind flags the image in question has, what the format is, and which of these flags is causing issues?

HansKristian-Work:
> Do we know which bind flags the image in question has, what the format is, and which of these flags is causing issues?

It's just plain SAMPLED + TRANSFER.

doitsujin (approved):

Issue #945, opened 2021-12-17
Aliasing Barrier problem in Horizon Zero Dawn


K0bin:
Horizon Zero Dawn recently got an update doing the following:

Create two textures that share the same memory.
- Use texture 1 in a render pass.
- Issue a ResourceBarrier call that transitions texture 1 from `D3D12_RESOURCE_RENDERTARGET` to `D3D12_RESOURCE_STATE_PIXEL_SHADER_RESOURCE` **and** an aliasing barrier from another to texture 2.
- Use contents of texture 1 in another render pass.

What ends up happening is that VKD3D-Proton does a `VK_IMAGE_LAYOUT_UNDEFINED` -> whatever common layout is barrier but the game relies on the contents of texture 1 from that earlier render pass.

I'm not sure what the correct fix is. Maybe treat aliasing barriers just as a wait & flush and rely on the game calling `DiscardResource`, `CopyResource` or any of the clear functions to re-initialize it with a proper VkImageLayout? Doing this fixes the black screen I was getting with FSR enabled.

It does **not** fix the broken TAA on vegetation. That's an unrelated issue. (Which seems to come down to fp issues with a depth prepass.)

PR #946, opened 2021-12-18
vkd3d: Don't do layout transition in aliasing barrier.

K0bin:
See #945

I'm not sure if this is correct so I'm looking forward to feedback.

Fixes black screen in HZD when enabling FSR on Nvidia Pascal. Might be broken in different ways on other HW.


K0bin:
Superseeded by 956.

PR #961, opened 2022-01-11
Wipe memory on the GPU

ishitatsuyuki:
This alleviates frame time oscillation in Ghostrunner main menu.

I don't know about the performance impact, but it doesn't seem to be too bad (comparing to Ghostrunner DX11).

### DIFF
--- a/libs/vkd3d/memory.c
+++ b/libs/vkd3d/memory.c
@@ -1038,6 +1038,7 @@ static HRESULT vkd3d_memory_allocator_flush_clears_locked(struct vkd3d_memory_al
     uint32_t queue_mask, queue_index;
     VkCommandBuffer vk_cmd_buffer;
     VkSubmitInfo submit_info;
+    VkBufferMemoryBarrier *buffer_memory_barriers;
     VkQueue vk_queue;
     VkResult vr;
     size_t i;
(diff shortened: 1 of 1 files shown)


ishitatsuyuki:
This is now unwip.

libs/vkd3d/memory.c
```diff
@@ -1038,6 +1038,7 @@ static HRESULT vkd3d_memory_allocator_flush_clears_locked(struct vkd3d_memory_al
     uint32_t queue_mask, queue_index;
     VkCommandBuffer vk_cmd_buffer;
     VkSubmitInfo submit_info;
+    VkBufferMemoryBarrier *buffer_memory_barriers;
```
doitsujin, line 1041: Don't use `VkBufferMemoryBarrier`, just use a single `VkMemoryBarrier`. Avoids both the memory allocation and the loop, there is no good reason to ever use `VkBufferMemoryBarrier` other than resource ownership transfers.

doitsujin (commented):
What kind of system was this observed on? I have a hard time believing that clearing system memory on the GPU and then stalling the CPU for it is faster just than doing a memset.

Only exception would be if the allocation in question is on `HOST_VISIBLE | DEVICE_LOCAL` memory which would be worth verifying.

ishitatsuyuki:
3700X, 5700XT, Arch Linux, own mesa-git build. It's really weird that it's taking so much of time to clear a 8MB buffer, but I'll need to debug more to see what's going on.

libs/vkd3d/memory.c
```diff
@@ -1159,19 +1178,41 @@ HRESULT vkd3d_memory_allocator_flush_clears(struct vkd3d_memory_allocator *alloc
...
+        else
+        {
+            /*
+             * Do it on the GPU. No need to flush memory ranges since this is a fresh allocation. From Vulkan
+             * specification 11.2.12:
+             * Unmapping non-coherent memory does not implicitly flush the host mapped memory, and host writes that have
+             * not been flushed may not ever be visible to the device. However, implementations must ensure that writes
+             * that have not been flushed do not become visible to any other memory.
+             */
+            clear_with_gpu = true;
```
HansKristian-Work, line 1206: This is quite questionable. Immediately stalling on the CPU is very bad. Why would potentially stalling for 10ms+ be preferable to a memset? How much data are we talking about here? How often? This is also clearing host memory, which means the clear is going to be PCI-e bound, which is far slower than DDR4 bandwidth.

HansKristian-Work (changes requested):
I won't accept this without an explanation why this is the better solution.

libs/vkd3d/memory.c
doitsujin: This could be HVV to be fair but even then it shouldn't be so slow, unless something is wrong with the memset implementation.

HansKristian-Work:
> Only exception would be if the allocation in question is on HOST_VISIBLE | DEVICE_LOCAL memory which would be worth verifying.

That might be something. If memset hits pathological behavior triggering hopelessly slow memset perf, but in that case we should make our own write-combined memory aware memset.

libs/vkd3d/memory.c
HansKristian-Work: Right, HVV could be an explanation ... Would be surprised if memset is that slow though. memset implementations like to use insane tricks for perf.

ishitatsuyuki:
It turns out that writing to BAR was slower than expected, and GPU clears is neither a good solution to this nor good for the general case. Closing. A fix for BAR writes should arrive later.

PR #975, opened 2022-01-24
SPIR-V caching in pipeline libraries - Phase 1

HansKristian-Work:
This PR focuses on slight refactors which will be required for saving any SPIR-V in cache.

- First, move over to RW locks for pipeline libraries. It was supposed to be a big optimization, but then I realized that the lock was unlocked before create_pipeline_state(), so it's not as useful as initially thought. Still nice to have though.
- Then, add a hash which should be used for storing SPIR-V and deduplicated blobs. It's the same one as used for DXBC shader blobs, just make it more reusable.
- Finally, add concept of shader interface key. Given equivalent DXBC + root signature + device UUID + shader interface key, we should guarantee a 1:1 match with SPIR-V.
- Store the shader interface key in pipeline library as well and validate it.

### DIFF
--- a/include/vkd3d_shader.h
+++ b/include/vkd3d_shader.h
@@ -54,16 +54,20 @@ enum vkd3d_shader_visibility
 
 typedef uint64_t vkd3d_shader_hash_t;
 
+enum vkd3d_shader_meta_flags
+{
+    VKD3D_SHADER_META_FLAG_REPLACED = 1 << 0,
+    VKD3D_SHADER_META_FLAG_USES_SUBGROUP_SIZE = 1 << 1,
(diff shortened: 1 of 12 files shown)


include/vkd3d_shader.h
```diff
@@ -63,7 +63,9 @@ struct vkd3d_shader_meta
     bool replaced;
     bool uses_subgroup_size;
     bool uses_native_16bit_operations;
+    bool padding;
```
doitsujin, line 66: could we turn the bools into a uint32 bitfield? Makes it easier to add stuff later if needed.

HansKristian-Work: I can fix that, yes.

HansKristian-Work: Fixed.

doitsujin (approved):

PR #977, opened 2022-01-27
Add tests for mesh shaders

doitsujin:
Doesn't test pipeline statistics queries with mesh shaders yet as that doesn't seem very interesting right now.

All tests pass on Nvidia, one is broken on AMD.


tests/d3d12_mesh_shader.c
```diff
@@ -0,0 +1,2326 @@
...
+    struct OutputVertex
+    {
+        float4 pos : SV_POSITION;
+    };
+
+    [numthreads(3,1,1)]
+    [outputtopology("triangle")]
+    void main(in uint tid : SV_GroupIndex, out vertices OutputVertex v[3]) {
+        SetMeshOutputCounts(3, 1);
+        v[tid].pos = float4(0.0f, 0.0f, 0.0f, 1.0f);
```
HansKristian-Work: Indices are not written here, is that expected to work?

tests/d3d12_mesh_shader.c
```diff
@@ -0,0 +1,2326 @@
...
+    ms_ia_pipeline_desc.primitive_topology = primitive_topology_subobject;
+    ms_ia_pipeline_desc.input_layout = input_layout_subobject;
+
+    hr = create_pipeline_state_from_stream(device2, &ms_ia_pipeline_desc, &pipeline_state);
+    ok(SUCCEEDED(hr), "Failed to create pipeline, hr %#x.\n", hr);
+    ID3D12PipelineState_Release(pipeline_state);
+
+    ID3D12RootSignature_Release(root_signature);
+    ID3D12Device2_Release(device2);
+    ID3D12Device_Release(device);
```
HansKristian-Work: Missing test context destroy?

tests/d3d12_mesh_shader.c
```diff
@@ -0,0 +1,2326 @@
...
+    }
+
+    hr = ID3D12Device_QueryInterface(device, &IID_ID3D12Device2, (void **)&device2);
+    ok(SUCCEEDED(hr), "Failed to query ID3D12Device2.\n");
+
+    root_signature_desc.NumParameters = 0;
+    root_signature_desc.pParameters = NULL;
+    root_signature_desc.NumStaticSamplers = 0;
+    root_signature_desc.pStaticSamplers = NULL;
+    root_signature_desc.Flags = D3D12_ROOT_SIGNATURE_FLAG_ALLOW_INPUT_ASSEMBLER_INPUT_LAYOUT;
```
HansKristian-Work: Is this required for mesh shaders? I thought input layout was just VS.

tests/d3d12_mesh_shader.c
doitsujin: There is no context here, we just create a raw device for this test.

tests/d3d12_mesh_shader.c
doitsujin: Hm, no, although we don't execute the shader here and it somehow compiles in D3D anyway. I can replace this with a different shader though.

tests/d3d12_mesh_shader.c
doitsujin: It's not, the point here is that pipeline compilation still succeeds and just ignores these subobjects.

tests/d3d12_mesh_shader.c
HansKristian-Work: I wonder if "triangle list" like output is inferred if we don't declare index outputs.

tests/d3d12_mesh_shader.c
HansKristian-Work: OK, fair enough.

tests/d3d12_mesh_shader.c
HansKristian-Work: True.

tests/d3d12_mesh_shader.c
```diff
@@ -0,0 +1,2326 @@
...
+        uint cull_mask = buf[gid.x];
+        uint primitive_index = ~0u;
+
+        if (cull_mask & (1u << tid))
+            InterlockedAdd(primitive_count, 1u, primitive_index);
+
+        GroupMemoryBarrierWithGroupSync();
+        SetMeshOutputCounts(3, primitive_count);
+
+        if (primitive_count == 0)
```
HansKristian-Work: It seems a bit spicy to emit 3 vertices + 0 primitives, but it should be legal, albeit bizarre.

doitsujin: The AMD driver hangs if we don't call `SetMeshOutputCounts` prior to returning, but yeah, this works on both drivers now and is probably something we should end up supporting.

HansKristian-Work (approved):
Tests pass on my laptop as well. A bit intense to review every line of code, but overall the coverage seems broad enough to be useful for mesh shader bringup. Thanks!

PR #979, opened 2022-02-01
vkd3d: Fix reporting of WriteBufferImmediateSupportFlags.

doitsujin:
Oversight from when we added bundles.


HansKristian-Work (approved):
LGTM.

PR #981, opened 2022-02-01
SPIR-V caching - Phase 2 - Rewrite pipeline library implementation

HansKristian-Work:
Builds on https://github.com/HansKristian-Work/vkd3d-proton/pull/980.

This became basically a rewrite in the end, and it got too awkward to
split these commits in any meaningful way.

The goals here were primarily to:

- Support serializing SPIR-V and load SPIR-V.
  To do this robustly requires a lot more validation and checks to make
  sure end up compiling the same SPIR-V that we load from cache.
  This is critical for performance when games have primed their pipeline
  libraries and expect that loading a PSO should be fast. Without this,
  we will hit vkd3d-shader for every PSO, causing very long load times.
- Implement the required validation for mismatched PSO descriptions.
- Rewrite the binary layout of the pipeline library for flexibility
  concerns and performance.
  If the pipeline library is mmap-ed from disk - which appears to be
  the intended use - we only need to scan through the TOC to fully parse
  the library contents.
  From a flexibility concern, a blob needs to support inlined data,
  but a library can use referential links. We introduce separate
  hashmaps which store deduplicated SPIR-V and pipeline cache blobs,
  which significantly drop memory and storage requirements.
  For future improvements, it should be fairly easy to add information
  which lets us avoid SPIR-V or pipeline cache data altogether if
  relevant changes to Vulkan/drivers are made.

### DIFF
--- a/libs/vkd3d/cache.c
+++ b/libs/vkd3d/cache.c
@@ -1,5 +1,6 @@
 /*
  * Copyright 2020 Philip Rebohle for Valve Corporation
+ * Copyright 2022 Hans-Kristian Arntzen for Valve Corporation
  *
  * This library is free software; you can redistribute it and/or
  * modify it under the terms of the GNU Lesser General Public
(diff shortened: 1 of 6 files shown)


libs/vkd3d/cache.c
```diff
@@ -35,75 +165,697 @@ VkResult vkd3d_create_pipeline_cache(struct d3d12_device *device,
...
+    {
+        ERR("Corrupt PSO cache blob entry found!\n");
+        /* Same rationale as above, avoid E_INVALIDARG, since that may confuse some games */
         return D3D12_ERROR_DRIVER_VERSION_MISMATCH;
+    }
+
+    /* Fetch compat info. */
+    chunk = find_blob_chunk(CONST_CAST_CHUNK_BASE(blob), payload_size, VKD3D_PIPELINE_BLOB_CHUNK_TYPE_PSO_COMPAT);
+    if (!chunk || chunk->size != sizeof(*pso_compat))
+        return E_FAIL;
```
doitsujin, line 317: I'm not entirely sure about the return codes here (and below), what's the rationale for those?

HansKristian-Work: The E_INVALIDARGs are supposed to catch the error in D3D12 API. If you pass mismatching descriptions compared to the blob, API requires us to fail the call.

The E_FAIL should be seen more as an assertion. I think I made it that error code since there was not other logical error code to use, but I can make that D3D12_ERROR_DRIVER_VERSION_MISMATCH I think.

HansKristian-Work: The validate call is made early in the Create*Pipeline calls, and it's supposed to catch any invalid usage early. We cannot keep trucking on and build SPIR-V files late.

HansKristian-Work: Fixed.

doitsujin (approved):

PR #983, opened 2022-02-02
vkd3d: Fold contiguous WriteBufferImmediate calls

rlocatti-nv:
Fold contiguous WriteBufferImmediate calls into unified vkCmdUpdateBuffer invocations.
Previous to this commit, each updated DWORD would be translated into a single vkCmdUpdateBuffer call.

This helps titles using many buffer updates in their command lists.

Signed-of-by: Rodrigo Locatti rlocatti@nvidia.com

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -9493,6 +9493,10 @@ static void STDMETHODCALLTYPE d3d12_command_list_WriteBufferImmediate(d3d12_comm
     const struct vkd3d_vk_device_procs *vk_procs = &list->device->vk_procs;
     VkPipelineStageFlagBits stage = VK_PIPELINE_STAGE_TOP_OF_PIPE_BIT;
     const struct vkd3d_unique_resource *resource;
+    VkBuffer payload_buffer = VK_NULL_HANDLE;
+    VkDeviceSize begin_offset = 0;
+    size_t payload_index = 0;
+    UINT32 payload[0x40];
     VkDeviceSize offset;
     unsigned int i;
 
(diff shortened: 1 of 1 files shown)


doitsujin:
Which game are we talking about here?

rlocatti-nv:
It's affecting Hitman 3.

libs/vkd3d/command.c
```diff
@@ -9496,6 +9496,11 @@ static void STDMETHODCALLTYPE d3d12_command_list_WriteBufferImmediate(d3d12_comm
     VkDeviceSize offset;
     unsigned int i;
 
+    VkBuffer payload_buffer = VK_NULL_HANDLE;
+    size_t payload_index = 0;
+    VkDeviceSize begin_offset = 0;
+    UINT32 payload[0x40];
```
DadSchoorse, line 9502: This belongs in the christmas tree above.

DadSchoorse: It's still not a christmas tree, you have to sort the lines by length. :frog:

doitsujin:
I think there's a bit of a conflict of interest here, `WriteBufferImmediate` is **NOT** meant to be a `vkUpdateBuffer` equivalent. If games use it this way anyway then we'll have to carefully isolate the cases where this makes sense, and do it on all GPUs, vs when apps actually want `vkCmdWriteBufferMarkerAMD` sematics.

I also don't know if this is really the best solution, copying the parameter array into a scratch buffer and dispatching a compute shader that performs BDA writes may be better in case some games do lots of small disjoint updates.

In any case, I'd like to hold off on merging this and revisit this later.

doitsujin:
Superseded by #1053.

PR #988, opened 2022-02-04
Fix Meson breakage on Windows CI

Saancreed:
If you can't beat them, avoid the confrontation?

I still have no idea why that one Meson call fails but whatever Meson wants to do there, the `vs2022` backend handles on its own.

The second commit pulls the acquisition of environment variables into the build step itself so they don't leak between x86 and x64 builds, otherwise MSBuild complains about paths being too long.


HansKristian-Work (approved):
LGTM, thanks

PR #993, opened 2022-02-08
Add breadcrumb instrumentation support

HansKristian-Work:
Idea is to greatly aid GPU crash debugging where we also have context of the D3D12 command list state around the crash site.


libs/vkd3d/breadcrumbs.c
```diff
@@ -0,0 +1,648 @@
...
+    D3D12_RESOURCE_DESC1 resource_desc;
+    HRESULT hr;
+    int rc;
+
+    memset(tracer, 0, sizeof(*tracer));
+
+    if ((rc = pthread_mutex_init(&tracer->lock, NULL)))
+        return hresult_from_errno(rc);
+
+    if (device->device_info.device_coherent_memory_features_amd.deviceCoherentMemory &&
```
werman, line 104: Is it possible to make `VK_AMD_device_coherent_memory` optional? In Turnip we don't have it implemented. On the other hand our memory by default is uncached | coherent, but writes to it from GPU may still have to be flushed from internal caches (the AMD extension seem to imply that any write to such memory bypasses all caches).

werman: > Can you guarantee that the marker write is coherent/uncached?

We have two means of writing the marker:
- For top of the pipe marker - via `CP_MEM_WRITE` which goes into a cache but we also dirty it so the next draw/blit would flush it.
- For others - through `CP_EVENT_WRITE` which isn't cached.

And the underlying memory itself is uncached and coherent.

HansKristian-Work: Fixed.

doitsujin (approved):

PR #995, opened 2022-02-11
Improve GPU hang debugging experience

HansKristian-Work:
Rewrites the debug ring to be useful in device lost scenarios.

Based on https://github.com/HansKristian-Work/vkd3d-proton/pull/993.


doitsujin (approved):
Not really familiar with what's going on here but it doesn't touch common code, so should be good to go.

PR #1010, opened 2022-02-25
Some performance workarounds for Elden Ring

HansKristian-Work:


libs/vkd3d/command.c
```diff
@@ -1587,7 +1587,36 @@ static ULONG STDMETHODCALLTYPE d3d12_command_allocator_Release(ID3D12CommandAllo
...
+            {
+                if (device->cached_command_allocator_count < ARRAY_SIZE(device->cached_command_allocators))
+                {
+                    /* Recycle the pool. Some games spam free/allocate pools,
+                     * even if it completely goes against the point of the API. */
+
+                    /* Have to free command buffers here if we're going to recycle,
+                     * otherwise DestroyCommandPool takes care of it. */
+                    VK_CALL(vkFreeCommandBuffers(device->vk_device, allocator->vk_command_pool,
+                            allocator->command_buffer_count, allocator->command_buffers));
```
doitsujin, line 1604: are we setting `àllocator->command_buffer_count` to 0 anywhere?

doitsujin (commented):
Do we really need to hide this behind an option? I suppose this might eat some host/hvv memory 
 in the worst case, but I don't expect well-behaved apps to destroy (m)any command pools in the first place.

libs/vkd3d/command.c
HansKristian-Work: This is in the path where allocator gets destroyed. command_buffer_count is set to 0 in command_allocator_Reset.

doitsujin: We'd be double-freeing command buffers there though:
```c
    if (allocator->command_buffer_count)
    {
        VK_CALL(vkFreeCommandBuffers(device->vk_device, allocator->vk_command_pool,
                allocator->command_buffer_count, allocator->command_buffers));
        allocator->command_buffer_count = 0;
    }
```

doitsujin (approved):

PR #1014, opened 2022-02-25
cache: Fix sign mismatch warning.

HansKristian-Work:


doitsujin:
What's the problem here and shouldn't we rather fix our headers?

HansKristian-Work:
Good point ... Fixed.

doitsujin (approved):

PR #1030, opened 2022-03-08
Implement correct behavior of multiple retrievals of pipeline state from libraries.

HansKristian-Work:
Both MHR and Elden Ring retrieve the same pipelines multiple times. The D3D12 pipeline library API is designed so that multiple retrievals is very fast and reuses an existing internally referenced state object. Avoids us having to validate, decompress, create SPIR-V objects, create pipeline cache blob, create pipeline, etc.

We still have to perform light validation, but pales in comparison to what we did ...

Also adds extensive test for the edge cases in ref-counting shenanigans.

### DIFF
--- a/tests/d3d12_pso_blob.c
+++ b/tests/d3d12_pso_blob.c
@@ -125,11 +125,15 @@ void test_pipeline_library(void)
     ID3D12PipelineLibrary *pipeline_library;
     ID3D12RootSignature *root_signature;
     struct test_context context;
+    ID3D12PipelineState *state3;
+    ID3D12PipelineState *state2;
     ID3D12PipelineState *state;
+    ULONG reference_refcount;
     size_t serialized_size;
     ID3D12Device1 *device1;
     void *serialized_data;
     ID3D12Device *device;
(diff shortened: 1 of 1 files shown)


libs/vkd3d/state.c
```diff
@@ -1874,6 +1880,17 @@ static ULONG STDMETHODCALLTYPE d3d12_pipeline_state_Release(ID3D12PipelineState
 
         d3d12_device_release(device);
```
doitsujin, line 1881: Do we test the device ref count behaviour? Normally we're supposed to release the device when the **public** ref count reaches 0, and AddRef the device when the public ref count goes from 0 to 1.

Probably not the end of the world here since the pipeline objects are owned by the pipeline library rather than the device itself, but generally this is done to avoid circular references.

HansKristian-Work: I can test this.

HansKristian-Work: Correct, we have to release the device reference, but not the private data. Just d3d12 memes ...

doitsujin (approved):

PR #1033, opened 2022-03-08
vkd3d: Remove unused renderpass remains. 

DadSchoorse:


libs/vkd3d/vulkan_procs.h
```diff
@@ -156,7 +152,6 @@ VK_DEVICE_PFN(vkGetImageSparseMemoryRequirements2)
 VK_DEVICE_PFN(vkGetImageSubresourceLayout)
 VK_DEVICE_PFN(vkGetPipelineCacheData)
 VK_DEVICE_PFN(vkGetQueryPoolResults)
-VK_DEVICE_PFN(vkGetRenderAreaGranularity)
```
doitsujin, line 159: Bit late but render areas are still very much a thing with KHR_dynamic_rendering.

We're not using this function though.

DadSchoorse: I thought this function can't be used without renderpasses.

PR #1034, opened 2022-03-09
vkd3d: Enable VK_KHR_create_renderpass2.

HansKristian-Work:
Required extension by VK_KHR_fragment_shading_rate and
VK_KHR_separate_depth_stencil_layouts, but we don't care about enabling
any features or use it directly.

Needed to silence validation errors.


DadSchoorse:
Ah sorry I didn't think about that.

doitsujin:
Seems a bit silly that this is required since we're not using the extension anymore, but well.

doitsujin (approved):

PR #1035, opened 2022-03-09
Avoid implicit DSV layout transitions with GENERAL layout when feasible

HansKristian-Work:
The idea behind this change is that some implementations don't really care if we use GENERAL layout, so might as well just use it to avoid having to carefully transition to correct image DSV layouts everywhere.

The key improvement here is in rendering passes which split depth-rendering over many parallel command lists. This is a very common pattern which we ideally want to avoid.

Also, since we're no longer doing transitions for DSV images, drop the somewhat redundant UNDEFINED transition in Clear*View() if we don't need to handle aliasing rules.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -2265,6 +2265,9 @@ static VkImageLayout dsv_plane_optimal_mask_to_layout(uint32_t plane_optimal_mas
         VK_IMAGE_LAYOUT_DEPTH_STENCIL_ATTACHMENT_OPTIMAL,
     };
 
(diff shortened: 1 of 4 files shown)


HansKristian-Work:
On RX 6800 this is a small win in HZD benchmarking, about 1% or so, and that benchmark is quite CPU heavy.
Some NVIDIA users on Discord also report a small, but noticable FPS gain.

As long as this does not regress performance (in fact it seems to be a small win), I would argue for us to move to GENERAL DSV where we can.

doitsujin (approved):

PR #1047, opened 2022-03-22
vkd3d: Do not create pipeline variants for NULL RTVs.

doitsujin:
Not sure if this is doing the right thing since this seemed a bit *too* simple, but it passes our tests on RADV.


HansKristian-Work (approved):
LGTM. Passes on NV as well.

PR #1049, opened 2022-03-23
Don't create pipeline variants for NULL DSV

doitsujin:
Similar to the RTV thing. Hope I caught everything.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -4489,11 +4489,6 @@ static void STDMETHODCALLTYPE d3d12_command_list_ClearState(d3d12_command_list_i
     d3d12_command_list_reset_api_state(list, pipeline_state);
 }
 
-static bool d3d12_graphics_pipeline_state_is_depth_stencil_sensitive(const struct d3d12_graphics_pipeline_state *graphics)
-{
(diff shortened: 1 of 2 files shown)


libs/vkd3d/command.c
```diff
@@ -8064,7 +8053,7 @@ static void STDMETHODCALLTYPE d3d12_command_list_OMSetRenderTargets(d3d12_comman
     {
         graphics = &list->state->graphics;
 
-        if (prev_dsv_format != next_dsv_format && d3d12_graphics_pipeline_state_is_depth_stencil_sensitive(graphics))
+        if (prev_dsv_format != next_dsv_format && d3d12_graphics_pipeline_state_has_unknown_dsv_format(graphics))
```
HansKristian-Work, line 8056: I think the code reads better if we also rename this to has_unknown_dsv_format_with_test_enabled or something like that to make it clear it's a workaround situation. Just "unknown_dsv_format" would normally mean it doesn't use depth-test at all, where "is depth stencil sensitive" is clearer for what it tries to tell.

HansKristian-Work (approved):
The PR itself is fine though, I'll just do the rename in a follow up PR.

PR #1053, opened 2022-03-29
Optimize WriteBufferImmediate

doitsujin:
But preserve marker semantics if we can support them, and also add buffer write tracking that was previously missing.

Supersedes #983. @rlocatti-nv can you test if this has the desired effect in the games you mentioned?


rlocatti-nv:
From some quick benchmarking, I see a ~2% improvement on Hitman 3. It has the desired effect.

HansKristian-Work (approved):
LGTM

PR #1059, opened 2022-04-01
vkd3d: Force non-invariant position for Serious Sam 4.

dbarkar:
Signed-off-by: Denis Barkar <dbarkar@nvidia.com>


HansKristian-Work (approved):
LGTM

PR #1060, opened 2022-04-01
vkd3d: Remove VKD3D_MAX_DYNAMIC_STATE_COUNT

Joshua-Ashton:
This was off by one, at some point, which could cause a stack buffer overrun which is naughty.

Replace this with just an ARRAY_SIZE on the dynamic_state_list for the array size.


HansKristian-Work (approved):
LGTM, good catch.

PR #1062, opened 2022-04-04
Rework ClearUnorderedAccessViewUint for R11G11B10

doitsujin:
Mirrors a recent [DXVK change](https://github.com/doitsujin/dxvk/commit/c87660480d8c18d94cee61a20276f70919639012). Could use this path for RGB9E5 in the future as well if needed.

TL;DR: Use buffer->image copy instead of creating an R32 view, but still use existing R11G11B10 view if the clear value is zero since the bit pattern will be preserved anyway. The expectation is that the fallback path will be hit extremely rarely, if ever, in real-world applications.


libs/vkd3d/command.c
```diff
@@ -7353,19 +7281,8 @@ static void STDMETHODCALLTYPE d3d12_command_list_ResourceBarrier(d3d12_command_l
...
-                                list->vk_queue_flags,
-                                false);
-                    }
-                    else
-                    {
-                        alias_src_access = vk_access_flags_all_possible_for_buffer(list->device,
-                                list->vk_queue_flags,
-                                false);
-                    }
+                    if (before)
```
HansKristian-Work, line 7284: I think this is a bug if resource is a buffer. This should be unconditionally added I think.

libs/vkd3d/command.c
```diff
@@ -8306,6 +8306,135 @@ static void d3d12_command_list_clear_uav(struct d3d12_command_list *list,
...
+            VK_PIPELINE_STAGE_TRANSFER_BIT,
+            0, 1, &barrier, 0, NULL, 0, NULL));
+
+    copy_region.sType = VK_STRUCTURE_TYPE_BUFFER_IMAGE_COPY_2_KHR;
+    copy_region.pNext = NULL;
+    copy_region.bufferOffset = scratch.offset;
+    copy_region.bufferRowLength = 0;
+    copy_region.bufferImageHeight = 0;
+    copy_region.imageOffset.z = 0;
+    copy_region.imageExtent.depth = 1;
```
HansKristian-Work, line 8391: This seems like it would break for 3D clears.

HansKristian-Work: nvm, check below ...

libs/vkd3d/command.c
```diff
@@ -8306,6 +8306,135 @@ static void d3d12_command_list_clear_uav(struct d3d12_command_list *list,
...
+    }
+
+    if (!d3d12_command_allocator_allocate_scratch_memory(list->allocator,
+            scratch_buffer_size, sizeof(uint32_t), &scratch))
+    {
+        ERR("Failed to allocate scratch memory for UAV clear");
+        return;
+    }
+
+    VK_CALL(vkCmdFillBuffer(list->vk_command_buffer, scratch.buffer,
```
HansKristian-Work, line 8369: Is it possible in theory to implement B10G11R11 clear by converting the 32-bit UINT payload back to a vec4 and clear "normally" with that?

libs/vkd3d/command.c
```diff
@@ -8306,6 +8306,135 @@ static void d3d12_command_list_clear_uav(struct d3d12_command_list *list,
...
+    VkMemoryBarrier barrier;
+
+    d3d12_command_list_track_resource_usage(list, resource, true);
+    d3d12_command_list_end_current_render_pass(list, false);
+
+    assert(args->has_view);
+    assert(d3d12_resource_is_texture(resource));
+
+    if (d3d12_resource_is_texture(resource))
+        miplevel_idx = args->u.view->info.texture.miplevel_idx;
```
HansKristian-Work, line 8330: We assert above that we have a texture.

libs/vkd3d/command.c
```diff
@@ -8306,6 +8306,135 @@ static void d3d12_command_list_clear_uav(struct d3d12_command_list *list,
...
+    {
+        scratch_buffer_size = 0;
+
+        for (i = 0; i < rect_count; i++)
+        {
+            if (rects[i].right > rects[i].left && rects[i].bottom > rects[i].top)
+            {
+                unsigned int w = rects[i].right - rects[i].left;
+                unsigned int h = rects[i].bottom - rects[i].top;
+                scratch_buffer_size = max(scratch_buffer_size, layer_count * w * h * sizeof(uint32_t));
```
HansKristian-Work, line 8353: Seems a bit wasteful to allocate for many layers if we can just broadcast the same layer buffer copy.

libs/vkd3d/command.c
```diff
@@ -8442,43 +8578,61 @@ static void STDMETHODCALLTYPE d3d12_command_list_ClearUnorderedAccessViewUint(d3
...
+                inline_view = args.u.view;
+            }
+            else
             {
-                ERR("Failed to create image view.\n");
-                return;
+                /* If the clear color is 0, we can safely use the existing view to perform the
+                 * clear since the bit pattern will not change. Otherwise, fill a scratch buffer
+                 * with the packed clear value and perform a buffer to image copy. */
+                if (color.uint32[0])
```
HansKristian-Work, line 8622: If we have 0 color here and we hit the normal clear path, wouldn't B10G11R11 end up looking at the 3 float values instead, which might not have been cleared to zero?

doitsujin: The fixup function zero-initializes its return value now.

libs/vkd3d/command.c
doitsujin: I don't think so, there are multiple NaN bit patterns that we'd lose and potential rounding errors.

libs/vkd3d/command.c

libs/vkd3d/command.c
doitsujin: Derp, relic of the past. Fixed.

libs/vkd3d/command.c
doitsujin: Added code to iterate over subresource layers now, still need this for 3D though.

libs/vkd3d/command.c
HansKristian-Work: NaN, ye ... that is problematic.

libs/vkd3d/command.c
HansKristian-Work: Right.

libs/vkd3d/command.c
```diff
@@ -8305,6 +8305,133 @@ static void d3d12_command_list_clear_uav(struct d3d12_command_list *list,
...
+            VK_PIPELINE_STAGE_TRANSFER_BIT | vk_queue_shader_stages(list->vk_queue_flags),
+            VK_PIPELINE_STAGE_TRANSFER_BIT,
+            0, 1, &barrier, 0, NULL, 0, NULL));
+
+    copy_region.sType = VK_STRUCTURE_TYPE_BUFFER_IMAGE_COPY_2_KHR;
+    copy_region.pNext = NULL;
+    copy_region.bufferOffset = scratch.offset;
+    copy_region.bufferRowLength = 0;
+    copy_region.bufferImageHeight = 0;
+    copy_region.imageOffset.z = 0;
```
HansKristian-Work, line 8384: Is there no front/back in D3D12_RECT?

doitsujin: No, that would be D3D_BOX.

HansKristian-Work: So the API is kinda broken, great <_<

HansKristian-Work (approved):
LGTM, thanks.

PR #1065, opened 2022-04-06
Adding support for profiling using Tracy.

dave-juicelabs:


subprojects/tracy.wrap
```diff
@@ -0,0 +1,5 @@
+[wrap-git]
```
HansKristian-Work, line 1: What is the purpose of this file? I don't know what a .wrap file is.

meson.build
```diff
@@ -1,4 +1,4 @@
-project('vkd3d-proton', ['c'], version : '2.6', meson_version : '>= 0.49', default_options : [
+project('vkd3d-proton', ['c'], version : '2.6', meson_version : '>= 0.55', default_options : [
```
HansKristian-Work, line 1: Is 0.55 required?

libs/vkd3d/command.c
```diff
@@ -11673,6 +11674,7 @@ static void *d3d12_command_queue_submission_worker_main(void *userdata)
     VKD3D_REGION_DECL(queue_signal);
     VKD3D_REGION_DECL(queue_execute);
 
+    VKD3D_PROFILE_THREAD_NAME("vkd3d_queue");
```
HansKristian-Work, line 11677: Couldn't this be part of vkd3d_set_thread_name()?

libs/vkd3d-common/profiling.c
```diff
@@ -192,4 +226,34 @@ void vkd3d_profiling_notify_work(unsigned int index,
     spinlock_release(lock);
 }
 
+#ifdef TRACY_ENABLE
+
+void ___vkd3d_set_thread_name( const char* name )
```
HansKristian-Work, line 231: Formatting is all over the place here. Do we need triple underscores? (reserved identifier)

libs/vkd3d-common/profiling.c
```diff
@@ -192,4 +226,34 @@ void vkd3d_profiling_notify_work(unsigned int index,
...
+
+void ___vkd3d_emit_frame_mark()
+{
+    if(pfn_tracy_emit_frame_mark)
+        pfn_tracy_emit_frame_mark(0);
+}
+
+TracyCZoneCtx ___vkd3d_emit_zone_begin( const struct ___tracy_source_location_data* srcloc, int active )
+{
+    static TracyCZoneCtx ctx;
```
HansKristian-Work, line 245: What if this is used from multiple threads concurrently?

libs/vkd3d-common/profiling.c
```diff
@@ -122,22 +122,56 @@ static void vkd3d_init_profiling_path(const char *path)
...
+
+static void (*pfn_tracy_set_thread_name)( const char* name );
+static void (*pfn_tracy_emit_frame_mark)( const char* name );
+static TracyCZoneCtx (*pfn_tracy_emit_zone_begin)( const struct ___tracy_source_location_data* srcloc, int active );
+static void (*pfn_tracy_emit_zone_end)( TracyCZoneCtx ctx );
+
+static void vkd3d_init_profiling_once(void)
+{
+#if defined(_WIN32)
+#define SONAME_TRACY "tracy.dll"
```
HansKristian-Work, line 143: Does this have to be a .dll? It looks like Trace is being built statically?

include/private/vkd3d_common.h
```diff
@@ -118,7 +118,7 @@ static inline unsigned int vkd3d_bitmask_tzcnt32(uint32_t mask)
 {
 #ifdef _MSC_VER
     unsigned long result;
-    _BitScanForward(&result, mask) ? result : 32;
+    _BitScanForward(&result, mask);
```
HansKristian-Work, line 121: This seems unrelated?

meson.build
```diff
@@ -69,6 +70,10 @@ if enable_breadcrumbs
   add_project_arguments('-DVKD3D_ENABLE_BREADCRUMBS', language : 'c')
 endif
 
+if enable_tracy
+  add_project_arguments('-DTRACY_ENABLE', language : 'c')
```
HansKristian-Work, line 74: enable_tracy should also imply enable_profiling.

subprojects/packagefiles/tracy-0.8/meson.build
```diff
@@ -0,0 +1,78 @@
+project('tracy', ['cpp'], version : '0.8')
+
+add_project_arguments('-DTRACY_EXPORTS', language : 'cpp')
+add_project_arguments('-DTRACY_ENABLE', language : 'cpp')
+add_project_arguments('-DTRACY_ONLY_LOCALHOST', language : 'cpp')
+add_project_arguments('-DTRACY_ON_DEMAND', language : 'cpp')
+add_project_arguments('-DTRACY_NO_CALLSTACK', language : 'cpp')
+add_project_arguments('-DTRACY_DELAYED_INIT', language : 'cpp')
+
+threads_dep = dependency('threads')
```
HansKristian-Work, line 10: Is all of this needed? I thought we're just adding tracing for vkd3d-proton here.

subprojects/tracy.wrap
DadSchoorse: It's a way to manage dependencies with meson: https://mesonbuild.com/Wrap-dependency-system-manual.html

HansKristian-Work: Ok, so this is something Meson supports as-is? Do we know if the Meson Proton builds with supports this?

DadSchoorse: I think Proton uses the soldier container, which is based on debian 10 (buster). So unless soldier has backported a new version, vkd3d-proton is stuck with meson 0.49.

eli-schwartz:
Upstream tracy includes a meson.build file which was once PRed by someone but only has a very basic version. Perhaps you should submit an improvement there? :)

Then you could rely on Meson's support for wrap-git since Meson 0.23 and `dependency('tracy', fallback: ['tracy', 'tracy_dep'])` since Meson 0.27

Use of patch_directory in this PR to apply your own alternative meson.build file greatly increases the required version though.

meson.build
dave-juicelabs: Only for the patch_directory portion of the .wrap file. I will find a way around that.

libs/vkd3d/command.c
dave-juicelabs: Yes it could. My thinking was to pass the static constant directly in to the function for performance but that is unnecessary for thread names.

libs/vkd3d-common/profiling.c
dave-juicelabs: I did that strictly to follow Tracys naming scheme.

libs/vkd3d-common/profiling.c
dave-juicelabs: Tracys functions are all thread safe. The default return value shouldn't be modified by the calling code. If the tracy dll is not found, ctx would not be passed to ___tracy_emit_zone_end either.

include/private/vkd3d_common.h
dave-juicelabs: Unrelated but was throwing a bunch of warnings in Visual Studio Clang. Can be removed from this PR if necessary.

meson.build
dave-juicelabs: Will update.

subprojects/packagefiles/tracy-0.8/meson.build
dave-juicelabs: The no_callstack and delayed_init were added to make Tracy work. The thread_local initialization and usage within Tracy was having difficulties on my machine.

dave-juicelabs:
I will update Tracys repo and remove the patch_directory requirement. In that case, Tracy can also become a submodule if desired.

dave-juicelabs:
image
image
image

dave-juicelabs:
I have created a PR in tracy but also set my fork here in the .wrap file for now until the Tracy PR is accepted.

dave-juicelabs:
PR for tracy has been merged.

meson.build
```diff
@@ -69,6 +71,15 @@ if enable_breadcrumbs
   add_project_arguments('-DVKD3D_ENABLE_BREADCRUMBS', language : 'c')
 endif
 
+if enable_tracy
+  add_project_arguments('-DVKD3D_ENABLE_PROFILING', language : 'c')
```
HansKristian-Work, line 75: This will probably conflict if enable_profiling is also set. enable_tracy should set that variable.

libs/vkd3d-common/profiling.c
```diff
@@ -192,4 +234,60 @@ void vkd3d_profiling_notify_work(unsigned int index,
     spinlock_release(lock);
 }
 
+#ifdef TRACY_ENABLE
+
+#ifdef DYNAMIC_TRACY
+
+void tracy_set_thread_name( const char* name )
```
HansKristian-Work, line 241: Formatting.

include/private/vkd3d_profiling.h
```diff
@@ -25,6 +25,31 @@
...
+#define VKD3D_PROFILE_FRAME() tracy_emit_frame_mark()
+
+void vkd3d_init_profiling(void);
+bool vkd3d_uses_profiling(void);
+
+#define VKD3D_STRINGIZE1(x) #x
+#define VKD3D_STRINGIZE(x) VKD3D_STRINGIZE1(x)
+
+#define VKD3D_REGION_DECL(name) TracyCZoneCtx _vkd3d_tracy_##name
+//#define TracyCZoneN( ctx, name, active ) static const struct ___tracy_source_location_data TracyConcat(__tracy_source_location,__LINE__) = { name, __func__,  __FILE__, (uint32_t)__LINE__, 0 }; TracyCZoneCtx ctx = ___tracy_emit_zone_begin( &TracyConcat(__tracy_source_location,__LINE__), active );
```
HansKristian-Work, line 47: Dead line.

include/private/vkd3d_common.h
HansKristian-Work: Which warning? This code will break since it doesn't handle the mask == 0 case.

subprojects/tracy.wrap
```diff
@@ -0,0 +1,3 @@
...
+url = https://github.com/wolfpld/tracy.git
+revision = 9b15870991da9ebf7d4e2b808691c3ce91af9356
```
HansKristian-Work, line 3: I tried building this on MinGW, but it fails with:

```
../subprojects/tracy/client/TracySysTrace.cpp: In function ‘void tracy::SetupVsync()’:
../subprojects/tracy/client/TracySysTrace.cpp:303:5: error: ‘EVENT_FILTER_EVENT_ID’ was not declared in this scope; did you mean ‘EVENT_FILTER_HEADER’?
  303 |     EVENT_FILTER_EVENT_ID fe = {};
      |     ^~~~~~~~~~~~~~~~~~~~~
      |     EVENT_FILTER_HEADER
../subprojects/tracy/client/TracySysTrace.cpp:304:5: error: ‘fe’ was not declared in this scope
  304 |     fe.FilterIn = TRUE;
      |     ^~
../subprojects/tracy/client/TracySysTrace.cpp:311:17: error: ‘EVENT_FILTER_TYPE_EVENT_ID’ was not declared in this scope; did you mean ‘EVENT_FILTER_TYPE_SCHEMATIZED’?
  311 |     desc.Type = EVENT_FILTER_TYPE_EVENT_ID;
      |                 ^~~~~~~~~~~~~~~~~~~~~~~~~~
      |                 EVENT_FILTER_TYPE_SCHEMATIZED
../subprojects/tracy/client/TracySysTrace.cpp:315:29: error: ‘EVENT_ENABLE_PROPERTY_IGNORE_KEYWORD_0’ was not declared in this scope
  315 |     params.EnableProperty = EVENT_ENABLE_PROPERTY_IGNORE_KEYWORD_0;
      |                             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

subprojects/tracy.wrap
HansKristian-Work: Tried building this PR in the Proton docker and it builds fine with default setup, so I think it's safe.

subprojects/tracy.wrap
```diff
@@ -0,0 +1,3 @@
+[wrap-git]
+url = https://github.com/wolfpld/tracy.git
```
HansKristian-Work, line 2: I also tried to build this, and this submodule is not cloned unless explicitly enabled, which is good.

meson.build
dave-juicelabs: Fixed

libs/vkd3d-common/profiling.c
dave-juicelabs: Fixed

include/private/vkd3d_common.h
dave-juicelabs: `In file included from ..\..\libs\vkd3d-shader/vkd3d_shader_private.h:48:
..\..\include\private\vkd3d_common.h(121,47): warning: expression result unused [-Wunused-value]
    _BitScanForward(&result, mask) ? result : 32;`

From what I can tell, this code is not checking the mask but rather a remnant from a previous implementation. I searched the rest of the codebase and find no other instances that is like this one.

This warning is from MSVC Clang.

subprojects/tracy.wrap
dave-juicelabs: PR has been merged.

HansKristian-Work:
CI complains here:

`meson.build:86:0: ERROR: Include dir ../../common/vulkan/include does not exist.`

include/private/vkd3d_common.h
HansKristian-Work: Yes, this is actually just a derpy bug, fun.

include/private/vkd3d_profiling.h
```diff
@@ -25,6 +25,32 @@
...
+#define VKD3D_PROFILE_FRAME() tracy_emit_frame_mark()
+
+void vkd3d_init_profiling(void);
+bool vkd3d_uses_profiling(void);
+
+#define VKD3D_STRINGIZE1(x) #x
+#define VKD3D_STRINGIZE(x) VKD3D_STRINGIZE1(x)
+
+#define VKD3D_REGION_DECL(name) TracyCZoneCtx _vkd3d_tracy_##name
+// Copied from definition of TracyCZoneN. In this implementation, _vkd3d_tracy_##name is expected
```
HansKristian-Work, line 47: `/* */` comment style is used.

meson.build
```diff
@@ -69,7 +71,15 @@ if enable_breadcrumbs
...
-vkd3d_external_includes = [ './subprojects/Vulkan-Headers/include', './subprojects/SPIRV-Headers/include' ]
+if enable_tracy
+  add_project_arguments('-DTRACY_ENABLE', language : 'c')
+
+  if dynamic_tracy
+    add_project_arguments('-DDYNAMIC_TRACY', language : 'c')
+  endif
+endif
+
+vkd3d_external_includes = [ '../../common/vulkan/include', './subprojects/SPIRV-Headers/include' ]
```
HansKristian-Work, line 82: `common/vulkan/include` what is that?

HansKristian-Work (changes requested):
Primary concerns now are getting CI to pass and fixing code style and other comments. Otherwise, the impact is fairly contained so I'm okay supporting this.

meson.build
dave-juicelabs: Code mistakenly added from Juice's private builds.

include/private/vkd3d_profiling.h
dave-juicelabs: Done

subprojects/tracy.wrap
HansKristian-Work:
```
Cloning into 'tracy'...
remote: Enumerating objects: 31151, done.
remote: Counting objects: 100% (7416/7416), done.
remote: Compressing objects: 100% (255/255), done.
remote: Total 31151 (delta 7198), reused 7284 (delta 7153), pack-reused 23735
Receiving objects: 100% (31151/31151), 17.89 MiB | 17.60 MiB/s, done.
Resolving deltas: 100% (22675/22675), done.
HEAD is now at 9b158709 Merge pull request #361 from Juice-Labs/meson-support

Executing subproject tracy 


../subprojects/tracy/meson.build:1:0: ERROR: In subproject tracy: Unknown options: "tracy:tracy_no_sys_trace"

```
Still getting this when trying to build.

HansKristian-Work:
Ping. Last reply was https://github.com/HansKristian-Work/vkd3d-proton/pull/1065#discussion_r855974659 and haven't seen any update in a good while.

HansKristian-Work:
> Sorry about that. Missed the last message. I have updated the code to pull the correct merged commit from tracy for building.

Tried checking out the current PR, but still getting this when running

`meson .. -Denable_tracy=true`

```
Executing subproject tracy 


../subprojects/tracy/meson.build:1:0: ERROR: In subproject tracy: Unknown options: "tracy:tracy_no_sys_trace"
```

dave-juicelabs:
I built an environment straight from the action used for building Linux utilizing an Ubuntu 20.04 VM with Joshua Ashtons arch-mingw-github-action dockerfile. From there I did not reproduce the error shown above but did find some linkage errors that I have a PR to fix in the tracy project. Will update here once the PR is accepted.

What is the environment you are running so I can better reproduce this issue?

I ran in to this issue before and it was the reason for the tracy_no_sys_trace option. mingw does not have a recent version of one of the header files which is causing this error. The tracy_no_sys_trace disables all of that code.

HansKristian-Work:
> I built an environment straight from the action used for building Linux utilizing an Ubuntu 20.04 VM with Joshua Ashtons arch-mingw-github-action dockerfile. From there I did not reproduce the error shown above but did find some linkage errors that I have a PR to fix in the tracy project. Will update here once the PR is accepted.
> 
> What is the environment you are running so I can better reproduce this issue?
> 
> I ran in to this issue before and it was the reason for the tracy_no_sys_trace option. mingw does not have a recent version of one of the header files which is causing this error. The tracy_no_sys_trace disables all of that code.

I see. Meson did not update the submodule checkout for whatever reason. There was a commit hash mismatch.

Got past that, so the remaining issue seems to be a missing link to winsock2:

```
/home/maister/git/vkd3d-standalone/vkd3d/build-tracy/../subprojects/tracy/common/TracySocket.cpp:673: undefined reference to `__imp_socket'
/usr/lib/gcc/x86_64-w64-mingw32/12.1.0/../../../../x86_64-w64-mingw32/bin/ld: /home/maister/git/vkd3d-standalone/vkd3d/build-tracy/../subprojects/tracy/common/TracySocket.cpp:681: undefined reference to `__imp_setsockopt'
/usr/lib/gcc/x86_64-w64-mingw32/12.1.0/../../../../x86_64-w64-mingw32/bin/ld: /home/maister/git/vkd3d-standalone/vkd3d/build-tracy/../subprojects/tracy/common/TracySocket.cpp:688: undefined reference to `__imp_setsockopt'
/usr/lib/gcc/x86_64-w64-mingw32/12.1.0/../../../../x86_64-w64-mingw32/bin/ld: /home/maister/git/vkd3d-standalone/vkd3d/build-tracy/../subprojects/tracy/common/TracySocket.cpp:695: undefined reference to `__imp_closesocket'
/usr/lib/gcc/x86_64-w64-mingw32/12.1.0/../../../../x86_64-w64-mingw32/bin/ld: /home/maister/git/vkd3d-standalone/vkd3d/build-tracy/../subprojects/tracy/common/TracySocket.cpp:704: undefined reference to `__imp_htons'
/usr/lib/gcc/x86_64-w64-mingw32/12.1.0/../../../../x86_64-w64-mingw32/bin/ld: /home/maister/git/vkd3d-standalone/vkd3d/build-tracy/../subprojects/tracy/common/TracySocket.cpp:707: undefined reference to `__imp_bind'
/usr/lib/gcc/x86_64-w64-mingw32/12.1.0/../../../../x86_64-w64-mingw32/bin/ld: /home/maister/git/vkd3d-standalone/vkd3d/build-tracy/../subprojects/tracy/common/TracySocket.cpp:710: undefined reference to `__imp_closesocket'
/usr/lib/gcc/x86_64-w64-mingw32/12.1.0/../../../../x86_64-w64-mingw32/bin/ld: subprojects/tracy/libtracy.a(TracyClient.cpp.obj): in function `tracy::UdpListen::Close()':
/home/maister/git/vkd3d-standalone/vkd3d/build-tracy/../subprojects/tracy/common/TracySocket.cpp:725: undefined reference to `__imp_closesocket'
/usr/lib/gcc/x86_64-w64-mingw32/12.1.0/../../../../x86_64-w64-mingw32/bin/ld: subprojects/tracy/libtracy.a(TracyClient.cpp.obj): in function `tracy::UdpListen::Read(unsigned long long&, tracy::IpAddress&, int)':
/home/maister/git/vkd3d-standalone/vkd3d/build-tracy/../subprojects/tracy/common/TracySocket.cpp:739: undefined reference to `WSAPoll'
/usr/lib/gcc/x86_64-w64-mingw32/12.1.0/../../../../x86_64-w64-mingw32/bin/ld: /home/maister/git/vkd3d-standalone/vkd3d/build-tracy/../subprojects/tracy/common/TracySocket.cpp:743: undefined reference to `__imp_recvfrom'
/usr/lib/gcc/x86_64-w64-mingw32/12.1.0/../../../../x86_64-w64-mingw32/bin/ld: subprojects/tracy/libtracy.a(TracyClient.cpp.obj): in function `tracy::__wsinit::__wsinit()':
/home/maister/git/vkd3d-standalone/vkd3d/build-tracy/../subprojects/tracy/common/TracySocket.cpp:55: undefined reference to `__imp_WSAStartup'
```

HansKristian-Work:
This quick and dirty hack did allow it to link, so I think a fix needs to be pushed to Tracy:

```
diff --git a/meson.build b/meson.build
index 5c349167..54423299 100644
--- a/meson.build
+++ b/meson.build
@@ -141,6 +141,9 @@ tracy_src = [
 
 tracy_public_include_dirs = include_directories('.')
 
+compiler = meson.get_compiler('cpp')
+lib_ws2_32  = compiler.find_library('ws2_32')
+
 if tracy_shared_libs
   tracy = shared_library('tracy', tracy_src, tracy_header_files,
     dependencies        : [ threads_dep ],
@@ -149,7 +152,7 @@ if tracy_shared_libs
     install             : true)
 else
   tracy = static_library('tracy', tracy_src, tracy_header_files,
-    dependencies        : [ threads_dep ],
+    dependencies        : [ threads_dep, lib_ws2_32 ],
     include_directories : tracy_public_include_dirs,
     override_options    : [ 'cpp_std=c++11' ],
     install             : true)
```

dave-juicelabs:
That is along the lines of the PR I have outstanding in tracy.

dave-juicelabs:
Currently, since Wine most likely does not support the required features to support using Tracy, Tracy support is really only useful when on Windows. Disabling Tracy support when on non-Windows platforms makes the most sense.

sherief:
> Currently, since Wine most likely does not support the required features to support using Tracy, Tracy support is really only useful when on Windows. Disabling Tracy support when on non-Windows platforms makes the most sense.

Tracy works just fine on Wine, I use Tracy to profile Win32 apps running on Steam Deck via Proton on a near-daily basis.

meson.build
```diff
@@ -35,6 +35,12 @@ else
   enable_trace = enable_trace == 'true'
 endif
 
+# Please see https://github.com/wolfpld/tracy/pull/428.
+# Wine probably does not support the required features for Tracy.
+if host_machine.system() != 'windows' and enable_tracy
+  error('Tracy is supported only when building on Windows.')
```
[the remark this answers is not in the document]
wolfpld: The problem with MinGW is that it targets a hopelessly outdated version of the Windows SDK. Something like Win2000, or XP, I can't remember exactly. The oldest version of Windows that Microsoft still supports is Win10 1507, and that's not even active support, just security fixes.

https://endoflife.date/windows

MinGW has been a constant source of complaints that some API introduced in the Vista era (i.e. 15 years ago) is not present in the header files, and so people can't compile Tracy. It's honestly hard to find a reasonable reason to actively support it.

wolfpld:
When I last checked, Wine was not able to support Tracy due to missing `RtlWalkFrameChain` implementation. It is still only a stub in current Wine's master:

https://github.com/wine-mirror/wine/blob/5a1d431b4bf93424a415d5e3d113b1f16431fcb1/dlls/ntdll/ntdll.spec#L1095

If you don't need to capture call stacks, it should work fine though.

The documented way to perform stack capture is to use `RtlCaptureStackBackTrace`: https://learn.microsoft.com/en-us/windows-hardware/drivers/ddi/ntifs/nf-ntifs-rtlcapturestackbacktrace

And this function is implemented by Wine.

https://github.com/wine-mirror/wine/commit/4e4aa467752db723cb36ea140860ea24efbe5719

Looking at the Windows implementation of `RtlCaptureStackBackTrace` reveals that the actual work is done by `RtlWalkFrameChain`, and `RtlCaptureStackBackTrace` just does some additional data rearangement and hash calculation that is not needed by Tracy. The obvious thing to do here is to skip the middleman, especially since this is executed in code that is profiled and the performance impact should be kept as low as possible.

HansKristian-Work:
> missing RtlWalkFrameChain implementation. It is still only a stub in current Wine's master:

Not entirely surprising given it's an undocumented private function. I don't see it on MSDN at least.

wolfpld:
The sad reality is that you eventually have to dig into what is actually being executed to make things work (either more performantly, or at all). This is true not only for Windows, but also for Linux.

eli-schwartz:
And the sadder reality is that sort of thing tends to get removed without warning on a total whim whenever upstream performs trivial internal refactoring.

Although granted once they're foolish enough to provide an actual runtime ABI for it once, they tend to keep that ABI around forever.

HansKristian-Work:
This PR is going nowhere, so closing.

PR #1070, opened 2022-04-13
Add ID3D12Resource sharing support

Guy1524:
The ID3D12Resource sharing is self explanatory, it uses the infrastructure in Proton Experimental with the same data structure as DXVK, allowing d3d11/d3d12 sharing.  Note that (although undocumented) Microsoft relies on the fact that KMT-style handles can be imported, which requires a change on the wine side to make KMT handles distinct from NT handles, which isn't in bleeding-edge yet.


libs/vkd3d/device.c
```diff
@@ -4311,18 +4312,131 @@ static HRESULT STDMETHODCALLTYPE d3d12_device_CreateSharedHandle(d3d12_device_if
         ID3D12DeviceChild *object, const SECURITY_ATTRIBUTES *attributes, DWORD access,
         const WCHAR *name, HANDLE *handle)
 {
-    FIXME("iface %p, object %p, attributes %p, access %#x, name %s, handle %p stub!\n",
+    struct d3d12_device *device = impl_from_ID3D12Device(iface);
```
Joshua-Ashton, line 4315: Merry Christmas!

libs/vkd3d/device.c
```diff
@@ -4311,18 +4312,131 @@ static HRESULT STDMETHODCALLTYPE d3d12_device_CreateSharedHandle(d3d12_device_if
...
 
+    if (IsEqualGUID(riid, &IID_ID3D12Resource))
+    {
+        struct DxvkSharedTextureMetadata metadata;
+        D3D12_RESOURCE_DESC1 desc;
+        D3D12_HEAP_PROPERTIES heap_props;
+        struct d3d12_resource *resource;
+        bool kmt_handle = false;
+
+        if ((ULONG_PTR)handle & 0x40000000 && ((ULONG_PTR)handle - 2) % 4 == 0)
```
Joshua-Ashton, line 4397: Split this out to a helper with a descriptive name, please.

libs/vkd3d/device.c
```diff
@@ -4311,18 +4312,131 @@ static HRESULT STDMETHODCALLTYPE d3d12_device_CreateSharedHandle(d3d12_device_if
...
+            desc.Flags |= D3D12_RESOURCE_FLAG_ALLOW_RENDER_TARGET;
+        desc.SamplerFeedbackMipRegion.Width = 0;
+        desc.SamplerFeedbackMipRegion.Height = 0;
+        desc.SamplerFeedbackMipRegion.Depth = 0;
+
+        heap_props.Type = D3D12_HEAP_TYPE_DEFAULT;
+        heap_props.CPUPageProperty = D3D12_CPU_PAGE_PROPERTY_UNKNOWN;
+        heap_props.MemoryPoolPreference = D3D12_MEMORY_POOL_UNKNOWN;
+        heap_props.CreationNodeMask = 0;
+        heap_props.VisibleNodeMask = 0; // ?
```
Joshua-Ashton, line 4446: What is your question? Looks fine to me.

also no // :frog:

libs/vkd3d/shared_metadata.c
```diff
@@ -0,0 +1,51 @@
+#define VKD3D_DBG_CHANNEL VKD3D_DBG_CHANNEL_API
+#include "vkd3d_private.h"
+
+#include "winioctl.h"
+
+#define IOCTL_SHARED_GPU_RESOURCE_SET_METADATA           CTL_CODE(FILE_DEVICE_VIDEO, 4, METHOD_BUFFERED, FILE_WRITE_ACCESS)
```
Joshua-Ashton, line 6: Put these IOCTL_SHARED_GPU_RESOURCE_SET_METADATA defines together

libs/vkd3d/shared_metadata.c
```diff
@@ -0,0 +1,51 @@
...
+    return DeviceIoControl(handle, IOCTL_SHARED_GPU_RESOURCE_SET_METADATA, buf, buf_size, NULL, 0, &ret_size, NULL);
+}
+
+#define IOCTL_SHARED_GPU_RESOURCE_GET_METADATA           CTL_CODE(FILE_DEVICE_VIDEO, 5, METHOD_BUFFERED, FILE_READ_ACCESS)
+
+bool vkd3d_get_shared_metadata(HANDLE handle, void *buf, uint32_t buf_size, uint32_t *metadata_size)
+{
+    DWORD ret_size;
+
+    bool ret = DeviceIoControl(handle, IOCTL_SHARED_GPU_RESOURCE_GET_METADATA, NULL, 0, buf, buf_size, &ret_size, NULL);
```
Joshua-Ashton, line 21: Why does ret_size exist? `lpBytesReturned` is optional, just pass metadata_size directly.

libs/vkd3d/shared_metadata.c
```diff
@@ -0,0 +1,51 @@
...
+
+#include "winioctl.h"
+
+#define IOCTL_SHARED_GPU_RESOURCE_SET_METADATA           CTL_CODE(FILE_DEVICE_VIDEO, 4, METHOD_BUFFERED, FILE_WRITE_ACCESS)
+
+bool vkd3d_set_shared_metadata(HANDLE handle, void *buf, uint32_t buf_size)
+{
+    DWORD ret_size;
+
+    return DeviceIoControl(handle, IOCTL_SHARED_GPU_RESOURCE_SET_METADATA, buf, buf_size, NULL, 0, &ret_size, NULL);
```
Joshua-Ashton, line 12: lpBytesReturned is optional, no need for `ret_size`

libs/vkd3d/shared_metadata.c
```diff
@@ -0,0 +1,51 @@
...
+}
+
+#define IOCTL_SHARED_GPU_RESOURCE_OPEN                   CTL_CODE(FILE_DEVICE_VIDEO, 1, METHOD_BUFFERED, FILE_WRITE_ACCESS)
+
+HANDLE vkd3d_open_kmt_handle(HANDLE kmt_handle)
+{
+    struct
+    {
+        unsigned int kmt_handle;
+        WCHAR name[1];
```
Joshua-Ashton, line 36: This `WCHAR name[1]` needs some explaining, I don't get it. Probably wrong(?) Did you mean to do like a VLA or something??

Joshua-Ashton (changes requested):
Reviewed the main shared resource stuff.

I am not familiar enough with the D3D12 fence impl + shared sync requirements to give a solid review of that part yet.

libs/vkd3d/shared_metadata.c
Guy1524: The same IOCTL is also used for opening named shared resources, when kmt_handle is 0.  The pattern of a 1 sized array is a pattern used in the wineserver interface indicating that the allocator can instead pass in a dynamically allocated structure with more string bytes.

libs/vkd3d/device.c
Guy1524: Ah yeah forgot to remove that.  At first I read it wrong and thought it had something to do with what command queues the resource could be used in.

libs/vkd3d/shared_metadata.c
Joshua-Ashton: Should put that in a comment.

libs/vkd3d/shared_metadata.c
Guy1524: The reason why the parameter is marked as optional is for use with lpOverlapped (async IO).  When going through the regular path, the parameter is required for some reason.

https://docs.microsoft.com/en-us/windows/win32/api/ioapiset/nf-ioapiset-deviceiocontrol
"If lpOverlapped is NULL, lpBytesReturned cannot be NULL. Even when an operation returns no output data and lpOutBuffer is NULL, DeviceIoControl makes use of lpBytesReturned. After such an operation, the value of lpBytesReturned is meaningless."

libs/vkd3d/shared_metadata.c
Guy1524: Same as above.

libs/vkd3d/shared_metadata.c
Joshua-Ashton: Thanks Microsoft

libs/vkd3d/device.c
```diff
@@ -4311,18 +4312,136 @@ static HRESULT STDMETHODCALLTYPE d3d12_device_CreateSharedHandle(d3d12_device_if
...
+        return vr ? E_FAIL : S_OK;
+    }
+
+    FIXME("Creating shared handle for type of object %p unsupported\n", object);
     return E_NOTIMPL;
 }
 
+static inline bool handle_is_kmt_style(HANDLE handle)
+{
+    return (ULONG_PTR)handle & 0x40000000 && ((ULONG_PTR)handle - 2) % 4 == 0;
```
HansKristian-Work, line 4382: This seems very spooky. Any documentation we can reference?

Also, I prefer `(a & b) && c`. I don't want to second guess operator precedence rules when reading bit fiddling code.

libs/vkd3d/device.c
```diff
@@ -4311,18 +4312,136 @@ static HRESULT STDMETHODCALLTYPE d3d12_device_CreateSharedHandle(d3d12_device_if
...
         HANDLE handle, REFIID riid, void **object)
 {
-    FIXME("iface %p, handle %p, riid %s, object %p stub!\n",
+    struct d3d12_device *device = impl_from_ID3D12Device(iface);
+    HRESULT hr;
+
+    TRACE("iface %p, handle %p, riid %s, object %p\n",
             iface, handle, debugstr_guid(riid), object);
 
+    if (IsEqualGUID(riid, &IID_ID3D12Resource))
```
HansKristian-Work, line 4394: What about ID3D12Resource1?

libs/vkd3d/shared_fence.c
```diff
@@ -0,0 +1,149 @@
+#include "shared_fence.h"
```
HansKristian-Work, line 1: Missing copyright header.

libs/vkd3d/command.c
```diff
@@ -10799,6 +10877,14 @@ static HRESULT STDMETHODCALLTYPE d3d12_command_queue_Wait(ID3D12CommandQueue *if
     sub.type = VKD3D_SUBMISSION_WAIT;
     sub.wait.fence = fence;
     sub.wait.value = value;
+    sub.wait.handle = NULL;
+
+    if (fence->is_shared)
+    {
+        sub.wait.handle = CreateEventW(NULL, TRUE, FALSE, NULL);
+        shared_fence_wait(fence->shared.fence, sub.wait.handle, value);
```
HansKristian-Work, line 10885: shared_fence_wait() implies that this is a blocking wait. Might be clearer if it's called `fence_add_waiter` or something.

libs/vkd3d/command.c
```diff
@@ -10799,6 +10877,14 @@ static HRESULT STDMETHODCALLTYPE d3d12_command_queue_Wait(ID3D12CommandQueue *if
     sub.type = VKD3D_SUBMISSION_WAIT;
     sub.wait.fence = fence;
     sub.wait.value = value;
+    sub.wait.handle = NULL;
+
+    if (fence->is_shared)
+    {
+        sub.wait.handle = CreateEventW(NULL, TRUE, FALSE, NULL);
```
HansKristian-Work, line 10884: Any reason for this to be manual-reset?

libs/vkd3d/shared_fence.c
```diff
@@ -0,0 +1,149 @@
...
+        return false;
+    }
+
+    if (handle != NULL && handle != INVALID_HANDLE_VALUE)
+    {
+        LeaveCriticalSection(fence->cs);
+        return true;
+    }
+
+    while (!waiter->signaled)
```
HansKristian-Work, line 108: Is this the path for when we have SetEventOnCompletion(NULL) which is blocking? Would be easier to understand the control flow if this was commented.

libs/vkd3d/shared_fence.c
```diff
@@ -0,0 +1,149 @@
...
+                waiter->waiting = false;
+            }
+            else
+            {
+                waiter->signaled = true;
+            }
+        }
+    }
+
+    WakeAllConditionVariable(fence->cv);
```
HansKristian-Work, line 145: Should only wake if we actually signalled something.

libs/vkd3d/shared_fence.c
```diff
@@ -0,0 +1,149 @@
...
+    fence->value = value;
+    for (i = 0; i < MAX_WAIT_COUNT; i++)
+    {
+        struct waiter_thread *waiter = &fence->waiters[i];
+
+        if (waiter->value <= value)
+        {
+            if (waiter->handle != NULL && waiter->handle != INVALID_HANDLE_VALUE)
+            {
+                SetEvent(waiter->handle);
```
HansKristian-Work, line 135: The handle should be cleared. We should not signal multiple times.

libs/vkd3d/shared_fence.c
```diff
@@ -0,0 +1,149 @@
...
+                waiter->waiting = false;
+            }
+            else
+            {
+                waiter->signaled = true;
```
HansKristian-Work, line 140: Do we need waiting and signaled? Seems like we'd only need one?

libs/vkd3d/shared_fence.c
```diff
@@ -0,0 +1,149 @@
...
+};
+
+HANDLE create_shared_fence(UINT64 initial_value)
+{
+    HANDLE mapping = CreateFileMappingW(INVALID_HANDLE_VALUE, NULL, PAGE_READWRITE | SEC_COMMIT, 0, sizeof(struct shared_fence), NULL);
+    struct shared_fence *fence = open_shared_fence(mapping);
+
+    fence->pid = GetCurrentProcessId();
+    fence->value = initial_value;
+    fence->cs = malloc(sizeof(*fence->cs));
```
HansKristian-Work, line 32: Is this a leak?

HansKristian-Work (changes requested):
Did a first pass over the PR.

libs/vkd3d/device.c
Guy1524: This is just going off what I saw when looking at the HANDLE values on Windows.  The ability to import KMT handles here is unfortunately undocumented, as far as I could tell.

libs/vkd3d/device.c
```diff
@@ -4311,18 +4312,136 @@ static HRESULT STDMETHODCALLTYPE d3d12_device_CreateSharedHandle(d3d12_device_if
...
+        switch (metadata.TextureLayout)
+        {
+            case D3D11_TEXTURE_LAYOUT_UNDEFINED: desc.Layout = D3D12_TEXTURE_LAYOUT_UNKNOWN; break;
+            case D3D11_TEXTURE_LAYOUT_ROW_MAJOR: desc.Layout = D3D12_TEXTURE_LAYOUT_ROW_MAJOR; break;
+            case D3D11_TEXTURE_LAYOUT_64K_STANDARD_SWIZZLE: desc.Layout = D3D12_TEXTURE_LAYOUT_64KB_STANDARD_SWIZZLE; break;
+            default: desc.Layout = D3D12_TEXTURE_LAYOUT_UNKNOWN;
+        }
+        desc.Flags = D3D12_RESOURCE_FLAG_ALLOW_SIMULTANEOUS_ACCESS;
+        if (metadata.BindFlags & D3D11_BIND_RENDER_TARGET)
+            desc.Flags |= D3D12_RESOURCE_FLAG_ALLOW_RENDER_TARGET;
```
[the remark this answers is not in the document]
Guy1524: Right, missed that.  Don't see any reason why it wouldn't be, I'll add the check.

libs/vkd3d/device.c
Guy1524: Yeah just forgot, I'll add that

libs/vkd3d/resource.c
```diff
@@ -2732,10 +2746,32 @@ HRESULT d3d12_resource_create_committed(struct d3d12_device *device, const D3D12
         else
             allocate_info.heap_flags |= D3D12_HEAP_FLAG_ALLOW_ONLY_NON_RT_DS_TEXTURES;
 
+        if (heap_flags & D3D12_HEAP_FLAG_SHARED)
+        {
+            use_dedicated_allocation = true;
+
+            if (shared_handle && shared_handle != INVALID_HANDLE_VALUE)
```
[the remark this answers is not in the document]
Guy1524: Yeah for legacy reasons both NULL and INVALID_HANDLE_VALUE are used as invalid handle markers.  https://devblogs.microsoft.com/oldnewthing/20040302-00/?p=40443

libs/vkd3d/shared_fence.c
Guy1524: closing conversation with new shared fence implementation

libs/vkd3d/device.c
Guy1524: Actually, looking at the documentation for ::OpenSharedHandle, it only mentions the ability to pass in the GUIDs for the base interfaces, I guess if it matters we could add a test to see whether this was just an oversight in the documentation.

https://docs.microsoft.com/en-us/windows/win32/api/d3d12/nf-d3d12-id3d12device-opensharedhandle

libs/vkd3d/device.c
```diff
@@ -4400,18 +4401,142 @@ static HRESULT STDMETHODCALLTYPE d3d12_device_CreateSharedHandle(d3d12_device_if
...
+        if (metadata.BindFlags & D3D11_BIND_DEPTH_STENCIL)
+            desc.Flags |= D3D12_RESOURCE_FLAG_ALLOW_DEPTH_STENCIL;
+        if (metadata.BindFlags & D3D11_BIND_UNORDERED_ACCESS)
+            desc.Flags |= D3D12_RESOURCE_FLAG_ALLOW_UNORDERED_ACCESS;
+        if (!(metadata.BindFlags & D3D11_BIND_SHADER_RESOURCE))
```
HansKristian-Work, line 4551: DENY_SHADER_RESOURCE is only allowed for ALLOW_DEPTH_STENCIL.

libs/vkd3d/device.c
HansKristian-Work: We can always fix it later if need be.

libs/vkd3d/resource.c
```diff
@@ -531,7 +531,7 @@ static HRESULT vkd3d_get_image_create_info(struct d3d12_device *device,
         external_info->pNext = NULL;
         external_info->handleTypes = VK_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_WIN32_BIT;
 
-        image_info->pNext = &external_info;
+        image_info->pNext = external_info;
```
HansKristian-Work, line 534: If this is a bugfix of the previous commit it should be fixed there.

HansKristian-Work (changes requested):
The resource sharing PR should be made standalone and build without Win32 needs to be fixed. We just won't support shared resources on native, that's fine.

libs/vkd3d/device.c
```diff
@@ -4400,18 +4403,214 @@ static HRESULT STDMETHODCALLTYPE d3d12_device_CreateSharedHandle(d3d12_device_if
...
+        {
+            WARN("Failed to get ID3D12Resource shared handle metadata.\n");
+            if (kmt_handle)
+                CloseHandle(handle);
+
+            *object = NULL;
+            return E_INVALIDARG;
+        }
+
+        desc.Dimension = D3D12_RESOURCE_DIMENSION_TEXTURE2D;
```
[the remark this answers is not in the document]
Guy1524: Unfortunately I didn't include a versioning mechanism for the shared metadata.  As far as I can see though, the D3D11 doc specifies that only 2D non mipmapable textures may be shared, but I'm not sure if this also applies for D3D12.  As far as the D3D12Heap goes, it's in a whole different domain so we can just use a completely different metadata structure.  (Which of course does preclude clean error checking for a type mismatch).

libs/vkd3d/device.c
Guy1524: Ah okay, removed the check entirely then.  Should I instead only set DENY_SHADER_RESOURCE when it's a depth stencil?

Guy1524:
Per request I moved the shared fence code out of this PR.

libs/vkd3d/device.c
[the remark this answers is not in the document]
HansKristian-Work: I suppose it would be possible to use a different DeviceIoControl thing to query for that?

libs/vkd3d/memory.c
```diff
@@ -327,23 +327,25 @@ static HRESULT vkd3d_import_host_memory(struct d3d12_device *device, void *host_
...
-    HRESULT hr;
+    HRESULT hr = S_OK;
 
     import_info.sType = VK_STRUCTURE_TYPE_IMPORT_MEMORY_HOST_POINTER_INFO_EXT;
     import_info.pNext = pNext;
     import_info.handleType = VK_EXTERNAL_MEMORY_HANDLE_TYPE_HOST_ALLOCATION_BIT_EXT;
     import_info.pHostPointer = host_address;
 
-    if (FAILED(hr = vkd3d_try_allocate_device_memory(device, size,
+    if (vkd3d_config_flags & VKD3D_CONFIG_FLAG_USE_HOST_IMPORT_FALLBACK ||
```
HansKristian-Work, line 337: `if ((vkd3d_config_flags & VKD3D_CONFIG_FLAG_USE_HOST_IMPORT_FALLBACK) || ...`

libs/vkd3d/resource.c
```diff
@@ -522,12 +524,22 @@ static HRESULT vkd3d_get_image_create_info(struct d3d12_device *device,
     image_info->sType = VK_STRUCTURE_TYPE_IMAGE_CREATE_INFO;
     image_info->pNext = NULL;
     image_info->flags = 0;
+
+    if (resource && (resource->heap_flags & D3D12_HEAP_FLAG_SHARED))
```
HansKristian-Work, line 528: Should validate here that we can only deal with simple 2D textures, otherwise just fail the call with E_NOTIMPL.

libs/vkd3d/device.c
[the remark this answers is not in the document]
Guy1524: Yeah sure, what kind of interface would you be thinking of, just an IOCTL to get and set a version integer value, and then just default to the current structure if getting the version returns an error?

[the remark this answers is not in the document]
doitsujin: We could always introduce a new metadata format and identify the old one by struct size or something, or by setting the first field (Width) to 0 or whatever.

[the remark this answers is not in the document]
HansKristian-Work: Ok, I guess we have a way forward if need be.

HansKristian-Work (approved):
LGTM

include/vkd3d.h
```diff
@@ -91,6 +91,7 @@ extern "C" {
 #define VKD3D_CONFIG_FLAG_ZERO_MEMORY_WORKAROUNDS_COMMITTED_BUFFER_UAV (1ull << 29)
 #define VKD3D_CONFIG_FLAG_ALLOW_SBT_COLLECTION (1ull << 30)
 #define VKD3D_CONFIG_FLAG_FORCE_NATIVE_FP16 (1ull << 31)
+#define VKD3D_CONFIG_FLAG_USE_HOST_IMPORT_FALLBACK (1ull << 32)
```
HansKristian-Work, line 94: Commit needs signed off by.

HansKristian-Work (changes requested):
Actually, minor commit nit, otherwise fine.

libs/vkd3d/resource.c
```diff
@@ -2754,10 +2771,36 @@ HRESULT d3d12_resource_create_committed(struct d3d12_device *device, const D3D12
...
+                import_info.handle = shared_handle;
+                import_info.name = NULL;
+                allocate_info.pNext = &import_info;
+            }
+            else
+            {
+                export_info.sType = VK_STRUCTURE_TYPE_EXPORT_MEMORY_ALLOCATE_INFO;
+                export_info.pNext = allocate_info.pNext;
+                export_info.handleTypes = VK_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_WIN32_BIT;
+                allocate_info.pNext = &export_info;
```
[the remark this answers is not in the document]
Guy1524: Nope, thanks for catching

libs/vkd3d/resource.c
Guy1524: I decided to instead put the logic checking this in ::CreateSharedHandle where we set the metadata, as for all we know it could be valid to shared non 2D textures within D3D12 and we just need an updated metadata structure.

HansKristian-Work (approved):
LGTM

PR #1072, opened 2022-04-18
build: Support custom Wine versions

forkbomb9:
This allows people to use custom wine versions by exporting the WINE env variable.

If undefined, it uses the `$PATH`'s `wine` using `command -v` instead of `which`.
This is POSIX compliant while `which` isn't, Debian 11 deprecated which.

The changes are the same as https://github.com/doitsujin/dxvk/pull/2617

### DIFF
--- a/setup_vkd3d_proton.sh
+++ b/setup_vkd3d_proton.sh
@@ -48,36 +48,36 @@ export WINEDEBUG=-all
 # wine gecko and mono
 export WINEDLLOVERRIDES="mscoree,mshtml="
 
-wine="wine"
-wine64="wine64"
-wineboot="wineboot"
+# support custom Wine versions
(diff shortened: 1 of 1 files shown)


SveSop:
1. Patch is tagged "build: ".. but is it? Maybe "setup: " would be better?
2. Why is it needed? "Easier than using $PATH variable!"... bit IS it really?
3. Who really cares what i think :frog:

forkbomb9:
1. The latest commit touching this file has the same tag (https://github.com/HansKristian-Work/vkd3d-proton/commit/1267b2a985e431031220737b653accc09a059c0e)
2. Yes, for a few custom scripts, exporting PATH is more messy than just pointing WINE to the custom wine

anyway dxvk rejected this patch, so this one's probably going to also 🤷🏻

Issue #1069, opened 2022-04-09
DCC on AMD HW is blocked due to aliasing images with incompatible format


Venemo:
I found this peculiar case when investigating *Ghostwire: Tokyo* game performance on RADV. We discussed this with Hans-Kristian, and here I'm providing a summary of that discussion.

What is happening?
--------------------------

The D3D12 app can use DCC, but the same app running on RADV through VKD3D-Proton can't use DCC with `DXGI_FORMAT_R16G16B16A16_FLOAT`.

As you can [see in the code here](https://github.com/HansKristian-Work/vkd3d-proton/blob/master/libs/vkd3d/utils.c#L191) the image is also accessed through the `DXGI_FORMAT_R16G16B16A16_UINT` format which is not compatible as far as DCC is concerned.

Why?
------

This is because VKD3D-Proton needs to ensure that the image works with the `ClearUnorderedAccessViewUint` function. VKD3D-Proton's implementation of this function uses a shader that writes the image through a uint format. This is done instead of reinterpreting the uint to float, in order to preserve NaN values.

What could be done to solve this?
---------------------------

The implementation of `ClearUnorderedAccessViewUint` could be changed to distinguish two cases:

1. When the uint reinterpreted as float is not NaN, use a shader that accesses the image in its proper format
2. When the uint is NaN, fill a buffer with the NaN value and use a buffer to image copy

We expect that case 1 is the most often used, and it's okay to have case 2 slower.

Note, a few things are also unclear:

* Does D3D12 expect the implementation to preserve these NaN values or not? If not, then case 2 above is not needed.
* Does a float image store with NaN preserve the NaN or not? If it does, then once again case 2 may not be necessary.

doitsujin:
`ClearUnorderedAccessViewUINT` is required to preserve the exact bit pattern. I don't know what a shader-based implementation using float stores would look like which would guarantee this without ever producing rounding errors during conversion.

I wasn't aware that UINT disables compression on FLOAT images even in bit-compatible formats, this is extremely annoying, and DXVK is affected in the same way. Moreover, **all** games that use TYPELESS formats are affected as well since those explicitly allow the application to create UINT views, and this is very common, even though the majority of games will never create such views.

Is there a list of DCC-incompatible formats somewhere? Especially on the DXVK side we may need to completely rethink our approach to implementing both typeless formats and UAV clears.

HansKristian-Work:
[...]

PR #1077, opened 2022-04-20
ClearUnorderedAccessViewUint rework (again)

doitsujin:
Addresses #1069. TL;DR instead of only supporting 32-bit formats for the fallback path, we now use a texel buffer view with the given UINT format to clear the scratch buffer to the desired clear value, then perform a buffer-to-image copy, allowing us to exclude the UINT format from the image's format compatibility list unless otherwise required.


libs/vkd3d/command.c
```diff
@@ -8307,21 +8307,30 @@ static void d3d12_command_list_clear_uav(struct d3d12_command_list *list,
 
 static void d3d12_command_list_clear_uav_with_copy(struct d3d12_command_list *list,
         const struct d3d12_desc_split *d, struct d3d12_resource *resource,
-        const struct vkd3d_clear_uav_info *args, uint32_t clear_value,
-        UINT rect_count, const D3D12_RECT *rects)
+        const struct vkd3d_clear_uav_info *args, const VkClearColorValue* clear_value,
```
HansKristian-Work, line 8310: ` *clear_value` nit.

libs/vkd3d/command.c
```diff
@@ -8675,7 +8687,7 @@ static void STDMETHODCALLTYPE d3d12_command_list_ClearUnorderedAccessViewUint(d3
                 /* If the clear color is 0, we can safely use the existing view to perform the
                  * clear since the bit pattern will not change. Otherwise, fill a scratch buffer
                  * with the packed clear value and perform a buffer to image copy. */
-                if (color.uint32[0])
+                if (color.uint32[0] || color.uint32[1] || color.uint32[2] || color.uint32[3])
```
HansKristian-Work, line 8690: Do we clear unused components to 0 somewhere to avoid hitting fallback due to garbage values?

HansKristian-Work: mask_uint_clear_color should probably clear out unused components?

libs/vkd3d/command.c
```diff
@@ -8550,6 +8543,25 @@ static const struct vkd3d_format *vkd3d_clear_uav_find_uint_format(struct d3d12_
...
+
+    if (resource_format->vk_format == uint_format->vk_format)
+        return true;
+
+    compat = &device->format_compatibility_lists[resource_format->dxgi_format];
+
+    for (i = 0; i < compat->format_count; i++)
+    {
+        if (compat->vk_formats[i] == uint_format->vk_format)
+            return true;
```
HansKristian-Work, line 8559: I assume TYPELESS will always pass this check?

HansKristian-Work (changes requested):
Small nits, otherwise LGTM.

libs/vkd3d/command.c

libs/vkd3d/command.c
doitsujin: Yes, all typeless formats have a corresponding UINT format.

libs/vkd3d/command.c
doitsujin: Done.

HansKristian-Work (approved):
LGTM, thanks!

PR #1078, opened 2022-04-21
Use VK_KHR_maintenance4.

doitsujin:
And make it a hard requirement since driver support overlaps fairly well with KHR_dynamic_rendering. Removes the need to create a whole bunch of dummy resources.


HansKristian-Work (approved):
LGTM, also agree on README, but could be separate PR.

PR #1083, opened 2022-04-24
vkd3d: Fix queue creation for queue family -1.

HansKristian-Work:
Fixes validation error on Intel where we are trying to create
CONCURRENT family with {0, -1}.

Signed-off-by: Hans-Kristian Arntzen <maister@archlinux.us>


doitsujin:
What's the issue exactly and why does it break on Intel?

HansKristian-Work:
The SPARSE QUEUE is detected with family -1, but the code which adds vkd3d_queue goes ahead and ends up pushing a queue family of -1 to the list. So we end up with CONCURRENT queue with family indices {0, -1}, which is nonsensical.

doitsujin (approved):

PR #1090, opened 2022-05-03
vkd3d: Add VKD3D_CONFIG option to force raw VA CBV descriptors.

HansKristian-Work:
For certain ExecuteIndirect() uses, we're forced to use this path
since we have no way to update push descriptors indirectly yet.

Also, useful for testing performance deltas easily on NV while we're
trying to figure out what is going on w.r.t. codegen.


HansKristian-Work:
Obsoleted.

PR #1091, opened 2022-05-03
Draft: vkd3d: Add descriptor_only_hvv for non-reBAR performance improvements.

ishitatsuyuki:
(Copied from updated README)

This option is the same as `no_upload_hvv`, except it allows vkd3d's internal descriptors heaps to be placed
on HVV. Useful for small BAR systems (systems without resizable BAR) to prevent UPLOAD heaps from evicting the
frequently read descriptor heap.

### DIFF
--- a/libs/vkd3d/memory.c
+++ b/libs/vkd3d/memory.c
@@ -77,7 +77,7 @@ static HRESULT vkd3d_select_memory_flags(struct d3d12_device *device, const D3D1
             *type_flags = VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT | VK_MEMORY_PROPERTY_HOST_COHERENT_BIT;
             if (vkd3d_config_flags & VKD3D_CONFIG_FLAG_FORCE_HOST_CACHED)
                 *type_flags |= VK_MEMORY_PROPERTY_HOST_CACHED_BIT;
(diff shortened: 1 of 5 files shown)


libs/vkd3d/memory.c
```diff
@@ -189,7 +189,8 @@ static HRESULT vkd3d_try_allocate_device_memory(struct d3d12_device *device,
     /* buffer_mask / sampled_mask etc will generally take care of this,
      * but for certain fallback scenarios where we select other memory
      * types, we need to mask here as well. */
-    type_mask &= device->memory_info.global_mask;
+    if (!ignore_hvv_restrictions)
+        type_mask &= device->memory_info.global_mask;
```
HansKristian-Work, line 193: Seeing this now, I think the bool here is pretty goofy and we should try to avoid it.

Thinking a bit more, I'm pondering if the global_mask is even meaningful here. Somehow I don't think it's actually useful here. I'll think a bit about if we can get rid of the global mask entirely. That would allow us to just use different memory_properties and everything will work fine.

HansKristian-Work:
Finally got around to testing this on my system and in HZD, 100% of the perf gains from ReBAR were from upload HVV sadly. Using descriptor_only_hvv regresses performance quite significantly for me. :\

PR #1092, opened 2022-05-04
Refactor away global_mask from memory_info.

HansKristian-Work:
The global mask was added with the intention to block PCI-e BAR memory
types from being used, however, this ends up being somewhat clumsy when
we want to allow PCI-e BAR for specific allocations.

Refactor the memory allocation scheme to never end up in a situation
where we accidentally allocate PCI-e BAR memory. We only attempt one
memory type. This makes is so that pure DEVICE cannot fallback to BAR,
and pure HOST allocations (if they somehow fail) cannot fallback to BAR.

The first eligible memory type is always selected.

Pre-work for https://github.com/HansKristian-Work/vkd3d-proton/pull/1091.

### DIFF
--- a/libs/vkd3d/memory.c
+++ b/libs/vkd3d/memory.c
@@ -592,8 +592,7 @@ static HRESULT vkd3d_memory_allocation_init(struct vkd3d_memory_allocation *allo
                 0, VK_WHOLE_SIZE, 0, &allocation->cpu_address))))
         {
             ERR("Failed to map memory, vr %d.\n", vr);
-            vkd3d_memory_allocation_free(allocation, device, allocator);
-            return hresult_from_vk_result(vr);
+            goto out_free_allocation;
         }
     }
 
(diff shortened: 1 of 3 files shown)


libs/vkd3d/memory.c
```diff
@@ -186,87 +185,87 @@ static HRESULT vkd3d_try_allocate_device_memory(struct d3d12_device *device,
...
+        return E_OUTOFMEMORY;
+    }
 
-        if (budget_sensitive)
+    /* Once we have found a suitable memory type, only attempt to allocate that memory type.
+     * This avoids some problems by design:
+     * - If we want to allocate DEVICE_LOCAL memory, we don't try to fallback to PCI-e BAR memory by mistake.
+     * - If we want to allocate system memory, we don't try to fallback to PCI-e BAR memory by mistake.
+     * - There is no reasonable scenario where we can expect one memory type to fail, and another memory type
+     *   with more memory property bits set to pass. This makes use of the rule where memory types which are a super-set
```
ishitatsuyuki, line 214: Didn't know that Vulkan spec actually had an ordering guarantee on this. Nice trick!
(Maybe it could cite the section or URL, but the meaning is clear either way)

ishitatsuyuki (approved):
`Reviewed-by: Tatsuyuki Ishi <ishitatsuyuki@gmail.com>`

Thanks for the cleanup. I will rebase #1091 soon.

HansKristian-Work:
Already merged.

PR #1100, opened 2022-05-23
Improve WSI robustness for back-to-back failures.

HansKristian-Work:


libs/vkd3d/swapchain.c
```diff
@@ -48,6 +48,28 @@
...
+{
+    unsigned int value = rand() & 15;
+    return value >= 8;
+}
+
+static bool VKD3D_SPURIOUS_OUT_OF_DATE(void)
+{
+    unsigned int value = rand() & 15;
+    return value >= 8;
+}
```
doitsujin, line 70: not a huge fan of the ifdef soup down below; if we actually want to ship this code, maybe we could do this in present/acquire wrappers?

HansKristian-Work: It might be better to just leave that commit out for now. It's trivial to re-add later. Was mostly to demonstrate what I did to test things.

HansKristian-Work: Removed commit from PR.

PR #1103, opened 2022-05-25
Enable and use KHR_fragment_shader_barycentric.

HansKristian-Work:


libs/vkd3d/vkd3d_private.h
```diff
@@ -3063,6 +3063,7 @@ struct vkd3d_physical_device_info
     VkPhysicalDeviceSeparateDepthStencilLayoutsFeaturesKHR separate_depth_stencil_layout_features;
     VkPhysicalDeviceShaderIntegerDotProductFeaturesKHR shader_integer_dot_product_features;
     VkPhysicalDeviceFragmentShaderBarycentricFeaturesNV barycentric_features_nv;
+    VkPhysicalDeviceFragmentShaderBarycentricFeaturesKHR barycentric_features_khr;
```
DadSchoorse, line 3066: These two structs are aliases of each other, so imo it would be better to just have the KHR one.

DadSchoorse: Ah wait, the OpExtension string is still different.

HansKristian-Work: Yes, we need to check later if we enabled KHR feature, so it's more straight forward to do it like this.

PR #1112, opened 2022-06-01
tests: Fix -Wstringop-overread warnings.

DadSchoorse:


K0bin:
What about the warnings in the clear test?

DadSchoorse:
Ah, I missed those after my clean build and then I thought I fixed all of them because the file wasn't recompiled. :frog:

PR #1117, opened 2022-06-06
Optimize consecutive CopyTextureRegion by batching the barriers

ishitatsuyuki:
The main idea is to introduce a "command list batch" system where it compares for the type of the previous and current command for every non render-pass command, and in the case there are consecutive CopyTextureRegion calls, batch them so we can issue layout transition barriers in bulk instead of interrupting after each copy.

See each commit for more details.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -48,6 +48,11 @@ static void d3d12_command_list_barrier_batch_add_layout_transition(
         struct d3d12_command_list *list,
         struct d3d12_command_list_barrier_batch *batch,
         const VkImageMemoryBarrier *image_barrier);
+static void d3d12_command_list_barrier_batch_add_global_transition(
+        struct d3d12_command_list *list,
+        struct d3d12_command_list_barrier_batch *batch,
+        VkAccessFlags srcAccessMask,
+        VkAccessFlags dstAccessMask);
 
 static uint32_t d3d12_command_list_promote_dsv_resource(struct d3d12_command_list *list,
         struct d3d12_resource *resource, uint32_t plane_optimal_mask);
@@ -6052,13 +6057,12 @@ static bool validate_d3d12_box(const D3D12_BOX *box)
 }
 
 static void d3d12_command_list_transition_image_layout_with_global_memory_barrier(struct d3d12_command_list *list,
+        struct d3d12_command_list_barrier_batch *batch,
         VkImage vk_image, const VkImageSubresourceLayers *vk_subresource,
         VkPipelineStageFlags src_stages, VkAccessFlags src_access, VkImageLayout old_layout,
         VkPipelineStageFlags dst_stages, VkAccessFlags dst_access, VkImageLayout new_layout,
         VkAccessFlags global_src_access, VkAccessFlags global_dst_access)
 {
-    const struct vkd3d_vk_device_procs *vk_procs = &list->device->vk_procs;
-    VkMemoryBarrier vk_global_barrier;
     VkImageMemoryBarrier vk_barrier;
     bool need_global_barrier;
 
@@ -6077,23 +6081,20 @@ static void d3d12_command_list_transition_image_layout_with_global_memory_barrie
 
     if (need_global_barrier)
     {
-        vk_global_barrier.sType = VK_STRUCTURE_TYPE_MEMORY_BARRIER;
-        vk_global_barrier.pNext = NULL;
-        vk_global_barrier.srcAccessMask = global_src_access;
-        vk_global_barrier.dstAccessMask = global_dst_access;
+        d3d12_command_list_barrier_batch_add_global_transition(list, batch, global_src_access, global_dst_access);
     }
-
-    VK_CALL(vkCmdPipelineBarrier(list->vk_command_buffer,
-            src_stages, dst_stages, 0, need_global_barrier, &vk_global_barrier, 0, NULL,
-            1, &vk_barrier));
+    d3d12_command_list_barrier_batch_add_layout_transition(list, batch, &vk_barrier);
+    batch->dst_stage_mask |= dst_stages;
+    batch->src_stage_mask |= src_stages;
 }
 
 static void d3d12_command_list_transition_image_layout(struct d3d12_command_list *list,
+        struct d3d12_command_list_barrier_batch *batch,
         VkImage vk_image, const VkImageSubresourceLayers *vk_subresource,
         VkPipelineStageFlags src_stages, VkAccessFlags src_access, VkImageLayout old_layout,
         VkPipelineStageFlags dst_stages, VkAccessFlags dst_access, VkImageLayout new_layout)
 {
-    d3d12_command_list_transition_image_layout_with_global_memory_barrier(list,
+    d3d12_command_list_transition_image_layout_with_global_memory_barrier(list, batch,
             vk_image, vk_subresource,
             src_stages, src_access, old_layout,
             dst_stages, dst_access, new_layout,
@@ -6210,6 +6211,7 @@ static bool d3d12_command_list_init_copy_texture_region(struct d3d12_command_lis
 }
 
 static void d3d12_command_list_before_copy_texture_region(struct d3d12_command_list *list,
+        struct d3d12_command_list_barrier_batch *batch,
         struct vkd3d_image_copy_info *info)
 {
     VkAccessFlags global_transfer_access;
@@ -6230,7 +6232,7 @@ static void d3d12_command_list_before_copy_texture_region(struct d3d12_command_l
         global_transfer_access = list->tracked_copy_buffer_count ? VK_ACCESS_TRANSFER_WRITE_BIT : 0;
         d3d12_command_list_reset_buffer_copy_tracking(list);
 
-        d3d12_command_list_transition_image_layout_with_global_memory_barrier(list, src_resource->res.vk_image,
+        d3d12_command_list_transition_image_layout_with_global_memory_barrier(list, batch, src_resource->res.vk_image,
                 &info->copy.buffer_image.imageSubresource, VK_PIPELINE_STAGE_TRANSFER_BIT, 0,
                 src_resource->common_layout, VK_PIPELINE_STAGE_TRANSFER_BIT, VK_ACCESS_TRANSFER_READ_BIT,
                 info->src_layout, global_transfer_access, global_transfer_access);
@@ -6239,7 +6241,7 @@ static void d3d12_command_list_before_copy_texture_region(struct d3d12_command_l
     {
         d3d12_command_list_track_resource_usage(list, dst_resource, !info->writes_full_subresource);
 
-        d3d12_command_list_transition_image_layout(list, dst_resource->res.vk_image,
+        d3d12_command_list_transition_image_layout(list, batch, dst_resource->res.vk_image,
                 &info->copy.buffer_image.imageSubresource, VK_PIPELINE_STAGE_TRANSFER_BIT, 0,
                 info->writes_full_subresource ? VK_IMAGE_LAYOUT_UNDEFINED : dst_resource->common_layout,
                 VK_PIPELINE_STAGE_TRANSFER_BIT, VK_ACCESS_TRANSFER_WRITE_BIT, info->dst_layout);
@@ -6251,6 +6253,7 @@ static void d3d12_command_list_before_copy_texture_region(struct d3d12_command_l
 }
 
 static void d3d12_command_list_copy_texture_region(struct d3d12_command_list *list,
+        struct d3d12_command_list_barrier_batch *batch,
         struct vkd3d_image_copy_info *info)
 {
     struct d3d12_resource *dst_resource, *src_resource;
@@ -6278,7 +6281,7 @@ static void d3d12_command_list_copy_texture_region(struct d3d12_command_list *li
 
         VK_CALL(vkCmdCopyImageToBuffer2KHR(list->vk_command_buffer, &copy_info));
 
-        d3d12_command_list_transition_image_layout_with_global_memory_barrier(list, src_resource->res.vk_image,
+        d3d12_command_list_transition_image_layout_with_global_memory_barrier(list, batch, src_resource->res.vk_image,
                 &info->copy.buffer_image.imageSubresource, VK_PIPELINE_STAGE_TRANSFER_BIT, 0,
                 info->src_layout, VK_PIPELINE_STAGE_TRANSFER_BIT, 0, src_resource->common_layout,
                 global_transfer_access, global_transfer_access);
@@ -6297,7 +6300,7 @@ static void d3d12_command_list_copy_texture_region(struct d3d12_command_list *li
 
         VK_CALL(vkCmdCopyBufferToImage2KHR(list->vk_command_buffer, &copy_info));
 
-        d3d12_command_list_transition_image_layout(list, dst_resource->res.vk_image,
+        d3d12_command_list_transition_image_layout(list, batch, dst_resource->res.vk_image,
                 &info->copy.buffer_image.imageSubresource, VK_PIPELINE_STAGE_TRANSFER_BIT,
                 VK_ACCESS_TRANSFER_WRITE_BIT, info->dst_layout, VK_PIPELINE_STAGE_TRANSFER_BIT,
                 0, dst_resource->common_layout);
@@ -6315,6 +6318,8 @@ static void STDMETHODCALLTYPE d3d12_command_list_CopyTextureRegion(d3d12_command
 {
     struct d3d12_command_list *list = impl_from_ID3D12GraphicsCommandList(iface);
     struct vkd3d_image_copy_info copy_info;
+    bool alias;
+    size_t i;
 
     TRACE("iface %p, dst %p, dst_x %u, dst_y %u, dst_z %u, src %p, src_box %p.\n",
             iface, dst, dst_x, dst_y, dst_z, src, src_box);
@@ -6329,6 +6334,49 @@ static void STDMETHODCALLTYPE d3d12_command_list_CopyTextureRegion(d3d12_command
         return;
 
     d3d12_command_list_ensure_transfer_batch(list, copy_info.batch_type);
+    alias = false;
+    for (i = 0; !alias && i < list->transfer_batch.batch_len; i++)
+    {
+        const struct vkd3d_image_copy_info *other_info = &list->transfer_batch.batch[i];
+        const VkImageSubresourceLayers *subres, *other_subres;
+        const struct d3d12_resource *res, *other_res;
+
+        switch (copy_info.batch_type)
+        {
+            case VKD3D_BATCH_TYPE_COPY_BUFFER_TO_IMAGE:
+                /* Test for destination aliasing as D3D12 requires serialization on overlapping copies (WAW hazard). */
+                subres = &copy_info.copy.buffer_image.imageSubresource;
+                other_subres = &other_info->copy.buffer_image.imageSubresource;
+                assert(subres->layerCount == 1 && other_subres->layerCount == 1);
+                alias = copy_info.dst.pResource == other_info->dst.pResource &&
+                         subres->aspectMask == other_subres->aspectMask &&
+                         subres->baseArrayLayer == other_subres->baseArrayLayer &&
+                         subres->mipLevel == other_subres->mipLevel;
+                break;
+            case VKD3D_BATCH_TYPE_COPY_IMAGE_TO_BUFFER:
+                /* Test for destination aliasing as D3D12 requires serialization on overlapping copies (WAW hazard). */
+                /* TODO: Do more granular alias testing or merge this with d3d12_command_list_mark_copy_buffer_write. */
+                res = impl_from_ID3D12Resource(copy_info.dst.pResource);
+                other_res = impl_from_ID3D12Resource(other_info->dst.pResource);
+                /* If either resource is sparse, consider it to alias with anything. */
+                alias = copy_info.dst.pResource == other_info->dst.pResource ||
+                        (res->flags & VKD3D_RESOURCE_RESERVED) || (other_res->flags & VKD3D_RESOURCE_RESERVED);
+                break;
+            case VKD3D_BATCH_TYPE_COPY_IMAGE:
+                /* TODO: Check for alias once we start batching barriers for image-image copies too */
+                break;
+            default:
+                assert(false);
+                return;
+        }
+    }
+    if (alias)
+    {
+        d3d12_command_list_end_transfer_batch(list);
+        /* end_transfer_batch resets the batch_type to NONE, so we need to restore it here. */
+        list->transfer_batch.batch_type = copy_info.batch_type;
+    }
+
     list->transfer_batch.batch[list->transfer_batch.batch_len++] = copy_info;
 
     VKD3D_BREADCRUMB_COMMAND(COPY);
@@ -6410,7 +6458,9 @@ static void STDMETHODCALLTYPE d3d12_command_list_CopyResource(d3d12_command_list
 
 static void d3d12_command_list_end_transfer_batch(struct d3d12_command_list *list)
 {
+    struct d3d12_command_list_barrier_batch barriers;
     size_t i;
+
     switch (list->transfer_batch.batch_type)
     {
         case VKD3D_BATCH_TYPE_NONE:
@@ -6419,12 +6469,14 @@ static void d3d12_command_list_end_transfer_batch(struct d3d12_command_list *lis
         case VKD3D_BATCH_TYPE_COPY_IMAGE_TO_BUFFER:
         case VKD3D_BATCH_TYPE_COPY_IMAGE:
             d3d12_command_list_end_current_render_pass(list, false);
-            /* TODO: once we check for aliasing and it's safe to batch things, transpose the loop. */
+            d3d12_command_list_barrier_batch_init(&barriers);
             for (i = 0; i < list->transfer_batch.batch_len; i++)
-            {
-                d3d12_command_list_before_copy_texture_region(list, &list->transfer_batch.batch[i]);
-                d3d12_command_list_copy_texture_region(list, &list->transfer_batch.batch[i]);
-            }
+                d3d12_command_list_before_copy_texture_region(list, &barriers, &list->transfer_batch.batch[i]);
+            d3d12_command_list_barrier_batch_end(list, &barriers);
+            d3d12_command_list_barrier_batch_init(&barriers);
+            for (i = 0; i < list->transfer_batch.batch_len; i++)
+                d3d12_command_list_copy_texture_region(list, &barriers, &list->transfer_batch.batch[i]);
+            d3d12_command_list_barrier_batch_end(list, &barriers);
             list->transfer_batch.batch_len = 0;
             break;
         default:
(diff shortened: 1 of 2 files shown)


libs/vkd3d/command.c
```diff
@@ -6234,26 +6234,28 @@ static void STDMETHODCALLTYPE d3d12_command_list_before_copy_texture_region(stru
...
-        d3d12_command_list_transition_image_layout_with_global_memory_barrier(list, batch, src_resource->res.vk_image,
-                                                                              &info->buffer_image_copy.imageSubresource,
-                                                                              VK_PIPELINE_STAGE_TRANSFER_BIT,
-                                                                              0, src_resource->common_layout,
-                                                                              VK_PIPELINE_STAGE_TRANSFER_BIT,
-                                                                              VK_ACCESS_TRANSFER_READ_BIT,
-                                                                              info->vk_layout,
-                                                                              global_transfer_access,
-                                                                              global_transfer_access);
+        if (!info->transition_coalesced)
```
HansKristian-Work, line 6241: Multi-line branch should be in braces.

libs/vkd3d/command.c
```diff
@@ -6328,6 +6332,8 @@ static void STDMETHODCALLTYPE d3d12_command_list_CopyTextureRegion(d3d12_command
 {
     struct d3d12_command_list *list = impl_from_ID3D12GraphicsCommandList(iface);
     struct vkd3d_image_copy_info copy_info;
+    bool alias;
+    bool format_match;
```
HansKristian-Work, line 6340: Inverse :christmas_tree:

libs/vkd3d/command.c
```diff
@@ -6344,6 +6350,42 @@ static void STDMETHODCALLTYPE d3d12_command_list_CopyTextureRegion(d3d12_command
         return;
 
     d3d12_command_list_ensure_batch(list, copy_info.batch_type);
+    alias = false;
+    format_match = true;
+    for (size_t i = 0; i < list->transfer_batch.ctr_batch_len; i++) {
```
HansKristian-Work, line 6359: This section is hard to reason about. I see it's disabling the alias checking for barriers, but I'm not sure if that's what we want here.

COPY_DEST resource state in D3D12 is kinda broken. Overlapping copies to same subresource must be synchronized implicitly by implementation. So, we'd need to either break batch, or add an internal TRANSFER memory barrier between two copies that overlap to resolve the write-after-write hazard.

libs/vkd3d/command.c
```diff
@@ -32,17 +32,20 @@ static void d3d12_command_queue_add_submission(struct d3d12_command_queue *queue
 static void d3d12_fence_inc_ref(struct d3d12_fence *fence);
 static void d3d12_fence_dec_ref(struct d3d12_fence *fence);
 
-#define MAX_BATCHED_IMAGE_BARRIERS 16
+
 struct d3d12_command_list_barrier_batch
 {
-    VkImageMemoryBarrier vk_image_barriers[MAX_BATCHED_IMAGE_BARRIERS];
+    VkImageMemoryBarrier *vk_image_barriers;
```
HansKristian-Work, line 38: I'm not sure this is a good idea. Do we really expect > 16 batching to gain anything? Heap allocations complicate things. The search algorithms where we check for overlap are quadratic in nature, so this can easily explode if we're not careful to limit it.

libs/vkd3d/command.c
```diff
@@ -7165,15 +7167,15 @@ static void d3d12_command_list_barrier_batch_add_layout_transition(
...
-    /* ResourceBarrier() in D3D12 behaves as if each transition happens in order.
-     * Vulkan memory barriers do not, so if there is a race condition, we need to split
-     * the barrier. */
-    for (i = 0; i < batch->image_barrier_count; i++)
-    {
-        if (vk_image_barrier_overlaps_subresource(image_barrier, &batch->vk_image_barriers[i]))
-        {
-            d3d12_command_list_barrier_batch_end(list, batch);
-            break;
+    if (batch->check_alias) {
```
HansKristian-Work, line 7226: The check_alias is not very obvious to reason about. I can see this being a thing for something like src_resource in a copy. If we're trying to record a common layout -> TRANSFER_SRC_OPTIMAL all the time, we only really need to do that once. It might make more sense to just eliminate redundant barriers instead, e.g. same old/newLayout on same subresource range is just dropped.

For common layout -> TRANSFER_DST_OPTIMAL. If that is recorded multiple times, it means we have overlapped writes, which needs to be resolved as a WAW hazard either way. It might be better to handle that outside when emitting the copy calls either way.

With that in mind, I don't think we'll need a check_alias bool at all.

HansKristian-Work (changes requested):
First pass of the review. The complexity here is fairly large, so it will be much easier for me to review if we go one patch at a time. There's fair amount of style nits that could be fixed first as well.

libs/vkd3d/command.c
ishitatsuyuki: For the old barrier batch the limit is still 16. For the new one, Forza Horizon can issue 100+ copies in a row so I’m using a larger batch size to maximize the benefit. For other games, the CTR batch only goes as long as the number of CTR the game does, and when it actually does lots of copies I’d expect the GPU bound gain to be more important.

As for heap allocation, they’re a little bit concerning but didn’t seem much of a problem in testing.

libs/vkd3d/command.c
```diff
@@ -6089,70 +6090,127 @@ static void d3d12_command_list_transition_image_layout(struct d3d12_command_list
...
-    vk_procs = &list->device->vk_procs;
+    memset(out, 0, sizeof(*out));
+    
+    out->src = *src;
+    out->dst = *dst;
 
     dst_resource = impl_from_ID3D12Resource(dst->pResource);
     src_resource = impl_from_ID3D12Resource(src->pResource);
 
-    d3d12_command_list_track_resource_usage(list, src_resource, true);
```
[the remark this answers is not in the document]
ishitatsuyuki: Ouch, looks like that is the case. I’ll put it back in before_ctr.

libs/vkd3d/command.c
ishitatsuyuki: Good catch, I'll remove the format_match condition.

As for why it has to be done here, it's because when there's an alias, we need to flush not only the barriers but also the copies and the post-barriers as well.

libs/vkd3d/command.c
```diff
@@ -65,6 +65,7 @@ static VkImageLayout d3d12_command_list_get_depth_stencil_resource_layout(const
 static void d3d12_command_list_decay_optimal_dsv_resource(struct d3d12_command_list *list,
         const struct d3d12_resource *resource, uint32_t plane_optimal_mask,
         struct d3d12_command_list_barrier_batch *batch);
+static void d3d12_command_list_end_batch(struct d3d12_command_list *list);
```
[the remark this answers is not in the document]
ishitatsuyuki: The intention is to prevent unnecessary command reordering, for example if an app does the copies in the beginning of the frame, and then does some compute before it draws, we don't want to defer the copies until right before the draw. 

That said, it's also possible that doing the flush in ResourceBarrier() could be sufficient, and I agree with the point on maintainability.

ishitatsuyuki (commented):
Thanks for the review. I will fix the semantical styling issues, but as for the formatting I'd like to fix them with a follow-up reformat commit if you don't mind.

libs/vkd3d/command.c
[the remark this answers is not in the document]
HansKristian-Work: Ending the copy batch in begin_render_pass and dispatch seems fine too. Just want to make sure we don't have to sprinkle it in *everywhere*.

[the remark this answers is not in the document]
HansKristian-Work: Games split their rendering over many command buffers too, so I doubt we should be that worried about extreme overlap.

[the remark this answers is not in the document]
ishitatsuyuki: Sounds good, will do so.

libs/vkd3d/command.c
HansKristian-Work: Maybe we can make a small_vector type optimization. Only allocate from heap if number of elements exceeds the array limit.

HansKristian-Work: However, given that is purely a tuning thing, it might be best to leave that out of this PR. A follow-up can do size tuning for Forza.

ishitatsuyuki: Not sure what you're suggesting here? Do you mean that we should leave the batch size as 16 so it can be in-place allocated?

The limits in this patch is the untouched for the memory barrier batch and unlimited for the newly added transfer batch for Forza (bounded by the 64 transfer batch limit).

HansKristian-Work: I think for this PR, it should be left as-is as an array with 16 elements. We can bump that later once this lands.

ishitatsuyuki: Actually, the patch is implemented in a way that requires unbounded barrier batch size, as it interleaves call to the Vulkan copy commands and addition to the post-barrier. I can further split out the barriers to a after_ctr function, but not sure that's what we want...

HansKristian-Work: Is it possible that barrier count > copy count? I'm not sure how that is possible.

ishitatsuyuki: For pre-barriers, there can be additional barriers coming from resource tracking. For post-barriers it probably won't exceed the copy count, but for the CTR batch I'm using a limit of 64 for reasons mentioned above.

HansKristian-Work: By resource tracking, do you mean `d3d12_command_list_track_resource_usage`? I think those are a separate system, and shouldn't affect barrier batching.

ishitatsuyuki: Yeah, sounds right. I guess we'll get the CTR batch limit lowered to 16 too for now.

libs/vkd3d/command.c
ishitatsuyuki: Actually, on second thought, this is quite troubling for Forza. It has multiple copies to the same image that doesn't overlap. But to check that we need to compare the coordinates as well. It does fit within the quadratic implementation though.

An alternative is to try to use render-based blits, which is guaranteed to be ordered, but that's going to be another massively overengineered hell...

ishitatsuyuki: Yes, exactly. I think I’ll just compromise here and go the less optimized path. It looks like the game developer is well aware of the implications as they interleave calls to different destination images; so while not having box overlap checks will make this suboptimal, it is still an improvement from having a barrier after each copy.

libs/vkd3d/command.c
ishitatsuyuki: The heap allocation commit is dropped. It's irrelevant now anyway now we're limited by the same-subresource aliasing.

ishitatsuyuki:
Updated. Should have addressed most of the review comments, plus:
- Refactored the `before_ctr` and `ctr` functions to branch on `info->batch_type` for simplicity.
- Fixed one place where `VKD3D_BATCH_TYPE_COPY_IMAGE_TO_BUFFER` and `VKD3D_BATCH_TYPE_COPY_BUFFER_TO_IMAGE` were switched around.
- Rewrote the alias checking logic to fix multiple correctness issues.

libs/vkd3d/command.c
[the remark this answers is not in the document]
ishitatsuyuki: The `end_batch` is now constrained to:

> 1. ResourceBarrier: This is where the transition from COPY_DEST to other
>    might happen, at which point the writes must be visible. This might
>    also transition away from COPY_SRC which invalidates the
>    precondition.
> 2. Copy operations. Copies to the same resource are implicitly ordered.
> 3. Draws and dispatches. These are not strictly necessary, but we don't
>    want too much command reordering so flushing here seems good.
> 4. Close. So that we don't throw commands into the void.

libs/vkd3d/command.c
```diff
@@ -6089,70 +6090,128 @@ static void d3d12_command_list_transition_image_layout(struct d3d12_command_list
...
+            WARN("Invalid format %#x.\n", src->PlacedFootprint.Footprint.Format);
+            return false;
+        }
+
+        vk_buffer_image_copy_from_d3d12(&out->copy.buffer_image, &src->PlacedFootprint,
+                dst->SubresourceIndex, &dst_resource->desc, out->src_format, out->dst_format, src_box, dst_x,
+                dst_y, dst_z);
+        out->copy.buffer_image.bufferOffset += src_resource->mem.offset;
+
+        out->vk_layout = d3d12_resource_pick_layout(dst_resource, VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL);
```
HansKristian-Work, line 6172: Writes_full_subresource is not set here. I don't think the original code did this either, but it should be added I think.

libs/vkd3d/command.c
```diff
@@ -6089,70 +6090,128 @@ static void d3d12_command_list_transition_image_layout(struct d3d12_command_list
...
+        if (out->dst_format->vk_aspect_mask != out->src_format->vk_aspect_mask &&
+                (out->copy.image.dstSubresource.aspectMask & VK_IMAGE_ASPECT_STENCIL_BIT) &&
+                !list->device->vk_info.EXT_shader_stencil_export)
+        {
+            FIXME("Destination depth-stencil format %#x is not supported for STENCIL dst copy with render pass fallback.\n",
+                    out->dst_format->dxgi_format);
+            return false;
+        }
+
+        out->writes_full_subresource = d3d12_image_copy_writes_full_subresource(dst_resource,
```
HansKristian-Work, line 6200: vk_layout is not set here. I think we need two. src_layout / dst_layout.

libs/vkd3d/command.c
```diff
@@ -5444,6 +5446,7 @@ static void STDMETHODCALLTYPE d3d12_command_list_DrawInstanced(d3d12_command_lis
             return;
     }
 
+    d3d12_command_list_end_batch(list);
```
HansKristian-Work, line 5455: This should be placed in begin_render_pass() I think. begin_render_pass is always done before a draw call.

libs/vkd3d/command.c
```diff
@@ -5544,6 +5548,8 @@ static void STDMETHODCALLTYPE d3d12_command_list_Dispatch(d3d12_command_list_ifa
...
```
HansKristian-Work, line 5557: Missing ExecuteIndirect as well I think?

libs/vkd3d/command.c
```diff
@@ -6353,6 +6363,7 @@ static void STDMETHODCALLTYPE d3d12_command_list_CopyResource(d3d12_command_list
     d3d12_command_list_track_resource_usage(list, src_resource, true);
 
     d3d12_command_list_end_current_render_pass(list, false);
+    d3d12_command_list_end_batch(list);
```
HansKristian-Work, line 6414: It would be clearer to just call this transfer_batch rather than just batch. Easier to read the code that way I think.

libs/vkd3d/command.c
```diff
@@ -6430,6 +6470,7 @@ static void STDMETHODCALLTYPE d3d12_command_list_CopyTiles(d3d12_command_list_if
             buffer, buffer_offset, flags);
 
     d3d12_command_list_end_current_render_pass(list, true);
+    d3d12_command_list_end_batch(list);
```
HansKristian-Work, line 6525: I think this is missing ResolveSubresource as well?

libs/vkd3d/command.c
```diff
@@ -6403,6 +6414,35 @@ static void STDMETHODCALLTYPE d3d12_command_list_CopyResource(d3d12_command_list
...
 
+static void d3d12_command_list_end_batch(struct d3d12_command_list *list)
+{
+    switch (list->transfer_batch.batch_type)
+    {
+        case VKD3D_BATCH_TYPE_COPY_BUFFER_TO_IMAGE:
+        case VKD3D_BATCH_TYPE_COPY_IMAGE_TO_BUFFER:
+        case VKD3D_BATCH_TYPE_COPY_IMAGE:
+            d3d12_command_list_end_current_render_pass(list, false);
+            for (size_t i = 0; i < list->transfer_batch.batch_len; i++)
```
HansKristian-Work, line 6475: I'm not sure it's safe to do this commit without also do the alias checks. If there are overlapped inputs, they will conflict I think, performing the same layout transition twice.

libs/vkd3d/command.c
```diff
@@ -6416,16 +6419,20 @@ static void STDMETHODCALLTYPE d3d12_command_list_CopyResource(d3d12_command_list
...
     {
         case VKD3D_BATCH_TYPE_COPY_BUFFER_TO_IMAGE:
         case VKD3D_BATCH_TYPE_COPY_IMAGE_TO_BUFFER:
         case VKD3D_BATCH_TYPE_COPY_IMAGE:
             d3d12_command_list_end_current_render_pass(list, false);
+            d3d12_command_list_barrier_batch_init(&barriers);
             for (size_t i = 0; i < list->transfer_batch.batch_len; i++)
-                d3d12_command_list_before_copy_texture_region(list, &list->transfer_batch.batch[i]);
+                d3d12_command_list_before_copy_texture_region(list, &barriers, &list->transfer_batch.batch[i]);
+            d3d12_command_list_barrier_batch_end(list, &barriers);
```
HansKristian-Work, line 6477: Don't we need an init after the end here?

libs/vkd3d/command.c
```diff
@@ -39,9 +39,10 @@ struct d3d12_command_list_barrier_batch
     VkMemoryBarrier vk_memory_barrier;
     uint32_t image_barrier_count;
     VkPipelineStageFlags dst_stage_mask, src_stage_mask;
+    bool check_alias;
```
HansKristian-Work, line 42: I still think the check_alias thing is a bad idea.

libs/vkd3d/command.c
```diff
@@ -6339,6 +6340,49 @@ static void STDMETHODCALLTYPE d3d12_command_list_CopyTextureRegion(d3d12_command
         return;
 
     d3d12_command_list_ensure_batch(list, copy_info.batch_type);
+    alias = false;
+    for (size_t i = 0; i < list->transfer_batch.batch_len; i++)
```
HansKristian-Work, line 6344: Just add `!alias` to the loop conditional.

libs/vkd3d/command.c
```diff
@@ -6339,6 +6340,49 @@ static void STDMETHODCALLTYPE d3d12_command_list_CopyTextureRegion(d3d12_command
...
+                // and WAW hazards.
+                subres = &copy_info.copy.buffer_image.imageSubresource;
+                other_subres = &other_info->copy.buffer_image.imageSubresource;
+                alias |= copy_info.dst.pResource == other_info->dst.pResource &&
+                         subres->aspectMask == other_subres->aspectMask &&
+                         subres->baseArrayLayer == other_subres->baseArrayLayer &&
+                         subres->mipLevel == other_subres->mipLevel;
+                break;
+            case VKD3D_BATCH_TYPE_COPY_IMAGE_TO_BUFFER:
+                // Test for source aliasing, needed to avoid issuing duplicate layout transitions which would
```
HansKristian-Work, line 6361: Source aliasing is fine. When we add image barriers, we'll avoid duplicate transitions anyways. I only think we need to consider aliasing for BUFFER_TO_IMAGE.

libs/vkd3d/vkd3d_private.h
```diff
@@ -2064,6 +2064,7 @@ struct vkd3d_image_copy_info
 {
     D3D12_TEXTURE_COPY_LOCATION src, dst;
     const struct vkd3d_format *src_format, *dst_format;
+    enum vkd3d_batch_type batch_type;
```
HansKristian-Work, line 2067: This should be folded into the commit that actually introduces the struct type.

HansKristian-Work (changes requested):

libs/vkd3d/command.c
```diff
@@ -6403,6 +6414,35 @@ static void STDMETHODCALLTYPE d3d12_command_list_CopyResource(d3d12_command_list
...
 }
 
+static void d3d12_command_list_end_batch(struct d3d12_command_list *list)
+{
+    switch (list->transfer_batch.batch_type)
+    {
+        case VKD3D_BATCH_TYPE_COPY_BUFFER_TO_IMAGE:
+        case VKD3D_BATCH_TYPE_COPY_IMAGE_TO_BUFFER:
+        case VKD3D_BATCH_TYPE_COPY_IMAGE:
+            d3d12_command_list_end_current_render_pass(list, false);
```
HansKristian-Work, line 6473: Could just skip this if batch_len is 0. This will be called unconditionally in many places.

ishitatsuyuki: The batch_type would be `NONE` in such cases. Does separating that test out of `switch` sounds good?

HansKristian-Work: That is true ... Just keep it as is then.

libs/vkd3d/command.c
ishitatsuyuki: I don't think so, since batch_end basically resets it to the fresh state after flush.

libs/vkd3d/command.c
ishitatsuyuki: Added to resolve_subresource.

libs/vkd3d/command.c
ishitatsuyuki: I've reorganized the commits so the first one introduces a batch without transposing the loop, and the second adds all the alias checks and transpose.

libs/vkd3d/command.c
ishitatsuyuki: I-to-I copies currently delegates to `d3d12_command_list_copy_image` and does not use these fields.

HansKristian-Work (changes requested):
Mostly nits and minor issues, the overall design looks good.

libs/vkd3d/command.c
```diff
@@ -6329,6 +6333,39 @@ static void STDMETHODCALLTYPE d3d12_command_list_CopyTextureRegion(d3d12_command
...
+                         subres->baseArrayLayer == other_subres->baseArrayLayer &&
+                         subres->mipLevel == other_subres->mipLevel;
+                break;
+            case VKD3D_BATCH_TYPE_COPY_IMAGE_TO_BUFFER:
+                /* Test for destination aliasing as D3D12 requires serialization on overlapping copies (WAW hazard). */
+                /* TODO: Do more granular alias testing */
+                alias |= copy_info.dst.pResource == other_info->dst.pResource;
+                break;
+            case VKD3D_BATCH_TYPE_COPY_IMAGE:
+                /* TODO: Check for alias once we start batching barriers for image-image copies too */
```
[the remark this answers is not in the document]
ishitatsuyuki: It's implicitly handled in the copy image calls. There's effectively no batching happening here, and `before_ctr` only submits the initial layout transition (which should be free of aliasing issues).

libs/vkd3d/command.c
```diff
@@ -6329,6 +6333,39 @@ static void STDMETHODCALLTYPE d3d12_command_list_CopyTextureRegion(d3d12_command
...
+                other_subres = &other_info->copy.buffer_image.imageSubresource;
+                alias |= copy_info.dst.pResource == other_info->dst.pResource &&
+                         subres->aspectMask == other_subres->aspectMask &&
+                         subres->baseArrayLayer == other_subres->baseArrayLayer &&
+                         subres->mipLevel == other_subres->mipLevel;
+                break;
+            case VKD3D_BATCH_TYPE_COPY_IMAGE_TO_BUFFER:
+                /* Test for destination aliasing as D3D12 requires serialization on overlapping copies (WAW hazard). */
+                /* TODO: Do more granular alias testing */
+                alias |= copy_info.dst.pResource == other_info->dst.pResource;
```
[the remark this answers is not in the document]
ishitatsuyuki: I added "copy_info res is sparse || other_info res is sparse" to the conditions.

libs/vkd3d/command.c
```diff
@@ -6397,6 +6408,37 @@ static void STDMETHODCALLTYPE d3d12_command_list_CopyResource(d3d12_command_list
     VKD3D_BREADCRUMB_COMMAND(COPY);
 }
 
+static void d3d12_command_list_end_transfer_batch(struct d3d12_command_list *list)
+{
+    switch (list->transfer_batch.batch_type)
+    {
+        case VKD3D_BATCH_TYPE_NONE:
```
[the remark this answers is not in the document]
ishitatsuyuki: Putting default: on the top seemed weird, so I changed the `NONE` branch to an early return and put a fallback `default:` at the bottom instead.

libs/vkd3d/command.c
```diff
@@ -7151,25 +7195,47 @@ static void d3d12_command_list_barrier_batch_add_layout_transition(
...
+             * different kind of overlaps.
+             * In CopyTextureRegion() we can have two copies on the same src or dst resource batched into one,
+             * resulting in an exact duplicate layout transition.
+             * In ResourceBarrier() we won't have such exact duplicates, mainly because doing the same transition twice
+             * is likely illegal (unless oldLayout == newLayout).
+             * The reason to test for overlap is because barriers in D3D12 behaves as if each transition happens in
+             * order. Vulkan memory barriers do not, so if there is a race condition, we need to split the barrier.
+             * As such, we need to eliminate duplicates like the former, while cutting the batch when we encounter an
+             * overlap like the latter. */
+            if (!(exact_match && image_barrier->oldLayout == batch->vk_image_barriers[i].oldLayout &&
```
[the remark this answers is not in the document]
ishitatsuyuki: Thanks, added early break and a variable for the condition.

libs/vkd3d/command.c
```diff
@@ -7106,26 +7209,50 @@ static void d3d12_command_list_barrier_batch_add_layout_transition(
...
+             * resulting in an exact duplicate layout transition.
+             * In ResourceBarrier() we won't have such exact duplicates, mainly because doing the same transition twice
+             * is illegal.
+             * The reason to test for overlap is that barriers in D3D12 behaves as if each transition happens in
+             * order. Vulkan memory barriers do not, so if there is a race condition, we need to split the barrier.
+             * As such, we need to eliminate duplicates like the former, while cutting the batch when we encounter an
+             * overlap like the latter. */
+            layout_match = image_barrier->oldLayout == batch->vk_image_barriers[i].oldLayout &&
+                    image_barrier->newLayout == batch->vk_image_barriers[i].newLayout;
+            if (!(exact_match && layout_match))
```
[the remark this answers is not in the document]
ishitatsuyuki: Sorry, this should be fixed now.

libs/vkd3d/command.c
```diff
@@ -6329,6 +6334,45 @@ static void STDMETHODCALLTYPE d3d12_command_list_CopyTextureRegion(d3d12_command
...
+            case VKD3D_BATCH_TYPE_COPY_IMAGE:
+                /* TODO: Check for alias once we start batching barriers for image-image copies too */
+                break;
+            default:
+                assert(false);
+                return;
+        }
+    }
+    if (alias)
+        d3d12_command_list_end_transfer_batch(list);
```
HansKristian-Work, line 6374: If we end the transfer batch here, wouldn't batch type be reset to NONE, and thus the copy_info we append here will be dropped?

ishitatsuyuki: Ouch, it doesn’t get dropped because we don’t reset length for NONE, which is why I didn’t catch it in testing, but it will cause legitimate issue when the batch length is 1. I’ll move the ensure() down to prevent that.

HansKristian-Work (approved):
LGTM.

PR #1136, opened 2022-06-23
Execute Indirect state - Phase 1

HansKristian-Work:


libs/vkd3d/vkd3d_private.h
```diff
@@ -2099,12 +2099,20 @@ struct d3d12_command_list
...
     bool has_replaced_shaders;
-    bool has_valid_index_buffer;
+
+    struct
+    {
+        VkBuffer buffer;
+        VkDeviceSize offset;
+        DXGI_FORMAT dxgi_format;
+        VkIndexType vk_type;
+        bool is_non_null;
```
doitsujin, line 2109: couldn't we use the VkBuffer being `VK_NULL_HANLDE` for this case instead? This seems somewhat redundant now.

libs/vkd3d/command.c
```diff
@@ -12442,37 +12446,136 @@ CONST_VTBL struct ID3D12CommandSignatureVtbl d3d12_command_signature_vtbl =
...
+    }
+
+    if (requires_root_signature && !root_signature)
+    {
+        ERR("Command signature requires root signature, but is not provided.\n");
+        return E_INVALIDARG;
+    }
+    else if (!requires_root_signature && root_signature)
+    {
+        ERR("Command signature requires root signature, root signature must be NULL.\n");
```
doitsujin, line 12570: "does not require root signature"?

doitsujin (changes requested):
Small nits.

libs/vkd3d/vkd3d_private.h
HansKristian-Work: Fixed.

libs/vkd3d/command.c

doitsujin (approved):

PR #1147, opened 2022-06-28
Do not attempt to place linear host visible textures on heaps.

HansKristian-Work:
Alignment and size requirements cannot be returned to application. Fall back to committed resources here instead.

Fixes weird GI flicker in Guardians of the Galaxy, since it ended up placing 3x 16x16x80 R32_UINT textures close to each other. In OPTIMAL tiling, the size was < 128 KiB, but ~300 KiB in linear tiling due to 256 byte aligned rows on AMD.


doitsujin (approved):

sygys:
This isn't fixed or broken again in proton 7

sygys:
No one reading this and fixing it?

K0bin:
People are on vacation right now. How do you know this isn't fixed?

r2rX:
> No one reading this and fixing it?

In fairness, not only are the devs hands full but it's also still a holiday. Give it some time.

mbriar:
Also, proton 7 is still on a pretty old version of vkd3d-proton (to support older drivers) that probably doesn't even have this fix, so make sure to test with experimental or newer, and make a bug report if it's still not fixed there.

PR #1156, opened 2022-07-06
Profiling improvements

HansKristian-Work:
RDTSC and --delta support in parsing tool.

--delta is useful for copying the profile before and after a test region to extract iteration / tick counts over the region instead of since bootup of application.

### DIFF
--- a/include/private/vkd3d_common.h
+++ b/include/private/vkd3d_common.h
@@ -311,4 +311,19 @@ static inline uint64_t vkd3d_get_current_time_ns(void)
 #endif
 }
 
+#ifdef _MSC_VER
+#pragma intrinsic(__rdtsc)
+#endif
+
+static inline uint64_t vkd3d_get_current_time_ticks(void)
+{
+#ifdef _MSC_VER
+    return __rdtsc();
+#elif defined(__i386__) || defined(__x86_64__)
+    return __builtin_ia32_rdtsc();
+#else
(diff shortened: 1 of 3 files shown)


include/private/vkd3d_common.h
```diff
@@ -311,4 +311,19 @@ static inline uint64_t vkd3d_get_current_time_ns(void)
...
+#ifdef _MSC_VER
+#pragma intrinsic(__rdtsc)
+#endif
+
+static inline uint64_t vkd3d_get_current_time_ticks(void)
+{
+#ifdef _MSC_VER
+    return __rdtsc();
+#elif defined(__i386__) || defined(__x86_64__)
+    return __builtin_ia32_rdtsc();
```
doitsujin, line 323: Depending on how accurate we want the measurement to be, we might want an lfence after/before rdtsc for begin/end respectively, since the instruction isn't serializing and measurements of short pieces of code will be affected by out-of-order execution otherwise.

HansKristian-Work: Considering how noisy the measurements are to begin with, I doubt it's that meaningful to lfence this.

doitsujin (approved):

include/private/vkd3d_common.h
doitsujin: fair.

PR #1157, opened 2022-07-08
Execute indirect - Phase 6 - Optimizations

HansKristian-Work:
- Use Index LUT feature.
- Try to reuse application indirect buffer when possible.

Builds on phase 5.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -9647,6 +9647,7 @@ static void d3d12_command_list_execute_indirect_state_template(
         struct d3d12_resource *count_buffer, UINT64 count_buffer_offset)
 {
     const struct vkd3d_vk_device_procs *vk_procs = &list->device->vk_procs;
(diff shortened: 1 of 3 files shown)


libs/vkd3d/command.c
```diff
@@ -9799,17 +9831,40 @@ static void d3d12_command_list_execute_indirect_state_template(
...
+        stream.buffer = arg_buffer->res.vk_buffer;
+        stream.offset = arg_buffer->mem.offset + arg_buffer_offset;
+    }
+
+#if 0
+    if (require_patch)
+        INFO("Template requires patching :(\n");
+    else
+        INFO("Template skips patching :)\n");
+#endif
```
doitsujin, line 9865: Couldn't we make this a `WARN` or something in case we need to patch?

doitsujin:
Nit since I'm not a huge fan of having disabled code, looks good otherwise.

doitsujin (approved):

PR #1172, opened 2022-07-18
Misc RT improvements

HansKristian-Work:
- Add some trace debug logs for RTAS info build.
- Check for index format != UNKNOWN instead of IBO VA != nullptr when deciding whether to use indexed or list build. UE5 hits this case.
- Update dxil-spirv to master with some CFG fixes.

### DIFF
--- a/libs/vkd3d/acceleration_structure.c
+++ b/libs/vkd3d/acceleration_structure.c
@@ -19,6 +19,8 @@
 #define VKD3D_DBG_CHANNEL VKD3D_DBG_CHANNEL_API
 #include "vkd3d_private.h"
 
+#define RT_TRACE TRACE
+
 void vkd3d_acceleration_structure_build_info_cleanup(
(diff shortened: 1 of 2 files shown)


libs/vkd3d/acceleration_structure.c
```diff
@@ -19,6 +19,8 @@
 #define VKD3D_DBG_CHANNEL VKD3D_DBG_CHANNEL_API
 #include "vkd3d_private.h"
 
+#define RT_TRACE TRACE
```
doitsujin, line 22: do we plan to decouple this at any point or is this just meant to be changed for local debugging?

doitsujin (approved):
LGTM, just a smol question.

libs/vkd3d/acceleration_structure.c
HansKristian-Work: Mostly just for local debug. We might want to consider a separate debug channel, but it's only really useful when doing deep dives.

PR #1174, opened 2022-07-19
Flush queued waiters before waiting for the sparse binding semaphore

Guy1524:
Discussed on Discord.


HansKristian-Work (approved):
LGTM, thanks for narrowing it down.

PR #1175, opened 2022-07-20
Implement ID3D12Fence sharing on top of D3D12-Fence exportable Vulkan timeline-semaphores.

Guy1524:


libs/vkd3d/command.c
```diff
@@ -392,6 +397,15 @@ static HRESULT vkd3d_create_timeline_semaphore(struct d3d12_device *device, uint
     type_info.semaphoreType = VK_SEMAPHORE_TYPE_TIMELINE_KHR;
     type_info.initialValue = initial_value;
 
+    if (shared)
```
HansKristian-Work, line 400: This is only required for exporting semaphores, not import. Also somewhat missing a check for support with vkGetPhysicalDeviceExternalSemaphoreProperties.

libs/vkd3d/command.c
```diff
@@ -1259,6 +1278,290 @@ HRESULT d3d12_fence_create(struct d3d12_device *device,
...
+    }
+
+    wait_info.sType = VK_STRUCTURE_TYPE_SEMAPHORE_WAIT_INFO;
+    wait_info.pNext = NULL;
+    wait_info.flags = 0;
+    wait_info.semaphoreCount = 1;
+    wait_info.pSemaphores = &fence->timeline_semaphore;
+    wait_info.pValues = &value;
+
+    if ((vr = VK_CALL(vkWaitSemaphoresKHR(fence->device->vk_device, &wait_info, -1))))
```
HansKristian-Work, line 1436: UINT64_MAX. Technically -1 will sign extend, but it jumps out as questionable.

libs/vkd3d/device.c
```diff
@@ -4443,6 +4447,37 @@ static HRESULT STDMETHODCALLTYPE d3d12_device_CreateSharedHandle(d3d12_device_if
         ID3D12Resource_Release(resource_iface);
         return vr ? E_FAIL : S_OK;
     }
+    if (SUCCEEDED(ID3D12DeviceChild_QueryInterface(object, &IID_ID3D12Fence, (void**)&fence_iface)))
+    {
+        VkSemaphoreGetWin32HandleInfoKHR win32_handle_info;
```
HansKristian-Work, line 4452: Needs win32 ifdeffing similar to resources.

libs/vkd3d/command.c
HansKristian-Work: Actually, it might be relevant for import as well. I think you can CreateSharedHandle from an OpenSharedHandle.

libs/vkd3d/device.c
```diff
@@ -4408,6 +4409,9 @@ static HRESULT STDMETHODCALLTYPE d3d12_device_CreateSharedHandle(d3d12_device_if
     struct d3d12_device *device = impl_from_ID3D12Device(iface);
     const struct vkd3d_vk_device_procs *vk_procs;
     ID3D12Resource *resource_iface;
+    ID3D12Fence *fence_iface;
+
+    vk_procs = &device->vk_procs;
```
HansKristian-Work, line 4414: Seems redundant to do `vk_procs` assignment here twice?

libs/vkd3d/command.c
```diff
@@ -11891,6 +12218,93 @@ static void d3d12_command_queue_signal(struct d3d12_command_queue *command_queue
...
+static void d3d12_command_queue_wait_shared(struct d3d12_command_queue *command_queue,
+        struct d3d12_shared_fence *fence, UINT64 value)
+{
+    assert(fence->timeline_semaphore);
+    vkd3d_queue_add_wait(command_queue->vkd3d_queue, &fence->ID3D12Fence_iface, fence->timeline_semaphore, value);
+
+    /* We can't get the physical value here for a shared fence,
+     * and waiting for the next submit would add even more time
+     * for a fence rewind to go unnoticed, so we flush. */
+    vkd3d_queue_flush_waiters(command_queue->vkd3d_queue, &command_queue->fence_worker, &command_queue->device->vk_procs);
```
HansKristian-Work, line 12230: Is there idea here that shared fences work like D3D12 timelines and they can rewind and everything and winevulkan resolves this on submit internally instead?

Does vkQueueSubmitKHR here basically translate into a blocking wait if it finds a timeline semaphore that is D3D12 shared?

libs/vkd3d/command.c
```diff
@@ -1259,6 +1278,290 @@ HRESULT d3d12_fence_create(struct d3d12_device *device,
...
+    struct d3d12_shared_fence *fence = shared_impl_from_ID3D12Fence1(iface);
+    const struct vkd3d_vk_device_procs *vk_procs = &fence->device->vk_procs;
+    VkSemaphoreWaitInfo wait_info;
+    VkResult vr;
+
+    TRACE("iface %p, value %#"PRIx64", event %p.\n", iface, value, event);
+
+    if (event)
+    {
+        FIXME("Signaling events on shared fence completion not supported.\n");
```
HansKristian-Work, line 1425: I would guess this is theoretically possible to support if we make a WINE vendor extension or Win32-only EXT? Ala `vkSemaphoreSetWin32EventOnCompletionEXT`?

libs/vkd3d/command.c
Guy1524: Yeah, that's the idea, but vkQueueSubmitKHR won't block, there's another queue worker thread inside winevulkan, similar to VKD3D-Proton's.  The only difference really is that winevulkan uses a mechanism to start the wait inside the vkQueueSubmitKHR call, in order to avoid missing any values that would signal that would then be rolled back by a signal.

If you'd like to see what that's like, here's my wine branch based off Proton Experimental: https://github.com/Guy1524/wine/commits/halo-infinite-fixes

libs/vkd3d/command.c
Guy1524: Yeah, if we run into this either of those two options should work.  DXVK uses a wait thread for ID3D11Fence, but the problem is that if you only want one thread per fence, it's not easy to support the rewind case.  Really the only working option would be a threadpool where you have a new worker for every wait, or as you mentioned a wine extension which would be pretty trivial.

libs/vkd3d/command.c
HansKristian-Work: Ok, that seems fine.

libs/vkd3d/command.c
HansKristian-Work: I'm not too concerned as long as we have a path forward if need be.

libs/vkd3d/command.c
```diff
@@ -12662,15 +13076,31 @@ static void *d3d12_command_queue_submission_worker_main(void *userdata)
...
+                d3d12_shared_fence_dec_ref(submission.wait.shared_fence);
+            }
             VKD3D_REGION_END(queue_wait);
             break;
 
         case VKD3D_SUBMISSION_SIGNAL:
             VKD3D_REGION_BEGIN(queue_signal);
-            d3d12_command_queue_signal(queue, submission.signal.fence, submission.signal.value);
-            d3d12_fence_dec_ref(submission.signal.fence);
+            if (submission.signal.fence)
```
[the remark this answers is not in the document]
Guy1524: I updated the PR and added this, not sure which way is neater.

libs/vkd3d/command.c
```diff
@@ -11879,6 +12199,88 @@ static void d3d12_command_queue_signal(struct d3d12_command_queue *command_queue
...
+    /* If there are pending waiters, we have to handle that since the signal must come after waiters. */
+    submit_info.waitSemaphoreCount = vkd3d_queue->wait_count;
+    submit_info.pWaitSemaphores = vkd3d_queue->wait_semaphores;
+    submit_info.pWaitDstStageMask = vkd3d_queue->wait_stages;
+    timeline_submit_info.waitSemaphoreValueCount = vkd3d_queue->wait_count;
+    timeline_submit_info.pWaitSemaphoreValues = vkd3d_queue->wait_values;
+
+    vr = VK_CALL(vkQueueSubmit(vk_queue, 1, &submit_info, VK_NULL_HANDLE));
+
+    vkd3d_queue_push_waiters_to_worker_locked(vkd3d_queue,
```
HansKristian-Work, line 12266: This looks obsolete.

libs/vkd3d/command.c
```diff
@@ -11879,6 +12199,88 @@ static void d3d12_command_queue_signal(struct d3d12_command_queue *command_queue
...
+    timeline_submit_info.signalSemaphoreValueCount = 2;
+
+    /* If there are pending waiters, we have to handle that since the signal must come after waiters. */
+    submit_info.waitSemaphoreCount = vkd3d_queue->wait_count;
+    submit_info.pWaitSemaphores = vkd3d_queue->wait_semaphores;
+    submit_info.pWaitDstStageMask = vkd3d_queue->wait_stages;
+    timeline_submit_info.waitSemaphoreValueCount = vkd3d_queue->wait_count;
+    timeline_submit_info.pWaitSemaphoreValues = vkd3d_queue->wait_values;
+
+    vr = VK_CALL(vkQueueSubmit(vk_queue, 1, &submit_info, VK_NULL_HANDLE));
```
HansKristian-Work, line 12264: I don't see where we queue up dec_ref for the shared fence once when the signal is satisfied.

libs/vkd3d/command.c
```diff
@@ -11879,6 +12199,88 @@ static void d3d12_command_queue_signal(struct d3d12_command_queue *command_queue
...
+    signal_semaphores[1] = vkd3d_queue->submission_timeline;
+
+    submit_info.pSignalSemaphores = signal_semaphores;
+    submit_info.signalSemaphoreCount = 2;
+
+    signal_values[0] = value;
+    signal_values[1] = vkd3d_queue->submission_timeline_count;
+
+    timeline_submit_info.pSignalSemaphoreValues = signal_values;
+    timeline_submit_info.signalSemaphoreValueCount = 2;
```
HansKristian-Work, line 12255: If we want submissions timeline to  come after the signal of shared fence, we need two submission batches to guarantee the order, otherwise the two semaphores here can signal out of order leading to a subtle race if we free the shared fence based on submission timeline signal.

libs/vkd3d/device.c
```diff
@@ -4580,6 +4617,41 @@ static HRESULT STDMETHODCALLTYPE d3d12_device_OpenSharedHandle(d3d12_device_ifac
 
         return return_interface(&resource->ID3D12Resource_iface, &IID_ID3D12Resource, riid, object);
     }
+    if (IsEqualGUID(riid, &IID_ID3D12Fence))
```
HansKristian-Work, line 4620: Blank line inbetween.

HansKristian-Work (changes requested):
Some minor issues.

HansKristian-Work (approved):
LGTM.

HansKristian-Work (changes requested):
Wait, another case of missing signed-off-by :frog:

HansKristian-Work (approved):
LGTM

PR #1182, opened 2022-07-25
Fall back to regular fences when shared timeline semaphores aren't supported by the host.

Guy1524:
Fixes Control in environment without shared fence support


HansKristian-Work (approved):
LGTM

PR #1207, opened 2022-08-31
Fix clearing attachments that are not currently active

doitsujin:
Apparently we hit a bug in Saint's Row where we clear an RTV that's bound to the D3D command list but wasn't used by any Vulkan pipeline, so it's not bound to the Vulkan command buffer.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -9473,6 +9473,42 @@ static void STDMETHODCALLTYPE d3d12_command_list_ClearUnorderedAccessViewFloat(d
     d3d12_command_list_clear_uav(list, &d, resource_impl, &args, &color, rect_count, rects);
 }
 
+static bool d3d12_command_list_is_subresource_bound_as_rtv_dsv(struct d3d12_command_list *list,
+        struct d3d12_resource *resource, const VkImageSubresource *subresource)
+{
+    unsigned int i;
+
+    if (!(list->rendering_info.state_flags & (VKD3D_RENDERING_ACTIVE | VKD3D_RENDERING_SUSPENDED)))
+        return false;
+
+    if (list->dsv.resource == resource)
+    {
+        const struct vkd3d_view *dsv = list->dsv.view;
+
+        if (subresource->mipLevel == dsv->info.texture.miplevel_idx &&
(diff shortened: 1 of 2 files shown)


libs/vkd3d/command.c
```diff
@@ -9473,21 +9473,55 @@ static void STDMETHODCALLTYPE d3d12_command_list_ClearUnorderedAccessViewFloat(d
...
 
+static bool d3d12_command_list_is_subresource_bound_as_rtv_dsv(struct d3d12_command_list *list,
+        struct d3d12_resource *resource, const VkImageSubresource *subresource)
+{
+    unsigned int i;
+
+    if (!(list->rendering_info.state_flags & (VKD3D_RENDERING_ACTIVE | VKD3D_RENDERING_SUSPENDED)))
+        return false;
+
+    if (list->dsv.resource == resource)
```
HansKristian-Work, line 9493: I assume this is okay even if the RTV mask / DSV mask does not bind every render target? I believe we perform the begin/end barriers either way regardless?

libs/vkd3d/command.c
```diff
@@ -9559,28 +9593,25 @@ static void STDMETHODCALLTYPE d3d12_command_list_DiscardResource(d3d12_command_l
...
-            else
-                has_unbound_subresource = true;
-        }
+        if (d3d12_command_list_is_subresource_bound_as_rtv_dsv(list, texture, &vk_subresource))
+            has_bound_subresource = true;
+        else
+            has_unbound_subresource = true;
+    }
 
+    if (all_subresource_full_discard)
```
HansKristian-Work, line 9620: It took a good minute to understand what this branch was supposed to do. It would help to bind this logic to something like `bool discard_partially_overlaps_render_targets` (or something less jank) I think

HansKristian-Work (changes requested):
Minor nit/question, otherwise LGTM.

libs/vkd3d/command.c
doitsujin: Yes, this is why this behaves differently than the clear stuff.

HansKristian-Work: OK.

HansKristian-Work (approved):
Probably fine as-is.

PR #1208, opened 2022-08-31
GetFrameLatencyWaitableObject should return a duplicated handle.

dave-juicelabs:


doitsujin:
It would be nice for any given pull request to state the motivation for any given change, otherwise it's hard to tell if a change is necessary or useful.

I *suppose* this is supposed to fix situations where an app calls CloseHandle on the returned handle? Has this been observed anywhere?

dave-juicelabs:
My apologies. This was observed when running Adobe After Effects. After Effects does call CloseHandle on the handle returned by GetFrameLatencyWaitableObject as stated in the documentation for the function. https://docs.microsoft.com/en-us/windows/win32/api/dxgi1_3/nf-dxgi1_3-idxgiswapchain2-getframelatencywaitableobject

As a result, during swapchain destruction in vkd3d, CloseHandle would cause an exception as the handle was now invalid having previously been closed.

libs/vkd3d/swapchain.c
```diff
@@ -2573,13 +2573,17 @@ static HRESULT STDMETHODCALLTYPE d3d12_swapchain_GetMaximumFrameLatency(dxgi_swa
...
     struct d3d12_swapchain *swapchain = d3d12_swapchain_from_IDXGISwapChain(iface);
+    HANDLE duplicated_handle;
 
     TRACE("iface %p.\n", iface);
 
     if (!(swapchain->desc.Flags & DXGI_SWAP_CHAIN_FLAG_FRAME_LATENCY_WAITABLE_OBJECT))
         return NULL;
 
-    return swapchain->frame_latency_event;
+    DuplicateHandle(GetCurrentProcess(), swapchain->frame_latency_event, GetCurrentProcess(), &duplicated_handle,
```
HansKristian-Work, line 2583: Are we required to return a different HANDLE for each call to this? Easy test is to call GetMaximumFrameLatency() twice and see if we get different HANDLE values.

HansKristian-Work: I can do a quick check ...

doitsujin (approved):

PR #1210, opened 2022-09-01
[WIP] Add VKD3D profiles to express vulkan requirements 

christophe-lunarg:
I created a Vulkan profiles JSON file to express VKD3D minimum requirements for each D3D level. This JSON files are defined by Vulkan Profiles schemas avaiable [here](https://schema.khronos.org/vulkan/) and can be used as part of the Vulkan SDK Profiles Toolset. ([PDF](https://www.lunarg.com/wp-content/uploads/2022/03/The-Vulkan-Profiles-Toolset-Solution-FEB2022.pdf), [Presentation](https://www.lunarg.com/wp-content/uploads/2022/08/Vulkan-SDK-Tools-to-Use-and-Create-Vulkan-Profiles-AUG-SIGGRAPH-2022.pdf))

I based myself on the repository source code as the main page requirements seem pretty sparsed.

This is an equivalent work from what we did for DXVK with this PR: https://github.com/doitsujin/dxvk/pull/2826

This PR is really for discussion purposes at this point so that I understand if you would be interested in such idea and the direction you would like it to take.

A couple of caveats :
- I had to create a new construct to express a "OR" of capability checks which I still need to . 
This is something we were pushing to avoid in the profiles concept because depending who produces and consumes a profile it can make things very hard for Vulkan application developers: "How to know what APIs to use when a profile is reported supported on a platform but has multiple variants to claim it reported?" The issue doesn't happen here because because it's VKD3D that creates the VkDevice instance.

Eg:
```
            "capabilities": [
                "vulkan10requirements",
                "vulkan11requirements",
                "d3d12_level_11_0",
                ["d3d12_level_11_0_bindless_uniform", "d3d12_level_11_0_bindless_storage"]
            ]
```
Means `VP_VKD3D_d3d12_level_11_0_baseline` requires "vulkan10requirements" AND "vulkan11requirements" AND "d3d12_level_11_0" AND ("d3d12_level_11_0_bindless_uniform" OR "d3d12_level_11_0_bindless_storage")

- VKD3D is a LOT more complex with capabilities checking than DXVK, at least partly due to D3D12 handling of features compared with D3D11. I am not quite sure it's really tracktable...
- I only created "baseline" profiles and no "optimal", "exhaustive" or "max" because it looks like VKD3D enables a lot of things my default based on the user system Vulkan capabilities and explicitly disable things it doesn't want (we can express disabled features actually). So it's not obvious to figure out what's actually used or not, I would have to do a very long investigation.

Looking forward your inputs!

### DIFF
--- a/VP_VKD3D_requirements.json
+++ b/VP_VKD3D_requirements.json
@@ -1,0 +1,778 @@
+{
+    "$schema": "https://schema.khronos.org/vulkan/profiles-0.8.0-224.json#",
+    "capabilities": {
+        "vulkan10requirements": {
+            "features": {
+                "VkPhysicalDeviceFeatures": {
+                    "robustBufferAccess": true
+                }
+            }
+        },
+        "vulkan11requirements": {
+            "features": {
+                "VkPhysicalDeviceVulkan11Features": {
+                    "multiview": true
+                }
+            },
+            "properties": {
+                "VkPhysicalDeviceVulkan11Properties": {
+                    "maxMultiviewViewCount": 6,
+                    "maxMultiviewInstanceIndex": 134217727
+                }
+            }
+        },
+        "d3d12_level_11_0": {
+            "extensions": {
+                "VK_KHR_surface": 1,
+                "VK_KHR_swapchain": 1,
+                "VK_KHR_timeline_semaphore": 1,
+                "VK_KHR_separate_depth_stencil_layouts": 1,
+                "VK_KHR_sampler_mirror_clamp_to_edge": 1,
+                "VK_EXT_robustness2": 1,
+                "VK_KHR_bind_memory2": 1,
+                "VK_KHR_copy_commands2": 1,
+                "VK_KHR_dynamic_rendering": 1,
+                "VK_EXT_extended_dynamic_state": 1,
+                "VK_EXT_extended_dynamic_state2": 1,
+                "VK_KHR_maintenance4": 1,
+                "VK_EXT_descriptor_indexing": 1
+            },
+            "features": {
+                "VkPhysicalDeviceSeparateDepthStencilLayoutsFeaturesKHR": {
+                    "separateDepthStencilLayouts": true
+                },
+                "VkPhysicalDeviceRobustness2FeaturesEXT": {
+                    "nullDescriptor": true
+                },
+                "VkPhysicalDeviceShaderDrawParametersFeatures": {
+                    "shaderDrawParameters": true
+                },
+                "VkPhysicalDeviceDynamicRenderingFeaturesKHR": {
+                    "dynamicRendering": true
+                },
+                "VkPhysicalDeviceExtendedDynamicStateFeaturesEXT": {
+                    "extendedDynamicState": true
+                },
+                "VkPhysicalDeviceExtendedDynamicState2FeaturesEXT": {
+                    "extendedDynamicState2": true
+                },
+                "VkPhysicalDeviceMaintenance4FeaturesKHR": {
+                    "maintenance4": true
+                },
+                "VkPhysicalDeviceDescriptorIndexingFeaturesEXT": {
+                    "runtimeDescriptorArray": true,
+                    "descriptorBindingPartiallyBound": true,
+                    "descriptorBindingUpdateUnusedWhilePending": true,
+                    "descriptorBindingVariableDescriptorCount": true,
+                    "descriptorBindingSampledImageUpdateAfterBind": true,
+                    "descriptorBindingUniformTexelBufferUpdateAfterBind": true,
+                    "shaderSampledImageArrayNonUniformIndexing": true,
+                    "shaderUniformTexelBufferArrayNonUniformIndexing": true
+                }
+            },
+            "properties": {
+                "VkPhysicalDeviceDescriptorIndexingPropertiesEXT": {
+                    "maxPerStageDescriptorUpdateAfterBindSampledImages": 1000000,
+                    "maxPerStageDescriptorUpdateAfterBindStorageImages": 1000000
+                }
+            }
+        },
+        "d3d12_level_11_0_bindless_uniform": {
+            "extensions": {
+                "VK_EXT_descriptor_indexing": 1
+            },
+            "properties": {
+                "VkPhysicalDeviceDescriptorIndexingPropertiesEXT": {
+                    "maxPerStageDescriptorUpdateAfterBindUniformBuffers": 1000000
+                }
+            }      
+        },
+        "d3d12_level_11_0_bindless_storage": {
+            "extensions": {
+                "VK_EXT_descriptor_indexing": 1
+            },
+            "properties": {
+                "VkPhysicalDeviceDescriptorIndexingPropertiesEXT": {
+                    "maxPerStageDescriptorUpdateAfterBindStorageBuffers": 1000000
+                }
+            }
+        },
+        "d3d12_level_11_0_optimal": {
+            "extensions": {
+                "VK_EXT_descriptor_indexing": 1,
+                "VK_EXT_shader_stencil_export": 1,
+                "VK_EXT_shader_viewport_index_layer": 1
+            },
+            "properties": {
+                "VkPhysicalDeviceDescriptorIndexingPropertiesEXT": {
+                    "robustBufferAccessUpdateAfterBind": true
+                }
+            }
+        },
+        "d3d12_level_11_1": {
+            "extensions": {
+            },
+            "features": {
+                "VkPhysicalDeviceFeatures": {
+                    "logicOp": true,
+                    "vertexPipelineStoresAndAtomics": true
+                }
+            },
+            "properties": {
+                "VkPhysicalDeviceProperties": {
+                    "limits": {
+                        "maxPerStageDescriptorStorageBuffers": 64,
+                        "maxPerStageDescriptorStorageImages": 64
+                    }
+                }
+            }
+        },
+        "d3d12_level_12_0": {
+            "extensions": {
+                "VK_EXT_sampler_filter_minmax": 1
+            },
+            "features": {
+                "VkPhysicalDeviceFeatures": {
+                    "sparseBinding": true,
+                    "sparseResidencyAliased": true,
+                    "sparseResidencyBuffer": true,
+                    "sparseResidencyImage2D": true,
+                    "shaderResourceResidency": true,
+                    "shaderResourceMinLod": true
+                }
+            },
+            "properties": {
+                "VkPhysicalDeviceProperties": {
+                    "sparseProperties": {
+                        "residencyStandard2DBlockShape": true,
+                        "residencyNonResidentStrict": true
+                    }
+                },
+                "VkPhysicalDeviceSamplerFilterMinmaxPropertiesEXT": {
+                    "filterMinmaxSingleComponentFormats": false
+                }
+            },
+            "queueFamiliesProperties": [
+                {
+                    "VkQueueFamilyProperties": {
+                        "queueFlags": [ "VK_QUEUE_SPARSE_BINDING_BIT" ],
+                        "queueCount": 1
+                    }
+                }
+            ]
+        },
+        "d3d12_level_12_0_properties3_read_without_format": {
+            "extensions": {
+                "VK_KHR_format_feature_flags2": 1
+            },
+            "formats": {
+                "VK_FORMAT_R32_SFLOAT": {
+                    "VkFormatProperties3KHR": {
+                        "linearTilingFeatures": [
+                            "VK_FORMAT_FEATURE_2_STORAGE_IMAGE_BIT_KHR", "VK_FORMAT_FEATURE_2_STORAGE_READ_WITHOUT_FORMAT_BIT_KHR"
+                        ],
+                        "optimalTilingFeatures": [
+                            "VK_FORMAT_FEATURE_2_STORAGE_IMAGE_BIT_KHR", "VK_FORMAT_FEATURE_2_STORAGE_READ_WITHOUT_FORMAT_BIT_KHR"
+                        ]
+                    }
+                },
+                "VK_FORMAT_R32_UINT": {
+                    "VkFormatProperties3KHR": {
+                        "linearTilingFeatures": [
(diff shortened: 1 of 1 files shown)


doitsujin:
> VKD3D is a LOT more complex with capabilities checking than DXVK, at least partly due to D3D12 handling of features compared with D3D11. I am not quite sure it's really tracktable...

I think the main problem here is that we need multiple code paths to implement various features, and therefore end up doing different things based on the driver we're running on, primarily because D3D12 has requirements that Vulkan drivers don't necessaril fulfill. D3D12 is also a moving target, so optional features that we'll use will constantly change (see mesh shaders).

I'm also expecting that we move to Vulkan 1.3 in the near-ish future to clean up the whole extension mess a bit and hard-require some features that are semi-optional right now (esp. BDA), so I'm not sure how useful it is to do the whole profile thing *right now*. Maybe we should gather baseline requirements for Feature Level 12_0 (in practice this is what new D3D12 games require to run at all) and start from there - "optimal" profiles are probably not useful here at all since we'll basically take what we can get, and vendor support for some features, especially regarding the binding model (i.e. descriptor indexing properties etc), is unlikely to converge any time soon.

christophe-lunarg:
Hi,
It's ok with me regarding : 
- Waiting for the Vulkan 1.3 requirement bump.
- Starting with baseline level 12_0.  This said, I think exposing level 11_0 is pretty trivial
- Only creating baseline profiles.

Thanks for you feedback !

HansKristian-Work:
Where would this profile be hosted? Is it intended that it's pulled automatically from our repo?

christophe-lunarg:
Hi @HansKristian-Work !
This profiles file would be hosted in this repository and kept in sync with VKD3D implementation.

versalinyaa:
Hi, any update on this MR? I'm asking because having a checked-in, maintained profile file would simplify the task of gathering Vulkan requirements for Chrome OS.

christophe-lunarg:
I haven't seen any which to Vulkan 1.3 at the moment. Should we more forward with Vulkan 1.1 + extensions?

HansKristian-Work:
@christophe-lunarg Things have been hectic and this never bubbled up. @doitsujin I suppose we should just do the transition to 1.3 now?

doitsujin:
Yeah, we should be able to do this now without too many problems.

christophe-lunarg:
Great! I'll be happy to follow this and work on profiles proposals. :)

HansKristian-Work:
Vulkan 1.3 cleanup has been merged now.

christophe-lunarg:
I started to update the file, you should have something soon!

VP_VKD3D_requirements.json
```diff
@@ -0,0 +1,759 @@
...
+            "features": {
+                "VkPhysicalDeviceFeatures": {
+                    "robustBufferAccess": true
+                }
+            }
+        },
+        "vulkan11requirements": {
+            "features": {
+                "VkPhysicalDeviceVulkan11Features": {
+                    "multiview": true
```
HansKristian-Work, line 14: I don't think we use multiview?

VP_VKD3D_requirements.json
```diff
@@ -0,0 +1,759 @@
...
+                    "maxMultiviewInstanceIndex": 134217727
+                }
+            }
+        },
+        "vulkan12requirements": {
+            "features": {
+                "VkPhysicalDeviceVulkan12Features": {
+                    "uniformBufferStandardLayout": true,
+                    "subgroupBroadcastDynamicId": true,
+                    "imagelessFramebuffer": true,
```
HansKristian-Work, line 29: Unused.

VP_VKD3D_requirements.json
```diff
@@ -0,0 +1,759 @@
...
+            "features": {
+                "VkPhysicalDeviceFeatures": {
+                    "logicOp": true,
+                    "vertexPipelineStoresAndAtomics": true
+                }
+            },
+            "properties": {
+                "VkPhysicalDeviceProperties": {
+                    "limits": {
+                        "maxPerStageDescriptorStorageBuffers": 64,
```
HansKristian-Work, line 236: This is somewhat redundant since we rely on update-after-bind.

VP_VKD3D_requirements.json
```diff
@@ -0,0 +1,759 @@
...
+                    "VkQueueFamilyProperties": {
+                        "queueFlags": [ "VK_QUEUE_SPARSE_BINDING_BIT" ],
+                        "queueCount": 1
+                    }
+                }
+            ]
+        },
+        "d3d12_level_12_1": {
+            "extensions": {
+                "VK_EXT_conservative_rasterization": 1
```
HansKristian-Work, line 284: Missing VK_EXT_fragment_shader_interlock with per-pixel and per-sample features.

VP_VKD3D_requirements.json
```diff
@@ -0,0 +1,759 @@
...
+                    "descriptorBuffer": true,
+                    "descriptorBufferPushDescriptors": true
+                }
+            },
+            "properties": {
+                "VkPhysicalDeviceProperties": {
+                    "limits": {
+                        "maxPerStageDescriptorStorageBuffers": 64,
+                        "maxPerStageDescriptorStorageImages": 64,
+                        "minStorageBufferOffsetAlignment": 4
```
HansKristian-Work, line 199: Is this an exact requirement or "must be less-or-equal to 4"?

VP_VKD3D_requirements.json
```diff
@@ -0,0 +1,759 @@
...
+                    "requiredSubgroupSizeStages": ["VK_SHADER_STAGE_COMPUTE_BIT"]
+                }
+            }
+        }
+    },
+    "profiles": {
+        "VP_VKD3D_d3d12_level_11_0_baseline": {
+            "version": 1,
+            "api-version": "1.3.204",
+            "label": "VKD3D D3D12 Level 11.0 Baseline profile",
```
HansKristian-Work, line 562: vkd3d-proton

HansKristian-Work:
Some overall comments:

- Which strategies do we have available to ensure the JSON is correct?
  - I can see the profiles layer which we could use to restrict a driver to 11.0/11.1/12.0/12.1 feature sets. We could then test D3D12CreateDevice with those feature levels and verify the device is created successfully.
  - Is there a JSON validator to ensure the syntax is correct? Given there is a schema linked, I suppose a general purpose validator would do?
  - Is there a way to inspect and load a profile in some tool and see what a given ICD supports and does not support? Just to ensure that relevant drivers report the feature sets that are expected.
  - Running against validation should ensure that we've enumerated relevant features if we use the profiles layer. Is there a way to ensure that we don't enable too much in the profile by accident? I can imagine if there is a way to dump a profile JSON based on our vkCreateDevice calls, and then compare that somehow to the profile JSON would help. It should be possible to observe any features that we have not accounted for.

Fixing all issues that comes up will likely take 20+ roundtrips to sort out all micro details, so finding a way to hand this over seems like a good idea.

VP_VKD3D_requirements.json
```diff
@@ -0,0 +1,759 @@
...
+                "VkPhysicalDeviceVulkan13Properties": {
+                    "requiredSubgroupSizeStages": ["VK_SHADER_STAGE_COMPUTE_BIT"]
+                }
+            }
+        }
+    },
+    "profiles": {
+        "VP_VKD3D_d3d12_level_11_0_baseline": {
+            "version": 1,
+            "api-version": "1.3.204",
```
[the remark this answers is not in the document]
christophe-lunarg: Because `api-version` refers to a specific Vulkan Header version, it indicates what's the minimum Vulkan API revision the Vulkan drivers must supports for the necessary to be potentially exposed and supported. 

204 is the first Vulkan 1.3 Vulkan Header version and this profile only requires Vulkan 1.3 and not additional extensions or capabilities added in following Vulkan Header revision.

This said, you can decide that you want to require new Vulkan API version if you want just to explicitly claim that you don't want to support Vulkan drivers that old.

I used the Profiles JSON schemas here https://schema.khronos.org/vulkan/ to figure out these `api-version` for each profile.

VP_VKD3D_requirements.json
christophe-lunarg: Fixed

VP_VKD3D_requirements.json
christophe-lunarg: Fixed

VP_VKD3D_requirements.json
christophe-lunarg: This comes from here:
```
static void d3d12_device_caps_init_feature_level(struct d3d12_device *device)
{
    const VkPhysicalDeviceFeatures *features = &device->device_info.features2.features;
    const struct vkd3d_vulkan_info *vk_info = &device->vk_info;
    struct d3d12_caps *caps = &device->d3d12_caps;

    caps->max_feature_level = D3D_FEATURE_LEVEL_11_0;

    if (caps->options.OutputMergerLogicOp && features->vertexPipelineStoresAndAtomics &&
            vk_info->device_limits.maxPerStageDescriptorStorageBuffers >= D3D12_UAV_SLOT_COUNT &&
            vk_info->device_limits.maxPerStageDescriptorStorageImages >= D3D12_UAV_SLOT_COUNT)
        caps->max_feature_level = D3D_FEATURE_LEVEL_11_1;
```

So I am wondering, if this code is correct but it's hard for me to really know.

christophe-lunarg:
> Some overall comments:
> 
> * Which strategies do we have available to ensure the JSON is correct?
>   
>   * I can see the profiles layer which we could use to restrict a driver to 11.0/11.1/12.0/12.1 feature sets. We could then test D3D12CreateDevice with those feature levels and verify the device is created successfully.
Yes, that's exactly the idea with the Profiles layer, running together with the validation layer to get error reports for Vulkan capabilities infringement. 

>   * Is there a JSON validator to ensure the syntax is correct? Given there is a schema linked, I suppose a general purpose validator would do?
Yes, I am using the Profiles schemas here https://schema.khronos.org/vulkan/ that we (LunarG) keep updating for each Vulkan Header revisions. It's validated with any generic JSON schema validator. I am using this one only https://www.jsonschemavalidator.net/ , https://github.com/tristanpenman/valijson in C++.

>   * Is there a way to inspect and load a profile in some tool and see what a given ICD supports and does not support? Just to ensure that relevant drivers report the feature sets that are expected.
You can use `vulkaninfo` JSON export or [Vulkan Hardware Capability Viewer](https://vulkan.gpuinfo.org/download.php) and download a device profile JSON file on each reports: https://vulkan.gpuinfo.org/displayreport.php?id=19738

>   * Running against validation should ensure that we've enumerated relevant features if we use the profiles layer. Is there a way to ensure that we don't enable too much in the profile by accident? I can imagine if there is a way to dump a profile JSON based on our vkCreateDevice calls, and then compare that somehow to the profile JSON would help. It should be possible to observe any features that we have not accounted for.
Unfortunately, we don't have a script at the moment to check if all the capabilities in one profile is present in another profile. I agree this could be useful.

> Fixing all issues that comes up will likely take 20+ roundtrips to sort out all micro details, so finding a way to hand this over seems like a good idea.
I agree, digging into VKD3D-Proton to really understand the feature requirements is not trivial as part of the way capabilities are handle if by considering the features are available by default and disabling features the project is not using. I am happy to hand this over and review changes to ensure things remains correct or to resolve specific issues.

HansKristian-Work:
> You can use vulkaninfo JSON export or [Vulkan Hardware Capability Viewer](https://vulkan.gpuinfo.org/download.php) and download a device profile JSON file on each reports: https://vulkan.gpuinfo.org/displayreport.php?id=19738

I think I need the opposite thing here kind of? For example, if a profile declares various tiers, I should be able to do something like get a report for every feature level and see if it's supported or not, and if not supported, which features are missing to make it supported.

I assume this is a normal use case for an IHV going "do I support the profile this app declares, what are we missing?"

HansKristian-Work:
Thanks, that is helpful. I'll experiment with this in some simpler projects first to make sure the flow works.

> This is probably not ideal in your scenario... I expect this is how IHVs do it because they asked for that a library...

Running custom code for this is fine, sounds like it would work for my purposes.

christophe-lunarg:
Ok! let me know how it goes!

christophe-lunarg:
Feathermore, you may want to automatically document your profiles in a human friendly matter and we do have such a tool.

You can find an instance of result here: https://vulkan.lunarg.com/doc/sdk/1.3.239.0/windows/profiles_definitions.html

This is done using the `gen_profiles_solution.py` script that ships in the Vulkan SDK and that is available in the Profiles repository: https://github.com/KhronosGroup/Vulkan-Profiles/tree/main/scripts

This is how the script is used:
```
gen_profiles_solution.py
  --registry vk.xml
  --input ./profiles/
  --output-doc ./PROFILES.md
```

HansKristian-Work:
Got around to experimenting with this finally.

Filed two bugs:
- https://github.com/KhronosGroup/Vulkan-Profiles/issues/394
- https://github.com/KhronosGroup/Vulkan-Profiles/issues/393

HansKristian-Work:
Got through the self-study session now.

Added profiles support to Granite so that if a program using Granite wants to use profiles to initialize stuff, it can. Used the generate_solution scripts for that. https://github.com/Themaister/Granite/pull/112.

Added a profile to parallel-rdp as well. https://github.com/Themaister/parallel-rdp/tree/vulkan-profile

Verified that I can use the profile layer to restrict feature sets (using the new default=false env-variable), and combine that with validation worked well enough to ensure the profile is sound.

I think the next step is to start making vkd3d-proton profiles and test them against the profile layer.

HansKristian-Work:
@christophe-lunarg https://github.com/HansKristian-Work/vkd3d-proton/pull/1567.

There were some issues in the profile layer:
- Overriding sparse properties does not work.
- The negative property "residencyAlignedMipSize" does not validate that the property is actually false it seems ...
- Overriding VkPhysicalDeviceMutableDescriptorTypeFeaturesEXT does not work. It's forced to false despite profile forcing it to true.

Using alternative caps `["foo", "bar"]` is broken. The script fails to run with an undefined variable error and the generated profile solution does not understand any of this.

christophe-lunarg:
What version of the script are you using? The Using alternative caps ["foo", "bar"] should work if you are using the "main" branch version.

I'll have at your profiles look tomorrow and investigate these issues, thanks!

HansKristian-Work:
Superseded by master.

PR #1211, opened 2022-09-01
Implement mesh shaders using VK_EXT_mesh_shader

doitsujin:
Draft for now because there's still a WIP commit (unfortunately we're going to need separate pipeline layouts for mesh/task shader stuff), but with headers now being public this should less painful to iterate on.

I'm also not sure about barriers, right now we're naively returning all stages from `vk_queue_shader_stages`, but we might need to start tracking which stages are actually being used in any given pipeline to not regress performance in regular applications. Hard to test right now though since we need winevulkan support.


libs/vkd3d/command.c
```diff
@@ -3334,6 +3334,13 @@ static VkPipelineStageFlags vk_queue_shader_stages(VkQueueFlags vk_queue_flags)
                 VK_PIPELINE_STAGE_TESSELLATION_EVALUATION_SHADER_BIT |
                 VK_PIPELINE_STAGE_GEOMETRY_SHADER_BIT |
                 VK_PIPELINE_STAGE_FRAGMENT_SHADER_BIT;
+
+        if (device->device_info.mesh_shader_features.meshShader &&
```
HansKristian-Work, line 3338: Given how NON_PIXEL_SHADER_RESOURCE always implies compute too, I don't imagine this will cause too many issues.

libs/vkd3d/state.c
```diff
@@ -1381,10 +1382,14 @@ static HRESULT d3d12_root_signature_init_global(struct d3d12_root_signature *roo
 
     root_signature->num_set_layouts = context.vk_set;
 
+    graphics_stages = VK_SHADER_STAGE_ALL_GRAPHICS;
+
+    if (device->device_info.mesh_shader_features.meshShader && device->device_info.mesh_shader_features.taskShader)
```
HansKristian-Work, line 1387: I think we can use the same schism for compute / traceRay to separate legacy / mesh here. We'd just need to apply the root signature type in SetPipeline and potentially invalidate root parameters.

tests/d3d12_mesh_shader.c
```diff
@@ -365,7 +365,10 @@ void test_mesh_shader_create_pipeline(void)
     ms_ps_pipeline_desc.rasterizer = rasterizer_subobject;
 
     hr = create_pipeline_state_from_stream(device2, &ms_ps_pipeline_desc, &pipeline_state);
-    ok(hr == E_INVALIDARG, "Unexpected result for pipeline creation, hr %#x.\n", hr);
+    todo ok(hr == E_INVALIDARG, "Unexpected result for pipeline creation, hr %#x.\n", hr);
```
HansKristian-Work, line 368: Could we rebase tests to come first?

tests/d3d12_mesh_shader.c
```diff
@@ -1879,6 +1879,7 @@ void test_mesh_shader_execute_indirect(void)
         ID3D12GraphicsCommandList6_ExecuteIndirect(command_list6, command_signature, 1, indirect_buffer, 0,
                 tests[i].indirect_count ? indirect_buffer : NULL, tests[i].indirect_count ? 12 : 0);
         transition_resource_state(context.list, context.render_target, D3D12_RESOURCE_STATE_RENDER_TARGET, D3D12_RESOURCE_STATE_COPY_SOURCE);
+        todo_if(tests[i].indirect_count)
```
HansKristian-Work, line 1882: Should be implementable.

HansKristian-Work (changes requested):
Looks pretty solid overall. My main concerns are:

- We can implement basic execute indirect. Was probably missed since it was added later to the EXT draft?
- We should deal properly with root signatures.

libs/vkd3d/command.c
doitsujin: My concern was that on AMD we'll have an implied semaphore for task shaders, but it seems that the RADV implementation doesn't do anything too crazy on the vkCmdPipelineBarrier call itself, just increment some counters.

libs/vkd3d/command.c
```diff
@@ -10576,6 +10577,25 @@ static void STDMETHODCALLTYPE d3d12_command_list_ExecuteIndirect(d3d12_command_l
                 }
                 break;
 
+            case D3D12_INDIRECT_ARGUMENT_TYPE_DISPATCH_MESH:
+                if (max_command_count != 1)
+                    FIXME("Ignoring command count %u.\n", max_command_count);
+
+                if (count_buffer)
```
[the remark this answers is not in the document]
doitsujin: Fixed locally and removed the commit that marks these tests as todo.

tests/d3d12_mesh_shader.c
doitsujin: Done.

libs/vkd3d/device.c
```diff
@@ -6066,8 +6076,7 @@ static void d3d12_device_caps_init_feature_options7(struct d3d12_device *device)
 {
     D3D12_FEATURE_DATA_D3D12_OPTIONS7 *options7 = &device->d3d12_caps.options7;
 
-    /* Not supported */
-    options7->MeshShaderTier = D3D12_MESH_SHADER_TIER_NOT_SUPPORTED;
+    options7->MeshShaderTier = d3d12_device_determine_mesh_shader_tier(device);
```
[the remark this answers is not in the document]
doitsujin: Done.

libs/vkd3d/state.c
```diff
@@ -3687,7 +3714,7 @@ static HRESULT d3d12_pipeline_state_init_graphics(struct d3d12_pipeline_state *s
      * be exceedingly rare. */
     can_compile_pipeline_early =
             (desc->primitive_topology_type != D3D12_PRIMITIVE_TOPOLOGY_TYPE_PATCH || graphics->patch_vertex_count != 0) &&
-            desc->primitive_topology_type != D3D12_PRIMITIVE_TOPOLOGY_TYPE_UNDEFINED;
+            (desc->primitive_topology_type != D3D12_PRIMITIVE_TOPOLOGY_TYPE_UNDEFINED || (graphics->stage_flags & VK_SHADER_STAGE_MESH_BIT_EXT));
```
HansKristian-Work, line 3717: Reading this again, there might be a problem if the PSO sets PRIMITIVE_TOPOLOGY_TYPE_PATCH here, since it should be ignored. Might be better to always allow early compile with MESH.

libs/vkd3d/state.c
```diff
@@ -4158,29 +4190,30 @@ VkPipeline d3d12_pipeline_state_create_pipeline_variant(struct d3d12_pipeline_st
...
     pipeline_desc.layout = graphics->pipeline_layout;
-    pipeline_desc.renderPass = VK_NULL_HANDLE;
-    pipeline_desc.subpass = 0;
-    pipeline_desc.basePipelineHandle = VK_NULL_HANDLE;
     pipeline_desc.basePipelineIndex = -1;
 
     if (d3d12_device_supports_variable_shading_rate_tier_2(device))
         pipeline_desc.flags |= VK_PIPELINE_CREATE_RENDERING_FRAGMENT_SHADING_RATE_ATTACHMENT_BIT_KHR;
 
+    if (!(graphics->stage_flags & VK_SHADER_STAGE_MESH_BIT_EXT))
```
HansKristian-Work, line 4210: We probably have to avoid some dynamic state in `d3d12_graphics_pipeline_state_init_dynamic_state` as well such as `VERTEX_BUFFER_STRIDE` and `TOPOLOGY`.

HansKristian-Work: `d3d12_pipeline_state_get_pipeline` can call `d3d12_pipeline_state_can_use_dynamic_stride` as well. I would assume that VBO binding count must be 0 for mesh PSOs?

libs/vkd3d/state.c
```diff
@@ -1387,6 +1388,19 @@ static HRESULT d3d12_root_signature_init_global(struct d3d12_root_signature *roo
...
+    if (device->device_info.mesh_shader_features.meshShader && device->device_info.mesh_shader_features.taskShader)
+    {
+        mesh_shader_stages = VK_SHADER_STAGE_MESH_BIT_EXT |
+                VK_SHADER_STAGE_TASK_BIT_EXT |
+                VK_SHADER_STAGE_FRAGMENT_BIT;
+
+        if (FAILED(hr = vkd3d_create_pipeline_layout_for_stage_mask(
+                device, root_signature->num_set_layouts, root_signature->set_layouts,
+                &root_signature->push_constant_range,
+                mesh_shader_stages, &root_signature->mesh)))
```
HansKristian-Work, line 1400: Is this freed anywhere?

libs/vkd3d/vkd3d_private.h
```diff
@@ -2333,6 +2339,7 @@ static inline struct vkd3d_pipeline_bindings *d3d12_command_list_get_bindings(
             break;
 
         case VKD3D_PIPELINE_TYPE_GRAPHICS:
+        case VKD3D_PIPELINE_TYPE_MESH:
```
HansKristian-Work, line 2345: MESH is still "graphics", probably want to rename GRAPHICS here.

libs/vkd3d/state.c
```diff
@@ -3141,14 +3145,20 @@ static HRESULT d3d12_pipeline_state_init_graphics(struct d3d12_pipeline_state *s
         {VK_SHADER_STAGE_TESSELLATION_CONTROL_BIT,    offsetof(struct d3d12_pipeline_state_desc, hs)},
         {VK_SHADER_STAGE_TESSELLATION_EVALUATION_BIT, offsetof(struct d3d12_pipeline_state_desc, ds)},
         {VK_SHADER_STAGE_GEOMETRY_BIT,                offsetof(struct d3d12_pipeline_state_desc, gs)},
+        {VK_SHADER_STAGE_TASK_BIT_EXT,                offsetof(struct d3d12_pipeline_state_desc, as)},
+        {VK_SHADER_STAGE_MESH_BIT_EXT,                offsetof(struct d3d12_pipeline_state_desc, ms)},
         {VK_SHADER_STAGE_FRAGMENT_BIT,                offsetof(struct d3d12_pipeline_state_desc, ps)},
     };
 
-    state->pipeline_type = VKD3D_PIPELINE_TYPE_GRAPHICS;
+    state->pipeline_type = desc->ms.pShaderBytecode
```
HansKristian-Work, line 3153: Might be better to use byte size rather than pointer if it's a dummy pointer with 0 size?

HansKristian-Work (changes requested):
Nits

libs/vkd3d/state.c
doitsujin: Reworked this.

HansKristian-Work (approved):
LGTM :D

PR #1220, opened 2022-09-06
Clean up bind point handling

doitsujin:
Introduces a new enum rather than using VkPipelineBindPoint everywhere, and cleans up how D3D12 binding info is stored and accessed.

The plan is to add another value to the `vkd3d_pipeline_type` enum for mesh shading pipelines, so we can distinguish those from regular graphics pipelines.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -5552,10 +5552,9 @@ static void d3d12_command_list_update_hoisted_descriptors(struct d3d12_command_l
     bindings->dirty_flags &= ~VKD3D_PIPELINE_DIRTY_HOISTED_DESCRIPTORS;
 }
 
-static void d3d12_command_list_update_descriptors(struct d3d12_command_list *list,
-        VkPipelineBindPoint bind_point)
+static void d3d12_command_list_update_descriptors(struct d3d12_command_list *list)
 {
-    struct vkd3d_pipeline_bindings *bindings = &list->pipeline_bindings[bind_point];
+    struct vkd3d_pipeline_bindings *bindings = d3d12_command_list_get_bindings(list, list->active_pipeline_type);
     const struct d3d12_root_signature *rs = bindings->root_signature;
     const struct d3d12_bind_point_layout *bind_point_layout;
     VkPipelineBindPoint vk_bind_point;
@@ -5608,7 +5607,7 @@ static bool d3d12_command_list_update_compute_state(struct d3d12_command_list *l
     if (!d3d12_command_list_update_compute_pipeline(list))
         return false;
 
-    d3d12_command_list_update_descriptors(list, VK_PIPELINE_BIND_POINT_COMPUTE);
+    d3d12_command_list_update_descriptors(list);
 
     return true;
 }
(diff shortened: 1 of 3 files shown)


libs/vkd3d/vkd3d_private.h
```diff
@@ -1370,6 +1370,31 @@ enum vkd3d_root_signature_flag
...
 
+enum vkd3d_pipeline_type
+{
+    VKD3D_PIPELINE_TYPE_NONE,
+    VKD3D_PIPELINE_TYPE_GRAPHICS,
+    VKD3D_PIPELINE_TYPE_COMPUTE,
+    VKD3D_PIPELINE_TYPE_RAY_TRACING,
+};
+
+inline VkPipelineBindPoint vk_bind_point_from_pipeline_type(enum vkd3d_pipeline_type pipeline_type)
```
HansKristian-Work, line 1379: Should be static inline. C linkage rules are fun.

libs/vkd3d/command.c
```diff
@@ -8306,15 +8302,14 @@ static void d3d12_command_list_set_root_descriptor_va(struct d3d12_command_list
 }
 
 static void d3d12_command_list_set_root_descriptor(struct d3d12_command_list *list,
-        VkPipelineBindPoint bind_point, unsigned int index, D3D12_GPU_VIRTUAL_ADDRESS gpu_address)
+        struct vkd3d_pipeline_bindings *bindings, unsigned int index, D3D12_GPU_VIRTUAL_ADDRESS gpu_address)
```
HansKristian-Work, line 8305: There are some warnings around enums still be used, and end up translating to NULL pointer :scream:

libs/vkd3d/vkd3d_private.h
```diff
@@ -1496,6 +1521,27 @@ HRESULT vkd3d_create_descriptor_set_layout(struct d3d12_device *device,
         VkDescriptorSetLayoutCreateFlags flags, unsigned int binding_count,
         const VkDescriptorSetLayoutBinding *bindings, VkDescriptorSetLayout *set_layout);
 
+inline const struct d3d12_bind_point_layout *d3d12_root_signature_get_layout(
+        const struct d3d12_root_signature *root_signature, enum vkd3d_pipeline_type pipeline_type)
+{
+    switch (pipeline_type)
+    {
+        case VKD3D_PIPELINE_TYPE_NONE:
```
HansKristian-Work, line 1527: Probably better to break here and let the outer scope return work. This pattern would easily trip warnings due to unreachable code.

libs/vkd3d/command.c
```diff
@@ -5557,25 +5557,17 @@ static void d3d12_command_list_update_descriptors(struct d3d12_command_list *lis
...
     VkPipelineLayout layout;
 
     if (!rs)
         return;
 
-    if (list->active_pipeline_type == VKD3D_PIPELINE_TYPE_RAY_TRACING)
-    {
-        /* We might have to emit to RT bind point,
-         * but we pretend we're in compute bind point. */
-        layout = bindings->rt_layout.vk_pipeline_layout;
```
HansKristian-Work, line 5568: Nice. This seems to actually fix a bug since we always used `bindings->layout` instead of `rt_layout` in vkCmdPushConstants above ... Wonder why this didn't explode ...

HansKristian-Work (changes requested):
Comments, otherwise seems like a good cleanup.

libs/vkd3d/command.c
doitsujin: I'm not seeing any; what's triggering those?

libs/vkd3d/vkd3d_private.h
doitsujin: Fixed.

HansKristian-Work (approved):
LGTM

PR #1224, opened 2022-09-08
dxil-spirv: Update submodule.

HansKristian-Work:
Notably, this applies the fix for OpEmitMeshTasksEXT where it is a terminator. This might break on drivers that ended up shipping with wrong semantics. This should only be a problem for day-1 beta drivers, and it's better to get this fixed ASAP rather than risk shipping broken shaders.


HansKristian-Work:
E.g.: 

- https://github.com/KhronosGroup/SPIRV-Tools/pull/4923
- https://github.com/KhronosGroup/SPIRV-Registry/pull/166

PR #1235, opened 2022-09-15
VK_EXT_shader_module_identifier - Phase 4/4

HansKristian-Work:
Goal here is to shrink the vkd3d-proton.cache by 95% or so. I verified that I'm hitting "successfully created from identifier" in various scenarios.

I also verified the feasibility with Fossilize.

The basic idea we need to validate is that if we have the shader identifier cache, and a Fossilize archive of the pipelines, we can achieve 100% cache hits after a FOZ replay. I verified this in our test suite.

In the validation run, everything works as expected. In real games, pipeline_library_log confirms that all pipelines are created from identifier. I also verified that on a driver update, we fail to create from identifier, and we fall back to creating SPIR-V. The next run after that, we hit in the shader caches again, as expected.

```
#!/bin/bash

rm vkd3d-proton.cache*
rm -f /tmp/test-suite.*

delete_shader_caches()
{
	rm -rf ~/.cache/mesa_shader_cache
	rm -rf ~/.nv
	rm -rf ~/.cache/nv
}

delete_shader_caches

echo "=== First run ==="
FOSSILIZE=1 FOSSILIZE_DUMP_PATH=/tmp/test-suite VKD3D_TEST_FILTER="$1" VKD3D_CONFIG=pipeline_library_log,shader_cache_sync ./tests/d3d12 2>&1 | grep IDENTIFIER
echo "=== First run done ... ==="

fossilize-merge-db /tmp/test-suite.foz /tmp/test-suite.*.foz

delete_shader_caches

fossilize-replay /tmp/test-suite.foz --progress --spirv-val
echo "=== Verification run ==="
VKD3D_TEST_FILTER="$1" VKD3D_CONFIG=pipeline_library_log,shader_cache_sync ./tests/d3d12 2>&1 | grep IDENTIFIER
echo "=== Done ==="
```

NOTE: The full test suite ends up consuming >256 VkInstance creations, so the write-only cache system fails eventually (normal applications will never hit this scenario). The upper limit can be increased with a simple hack to Fossilize:

```
diff --git a/fossilize_db.cpp b/fossilize_db.cpp
index ce75f84..c270335 100644
--- a/fossilize_db.cpp
+++ b/fossilize_db.cpp
@@ -1883,7 +1883,7 @@ struct ConcurrentDatabase : DatabaseInterface
         {
             // Lazily create a new database. Open the database file exclusively to work concurrently with other processes.
             // Don't try forever.
-            for (unsigned index = 1; index < 256 && !writeonly_interface; index++)
+            for (unsigned index = 1; index < 1024 && !writeonly_interface; index++)
             {
```

### DIFF
--- a/libs/vkd3d/cache.c
+++ b/libs/vkd3d/cache.c
@@ -198,6 +198,8 @@ enum vkd3d_pipeline_blob_chunk_type
     /* VkShaderStage is stored in upper 16 bits. */
     VKD3D_PIPELINE_BLOB_CHUNK_TYPE_SHADER_META = 4,
     VKD3D_PIPELINE_BLOB_CHUNK_TYPE_PSO_COMPAT = 5,
+    /* VkShaderStage is stored in upper 16 bits. */
+    VKD3D_PIPELINE_BLOB_CHUNK_TYPE_SHADER_IDENTIFIER = 6,
     VKD3D_PIPELINE_BLOB_CHUNK_TYPE_MASK = 0xffff,
     VKD3D_PIPELINE_BLOB_CHUNK_INDEX_SHIFT = 16,
 };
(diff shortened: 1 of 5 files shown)


doitsujin:
Mostly hypothetical, but can there ever be a scenario where we might end up loading and accepting a cache with the identifier flag on an implementation that doesn't support the extension? We're checking the algorithm UUID, which **shouldn't** be 0, but someone might intentionally try to break it and I'm not sure how robust we're against that.

Other than that, LGTM.

HansKristian-Work:
> accepting a cache with the identifier flag on an implementation that doesn't support the extension

If the impl doesn't support the flag we won't try to load the module identifiers to begin with, so the cache will theoretically work, but be useless since it won't contain anything useful (no SPIR-V and identifiers cannot be used anyways).

> which shouldn't be 0

The spec is worded with strong discouragement against "dummy"/easily conflicting hard-coded values for UUID, and I believe it would trip a CTS warning.

doitsujin (approved):

PR #1238, opened 2022-09-18
Warning message shows expected block size.

sherief:
This could've helped when I ran into the fact that Windows incorrectly accepts a 1x1 BCn texture and returnes valid copyable footprints.


HansKristian-Work:
This check was removed a few days ago. The check itself is wrong since D3D12 now allows unaligned compressed textures, just like Vulkan. See OPTIONS8 struct.

sherief:
Good to know, thanks!

PR #1242, opened 2022-09-21
vkd3d: Add missing descriptor type to local descriptor pools.

doitsujin:
This whole situation was reported as a Vulkan CTS and then Mesa bug rather than Proton, this should hopefully fix things.

Storage buffers are used for query resolves.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -1794,6 +1794,7 @@ static VkDescriptorPool d3d12_command_allocator_allocate_descriptor_pool(
         {VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER, 1024},
         {VK_DESCRIPTOR_TYPE_UNIFORM_TEXEL_BUFFER, 1024},
         {VK_DESCRIPTOR_TYPE_SAMPLED_IMAGE, 1024},
+        {VK_DESCRIPTOR_TYPE_STORAGE_BUFFER, 1024},
         {VK_DESCRIPTOR_TYPE_STORAGE_TEXEL_BUFFER, 1024},
         {VK_DESCRIPTOR_TYPE_STORAGE_IMAGE, 1024},
         /* must be last in the array */


HansKristian-Work (approved):

ShabbyX:
Have you checked how well this works memory-wise?

```
    static const VkDescriptorPoolSize pool_sizes[] =
    {
        {VK_DESCRIPTOR_TYPE_SAMPLER, 2048},
        {VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER, 1024},
        {VK_DESCRIPTOR_TYPE_UNIFORM_TEXEL_BUFFER, 1024},
        {VK_DESCRIPTOR_TYPE_SAMPLED_IMAGE, 1024},
        {VK_DESCRIPTOR_TYPE_STORAGE_BUFFER, 1024},
        {VK_DESCRIPTOR_TYPE_STORAGE_TEXEL_BUFFER, 1024},
        {VK_DESCRIPTOR_TYPE_STORAGE_IMAGE, 1024},
        /* must be last in the array */
        {VK_DESCRIPTOR_TYPE_INLINE_UNIFORM_BLOCK_EXT, 65536}
    };
```

I recall when we (ANGLE) had an issue long ago where the pools would consume the memory necessary for all these descriptors (per the declaration), but then fail allocation when _one_ of them is exhausted (wasting memory allocated for the rest of them.

In other words, the pool created with `pool_sizes` is large (can hold like 70K descriptors), but if you allocate 2k samplers from it and then attempt to allocate another sampler, it can fail allocation and you'll have to create a new one (wasting 68K worth of descriptor memory).

What we ended up doing in ANGLE is to have a pool per pipeline layout, where the pool size is `N` x whatever is in the layout. Then we create a new desc set after `N` allocations because we know the desc set is exhausted and we also no there is no memory waste.

---

What you have probably works fine on (some) desktop vendors because they don't actually care about the descriptor types and just allocate generic memory, so like you could allocate 70K or 35K samplers from the descriptor despite it technically being limited to 2K.

ShabbyX:
> This code is dead. We removed it when moving to push descriptors.

Oops, somebody linked to this in relation to a recent spec MR, I didn't pay attention to the date.

> Generally this approach is questionable for emulation layers since there is zero control of what shaders look like. If e.g. D3D11 has 100k-ish shaders, you could end up with 1000s+ unique pipeline layouts and pools that have to be managed.

From GL's perspective at least, if you have 100k GL programs, 1K desc sets is nothing, lol

But yeah, VK_EXT_descriptor_buffer is what should have always been

doitsujin:
> From GL's perspective at least, if you have 100k GL programs, 1K desc sets is nothing, lol

Getting a bit off-topic, but the problem is more that descriptor *pools* are very heavy-weight (memory) objects on most/all drivers.

`VK_EXT_descriptor_buffer` makes the allocation part a lot nicer, but comes with other drawbacks - some descriptor types (mostly UBOs) are just straight-up slow on some hardware when used with descriptor buffers so push descriptors **must** be used, but push descriptors are not really viable in combination with `EXT_graphics_pipeline_library` when we don't know the full pipeline layout in advance. It's great for D3D12 emulation, but not really something that we're going to use in DXVK for example.

PR #1244, opened 2022-09-22
Remove some old Wine code we cannot track down

HansKristian-Work:
- trace.c was effectively dead code, so just remove it.
- checksum.c is technically app visible when serializing root signatures, and an application might rely on being able to load our serialized root sigs on native drivers, so we cannot trivially just remove it either. Just reuses RenderDoc's implementation here since it's MIT.
- Wrote a simple test to verify that checksum output is 1:1 before and after the change.


libs/vkd3d-shader/checksum.c
```diff
@@ -1,303 +1,118 @@
 /*
- * Copyright (C) 2001 Nikos Mavroyanopoulos
- * Copyright (C) 2004 Hans Leidekker
+ * Copyright (c) 2022 Hans-Kristian Arntzen for Valve Corporation
  *
- * This library is free software; you can redistribute it and/or
- * modify it under the terms of the GNU Lesser General Public
- * License as published by the Free Software Foundation; either
- * version 2.1 of the License, or (at your option) any later version.
+ * SPDX-License-Identifier: MIT
```
Joshua-Ashton, line 4: If we are adding a SPDX-License-Identifier... Can we just drop our whole license text past where that is.

Kernel, some new parts of Mesa and Khronos stuff does that.

(obv cant do it from the part taken from RDoc)

Joshua-Ashton (changes requested):

libs/vkd3d-shader/checksum.c
HansKristian-Work: Fixed.

Joshua-Ashton (approved):
LGTM!

PR #1245, opened 2022-09-22
Push descriptor rewrite - BDA cleanup

HansKristian-Work:


libs/vkd3d/vkd3d_private.h
```diff
@@ -2875,7 +2875,6 @@ enum vkd3d_bindless_set_flag
     VKD3D_BINDLESS_SET_UAV        = (1u << 3),
     VKD3D_BINDLESS_SET_IMAGE      = (1u << 4),
     VKD3D_BINDLESS_SET_BUFFER     = (1u << 5),
-    VKD3D_BINDLESS_SET_AUX_BUFFER = (1u << 6),
     VKD3D_BINDLESS_SET_RAW_SSBO   = (1u << 7),
```
Nefsen402, line 2878: Tightly pack the bit field?

libs/vkd3d/command.c
```diff
@@ -5489,20 +5473,20 @@ static void d3d12_command_list_update_root_descriptors(struct d3d12_command_list
...
     }
 
-    if (root_signature->flags & VKD3D_ROOT_SIGNATURE_USE_INLINE_UNIFORM_BLOCK)
+    if (root_signature->flags & VKD3D_ROOT_SIGNATURE_USE_PUSH_CONSTANT_UNIFORM_BLOCK)
     {
-        d3d12_command_list_fetch_inline_uniform_block_data(list, bindings, &root_parameter_data);
-
-        vk_write_descriptor_set_and_inline_uniform_block(&descriptor_writes[descriptor_write_count],
-                &inline_uniform_block_write, descriptor_set, root_signature, &root_parameter_data);
+        d3d12_command_list_fetch_inline_uniform_block_data(list, bindings, ptr_root_parameter_data);
```
doitsujin, line 5314: Might actually want to rename `d3d12_command_list_fetch_inline_uniform_block_data` to... something as well since we're no longer using inline uniform blocks.

libs/vkd3d/command.c
```diff
@@ -4045,14 +3876,11 @@ static bool d3d12_command_list_gather_pending_queries(struct d3d12_command_list
         map_buffer.range = entry_buffer_size;
 
         for (j = 0; j < ARRAY_SIZE(vk_writes); j++)
-            vk_writes[j].dstSet = vk_set;
-
-        VK_CALL(vkUpdateDescriptorSets(list->device->vk_device,
-                ARRAY_SIZE(vk_writes), vk_writes, 0, NULL));
+            vk_writes[j].dstSet = VK_NULL_HANDLE;
```
doitsujin, line 3879: This can be moved into the loop above that initializes `vk_writes`, no reason to have this in a nested loop if the set is always null.

doitsujin (approved):

doitsujin:
Just nits really but overall LGTM.

PR #1249, opened 2022-09-28
vkd3d: Rewrite how buffers are stored in desc heap metadata

HansKristian-Work:
Rather than storing VkBuffer / offset / size, we now store:
    
    - VA
    - Range (32-bit)
    - Format
    
    This accomplishes several things:
    
    - Prepares for a world where VkBufferView is not created in the heap itself (i.e. no offset buffer).
    - The metadata struct can now be shrunk to a neat 16 bytes instead of 32 before the refactor began.
    - Allow UAV clears to operate without a dependency on offset buffers.
    
    Hoisted CBV table descriptors (barely used, if ever) need to deref the VA late, but it's okay.
    Hopefully we get the option to push descriptors with VA later, but it's not critically important now.
    
    ClearUAV paths need to be rewritten to account for this.
    Rather than reusing VkBufferViews and reading the offset buffer,
    we can just create the buffer view on-demand instead.
    A VA deref is fortunately not needed since we get the ID3D12Resource inline.
    We already had to create fallback views for fallback clear paths anyways.
    
Builds on https://github.com/HansKristian-Work/vkd3d-proton/pull/1245

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -9178,7 +9178,7 @@ static void STDMETHODCALLTYPE d3d12_command_list_ClearUnorderedAccessViewUint(d3
     {
         /* Theoretically possibly for buggy application that tries to clear a buffer view with a texture resource.
          * Safeguard against crash. */
-        ERR("Attempted to clear buffer with image resource.\n");
+        WARN("Attempted to clear buffer with image resource.\n");
         return;
     }
 
@@ -9300,7 +9300,7 @@ static void STDMETHODCALLTYPE d3d12_command_list_ClearUnorderedAccessViewFloat(d
     {
         /* Theoretically possibly for buggy application that tries to clear a buffer view with a texture resource.
          * Safeguard against crash. */
-        ERR("Attempted to clear buffer with image resource.\n");
+        WARN("Attempted to clear buffer with image resource.\n");
         return;
     }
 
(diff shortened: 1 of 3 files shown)


libs/vkd3d/command.c
```diff
@@ -9154,86 +9174,96 @@ static void STDMETHODCALLTYPE d3d12_command_list_ClearUnorderedAccessViewUint(d3
     if (!vkd3d_clear_uav_info_from_desc(&args, &d))
         return;
 
-    if (args.has_view)
-        color = vkd3d_fixup_clear_uav_swizzle(list->device, d.view->info.view->format->dxgi_format, color);
+    if (d3d12_resource_is_texture(resource_impl) && !args.has_view)
+    {
+        /* Theoretically possibly for buggy application that tries to clear a buffer view with a texture resource.
+         * Safeguard against crash. */
+        ERR("Attempted to clear buffer with image resource.\n");
```
doitsujin, line 9181: `WARN`? That's what we usually do for potential game bugs to not spam messages by default.

libs/vkd3d/command.c
```diff
@@ -9265,10 +9296,33 @@ static void STDMETHODCALLTYPE d3d12_command_list_ClearUnorderedAccessViewFloat(d
...
```
doitsujin, line 9303: same here

doitsujin (approved):
nit, otherwise LGTM.

PR #1257, opened 2022-10-05
Add PREALLOCATE_SRV_MIP_CLAMPS workaround for Halo Infinite

esullivan-nvidia:
Halo Infinite currently has a race condition that can result in a GPU fault on NVIDIA hardware. The root cause of this race condition is the application overwriting existing SRV descriptors with a new mip clamp value asynchronously to the applications GPU submission timeline. This is problematic for vkd3d because it requires creating a new VkImageView to represent the SRV with the updated mip clamp value. If this SRV is being created after work has been submitted that accesses it, the GPU may not see the correct descriptor state possibly resulting in a fault.

This commit adds the VKD3D_CONFIG_FLAG_PREALLOCATE_SRV_MIP_CLAMPS configuration as a workaround to the race condition. It works by preallocating all the VkImageViews for each possible mip clamp level the application may attempt to create for the resource. Technically this is still a race condition but the end result isn't a GPU fault if the race condition is lost. Instead the shader will just sample from the texture with the previous SRVs mip clamp level rather than the updated one. These are the same outcomes that are possible on RADV without this workaround.


libs/vkd3d/resource.c
```diff
@@ -4432,6 +4432,25 @@ static void vkd3d_create_texture_srv(vkd3d_cpu_descriptor_va_t desc_va,
     if (key.u.texture.miplevel_count == VK_REMAINING_MIP_LEVELS)
         key.u.texture.miplevel_count = resource->desc.MipLevels - key.u.texture.miplevel_idx;
 
+    if ((vkd3d_config_flags & VKD3D_CONFIG_FLAG_PREALLOCATE_SRV_MIP_CLAMPS) &&
+        (desc->ViewDimension != D3D12_SRV_DIMENSION_TEXTURE2DMS) &&
+        (desc->ViewDimension != D3D12_SRV_DIMENSION_TEXTURE2DMSARRAY) &&
+        (hash_map_find(&resource->view_map.map, &key) == NULL))
+    {
+        uint32_t starting_mip = key.u.texture.miplevel_idx;
```
HansKristian-Work, line 4440: Starting mip should be +1 here since we create the base level below.

libs/vkd3d/resource.c
```diff
@@ -4432,6 +4432,25 @@ static void vkd3d_create_texture_srv(vkd3d_cpu_descriptor_va_t desc_va,
...
+        uint32_t mip_count = key.u.texture.miplevel_count != UINT32_MAX ?
+            key.u.texture.miplevel_count :
+            resource->desc.MipLevels - starting_mip;
+
+        struct vkd3d_view_key preallocate_key = key;
+
+        for (uint32_t i = starting_mip; i < mip_count; i++)
```
HansKristian-Work, line 4447: We use C89 style for the loop initializer, so it should be done outside the loop.

libs/vkd3d/resource.c
```diff
@@ -4432,6 +4432,25 @@ static void vkd3d_create_texture_srv(vkd3d_cpu_descriptor_va_t desc_va,
     if (key.u.texture.miplevel_count == VK_REMAINING_MIP_LEVELS)
         key.u.texture.miplevel_count = resource->desc.MipLevels - key.u.texture.miplevel_idx;
 
+    if ((vkd3d_config_flags & VKD3D_CONFIG_FLAG_PREALLOCATE_SRV_MIP_CLAMPS) &&
+        (desc->ViewDimension != D3D12_SRV_DIMENSION_TEXTURE2DMS) &&
+        (desc->ViewDimension != D3D12_SRV_DIMENSION_TEXTURE2DMSARRAY) &&
+        (hash_map_find(&resource->view_map.map, &key) == NULL))
```
HansKristian-Work, line 4438: !hash_map_find(...) is more in-style.

libs/vkd3d/resource.c
```diff
@@ -4432,6 +4432,25 @@ static void vkd3d_create_texture_srv(vkd3d_cpu_descriptor_va_t desc_va,
     if (key.u.texture.miplevel_count == VK_REMAINING_MIP_LEVELS)
         key.u.texture.miplevel_count = resource->desc.MipLevels - key.u.texture.miplevel_idx;
 
+    if ((vkd3d_config_flags & VKD3D_CONFIG_FLAG_PREALLOCATE_SRV_MIP_CLAMPS) &&
+        (desc->ViewDimension != D3D12_SRV_DIMENSION_TEXTURE2DMS) &&
```
HansKristian-Work, line 4436: Can drop the parens here. and the one below.

libs/vkd3d/resource.c
```diff
@@ -4432,6 +4432,25 @@ static void vkd3d_create_texture_srv(vkd3d_cpu_descriptor_va_t desc_va,
...
+        (desc->ViewDimension != D3D12_SRV_DIMENSION_TEXTURE2DMSARRAY) &&
+        (hash_map_find(&resource->view_map.map, &key) == NULL))
+    {
+        uint32_t starting_mip = key.u.texture.miplevel_idx;
+        uint32_t mip_count = key.u.texture.miplevel_count != UINT32_MAX ?
+            key.u.texture.miplevel_count :
+            resource->desc.MipLevels - starting_mip;
```
HansKristian-Work, line 4443: I think this is wrong. starting_mip is relative to the resource itself and we're iterating from starting_mip to MipLevels here.

HansKristian-Work (changes requested):
Some nits.

libs/vkd3d/resource.c
```diff
@@ -4432,6 +4432,25 @@ static void vkd3d_create_texture_srv(vkd3d_cpu_descriptor_va_t desc_va,
     if (key.u.texture.miplevel_count == VK_REMAINING_MIP_LEVELS)
         key.u.texture.miplevel_count = resource->desc.MipLevels - key.u.texture.miplevel_idx;
 
+    if ((vkd3d_config_flags & VKD3D_CONFIG_FLAG_PREALLOCATE_SRV_MIP_CLAMPS) &&
```
HansKristian-Work, line 4435: Given that we only add clamps for integer mips, we should probably floor() the mip clamp when this workaround is enabled.

libs/vkd3d/resource.c
esullivan-nvidia: I don't think this code can assume the application will always initially create the SRV with the mip clamp set to the same level as MostDetailedMip. Any value in the range [MostDetailedMip, MostDetailedMip + MipLevels] would be valid.

libs/vkd3d/resource.c
esullivan-nvidia: Fixed

libs/vkd3d/resource.c
esullivan-nvidia: Fixed

libs/vkd3d/resource.c
esullivan-nvidia: Fixed

esullivan-nvidia:
> This is truly cursed. This must have been a true nightmare to debug. Thanks a lot for this. Can you confirm this contribution is done on behalf of NVIDIA given your commit email (w.r.t the MIT relicense process we're working through)?

Yes I can. This contribution is done on behalf of NVIDIA.

esullivan-nvidia:
> FWIW, this is clearly a game bug. It's not allowed to do this in D3D12 either, has the bug been reported to the developers already?

Yes, it has been reported to the developer.

libs/vkd3d/resource.c
HansKristian-Work: Fair enough.

PR #1266, opened 2022-10-18
Swapchain rewrite - Phase 3 / 3

HansKristian-Work:
Focuses on the removal of vkd3d-utils, and rewrites the demo apps to use the new swapchain interface.

Expose libvkd3d-proton-d3d12.so native D3D12 implementation directly on Linux.


libs/vkd3d/swapchain_new.c
```diff
@@ -0,0 +1,1627 @@
...
+
+#ifdef VK_KHR_win32_surface
+    create_info.sType = VK_STRUCTURE_TYPE_WIN32_SURFACE_CREATE_INFO_KHR;
+    create_info.pNext = NULL;
+    create_info.hwnd = hwnd;
+    create_info.hinstance = GetModuleHandleA("d3d12.dll");
+    create_info.flags = 0;
+    vr = VK_CALL(vkCreateWin32SurfaceKHR(vk_instance, &create_info, NULL, &chain->vk_surface));
+#else
+    /* TODO: With dxvk-native integration, we can modify this as needed. */
```
Joshua-Ashton, line 730: I think we should let DXVK create the surface -- even for non-native builds.

There are two options here:
 - DXVK DXGI owns the surface but we need to share a VkInstance
 - We own the surface and ask for it to be made with a specific instance

 I am kind of in favour of the former as it seems cleaner overall.

I dont want to duplicate any logic here as I eventually I do want to add non-SDL WSI to DXVK and I don't want to have copy pasta when we can solve this elegantly.

HansKristian-Work: Sharing VkInstance could get annoying due to debug extensions, DXVK would have to know about those. Having DXVK create the surface would be cleaner.

HansKristian-Work: We're using the new DXVK SurfaceFactory interface, so this should be resolved now.

libs/d3d12/main.c
```diff
@@ -200,26 +225,35 @@ static VkPhysicalDevice d3d12_find_physical_device(struct vkd3d_instance *instan
...
+    IDXGIAdapter *dxgi_adapter;
+#endif
+
     static const char * const instance_extensions[] =
     {
         VK_KHR_SURFACE_EXTENSION_NAME,
+#ifdef _WIN32
         VK_KHR_WIN32_SURFACE_EXTENSION_NAME,
+#else
+        "VK_KHR_xcb_surface",
```
Joshua-Ashton, line 250: Good enough for now, but definitely worth adding a to-do here... DXVK DXGI uses SDL2 to get the extension to use here, so this will conflict eventually if people force Wayland, etc.

Joshua-Ashton (changes requested):

libs/d3d12/main.c
Joshua-Ashton: (we also probably want to add a bit more flexibility here on the DXVK side too, outside of just SDL2... just haven't had a huge need yet)

Joshua-Ashton:
Also the fact that the DXGI adapter stuff isn't implemented at all in native d3d12 is kinda oof right now.

I don't mind it for this PR, but we should definitely just like `dlopen` dxvk dxgi at some point and just make things work.

doitsujin (approved):
As discussed in Discord, we should probably do a sample app that pulls in vkd3d-proton and dxvk's DXGI in a separate repo. The current situation with the demo apps isn't great but will do for now.

PR #1267, opened 2022-10-18
Force initial placed transition for Assassin's Creed Valhalla

Saancreed:
Caught some "Application uses placed resource that must be initialized explicitly" while trying to debug CPU hangs on NV. Does not fix that particular issue, but maybe it will let us avoid other headaches in this game.

[ACValhalla.log](https://github.com/HansKristian-Work/vkd3d-proton/files/9815401/ACValhalla.log)


HansKristian-Work:
Color images are not known to cause issues, and I'd only add issues when there are known crash bugs that are explained by this bug.

Saancreed:
ACK. I can't tell for sure if this does or doesn't fix anything further on because right now the game appears to be borked by (presumably) some NV driver bug anyway. On the other hand, there are no stability issues on the Steam Deck so perhaps this is indeed harmless.

Feel free to close it then, I can reopen this once I find an actual issue that would be fixed by this workaround.

PR #1284, opened 2022-11-07
vkd3d: Implement SetEventOnCompletion for shared fences

### LINKED ISSUE #1281: Gears 5 hangs on launch (Bisected regression)
First initially reported by @GloriousEggroll. The game hangs before presenting a single frame.

The game needs ProtonGE to work around EAC for the single player campaign.

I've bisected the hang down to df1829e40770482597ab168559ef2d68ab3ec707 and confirmed that by reverting the commit on top of master.
```
df1829e40770482597ab168559ef2d68ab3ec707 is the first bad commit
commit df1829e40770482597ab168559ef2d68ab3ec707
Author: Derek Lesho <dlesho@codeweavers.com>
Date:   Fri Jul 15 15:35:42 2022 -0400

    vkd3d: Implement ID3D12Fence sharing on top of D3D12-Fence exportable Vulkan timeline semaphores.
    
    Signed-off-by: Derek Lesho <dlesho@codeweavers.com>

 libs/vkd3d/command.c       | 460 ++++++++++++++++++++++++++++++++++++++++++---
 libs/vkd3d/device.c        |  90 ++++++++-
 libs/vkd3d/vkd3d_private.h |  61 +++++-
 libs/vkd3d/vulkan_procs.h  |   7 +
 4 files changed, 579 insertions(+), 39 deletions(-)
```

There's also this interesting warning in the log file:
```
0130:fixme:d3d12_shared_fence_SetEventOnCompletion: Signaling events on shared fence completion not supported.
```

K0bin:
Fixes #1281

It's good enough to make Gears 5 launch but mostly untested beyond that.
Mostly a port from the same DXVK code.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -1314,10 +1314,34 @@ static void d3d12_shared_fence_inc_ref(struct d3d12_shared_fence *fence)
 static void d3d12_shared_fence_dec_ref(struct d3d12_shared_fence *fence)
 {
     ULONG refcount_internal = InterlockedDecrement(&fence->refcount_internal);
+    struct vkd3d_shared_fence_waiting_event *current, *e;
     const struct vkd3d_vk_device_procs *vk_procs;
+    bool is_running;
 
     if (!refcount_internal)
     {
+        pthread_mutex_lock(&fence->mutex);
+        is_running = fence->is_running;
+        if (is_running)
+        {
+            fence->is_running = false;
+            pthread_cond_signal(&fence->cond_var);
+        }
+        pthread_mutex_unlock(&fence->mutex);
+
+        if (is_running)
+        {
+            pthread_join(fence->thread, NULL);
+        }
+
+        LIST_FOR_EACH_ENTRY_SAFE(current, e, &fence->events, struct vkd3d_shared_fence_waiting_event, entry)
+        {
+            vkd3d_free(current);
+        }
+
+        pthread_mutex_destroy(&fence->mutex);
+        pthread_cond_destroy(&fence->cond_var);
+
         vk_procs = &fence->device->vk_procs;
         VK_CALL(vkDestroySemaphore(fence->device->vk_device, fence->timeline_semaphore, NULL));
 
@@ -1337,6 +1361,7 @@ static ULONG STDMETHODCALLTYPE d3d12_shared_fence_Release(d3d12_fence_iface *ifa
     if (!refcount)
     {
         struct d3d12_device *device = fence->device;
+
         d3d12_shared_fence_dec_ref(fence);
         d3d12_device_release(device);
     }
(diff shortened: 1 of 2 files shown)


libs/vkd3d/command.c
```diff
@@ -1539,8 +1633,15 @@ HRESULT d3d12_shared_fence_create(struct d3d12_device *device,
...
     }
 
+    pthread_mutex_init(&object->mutex, NULL);
+    list_init(&object->events);
+    object->last_value = 0;
+    object->stop = 0;
+
     d3d12_device_add_ref(object->device = device);
 
+    pthread_create(&object->thread, NULL, vkd3d_shared_fence_worker_main, object);
```
HansKristian-Work, line 1643: I would prefer that a thread is spawned on first waiter that actually requires a thread.

libs/vkd3d/command.c
```diff
@@ -1412,35 +1421,117 @@ static HRESULT STDMETHODCALLTYPE d3d12_shared_fence_GetDevice(d3d12_fence_iface
...
+        pthread_mutex_lock(&fence->mutex);
+
+        if (!(waiting_event = vkd3d_malloc(sizeof(*waiting_event))))
+        {
+            ERR("Failed to register device singleton for adapter.");
+            pthread_mutex_unlock(&fence->mutex);
+            return E_OUTOFMEMORY;
+        }
+        waiting_event->value = value;
+        waiting_event->event = event;
```
HansKristian-Work, line 1530: Need to handle the scenario where event is NULL, in which case we just need to block in this call until it's done. Could just call WaitSemaphoresKHR directly here.

libs/vkd3d/command.c
```diff
@@ -1412,35 +1421,117 @@ static HRESULT STDMETHODCALLTYPE d3d12_shared_fence_GetDevice(d3d12_fence_iface
 static UINT64 STDMETHODCALLTYPE d3d12_shared_fence_GetCompletedValue(d3d12_fence_iface *iface)
 {
     struct d3d12_shared_fence *fence = shared_impl_from_ID3D12Fence1(iface);
-    const struct vkd3d_vk_device_procs *vk_procs = &fence->device->vk_procs;
+
+    TRACE("iface %p\n", iface);
+    return vkd3d_atomic_uint64_load_explicit(&fence->last_value, vkd3d_memory_order_acquire);
```
HansKristian-Work, line 1426: Is this correct? A different process can update the counter, so a direct call to vkGetSemaphoreValue should be made here I think.

libs/vkd3d/command.c
```diff
@@ -1412,35 +1421,117 @@ static HRESULT STDMETHODCALLTYPE d3d12_shared_fence_GetDevice(d3d12_fence_iface
...
+}
+
+static void vkd3d_shared_fence_signal_events(struct d3d12_shared_fence *fence, uint64_t value)
```
HansKristian-Work, line 1429: Is this even needed? If the shared fence thread waits for a semaphore value, it should be unblocked when the vkSignalSemaphore is called on host.

libs/vkd3d/command.c
```diff
@@ -1337,9 +1337,18 @@ static void d3d12_shared_fence_dec_ref(struct d3d12_shared_fence *fence)
...
+    struct vkd3d_shared_fence_waiting_event *current, *e;
 
     if (!refcount_internal)
     {
+        vkd3d_atomic_uint64_store_explicit(&fence->stop, 1, vkd3d_memory_order_release);
+        pthread_join(fence->thread, NULL);
+
+        LIST_FOR_EACH_ENTRY_SAFE(current, e, &fence->events, struct vkd3d_shared_fence_waiting_event, entry)
+        {
+            vkd3d_free(current);
```
HansKristian-Work, line 1349: Is this safe? Wouldn't current->next be de-referenced right after?

libs/vkd3d/command.c
```diff
@@ -1337,9 +1337,18 @@ static void d3d12_shared_fence_dec_ref(struct d3d12_shared_fence *fence)
 {
     ULONG refcount_internal = InterlockedDecrement(&fence->refcount_internal);
     const struct vkd3d_vk_device_procs *vk_procs;
+    struct vkd3d_shared_fence_waiting_event *current, *e;
 
     if (!refcount_internal)
     {
+        vkd3d_atomic_uint64_store_explicit(&fence->stop, 1, vkd3d_memory_order_release);
```
HansKristian-Work, line 1344: Does this have to be a 64-bit atomic?

libs/vkd3d/command.c
```diff
@@ -1412,35 +1421,117 @@ static HRESULT STDMETHODCALLTYPE d3d12_shared_fence_GetDevice(d3d12_fence_iface
...
+            ERR("Failed to get shared fence counter value, error %d.\n", vr);
+            return NULL;
+        }
+
+        vkd3d_shared_fence_signal_events(fence, completed_value);
+
+        if (!!vkd3d_atomic_uint32_load_explicit(&fence->stop, vkd3d_memory_order_acquire))
+            return NULL;
+
+        completed_value++;
```
HansKristian-Work, line 1487: This is questionable in face of rewind. I would expect that you'd want to push wait requests to a queue. The wait thread wakes up, blocks on vkGetSemaphoreCounterValueKHR, signals the event and goes to sleep again.

HansKristian-Work (changes requested):
I think this needs to be rewritten.

- The thread should be spawned on first use where an event actually needs to be signalled.
- Signal requests should be pushed into a work queue. A simple array is good enough here similar to the fence worker threads.
- A work item blocks on the semaphore, then signals events.
- When pushing a fence to the queue, it needs to take a private reference on it.

libs/vkd3d/command.c
K0bin: That's exactly how this works at the moment. This is inside an `if(event)`, so if the event is null, we just go to the `vkWaitSemaphores` call at the end of the function.

libs/vkd3d/command.c
K0bin: Oops, should be 32bit.

libs/vkd3d/command.c
HansKristian-Work: Oh, I see ... Ye, the early return was a bit surprising here.

libs/vkd3d/command.c
K0bin: The worker thread would notice it and update the atomic.

libs/vkd3d/command.c
K0bin: The `LIST_FOR_EACH_ENTRY_SAFE` version copies the list to each step (hence why it has 2 ""iterators""). This is also done somewhere else in the code base, so I'm pretty sure it is safe.

K0bin:
Addressed the review feedback.

libs/vkd3d/command.c
```diff
@@ -1331,12 +1331,32 @@ static ULONG STDMETHODCALLTYPE d3d12_shared_fence_Release(d3d12_fence_iface *ifa
...
+    struct vkd3d_shared_fence_waiting_event *current, *e;
+    struct d3d12_device *device = fence->device;
+    BOOL is_running;
 
     TRACE("%p decreasing refcount to %u.\n", fence, refcount);
 
     if (!refcount)
     {
-        struct d3d12_device *device = fence->device;
+        pthread_mutex_lock(&fence->mutex);
```
HansKristian-Work, line 1342: This needs to happen when private ref-count hits zero, not public ref-count.

libs/vkd3d/command.c
```diff
@@ -1331,12 +1331,32 @@ static ULONG STDMETHODCALLTYPE d3d12_shared_fence_Release(d3d12_fence_iface *ifa
...
+        is_running = fence->is_running;
+        fence->is_running = false;
+        pthread_mutex_unlock(&fence->mutex);
+        if (is_running)
+        {
+            pthread_cond_signal(&fence->cond_var);
```
HansKristian-Work, line 1348: While pthread_cond_signal outside the lock here is technically OK, it is frowned upon:

https://stackoverflow.com/questions/4544234/calling-pthread-cond-signal-without-locking-mutex.

It is best practice to signal the cond variable while holding a lock to the associated mutex.

libs/vkd3d/command.c
```diff
@@ -1405,36 +1425,143 @@ static UINT64 STDMETHODCALLTYPE d3d12_shared_fence_GetCompletedValue(d3d12_fence
...
+        while (fence->is_running && list_empty(&fence->events))
+        {
+            pthread_cond_wait(&fence->cond_var, &fence->mutex);
+        }
+
+        vr = VK_CALL(vkGetSemaphoreCounterValueKHR(fence->device->vk_device, fence->timeline_semaphore, &completed_value));
+        if (vr != VK_SUCCESS)
+        {
+            ERR("Failed to get shared fence counter value, error %d.\n", vr);
+            return NULL;
```
HansKristian-Work, line 1459: Lock is still being held.

libs/vkd3d/command.c
```diff
@@ -1405,36 +1425,143 @@ static UINT64 STDMETHODCALLTYPE d3d12_shared_fence_GetCompletedValue(d3d12_fence
...
+            if (current->value <= completed_value)
+            {
+                vkd3d_native_sync_handle_signal(current->handle);
+                list_remove(&current->entry);
+                vkd3d_free(current);
+            }
+        }
+
+        if (!fence->is_running)
+            return NULL;
```
HansKristian-Work, line 1473: Lock is still being held.

libs/vkd3d/command.c
```diff
@@ -1405,36 +1425,143 @@ static UINT64 STDMETHODCALLTYPE d3d12_shared_fence_GetCompletedValue(d3d12_fence
...
+
+        if (list_empty(&fence->events))
+            continue;
```
HansKristian-Work, line 1476: Lock is still being held.

libs/vkd3d/command.c
```diff
@@ -1405,36 +1425,143 @@ static UINT64 STDMETHODCALLTYPE d3d12_shared_fence_GetCompletedValue(d3d12_fence
...
+        if (completed_value >= value)
+        {
+            vkd3d_native_sync_handle_signal(handle);
+            return S_OK;
+        }
+        else
+        {
+            pthread_mutex_lock(&fence->mutex);
+
+            if (!(waiting_event = vkd3d_malloc(sizeof(*waiting_event))))
```
HansKristian-Work, line 1517: No need to hold lock while allocating.

libs/vkd3d/command.c
```diff
@@ -1405,36 +1425,143 @@ static UINT64 STDMETHODCALLTYPE d3d12_shared_fence_GetCompletedValue(d3d12_fence
...
+
+            if (!fence->is_running)
+            {
+                fence->is_running = true;
+                pthread_create(&fence->thread, NULL, vkd3d_shared_fence_worker_main, fence);
+            }
+
+            pthread_mutex_unlock(&fence->mutex);
+
+            pthread_cond_signal(&fence->cond_var);
```
HansKristian-Work, line 1535: Signal after unlock, bad form.

HansKristian-Work: Also, no need to signal here, the thread will start with an entry to work on.

libs/vkd3d/command.c
```diff
@@ -1331,12 +1331,32 @@ static ULONG STDMETHODCALLTYPE d3d12_shared_fence_Release(d3d12_fence_iface *ifa
 {
     struct d3d12_shared_fence *fence = shared_impl_from_ID3D12Fence1(iface);
     ULONG refcount = InterlockedDecrement(&fence->refcount);
+    struct vkd3d_shared_fence_waiting_event *current, *e;
+    struct d3d12_device *device = fence->device;
+    BOOL is_running;
```
HansKristian-Work, line 1336: why BOOL and not bool?

HansKristian-Work (changes requested):

libs/vkd3d/command.c
```diff
@@ -1405,36 +1430,147 @@ static UINT64 STDMETHODCALLTYPE d3d12_shared_fence_GetCompletedValue(d3d12_fence
...
+        {
+            if (current->value <= completed_value)
+            {
+                vkd3d_native_sync_handle_signal(current->handle);
+                list_remove(&current->entry);
+                vkd3d_free(current);
+            }
+        }
+
+        if (!fence->is_running) {
```
HansKristian-Work, line 1478: :frog:

libs/vkd3d/command.c
```diff
@@ -1405,36 +1430,147 @@ static UINT64 STDMETHODCALLTYPE d3d12_shared_fence_GetCompletedValue(d3d12_fence
...
+            pthread_mutex_lock(&fence->mutex);
+            return NULL;
+        }
+
+        if (list_empty(&fence->events)) {
```
HansKristian-Work, line 1483: brace

HansKristian-Work (approved):
LGTM

Issue #1279, opened 2022-11-04
Age of Empires IV GPU hang with RADV


hakzsam:
Firstly, I bisected to a ACO change which is https://gitlab.freedesktop.org/mesa/mesa/-/commit/0e783d687a3e13284eeae4081f16ee22033ff89b but it turns out it's a game bug. The game does OOB access of private (or scratch) memory in a compute shader.

This Mesa branch https://gitlab.freedesktop.org/hakzsam/mesa/-/commits/nir_bound_check_scratch_aoe4 workarounds the issue by implementing OOB checks for scratch in NIR. I reported a similar issue for Dirt Rally and DXVK earlier this year.

Original Mesa issue (for steps to reproduce) is https://gitlab.freedesktop.org/mesa/mesa/-/issues/7532

I guess vkd3d-proton should implement a workaround.

PR #1285, opened 2022-11-07
Implement robustness for indexable temps in DXBC.

HansKristian-Work:
Should fix https://github.com/HansKristian-Work/vkd3d-proton/issues/1279 when done.

Undefined behavior, but games rely on this :(


doitsujin:
Implementation looks good but not sure if it's a great idea to make this opt-in, since OOB access isn't *that* uncommon and works well on all D3D drivers in practice. DXVK also unconditionally enables bounds checking.

doitsujin (approved):

PR #1292, opened 2022-11-12
build: Declare setup command explicitely

jp7677:
This resolves the following meson warning:

```
WARNING: Running the setup command as `meson [options]` instead of `meson setup [options]` is ambiguous and deprecated.
```


HansKristian-Work (approved):
LGTM

PR #1297, opened 2022-11-17
Implement VK_EXT_descriptor_buffer

HansKristian-Work:
Final PR for descriptor buffers:

- Allocate VkBuffer or malloc instead of CreatePool/AllocateSets.
- Implement NULL descriptor copies and templates as memcpy from static buffers.
- Implement descriptor updates through vkGetDescriptorEXT.
- Update meta shaders to consider PUSH | DESCRIPTOR_BUFFER usage flags where needed.
- Enables the feature.

### DIFF
--- a/libs/vkd3d/device.c
+++ b/libs/vkd3d/device.c
@@ -3070,6 +3070,7 @@ static void d3d12_device_destroy(struct d3d12_device *device)
     d3d12_device_destroy_vkd3d_queues(device);
     vkd3d_memory_allocator_cleanup(&device->memory_allocator, device);
     vkd3d_memory_transfer_queue_cleanup(&device->memory_transfers);
+    vkd3d_global_descriptor_buffer_cleanup(&device->global_descriptor_buffer, device);
     /* Tear down descriptor global info late, so we catch last minute faults after we drain the queues. */
     vkd3d_descriptor_debug_free_global_info(device->descriptor_qa_global_info, device);
 
@@ -6779,8 +6780,11 @@ static HRESULT d3d12_device_init(struct d3d12_device *device,
     if (FAILED(hr = vkd3d_memory_info_init(&device->memory_info, device)))
         goto out_cleanup_format_info;
 
-    if (FAILED(hr = vkd3d_bindless_state_init(&device->bindless_state, device)))
+    if (FAILED(hr = vkd3d_global_descriptor_buffer_init(&device->global_descriptor_buffer, device)))
         goto out_cleanup_memory_info;
+
+    if (FAILED(hr = vkd3d_bindless_state_init(&device->bindless_state, device)))
+        goto out_cleanup_global_descriptor_buffer;
 
     if (FAILED(hr = vkd3d_view_map_init(&device->sampler_map)))
         goto out_cleanup_bindless_state;
@@ -6843,6 +6847,8 @@ out_cleanup_view_map:
     vkd3d_view_map_destroy(&device->sampler_map, device);
 out_cleanup_bindless_state:
     vkd3d_bindless_state_cleanup(&device->bindless_state, device);
+out_cleanup_global_descriptor_buffer:
+    vkd3d_global_descriptor_buffer_cleanup(&device->global_descriptor_buffer, device);
 out_cleanup_memory_info:
     vkd3d_memory_info_cleanup(&device->memory_info, device);
 out_cleanup_format_info:


libs/vkd3d/resource.c
```diff
@@ -6704,3 +7151,75 @@ HRESULT vkd3d_memory_info_init(struct vkd3d_memory_info *info,
...
+HRESULT vkd3d_global_descriptor_buffer_init(struct vkd3d_global_descriptor_buffer *global_descriptor_buffer,
+        struct d3d12_device *device)
+{
+    VkBufferUsageFlags vk_usage_flags;
+    HRESULT hr;
+
+    /* Don't bother with descriptor buffers if we need to keep offset buffer around. */
+    if (!device->device_info.descriptor_buffer_features.descriptorBuffer ||
+            !device->device_info.descriptor_buffer_features.descriptorBufferPushDescriptors ||
+            device->device_info.properties2.properties.limits.minStorageBufferOffsetAlignment > 4)
```
werman, line 7164: Tried testing Turnip's implementation of `VK_EXT_descriptor_buffer` with vkd3d, but Turnip has `minStorageBufferOffsetAlignment == 64` so we don't pass this check (unfortunately it's a hardware limit, in a descriptor the buffer address is encoded without 5 lowest bits).

> /* Don't bother with descriptor buffers if we need to keep offset buffer around. */

So the high alignment values are worked around by having a separate buffer with per-descriptor offsets, and keeping the offsets buffer around isn't worth due to perf, code complexity, or both?

HansKristian-Work: 64?!?!?! Dear lord. How would this work on DX12 then? I think it would "just work" to use offset buffers here, it's just not tested.

HansKristian-Work: If you can remove the check and stuff is running, we can remove the check.

werman: With  `VK_EXT_descriptor_buffer` enabled there are two additional tests failures: `test_undefined_read_typed_buffer_as_untyped_dxbc`  and `test_bindless_uav_counter_sm51`

libs/vkd3d/state.c
```diff
@@ -5202,8 +5355,10 @@ static uint32_t vkd3d_bindless_state_get_bindless_flags(struct d3d12_device *dev
             flags |= VKD3D_SSBO_OFFSET_BUFFER;
     }
 
-    /* Always use a typed offset buffer. Otherwise, we risk ending up with unbounded size on view maps. */
-    flags |= VKD3D_TYPED_OFFSET_BUFFER;
+    /* Always use a typed offset buffer. Otherwise, we risk ending up with unbounded size on view maps.
+     * Fortunately, we can place descriptors directly if we have descriptor buffers, so this is not required. */
+    if (!d3d12_device_uses_descriptor_buffers(device))
```
werman, line 5360: ` || device_info->properties2.properties.limits.minStorageBufferOffsetAlignment > 4`

libs/vkd3d/resource.c
```diff
@@ -4644,14 +4788,41 @@ static void vkd3d_create_buffer_uav(vkd3d_cpu_descriptor_va_t desc_va, struct d3
...
+            addr_info.sType = VK_STRUCTURE_TYPE_DESCRIPTOR_ADDRESS_INFO_EXT;
+            addr_info.pNext = NULL;
+            addr_info.address = d.view->info.buffer.va;
+            addr_info.range = d.view->info.buffer.range;
+            addr_info.format = vkd3d_internal_get_vk_format(device, d.view->info.buffer.dxgi_format);
+            /* If we really intended to emit raw buffers, the fallback will be inferred as R32_UINT. */
+            if (addr_info.format == VK_FORMAT_UNDEFINED)
+                addr_info.format = VK_FORMAT_R32_UINT;
+            payload = d3d12_descriptor_heap_get_mapped_payload(d.heap, binding.set, d.offset);
+            VK_CALL(vkGetDescriptorEXT(device->vk_device, &get_info, payload.stride, payload.ptr));
```
werman, line 4809: `bound_range` should be updated.

werman (commented):
Making mentioned changes allows `test_bindless_uav_counter_sm51` to pass on Turnip, though `test_undefined_read_typed_buffer_as_untyped_dxbc` still fails.

I think I missed some update of the offset buffer, because GPU packets are the same with and without `VK_EXT_descriptor_buffer` extension, so the difference should be somewhere in buffers.

HansKristian-Work:
Actually, with SSBO alignment  > 16, we cannot even use SSBOs, due to robustness rules for byte address buffer, so we should always be falling back to texel buffers (and block 16-bit support).

cwabbott0:
Seems there's a start offset field we missed in the descriptor, so we should be able to do 4-byte alignment for SSBOs actually. No idea why the blob vulkan driver doesn't use it.

HansKristian-Work:
> Seems there's a start offset field we missed in the descriptor, so we should be able to do 4-byte alignment for SSBOs actually. No idea why the blob vulkan driver doesn't use it.

Why on earth doesn't the descriptor just add those bits to the VA then? :|

cwabbott0:
> > Seems there's a start offset field we missed in the descriptor, so we should be able to do 4-byte alignment for SSBOs actually. No idea why the blob vulkan driver doesn't use it.
> 
> Why on earth doesn't the hardware descriptor layout just add those bits to the VA then? :| Oh well. 4-byte SSBO alignment would be ideal for sure.

The alignment for images/textures (i.e. everything other than `texelBuffer` and storage buffers) is 64 bytes, and the offset field only appears if you set the `BUFFER` type in the descriptor, but there's this sort-of tacked on `TEXELOFFSET` thing that just adds to the texel offset specified in the shader, going where the image pitch normally would be. Dunno why they needed to add that...

doitsujin (approved):

PR #1303, opened 2022-11-21
vkd3d: Remove IdentifierCreateInfo in late compile.

ishitatsuyuki:
Fixes validation warning.

Signed-off-by: Tatsuyuki Ishi <ishitatsuyuki@gmail.com>

### DIFF
--- a/libs/vkd3d/state.c
+++ b/libs/vkd3d/state.c
@@ -2474,6 +2474,11 @@ static HRESULT vkd3d_late_compile_shader_stages(struct d3d12_pipeline_state *sta
             {
                 break;
             }
+
+            /* When a shader module is specified, VkPipelineShaderStageModuleIdentifierCreateInfoEXT must have an
+             * identifierSize of 0 or be absent. */
+            vk_remove_struct(&graphics->stages[i],
+                    VK_STRUCTURE_TYPE_PIPELINE_SHADER_STAGE_MODULE_IDENTIFIER_CREATE_INFO_EXT);
         }
 
         /* We'll keep the module around here, no need to keep code/size pairs around for this.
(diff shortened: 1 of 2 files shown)


libs/vkd3d/state.c
```diff
@@ -2474,6 +2488,9 @@ static HRESULT vkd3d_late_compile_shader_stages(struct d3d12_pipeline_state *sta
             {
                 break;
             }
+
+            // Vulkan requires identifier to be not present when shader module is specified.
+            vkd3d_shader_stage_remove_identifier_pnext(&graphics->stages[i]);
```
HansKristian-Work, line 2493: This is not true, identifierSize must be 0 instead.

ishitatsuyuki: You're right. Should I just update the comment or also modify the helper to set size to 0?

HansKristian-Work: Just setting identifierSize to 0 is problematic too actually, because that identifier might be read by the disk cache threads. Let's go with pNext removal instead.

libs/vkd3d/state.c
```diff
@@ -2430,6 +2430,20 @@ static bool vkd3d_shader_stages_require_work_locked(struct d3d12_pipeline_state
     return false;
 }
 
+static void vkd3d_shader_stage_remove_identifier_pnext(VkPipelineShaderStageCreateInfo *stage)
+{
+    VkBaseInStructure **chain;
+
+    for (chain = (VkBaseInStructure **)&stage->pNext; chain; chain = (VkBaseInStructure **)&(*chain)->pNext)
```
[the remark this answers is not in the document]
HansKristian-Work: This double-pointer for loop with casting is very hard to read, there should be a way to simplify it a bit.

libs/vkd3d/state.c
HansKristian-Work: Can update the comment to reflect this.

libs/vkd3d/state.c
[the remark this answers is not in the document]
HansKristian-Work: Might as well make it more generic by taking removal sType, and `VkBaseInStructure *` as input.

libs/vkd3d/state.c
```diff
@@ -2643,6 +2660,9 @@ static HRESULT vkd3d_create_compute_pipeline(struct d3d12_pipeline_state *state,
                 spirv_code)))
             return hr;
 
+        // Vulkan requires identifier to be not present when shader module is specified.
```
HansKristian-Work, line 2663: C89 comments.

libs/vkd3d/state.c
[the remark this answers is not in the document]
ishitatsuyuki: I did another try, please see `vk_remove_struct`.

HansKristian-Work (approved):
LGTM

PR #1306, opened 2022-11-22
Refactor how we allocate ReBAR memory.

### LINKED ISSUE #1273: games slowing down with reBAR enabled
i have a RX 560 that has only 4 GB's of VRAM, when reBAR is enabled all my dx12 games slow down when: switching scenarios or changing graphics settings

here is RE village with rebar enabled in bios:
rebarenabled

here is with rebar disabled: 
norebar

here is with rebar enabled but with no_upload_hvv:
noupload

edit1: theres also massive performance loss when changing a game from fullscreen to borderless full screen with rebar enabled

HansKristian-Work:
For descriptor buffers to work optimally, we need to consider:

- Applying the ReBAR budget is problematic. We really want descriptors in ReBAR regardless of what app did with UPLOAD heaps.
- Remove the global mask. Allows some allocations to use BAR and some to skip.
- Add ~5GB limit. Avoids OOMing 4 GB cards in some cases where they work with no_upload_hvv.
- Refactor memory properties for UPLOAD / descriptor heaps to be resolved up front.
- Rename vkd3d_allocate_buffer_memory is only used for internal private allocations, so clarify that in code.
- For smol BAR, do not use BAR for UPLOAD heap, but allow it for internal descriptor allocations.

Fix #1273.

### DIFF
--- a/libs/vkd3d/memory.c
+++ b/libs/vkd3d/memory.c
@@ -1566,31 +1566,3 @@ HRESULT vkd3d_allocate_buffer_memory(struct d3d12_device *device, VkBuffer vk_bu
 
     return hr;
 }
-
-HRESULT vkd3d_allocate_image_memory(struct d3d12_device *device, VkImage vk_image,
-        VkMemoryPropertyFlags type_flags,
-        struct vkd3d_device_memory_allocation *allocation)
-{
-    const struct vkd3d_vk_device_procs *vk_procs = &device->vk_procs;
-    VkMemoryRequirements memory_requirements;
-    VkBindImageMemoryInfo bind_info;
-    VkResult vr;
-    HRESULT hr;
-
-    VK_CALL(vkGetImageMemoryRequirements(device->vk_device, vk_image, &memory_requirements));
-
-    if (FAILED(hr = vkd3d_allocate_device_memory(device, memory_requirements.size,
-            type_flags, memory_requirements.memoryTypeBits, NULL, allocation)))
-        return hr;
-
-    bind_info.sType = VK_STRUCTURE_TYPE_BIND_IMAGE_MEMORY_INFO;
-    bind_info.pNext = NULL;
-    bind_info.image = vk_image;
-    bind_info.memory = allocation->vk_memory;
-    bind_info.memoryOffset = 0;
-
-    if (FAILED(vr = VK_CALL(vkBindImageMemory2KHR(device->vk_device, 1, &bind_info))))
-        return hresult_from_vk_result(vr);
-
-    return hr;
-}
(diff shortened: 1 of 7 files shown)


kassindornelles:
> Add ~5GB limit. Avoids OOMing 4 GB cards in some cases where they work with no_upload_hvv.

thx 🐸 👍

libs/vkd3d/memory.c
```diff
@@ -173,14 +169,10 @@ void vkd3d_free_device_memory(struct d3d12_device *device, const struct vkd3d_de
 }
 
 static HRESULT vkd3d_try_allocate_device_memory(struct d3d12_device *device,
-        VkDeviceSize size, VkMemoryPropertyFlags type_flags, uint32_t base_type_mask,
-        void *pNext, struct vkd3d_device_memory_allocation *allocation)
+        VkDeviceSize size, VkMemoryPropertyFlags type_flags, const uint32_t base_type_mask,
+        void *pNext, bool budgeted, struct vkd3d_device_memory_allocation *allocation)
```
ishitatsuyuki, line 173: Not entirely sure but `budgeted` feels like the wrong word (IIUC it means something that was drafted into the budget). Maybe `use_budget`?

libs/vkd3d/resource.c
```diff
@@ -6430,47 +6426,92 @@ static void vkd3d_memory_info_get_topology(struct vkd3d_memory_topology *topolog
...
+            return VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT |
+                    VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT |
+                    VK_MEMORY_PROPERTY_HOST_COHERENT_BIT;
         }
     }
-
-    return ~mask;
+    else
+    {
+        INFO("Topology: Device heaps are split. Assuming small BAR situation. Using DEVICE_LOCAL | HOST_COHERENT.\n");
```
ishitatsuyuki, line 6512: Should we have a debug flag (the opposite to `no_upload_hvv`) to force HVV use even on small BAR systems?

ishitatsuyuki (commented):
Nice changes!

It's great to no longer have upload usage fighting with descriptors on small BAR system. Might want to get this documented / announced in some way though. (It wasn't obvious that BAR uploads are now effectively disabled for small BAR system.)

HansKristian-Work:
> (It wasn't obvious that BAR uploads are now effectively disabled for small BAR system.)

I believe that was always the case, using the global_mask.

libs/vkd3d/resource.c
HansKristian-Work: I don't think so.

ishitatsuyuki (approved):
LGTM

PR #1307, opened 2022-11-25
Swapchain rewrite - Phase 2

HansKristian-Work:
This is the actual implementation. It wasn't particularly feasible to split this up unfortunately, but I hope I've made the implementation far more readable than the previous one.

The core idea behind this implementation is that Present() just queues up work using the queue callback mechanism.

This task will try its best to perform a present operation which is the usual:
- AcquireNextImage
- Blit
- Present

There are ways for the main thread to monitor if the queued present tasks are complete. This is particularly useful for ResizeBuffers, where we need to make sure that all blit operations have completed on GPU. This is nice since we don't need a full QueueWaitIdle anymore to do this.

On the end of each present task, we are required to signal latency semaphores. There are two mechanisms here:
- If present wait is supported, and we're using FIFO mode, push a wait ID to the wait thread, which will then signal latency semaphores. If we are in a fallback situation, we do the usual CommandQueue::Signal(fence) approach we used in legacy swapchain. For IMMEDIATE, this is a reasonable approach anyways.

- On top of the application frame latency handle, we employ our own latency semaphore. Using VKD3D_SWAPCHAIN_LATENCY_FRAMES env-variable, we can force a specific maximum latency. This aims to mimic the approximate latency we used to get in the legacy swapchain where we did eager vkAcquireNextImageKHR. For minImageCount == 3 -> approx 2 frames latency.

- If presentWait is not supported, we end up with a similar path that we used to get in the legacy swapchain. A Present() will block until we have observed that the task has completed. This does not really solve the deadlock situation, but to combat this there is a reasonably small timeout on the wait. It's never a problem not to wait, it just risks larger latency. Waiting on the present task is only a way to keep latency under control since we no longer do the present inline in a blocking fashion. Having a small-ish timeout can also avoid problems where the present task takes a long time when resizing or changing present modes.

The new swapchain also works fine on Linux native, so it's always included in the build. Following PR will refactor the demos and remove vkd3d-utils to be able to use the new DXGIVk interfaces directly.


libs/vkd3d/swapchain_new.c
```diff
@@ -0,0 +1,1785 @@
...
+        VK_CALL(vkAllocateCommandBuffers(vk_device, &allocate_info,
+                    &chain->present.vk_blit_command_buffers[swapchain_index]));
+    }
+
+    if (chain->present.vk_blit_fences[swapchain_index])
+    {
+        vr = VK_CALL(vkWaitForFences(vk_device, 1, &chain->present.vk_blit_fences[swapchain_index], VK_TRUE, UINT64_MAX));
+        VKD3D_DEVICE_REPORT_BREADCRUMB_IF(chain->queue->device, vr == VK_ERROR_DEVICE_LOST);
+        if (vr < 0)
+            return false;
```
doitsujin, line 1442: this seems a tad inconsistent with error messages, we show one when semaphore creation above fails but not here or below in case fence creation goes wrong.

I guess this will only ever fail on `DEVICE_LOST`,  but might be worth priting anyway.

doitsujin (commented):
This is obviously a lot of code all at once but makes sense at a first glance. Nit wrt some error messages, we should probably be verbose when things fail unexpectedly.

Joshua-Ashton (approved):
Looks fine in theory. I am sure we will probably find things later though as it's a lot of code :frog:

PR #1321, opened 2022-12-05
Implement VK_EXT_descriptor_buffer - Phase 3

HansKristian-Work:
Reimplements SetDescriptorHeaps.

Rather than binding N descriptor sets, binds descriptor buffers, and sets up N offsets.
The N offsets are always bound in one go. We don't have to consider NULL descriptor set or any holes here since we always bind a descriptor buffer (fallback to global), so we don't need the bit iteration we have to do for plain descriptor sets.

Also adds a descriptor buffer path for setting up descriptor set layouts. The only real difference here is that we just use VARIABLE_COUNT instead of UPDATE_AFTER_BIND, drop the host layout entirely and reuse the VALVE_host_mapping path to set up mapping offsets and strides.

Also, set up null descriptor templates, which is just queried once per descriptor types which we can then rapidly memcpy in as needed.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -4761,6 +4761,17 @@ bool d3d12_command_list_reset_query(struct d3d12_command_list *list,
     return true;
 }
 
+static void d3d12_command_list_init_default_descriptor_buffers(struct d3d12_command_list *list)
+{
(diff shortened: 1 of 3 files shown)


doitsujin (approved):
Haven't tested, but LGTM.

PR #1324, opened 2022-12-07
Add more extensive validation of SPIR-V capabilities against exposed features

HansKristian-Work:


doitsujin:
Maybe it's a better idea to scan the OpCapability instructions of the compiled shaders instead; that way we'd get this for both DXIL an DXBC shaders and this should also work fine since SPV capabilities are usually tied directly to Vulkan features. Technically there's a **lot** we need to check for:
- 64/16 bit support
- Sparse stuff
- Stencil export
- Layer export without GS
- ROV stuff once we implement those
- And all the optional stuff in DXIL

HansKristian-Work:
I guess we can technically do that too.

djdeath:
I can confirm this gets rid of unexpected stuff in our NIR lowering code.

doitsujin (approved):

PR #1331, opened 2022-12-12
Rewrite ReadFromSubresource / WriteToSubresource (attempt 2)

doitsujin:
Different approach this time which should work around all the aliasing issues, but increases VRAM usage (rather than system memory usage, like the current implementation).

This is still somewhat buggy with `ReadFromSubresource` on UAVs with `SIMULTANEOUS_ACCESS`, since we can't reliably detect UAV writes on the CPU side. We'll print a warning for now; if we ever find an app relying on this, we can dispatch a copy and do a GPU sync at `ReadFromSubresource` time, however, if we do that, it should be opt-in due to the performance impact.

Tested on AMD in a number of games (not Deathloop though). Haven't tested Nvidia yet.


libs/vkd3d/memory.c
```diff
@@ -309,28 +345,16 @@ static void vkd3d_memory_transfer_queue_clear_allocation(struct vkd3d_memory_tra
...
-        pthread_mutex_lock(&queue->mutex);
-
-        if (!vkd3d_array_reserve((void**)&queue->allocations, &queue->allocations_size,
-                queue->allocations_count + 1, sizeof(*queue->allocations)))
-        {
-            ERR("Failed to insert free range.\n");
-            pthread_mutex_unlock(&queue->mutex);
-            return;
-        }
-
```
HansKristian-Work, line 321: We're sampling queue->next_signal_value here outside the lock, is that safe?

HansKristian-Work: It might be safer to take lock manually then vkd3d_memory_transfer_queue_execute_transfer_locked() instead.

libs/vkd3d/resource.c
```diff
@@ -875,6 +879,14 @@ HRESULT vkd3d_get_image_allocation_info(struct d3d12_device *device,
         allocation_info->Alignment = target_alignment;
     }
 
+    /* We can't know at this point whether the image is going to be CPU-visible
+     * or not, so always make sure we reserve enough space to store the linear
+     * subresource data as well. */
+    linear_size = vkd3d_compute_resource_layouts_from_desc(device, desc, NULL);
```
HansKristian-Work, line 827: I suppose the only plausible scenario for linear_size to be > optimal size is for D24S8? I suppose since we have to split the planes we end up with 32-bit + 8-bit?

libs/vkd3d/heap.c
```diff
@@ -80,7 +85,10 @@ static ULONG STDMETHODCALLTYPE d3d12_heap_Release(d3d12_heap_iface *iface)
     TRACE("%p decreasing refcount to %u.\n", heap, refcount);
 
     if (!refcount)
-        d3d12_heap_destroy(heap);
+    {
+        d3d12_heap_decref(heap);
+        d3d12_device_release(heap->device);
```
HansKristian-Work, line 90: Bug to access heap->device after d3d12_heap_decref.

libs/vkd3d/resource.c
doitsujin: depth-stencil isn't actually allowed by the API here.

We shouldn't *really* be hitting this in practice with any of the formats we support, it's more a safeguard in case the driver is doing something really weird behind our backs. Can remove if you want.

HansKristian-Work: Is separate depth/stencil allowed?

libs/vkd3d/memory.c
doitsujin: Where is that access exactly?

HansKristian-Work: Right below here.

```
        allocation->clear_semaphore_value = queue->next_signal_value;

        if (allocation->chunk)
            allocation->chunk->allocation.clear_semaphore_value = queue->next_signal_value;
```

next_signal_value can bump if another thread flushes the clear queue I think?

doitsujin: nvm github just made this hard to find.

libs/vkd3d/heap.c
doitsujin: fixed

libs/vkd3d/command.c
```diff
@@ -3559,11 +3559,129 @@ static VkPipelineStageFlags vk_render_pass_barrier_from_view(struct d3d12_comman
...
+        for (i = 0; i < list->subresource_tracking_count; )
+        {
+            entry = &list->subresource_tracking[i];
+
+            if (entry->resource == resource && entry->subresource.aspectMask == aspect && entry->subresource.mipLevel == subresource.mipLevel)
+            {
+                if (entry->subresource.baseArrayLayer <= subresource.baseArrayLayer &&
+                        entry->subresource.baseArrayLayer + entry->subresource.layerCount >= subresource.baseArrayLayer + subresource.layerCount)
+                {
+                    return;
```
HansKristian-Work, line 3585: Should this really be a return? If there are multiple aspects we might get a full coverage here, but not in other aspects.

libs/vkd3d/command.c
```diff
@@ -3559,11 +3559,129 @@ static VkPipelineStageFlags vk_render_pass_barrier_from_view(struct d3d12_comman
...
+
+        VK_CALL(vkCmdCopyImageToBuffer2KHR(list->vk_command_buffer, &copy_info));
+    }
+
+    list->subresource_tracking_count = 0;
+
+    /* Do not emit another barrier here, let the caller handle that as
+     * necessary. If this is used at the end of a command buffer, no
+     * barrier is necessary as submissions will emit a host barrier. */
+    return true;
```
HansKristian-Work, line 3674: Is the return value ever checked?

libs/vkd3d/resource.c
```diff
@@ -682,6 +675,9 @@ static HRESULT vkd3d_get_image_create_info(struct d3d12_device *device,
             (image_info->usage & VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT))
         image_info->flags |= VK_IMAGE_CREATE_2D_ARRAY_COMPATIBLE_BIT;
 
+    if (heap_properties && is_cpu_accessible_heap(heap_properties))
+        use_concurrent = true;
```
HansKristian-Work, line 679: Is this meaningful code? It's overwritten right after.

libs/vkd3d/resource.c
```diff
@@ -711,53 +707,10 @@ static HRESULT vkd3d_get_image_create_info(struct d3d12_device *device,
...
-        image_info->initialLayout = VK_IMAGE_LAYOUT_UNDEFINED;
-    }
-
     if ((image_info->flags & VK_IMAGE_CREATE_MUTABLE_FORMAT_BIT) &&
             !vkd3d_format_check_usage_support(device, format->vk_format, image_info->usage, image_info->tiling))
         image_info->flags |= VK_IMAGE_CREATE_EXTENDED_USAGE_BIT;
 
-    if (image_info->tiling == VK_IMAGE_TILING_LINEAR)
-    {
-        bool supported = vkd3d_is_linear_tiling_supported(device, image_info);
```
HansKristian-Work, line 740: Do we remember to remove this function? It should be dead code at this point.

HansKristian-Work: In a similar vein, do we fix up memory_info for host visible to only consider that we need to place VkBuffers?

libs/vkd3d/resource.c
```diff
@@ -6829,13 +6829,6 @@ HRESULT vkd3d_memory_info_init(struct vkd3d_memory_info *info,
     info->cpu_accessible_domain.sampled_type_mask = sampled_type_mask_cpu & host_visible_mask;
     info->cpu_accessible_domain.rt_ds_type_mask = rt_ds_type_mask_cpu & host_visible_mask;
 
-    /* If we cannot support linear render targets, this is fine.
```
HansKristian-Work, line 6891: I think info->cpu_accessible_domain for sampled and rt_ds_type_mask just need to inherit buffer_type_mask.

libs/vkd3d/resource.c
```diff
@@ -631,6 +631,8 @@ static HRESULT vkd3d_get_image_create_info(struct d3d12_device *device,
 
     image_info->mipLevels = min(desc->MipLevels, max_miplevel_count(desc));
     image_info->samples = vk_samples_from_dxgi_sample_desc(&desc->SampleDesc);
+    image_info->tiling = VK_IMAGE_TILING_OPTIMAL;
```
HansKristian-Work, line 634: Do we check for depth-stencil usage + host image and report a FIXME and E_NOTIMPL?

libs/vkd3d/resource.c
HansKristian-Work: Hitting `Found no suitable memory type for requested type flags #6` on NV with this PR.

HansKristian-Work (changes requested):
Comments.

libs/vkd3d/command.c
doitsujin: oversight, this didn't iterate over the aspects at first.

HansKristian-Work (approved):
LGTM, thanks!

PR #1336, opened 2022-12-14
vkd3d: Clamp maxAnisotropy.

K0bin:
Witcher 3 uses a maxAnisotropy value of 0.0 which is illegal and crashes the Nvidia Vulkan driver.

Docs:
> Clamping value used if D3D12_FILTER_ANISOTROPIC or D3D12_FILTER_COMPARISON_ANISOTROPIC is specified in Filter. Valid values are between 1 and 16.

Alternatively we could also disable anisotropy when a game passes a value of 0.0.


doitsujin:
> Alternatively we could also disable anisotropy when a game passes a value of 0.0.

That's probably a good idea, an anisotropy value of 1 isn't meaningful to begin with.

HansKristian-Work (approved):

PR #1337, opened 2022-12-14
vkd3d: Account for null root signature in SetDescriptorTable.

doitsujin:
Looks like we can't assume that any root signature is active by the time the app changes root parameters.

This is technically a problem for `set_root_descriptor` as well, but we haven't run into issues there yet.


HansKristian-Work:
Technically this is a spec violation, but if it's easy to work around and doesn't crash real drivers, sure.

PR #1343, opened 2022-12-16
Various fixes for Witcher 3

K0bin:
Most of this just fixes validation errors that seem to be benign in practice.

586d228 fixes Hairworks.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -8956,7 +8956,7 @@ static void STDMETHODCALLTYPE d3d12_command_list_SOSetTargets(d3d12_command_list
             list->so_counter_buffers[start_slot + i] = VK_NULL_HANDLE;
             list->so_counter_buffer_offsets[start_slot + i] = 0;
 
-            WARN("Trying to unbind transform feedback buffer %u. Ignoring.\n", start_slot + i);
+            TRACE("Trying to unbind transform feedback buffer %u. Ignoring.\n", start_slot + i);
         }
     }
 
@@ -9017,7 +9017,7 @@ static void STDMETHODCALLTYPE d3d12_command_list_OMSetRenderTargets(d3d12_comman
 
         if (!rtv_desc || !rtv_desc->resource)
         {
-            WARN("RTV descriptor %u is not initialized.\n", i);
+            TRACE("RTV descriptor %u is not initialized.\n", i);
             continue;
         }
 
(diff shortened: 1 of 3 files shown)


libs/vkd3d/command.c
```diff
@@ -8956,7 +8956,7 @@ static void STDMETHODCALLTYPE d3d12_command_list_SOSetTargets(d3d12_command_list
             list->so_counter_buffers[start_slot + i] = VK_NULL_HANDLE;
             list->so_counter_buffer_offsets[start_slot + i] = 0;
 
-            WARN("Trying to unbind transform feedback buffer %u. Ignoring.\n", start_slot + i);
+            INFO("Trying to unbind transform feedback buffer %u. Ignoring.\n", start_slot + i);
```
HansKristian-Work, line 8961: Should be TRACE. INFO is for affirming that things are enabled, etc.

libs/vkd3d/command.c
```diff
@@ -6537,31 +6537,33 @@ static void vk_image_copy_from_d3d12(VkImageCopy2KHR *image_copy,
...
+    image_copy->dstOffset.x = min(dst_x, dstExtent.width);
+    image_copy->dstOffset.y = min(dst_y, dstExtent.height);
+    image_copy->dstOffset.z = min(dst_z, dstExtent.depth);
     if (src_box)
     {
-        image_copy->extent.width = src_box->right - src_box->left;
-        image_copy->extent.height = src_box->bottom - src_box->top;
-        image_copy->extent.depth = src_box->back - src_box->front;
+
+        image_copy->extent.width = min(src_box->right - src_box->left, min(srcExtent.width, dstExtent.width));
```
HansKristian-Work, line 6561: Shouldn't it be `min(srcExtent.width - srcOffset.x, dstExtent.width - dstOffset.x)`?

K0bin: That would ignore the `src_box` parameter.

HansKristian-Work: No, I mean `min(src_box->right, min(srcExtent.width, dstExtent.width))) - src_box->left`. If offset is not != 0, the extent cannot be the full miplevel width/height.

HansKristian-Work: Either way, this particular commit seems a bit too hairy I'm not comfortable merging it blind without more testing unless it fixes a real crash.

K0bin: Okay, I'll drop it.

HansKristian-Work: A draft PR with that commit would be good so we can revisit it later.

HansKristian-Work (approved):

PR #1353, opened 2022-12-21
vkd3d: Set descriptor_size for each entry point.

K0bin:
Fixes out of bounds descriptor heap accesses in Witcher 3 with ray tracing enabled.

Unfortunately, it still hangs. :(


K0bin:
The remaining hang seems to come down to a missing TRANSFER->ALL_GRAPHICS barrier somewhere. If I insert such a barrier at the end of `ResourceBarrier`, it works. Doing a full fat ALL_COMMANDS, MEMORY_READ/MEMORY_WRITE barrier before and after every `Dispatch` and `DispatchRays` call did not fix it.

PR #1368, opened 2023-01-07
vkd3d: Take into account heap offset for image offset alignment.

BNieuwenhuizen:
We can't make the assumption that the heap will be aligned to the image alignment in case the image alignment needs to be larger than the default alignment.

Signed-off-by: Bas Nieuwenhuizen <bas@basnieuwenhuizen.nl>


Joshua-Ashton (approved):
Makes sense

PR #1372, opened 2023-01-12
Witcher 3 RT barrier workaround

K0bin:
This fixes a GPU hang when loading into the game with ray tracing enabled.

I've tried a lot of different things including lots of different tracking within a single command list and none of it worked.
I also noticed that they create the command lists to build acceleration structures on different threads, so my best guess is that they don't synchronize properly *across* command lists.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -11468,6 +11468,8 @@ static void STDMETHODCALLTYPE d3d12_command_list_BuildRaytracingAccelerationStru
     build_info.build_info.scratchData.deviceAddress = desc->ScratchAccelerationStructureData;
 
     d3d12_command_list_end_current_render_pass(list, true);
+    d3d12_command_list_end_transfer_batch(list);
+
     VK_CALL(vkCmdBuildAccelerationStructuresKHR(list->vk_command_buffer, 1,
             &build_info.build_info, build_info.build_range_ptrs));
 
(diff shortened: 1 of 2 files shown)


libs/vkd3d/command.c
```diff
@@ -8291,6 +8304,8 @@ static void STDMETHODCALLTYPE d3d12_command_list_ResourceBarrier(d3d12_command_l
                 if (!preserve_resource || !d3d12_resource_is_acceleration_structure(preserve_resource))
                     state_mask |= D3D12_RESOURCE_STATE_UNORDERED_ACCESS;
 
+                list->has_unsynced_accel_struct_build = false;
```
HansKristian-Work, line 8307: It's also possible to use TRANSITION barriers.

libs/vkd3d/command.c
```diff
@@ -11523,6 +11540,8 @@ static void STDMETHODCALLTYPE d3d12_command_list_CopyRaytracingAccelerationStruc
 
     d3d12_command_list_end_current_render_pass(list, true);
     d3d12_command_list_end_transfer_batch(list);
+    list->has_unsynced_accel_struct_build = true;
```
HansKristian-Work, line 11543: Does CopyRTAS use scratch?

libs/vkd3d/command.c
```diff
@@ -4803,6 +4803,18 @@ static HRESULT STDMETHODCALLTYPE d3d12_command_list_Close(d3d12_command_list_ifa
     /* If there are pending subresource updates, execute them now that all other operations have completed */
     d3d12_command_list_flush_subresource_updates(list);
 
+    if (list->has_unsynced_accel_struct_build)
```
HansKristian-Work, line 4806: I'm not comfortable adding a hack like this unconditionally. If anything, has_unsynced_accel_struct_build should be in an anonymous "workaround" struct, to make it clear what the intention is. It should also be gated behind a vkd3d_config if anything.

PR #1374, opened 2023-01-12
Added Agility SDK 1.608 level feature checks.

sherief:
A Win32 app on Win32 bundling Agility SDK 1.608 would see feature checks for up to OPTIONS16 never fail, but currently when running that app under Proton checks for anything beyond OPTIONS11 would return a failure HRESULT - something that never happens on Win32. This patch prevents seemingly unreachable code paths from triggering under Proton.


Joshua-Ashton (changes requested):

Joshua-Ashton (approved):
LGTM!

HansKristian-Work:
Before merging this, do you consent to this being re-licensed to MIT?

sherief:
I consent to re-licensing this to MIT or Public Domain, whichever you prefer.

PR #1379, opened 2023-01-16
ASan fixes

ishitatsuyuki:
One leak on error path and one memcpy against NULL issue (technically UB), found with ASan while running tests.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -12436,8 +12436,9 @@ static void STDMETHODCALLTYPE d3d12_command_queue_ExecuteCommandLists(ID3D12Comm
         for (i = 0; i < command_list_count; ++i)
         {
             cmd_list = unsafe_impl_from_ID3D12CommandList(command_lists[i]);
-            memcpy(transitions, cmd_list->init_transitions,
-                   cmd_list->init_transitions_count * sizeof(*transitions));
+            if (cmd_list->init_transitions_count)
+                memcpy(transitions, cmd_list->init_transitions,
+                        cmd_list->init_transitions_count * sizeof(*transitions));
             transitions += cmd_list->init_transitions_count;
         }
     }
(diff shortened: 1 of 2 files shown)


libs/vkd3d/command.c
```diff
@@ -12436,8 +12436,9 @@ static void STDMETHODCALLTYPE d3d12_command_queue_ExecuteCommandLists(ID3D12Comm
         for (i = 0; i < command_list_count; ++i)
         {
             cmd_list = unsafe_impl_from_ID3D12CommandList(command_lists[i]);
-            memcpy(transitions, cmd_list->init_transitions,
-                   cmd_list->init_transitions_count * sizeof(*transitions));
+            if (cmd_list->init_transitions_count)
```
HansKristian-Work, line 12439: This is a false positive isn't it?

ishitatsuyuki: I did see a runtime warning when I discovered this. Looking at the code, `init_transitions_count` is initialized to NULL and reserved on demand with `vkd3d_array_reserve`, so I think it's indeed possible to go through this code path with a NULL pointer when there are multiple command lists and one of them doesn't have init transitions.

If "false positive" means calling memcpy with src=NULL and len=0, it's technically considered UB, per https://stackoverflow.com/questions/5243012/is-it-guaranteed-to-be-safe-to-perform-memcpy0-0-0.

HansKristian-Work: ... welp. I suppose this means we likely have similar "bugs" elsewhere as well. Did ASan just trip up here?

ishitatsuyuki: Yes, I just reran the entire test suite with `./build-native/tests/d3d12` on this branch and there seems to be nothing else that trips ASan.

HansKristian-Work: Ok, can't hurt to merge this I guess.

PR #1391, opened 2023-01-25
vkd3d: Hold reference to ID3D12Heap in placed resource.

HansKristian-Work:
Apparently, we are required to do this (but not for reserved resources). Docs seem to have been updated at some point to require this.

### DIFF
--- a/libs/vkd3d/resource.c
+++ b/libs/vkd3d/resource.c
@@ -2594,6 +2594,8 @@ static void d3d12_resource_destroy(struct d3d12_resource *resource, struct d3d12
         VK_CALL(vkDestroyImageView(device->vk_device, resource->vrs_view, NULL));
 
     vkd3d_private_store_destroy(&resource->private_store);
+    if (resource->heap)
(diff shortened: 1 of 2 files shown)


Joshua-Ashton (approved):

doitsujin (approved):

HansKristian-Work:
fwiw, verified on native that there is no public refcount here on Agility SDK.

libs/vkd3d/resource.c
```diff
@@ -3000,7 +3002,9 @@ HRESULT d3d12_resource_create_placed(struct d3d12_device *device, const D3D12_RE
             &heap->desc.Properties, heap->desc.Flags, initial_state, optimized_clear_value, &object)))
         return hr;
 
-    object->heap = heap;
+    /* https://learn.microsoft.com/en-us/windows/win32/api/d3d12/nf-d3d12-id3d12device-createheap#remarks.
```
saboya, line 3005: Super nit, but the `.` at the end of the URL prevents the anchor from actually working (does not scroll to remarks)

PR #1392, opened 2023-01-25
vkd3d: Get rid of held SPIR-V code in 99.9% in cases.

HansKristian-Work:
Saves a ton of memory on first run of an application.

Two pieces are required to make this work:
- Assume that dynamic VBO stride problems are not real in practice. They don't work quite like how you expect on native AMD drivers either.
- Handle dynamic patch control points. We can fallback and retain SPIR-V if we are at risk of fallback compile.


HansKristian-Work:
Spot checked RADV and NV and they pass test suite as-is.

doitsujin (approved):

PR #1399, opened 2023-01-28
Workaround broken VRS in Dead Space remaster

### LINKED ISSUE #1398: Dead Space: RSSetShadingRateImage called with invalid resource for VRS.
This warning is spammed in Dead Space. The resource is rejected for VRS because it was created with the flag `ALLOW_RENDER_TARGET`. According to the docs this flag is not allowed for VRS images.

> These flags are not permitted.
> ALLOW_RENDER_TARGET
> ALLOW_DEPTH_STENCIL
> ALLOW_CROSS_ADAPTER
> ALLOW_SIMULTANEOUS_ACCESS
> VIDEO_DECODE_REFERENCE_ONLY

https://learn.microsoft.com/en-us/windows/win32/direct3d12/vrs

There's three options here:

- this isn't enforced and works fine despite what the docs say
- this changed with a newer Agility SDK release and the docs are out of date
- this is a game bug and doesn't work on Windows either.

I've tried removing `ALLOW_RENDER_TARGET` from the VRS check and the game worked fine and was slightly faster.

HansKristian-Work:
Fix #1398.

### DIFF
--- a/libs/vkd3d/resource.c
+++ b/libs/vkd3d/resource.c
@@ -376,6 +376,11 @@ static bool vkd3d_sparse_image_may_have_mip_tail(const D3D12_RESOURCE_DESC1 *des
 static bool vkd3d_resource_can_be_vrs(struct d3d12_device *device,
         const D3D12_HEAP_PROPERTIES *heap_properties, const D3D12_RESOURCE_DESC1 *desc)
 {
+    /* Docs say that RTV should not be allowed for fragment shading rate images, yet it works on native,
+     * Dead Space 2023 relies on it, and D3D12 debug layers don't complain.
+     * Technically, it does not seem to care about SIMULTANEOUS_ACCESS either,
+     * but we only workaround when it's proven to be required.
+     * It would complicate things since it affects layouts, etc. */
     return device->device_info.fragment_shading_rate_features.attachmentFragmentShadingRate &&
             desc->Format == DXGI_FORMAT_R8_UINT &&
             desc->Dimension == D3D12_RESOURCE_DIMENSION_TEXTURE2D &&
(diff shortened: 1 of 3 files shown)


libs/vkd3d/resource.c
```diff
@@ -778,6 +782,16 @@ HRESULT vkd3d_get_image_allocation_info(struct d3d12_device *device,
...
     allocation_info->Alignment = requirements.memoryRequirements.alignment;
 
+    /* If we might create an image with VRS usage, need to also check memory requirements without VRS usage.
+     * VRS usage can depend on heap properties and this can affect compression, tile layouts, etc. */
+    if (create_info.image_info.usage & VK_IMAGE_USAGE_FRAGMENT_SHADING_RATE_ATTACHMENT_BIT_KHR)
+    {
+        create_info.image_info.usage &= ~VK_IMAGE_USAGE_FRAGMENT_SHADING_RATE_ATTACHMENT_BIT_KHR;
+        VK_CALL(vkGetDeviceImageMemoryRequirementsKHR(device->vk_device, &requirement_info, &requirements));
+        allocation_info->SizeInBytes = max(requirements.memoryRequirements.size, allocation_info->SizeInBytes);
+        allocation_info->Alignment = max(requirements.memoryRequirements.alignment, allocation_info->Alignment);
```
doitsujin, line 792: We might need to realign the reported size to the new alignment as well in case the VRS alignment is smaller but size is larger than the old values.

HansKristian-Work: Not sure I follow we do max here for both cases. What are the potential returned values that can cause issues here?

doitsujin: Say old alignment is 64k, old size is 64k, VRS alignment is 32k and VRS size is 96k, then we end up with new alignment 64k but new size 96k, but we'd need to return a size of 128k.

Probably won't ever happen in practice though.

HansKristian-Work: So the requirement is that returned size must align? Not sure where that requirement is stated.

doitsujin: Turns out we're already doing this elsewhere anyway, so nevermind.

doitsujin (approved):

PR #1411, opened 2023-02-02
Batch query resolves and WBI

doitsujin:
Reportedly helps Elden Ring on some GPUs.

Next thing to look into would be to get rid of unnecessary compute invocations where possible, but it's probably useful to land this first.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -75,6 +75,7 @@ static void d3d12_command_list_decay_optimal_dsv_resource(struct d3d12_command_l
         const struct d3d12_resource *resource, uint32_t plane_optimal_mask,
         struct d3d12_command_list_barrier_batch *batch);
 static void d3d12_command_list_end_transfer_batch(struct d3d12_command_list *list);
+static void d3d12_command_list_end_wbi_batch(struct d3d12_command_list *list);
 static inline void d3d12_command_list_ensure_transfer_batch(struct d3d12_command_list *list, enum vkd3d_batch_type type);
 
 static HRESULT vkd3d_create_binary_semaphore(struct d3d12_device *device, VkSemaphore *vk_semaphore)
@@ -4315,6 +4316,8 @@ static void d3d12_command_list_end_current_render_pass(struct d3d12_command_list
 
         list->xfb_enabled = false;
     }
+
+    d3d12_command_list_end_wbi_batch(list);
 }
 
 static void d3d12_command_list_invalidate_push_constants(struct vkd3d_pipeline_bindings *bindings)
@@ -5090,6 +5093,7 @@ static void d3d12_command_list_reset_internal_state(struct d3d12_command_list *l
     list->dsv_resource_tracking_count = 0;
     list->subresource_tracking_count = 0;
     list->tracked_copy_buffer_count = 0;
+    list->wbi_batch.batch_len = 0;
 
     list->rendering_info.state_flags = 0;
     list->execute_indirect.has_emitted_indirect_to_compute_barrier = false;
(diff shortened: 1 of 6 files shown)


libs/vkd3d/command.c
```diff
@@ -10214,6 +10307,41 @@ static void d3d12_command_list_execute_query_resolve(struct d3d12_command_list *
...
+        return;
+
+    d3d12_command_list_update_descriptor_buffers(list);
+
+    for (i = 0; i < list->query_resolve_count; i++)
+        d3d12_command_list_execute_query_resolve(list, &list->query_resolves[i]);
+
+    list->query_resolve_count = 0;
+
+    hash_map_clear(&list->query_resolve_lut);
```
HansKristian-Work, line 10324: We might want a way to avoid freeing the memory and allocate it over and over. hash_map_reset() perhaps?

libs/vkd3d/command.c
```diff
@@ -10118,6 +10174,174 @@ static void STDMETHODCALLTYPE d3d12_command_list_DiscardResource(d3d12_command_l
...
+        const struct vkd3d_query_resolve_entry *entry)
+{
+    uint32_t query_index = entry->query_index;
+    uint32_t query_count = entry->query_count;
+    uint32_t advance, bucket, shift;
+    uint64_t query_mask;
+
+    while (query_count)
+    {
+        query_mask = ~0ull;
```
HansKristian-Work, line 10240: This is pretty clever. Would be helpful to have a comment to explain the algorithm. From what I can tell, groups of 64 query indices occupy one hash map entry and the uint64_t bitfield is there to signal individual bits?

HansKristian-Work (changes requested):
Smol nits and will want to have a second look when I'm a bit more awake.

libs/vkd3d/command.c
```diff
@@ -11369,58 +11414,47 @@ static void STDMETHODCALLTYPE d3d12_command_list_WriteBufferImmediate(d3d12_comm
...
+        if ((do_flush && i + 1 == count) || list->wbi_batch.batch_len == VKD3D_MAX_WBI_BATCH_SIZE)
+        {
+            if (list->rendering_info.state_flags & VKD3D_RENDERING_ACTIVE)
+            {
+                /* Implicitly calls end_wbi_batch. We cannot have any pending transfers
+                 * while inside a render pass instance. */
+                d3d12_command_list_end_current_render_pass(list, true);
+
+                /* Relevant if we exceeded the maximum batch size */
+                do_flush = true;
```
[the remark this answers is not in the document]
doitsujin: No, the idea is that we will only defer WBI batches if we're inside a render pass instance, but since we're ending the render pass instance here we have to flush any subsequent batch from the current WBI command as well.

libs/vkd3d/command.c
doitsujin: Done, `clear` now only clears while `free` deallocates memory.

libs/vkd3d/command.c
doitsujin: Yeah, that's how it works. Pretty simple, added a small comment.

libs/vkd3d/command.c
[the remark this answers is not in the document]
HansKristian-Work: Ok, right ... That interaction is pretty subtle. Worth a comment I think.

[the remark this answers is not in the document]
HansKristian-Work: I guess the comment was changed from when I last reviewed, now it makes more sense.

HansKristian-Work (approved):
LGTM, thanks

PR #1414, opened 2023-02-03
Implement some RTPSO subobject edge cases in object assignment

HansKristian-Work:
If an object is used for explicit assignment it takes a lower precedence than objects that have no assignments associated with them.


HansKristian-Work:
Waiving review due to inactivity.

PR #1429, opened 2023-02-16
meta: Create a simple issue template

Blisto91:
A simple issue template based on the one used for dxvk. Can be expanded as needed


K0bin:
Nice, so people can ignore that one too.

doitsujin (approved):

PR #1430, opened 2023-02-16
Add support for VK_EXT_image_sliced_view_of_3d

HansKristian-Work:
Fix some other issues while at it.

- Found 3D bug in UAV fallback copy path. Use imageOffset.z instead of baseLayer.
- Use 4x4x4 workgroup when UAV clearing 3D images.
- Consider WSize/WOffset when clearing UAV.

test_uav_3d_sliced_view now Passes RADV.


doitsujin (approved):
LGTM.

PR #1439, opened 2023-02-22
Use graphics pipeline libraries for fallback pipelines

doitsujin:
Catches things like UNKNOWN DSV with depth test as well as undefined primitive topology.

Tests pass on Nvidia and SotTR works.

### DIFF
--- a/libs/vkd3d/device.c
+++ b/libs/vkd3d/device.c
@@ -3041,6 +3041,16 @@ static ULONG STDMETHODCALLTYPE d3d12_device_AddRef(d3d12_device_iface *iface)
     return refcount;
 }
 
+static void d3d12_device_free_pipeline_libraries(struct d3d12_device *device)
+{
+    hash_map_iter(&device->vertex_input_pipelines, vkd3d_vertex_input_pipeline_free, device);
+    hash_map_free(&device->vertex_input_pipelines);
+
+    hash_map_iter(&device->fragment_output_pipelines, vkd3d_fragment_output_pipeline_free, device);
+    hash_map_free(&device->fragment_output_pipelines);
+}
+
+
 static void d3d12_device_destroy(struct d3d12_device *device)
 {
     const struct vkd3d_vk_device_procs *vk_procs = &device->vk_procs;
@@ -3076,6 +3086,7 @@ static void d3d12_device_destroy(struct d3d12_device *device)
     vkd3d_memory_allocator_cleanup(&device->memory_allocator, device);
     vkd3d_memory_transfer_queue_cleanup(&device->memory_transfers);
     vkd3d_global_descriptor_buffer_cleanup(&device->global_descriptor_buffer, device);
+    d3d12_device_free_pipeline_libraries(device);
     /* Tear down descriptor global info late, so we catch last minute faults after we drain the queues. */
     vkd3d_descriptor_debug_free_global_info(device->descriptor_qa_global_info, device);
 
@@ -3085,6 +3096,8 @@ static void d3d12_device_destroy(struct d3d12_device *device)
 #endif
 
     VK_CALL(vkDestroyDevice(device->vk_device, NULL));
+    rwlock_destroy(&device->fragment_output_lock);
+    rwlock_destroy(&device->vertex_input_lock);
     pthread_mutex_destroy(&device->mutex);
     if (device->parent)
         IUnknown_Release(device->parent);
@@ -6897,8 +6910,20 @@ static HRESULT d3d12_device_init(struct d3d12_device *device,
     
     device->ID3D12DeviceExt_iface.lpVtbl = &d3d12_device_vkd3d_ext_vtbl;
 
-    if (FAILED(hr = vkd3d_create_vk_device(device, create_info)))
+    if ((rc = rwlock_init(&device->vertex_input_lock)))
+    {
+        hr = hresult_from_errno(rc);
         goto out_free_mutex;
+    }
+
+    if ((rc = rwlock_init(&device->fragment_output_lock)))
+    {
+        hr = hresult_from_errno(rc);
+        goto out_free_vertex_input_lock;
+    }
+
+    if (FAILED(hr = vkd3d_create_vk_device(device, create_info)))
+        goto out_free_fragment_output_lock;
 
     if (FAILED(hr = vkd3d_private_store_init(&device->private_store)))
         goto out_free_vk_resources;
@@ -6945,6 +6970,16 @@ static HRESULT d3d12_device_init(struct d3d12_device *device,
                 VKD3D_DESCRIPTOR_DEBUG_DEFAULT_NUM_COOKIES, device)))
             goto out_cleanup_breadcrumb_tracer;
     }
+
+    hash_map_init(&device->vertex_input_pipelines,
+            vkd3d_vertex_input_pipeline_desc_hash,
+            vkd3d_vertex_input_pipeline_desc_compare,
+            sizeof(struct vkd3d_vertex_input_pipeline));
+
+    hash_map_init(&device->fragment_output_pipelines,
+            vkd3d_fragment_output_pipeline_desc_hash,
+            vkd3d_fragment_output_pipeline_desc_compare,
+            sizeof(struct vkd3d_fragment_output_pipeline));
 
     if ((device->parent = create_info->parent))
         IUnknown_AddRef(device->parent);
@@ -6999,6 +7034,10 @@ out_free_vk_resources:
     VK_CALL(vkDestroyDevice(device->vk_device, NULL));
 out_free_instance:
     vkd3d_instance_decref(device->vkd3d_instance);
+out_free_fragment_output_lock:
+    rwlock_destroy(&device->fragment_output_lock);
+out_free_vertex_input_lock:
+    rwlock_destroy(&device->vertex_input_lock);
 out_free_mutex:
     pthread_mutex_destroy(&device->mutex);
     return hr;
@@ -7164,6 +7203,70 @@ void d3d12_device_mark_as_removed(struct d3d12_device *device, HRESULT reason,
     va_end(args);
 
     device->removed_reason = reason;
+}
+
+VkPipeline d3d12_device_get_or_create_vertex_input_pipeline(struct d3d12_device *device,
+        const struct vkd3d_vertex_input_pipeline_desc *desc)
+{
+    struct vkd3d_vertex_input_pipeline pipeline, *entry;
+
+    memset(&pipeline, 0, sizeof(pipeline));
+
+    rwlock_lock_read(&device->vertex_input_lock);
+    entry = (void*)hash_map_find(&device->vertex_input_pipelines, desc);
+
+    if (entry)
+        pipeline.vk_pipeline = entry->vk_pipeline;
+
+    rwlock_unlock_read(&device->vertex_input_lock);
+
+    if (!pipeline.vk_pipeline)
+    {
+        rwlock_lock_write(&device->vertex_input_lock);
+        pipeline.desc = *desc;
+
+        entry = (void*)hash_map_insert(&device->vertex_input_pipelines, desc, &pipeline.entry);
+
+        if (!entry->vk_pipeline)
+            entry->vk_pipeline = vkd3d_vertex_input_pipeline_create(device, desc);
+
+        pipeline.vk_pipeline = entry->vk_pipeline;
+        rwlock_unlock_write(&device->vertex_input_lock);
+    }
+
+    return pipeline.vk_pipeline;
+}
+
+VkPipeline d3d12_device_get_or_create_fragment_output_pipeline(struct d3d12_device *device,
+        const struct vkd3d_fragment_output_pipeline_desc *desc)
+{
+    struct vkd3d_fragment_output_pipeline pipeline, *entry;
+
+    memset(&pipeline, 0, sizeof(pipeline));
+
+    rwlock_lock_read(&device->fragment_output_lock);
+    entry = (void*)hash_map_find(&device->fragment_output_pipelines, desc);
+
+    if (entry)
+        pipeline.vk_pipeline = entry->vk_pipeline;
+
+    rwlock_unlock_read(&device->fragment_output_lock);
+
+    if (!pipeline.vk_pipeline)
+    {
+        rwlock_lock_write(&device->fragment_output_lock);
+        pipeline.desc = *desc;
+
+        entry = (void*)hash_map_insert(&device->fragment_output_pipelines, desc, &pipeline.entry);
+
+        if (!entry->vk_pipeline)
+            entry->vk_pipeline = vkd3d_fragment_output_pipeline_create(device, desc);
+
+        pipeline.vk_pipeline = entry->vk_pipeline;
+        rwlock_unlock_write(&device->fragment_output_lock);
+    }
+
+    return pipeline.vk_pipeline;
 }
 
 IUnknown *vkd3d_get_device_parent(ID3D12Device *device)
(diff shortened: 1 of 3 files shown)


HansKristian-Work (changes requested):
Minor nits. To make review simpler I'd prefer if the pure refactoring commits be moved a separate PR.

The range from 4f9c6d8 to bade04c inclusive.

libs/vkd3d/state.c
```diff
@@ -3222,6 +3222,102 @@ static bool vk_blend_attachment_needs_blend_constants(const VkPipelineColorBlend
...
+{
+    return type == D3D12_PRIMITIVE_TOPOLOGY_TYPE_TRIANGLE ||
+           type == D3D12_PRIMITIVE_TOPOLOGY_TYPE_LINE;
+}
+
+static bool vkd3d_topology_can_restart(VkPrimitiveTopology topology)
+{
+    switch (topology)
+    {
+    case VK_PRIMITIVE_TOPOLOGY_POINT_LIST:
```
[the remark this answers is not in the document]
doitsujin: yeah this is just copy-pasted from below.

doitsujin:
Opened refactor PR in #1441.

libs/vkd3d/state.c
```diff
@@ -4663,21 +4670,42 @@ VkPipeline d3d12_pipeline_state_create_pipeline_variant(struct d3d12_pipeline_st
...
+    {
+        TRACE("Compiling pipeline library for %p with flags %#x.\n", state, library_flags);
+
+        library_create_info.sType = VK_STRUCTURE_TYPE_GRAPHICS_PIPELINE_LIBRARY_CREATE_INFO_EXT;
+        /* Explicit cast to silence a constness warning, this seems to be a Vulkan header bug */
+        library_create_info.pNext = (void*)pipeline_desc.pNext;
+        library_create_info.flags = library_flags;
+
+        pipeline_desc.pNext = &library_create_info;
+        pipeline_desc.flags |= VK_PIPELINE_CREATE_RETAIN_LINK_TIME_OPTIMIZATION_INFO_BIT_EXT;
```
HansKristian-Work: Missing VK_PIPELINE_CREATE_LIBRARY_BIT_KHR.

libs/vkd3d/state.c
```diff
@@ -4632,6 +4634,11 @@ VkPipeline d3d12_pipeline_state_create_pipeline_variant(struct d3d12_pipeline_st
     *dynamic_state_flags = d3d12_graphics_pipeline_state_init_dynamic_state(state, &dynamic_create_info,
             dynamic_state_buffer, key);
 
+    has_vertex_input_state = (!(graphics->stage_flags & VK_SHADER_STAGE_MESH_BIT_EXT)) &&
+            (!library_flags || (library_flags & VK_GRAPHICS_PIPELINE_LIBRARY_VERTEX_INPUT_INTERFACE_BIT_EXT));
+
+    has_fragment_output_state = (!library_flags || (library_flags & VK_GRAPHICS_PIPELINE_LIBRARY_FRAGMENT_OUTPUT_INTERFACE_BIT_EXT));
```
HansKristian-Work, line 4875: Nit: Redundant outer parentheses.

HansKristian-Work: Same above.

libs/vkd3d/state.c
```diff
@@ -4663,21 +4670,42 @@ VkPipeline d3d12_pipeline_state_create_pipeline_variant(struct d3d12_pipeline_st
...
     pipeline_desc.pDynamicState = &dynamic_create_info;
     pipeline_desc.layout = graphics->pipeline_layout;
     pipeline_desc.basePipelineIndex = -1;
 
     if (d3d12_device_supports_variable_shading_rate_tier_2(device))
         pipeline_desc.flags |= VK_PIPELINE_CREATE_RENDERING_FRAGMENT_SHADING_RATE_ATTACHMENT_BIT_KHR;
 
-    if (!(graphics->stage_flags & VK_SHADER_STAGE_MESH_BIT_EXT))
+    if (graphics->stage_flags & VK_SHADER_STAGE_TESSELLATION_CONTROL_BIT)
+        pipeline_desc.pTessellationState = &tessellation_info;
```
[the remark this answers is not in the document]
doitsujin: Tessellation state is part of pre-raster state, not vertex input state.

libs/vkd3d/state.c
doitsujin: Urg, why does validation not catch this.

doitsujin: nvm I added this but the change is in another commit.

libs/vkd3d/vkd3d_private.h
```diff
@@ -1807,6 +1807,8 @@ struct d3d12_graphics_pipeline_state
 
     VkPipelineLayout pipeline_layout;
     VkPipeline pipeline;
+    VkPipeline library;
```
HansKristian-Work, line 1836: This seems to be unused?

libs/vkd3d/state.c
```diff
@@ -3458,6 +3519,69 @@ void vkd3d_fragment_output_pipeline_desc_prepare(struct vkd3d_fragment_output_pi
...
+uint32_t vkd3d_fragment_output_pipeline_desc_hash(const void *key)
+{
+    return hash_data(key, sizeof(struct vkd3d_fragment_output_pipeline_desc));
+}
+
+bool vkd3d_fragment_output_pipeline_desc_compare(const void *key, const struct hash_map_entry *entry)
+{
+    const struct vkd3d_fragment_output_pipeline *pipeline = (const void*)entry;
+    const struct vkd3d_fragment_output_pipeline_desc *desc = key;
+    // We zero-initialize these so memcmp is fine
```
HansKristian-Work, line 3532: C++ comment.

HansKristian-Work: Pointers are not set either, that might be worth commenting about as well. I was initially a bit confused why pointers were not set early but instead deferred to the "prepare" command.

libs/vkd3d/vkd3d_private.h
doitsujin: It's used in a later commit, I'll move the change.

libs/vkd3d/state.c
doitsujin: derp, fixed

doitsujin: fwiw we do have a comment in `vkd3d_vertex_input_pipeline_desc_init` about pointers not being set.

libs/vkd3d/state.c
```diff
@@ -3458,6 +3519,69 @@ void vkd3d_fragment_output_pipeline_desc_prepare(struct vkd3d_fragment_output_pi
...
+    memset(&library_create_info, 0, sizeof(library_create_info));
+    library_create_info.sType = VK_STRUCTURE_TYPE_GRAPHICS_PIPELINE_LIBRARY_CREATE_INFO_EXT;
+    library_create_info.pNext = &desc_copy.rt_info;
+    library_create_info.flags = VK_GRAPHICS_PIPELINE_LIBRARY_FRAGMENT_OUTPUT_INTERFACE_BIT_EXT;
+
+    memset(&create_info, 0, sizeof(create_info));
+    create_info.sType = VK_STRUCTURE_TYPE_GRAPHICS_PIPELINE_CREATE_INFO;
+    create_info.pNext = &library_create_info;
+    create_info.flags = VK_PIPELINE_CREATE_LIBRARY_BIT_KHR | VK_PIPELINE_CREATE_RETAIN_LINK_TIME_OPTIMIZATION_INFO_BIT_EXT;
+    create_info.pColorBlendState = &desc_copy.cb_info;
```
HansKristian-Work, line 3557: Missing pDepthStencil?

HansKristian-Work: nvm, was looking the wrong place in spec ...

libs/vkd3d/device.c
```diff
@@ -7166,6 +7187,52 @@ void d3d12_device_mark_as_removed(struct d3d12_device *device, HRESULT reason,
     device->removed_reason = reason;
 }
 
+VkPipeline d3d12_device_get_or_create_vertex_input_pipeline(struct d3d12_device *device,
+        const struct vkd3d_vertex_input_pipeline_desc *desc)
+{
+    struct vkd3d_vertex_input_pipeline pipeline, *entry;
+
+    pthread_mutex_lock(&device->mutex);
```
HansKristian-Work, line 7195: Should we use RW lock here similar to fallback pipeline cache? Using the full device mutex seems a bit pessimistic.

libs/vkd3d/state.c
doitsujin: depth-stencil state is part of fragment shader state, not fragment output..

libs/vkd3d/state.c
```diff
@@ -4729,6 +4728,78 @@ static bool d3d12_pipeline_state_put_pipeline_to_cache(struct d3d12_pipeline_sta
...
+    create_info.basePipelineIndex = -1;
+
+    if (d3d12_device_uses_descriptor_buffers(state->device))
+        create_info.flags |= VK_PIPELINE_CREATE_DESCRIPTOR_BUFFER_BIT_EXT;
+
+    /* Only use LINK_TIME_OPTIMIZATION for the primary pipeline for now,
+     * accept a small runtime perf hit on subsequent compiles in order
+     * to avoid stutter. */
+    if (!key)
+      create_info.flags |= VK_PIPELINE_CREATE_LINK_TIME_OPTIMIZATION_BIT_EXT;
```
HansKristian-Work, line 4812: Weird indent.

libs/vkd3d/device.c
doitsujin: Can do, correctly managing device-scope objects like that is just rather annoying.

libs/vkd3d/state.c
```diff
@@ -4729,6 +4728,78 @@ static bool d3d12_pipeline_state_put_pipeline_to_cache(struct d3d12_pipeline_sta
...
+
+    /* If the shader library was compiled with identifiers, we need
+     * to set the appropriate bit on the linked pipelines as well
+     * and also explicitly pass the identifiers for linking */
+    for (i = 0; i < graphics->stage_count; i++)
```
HansKristian-Work, line 4817: I don't think this is correct. Sounds like validation bug.

libs/vkd3d/state.c
```diff
@@ -4729,6 +4728,78 @@ static bool d3d12_pipeline_state_put_pipeline_to_cache(struct d3d12_pipeline_sta
...
+    struct vkd3d_fragment_output_pipeline_desc fragment_output_desc;
+    struct vkd3d_vertex_input_pipeline_desc vertex_input_desc;
+    VkPipelineLibraryCreateInfoKHR library_info;
+    VkGraphicsPipelineCreateInfo create_info;
+    VkPipeline vk_libraries[3];
+    uint32_t library_count = 0;
+    unsigned int i;
+    VkResult vr;
+
+    vk_libraries[library_count++] = graphics->library;
```
HansKristian-Work, line 4779: If graphics->library was created with identifiers, this link must also use VK_PIPELINE_CREATE_FAIL_ON_COMPILE_REQUIRED_BIT. If we fail to link, we can convert to SPIR-V and try again.

HansKristian-Work: nvm, it's added below.

doitsujin: what exactly is this referring to? We are setting `VK_PIPELINE_CREATE_FAIL_ON_COMPILE_REQUIRED_BIT` for linking.

libs/vkd3d/state.c
```diff
@@ -4827,7 +4905,8 @@ VkPipeline d3d12_pipeline_state_create_pipeline_variant(struct d3d12_pipeline_st
         library_create_info.flags = library_flags;
 
         pipeline_desc.pNext = &library_create_info;
-        pipeline_desc.flags |= VK_PIPELINE_CREATE_RETAIN_LINK_TIME_OPTIMIZATION_INFO_BIT_EXT;
+        pipeline_desc.flags |= VK_PIPELINE_CREATE_LIBRARY_BIT_KHR |
```
HansKristian-Work, line 4941: This should be moved to the first commit to avoid confusion.

doitsujin: done

libs/vkd3d/state.c
doitsujin: fixed

libs/vkd3d/state.c
doitsujin: clarified comment

libs/vkd3d/state.c
```diff
@@ -4758,6 +4829,13 @@ VkPipeline d3d12_pipeline_state_create_pipeline_variant(struct d3d12_pipeline_st
     *dynamic_state_flags = d3d12_graphics_pipeline_state_init_dynamic_state(state, &dynamic_create_info,
             dynamic_state_buffer, key);
 
+    if (!library_flags && graphics->library)
+    {
+        if (d3d12_pipeline_state_link_pipeline_variant(state, key, dsv_format,
+                vk_cache, *dynamic_state_flags, &vk_pipeline) == VK_SUCCESS)
+            return vk_pipeline;
```
HansKristian-Work, line 4869: If this fails (due to e.g. identifiers), we might have some problems since we're falling back to a non-library path. Is that intentional? I'm not sure if we retain SPIR-V or DXBC at this stage.

It might be possible to synthesize this behavior by always returning FAIL_ON_COMPILE_REQUIRED in this call and making sure everything makes sense.

libs/vkd3d/state.c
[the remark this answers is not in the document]
HansKristian-Work: Right, noticed. Was a bit surprised, but ok.

libs/vkd3d/state.c
doitsujin: We should be retaining DXBC since there are no changes to the `is_fully_dynamic` logic which governs this.

libs/vkd3d/state.c
```diff
@@ -4663,21 +4905,43 @@ VkPipeline d3d12_pipeline_state_create_pipeline_variant(struct d3d12_pipeline_st
...
+        pipeline_desc.pMultisampleState = &graphics->ms_desc;
+
+    if (has_fragment_output_state)
+        pipeline_desc.pColorBlendState = &fragment_output_desc.cb_info;
+
+    if (library_flags)
+    {
+        TRACE("Compiling pipeline library for %p with flags %#x.\n", state, library_flags);
+
+        library_create_info.sType = VK_STRUCTURE_TYPE_GRAPHICS_PIPELINE_LIBRARY_CREATE_INFO_EXT;
```
HansKristian-Work, line 4935: If we're building as a library, do we need to remove dynamic states that are not relevant to library_flags?

doitsujin: According to spec, dynamic states that do not affect any of the pipeline state present must be ignored by the implementation.

That said, we should currently never run into a situation where dynamic states for any of the linked pipelines would differ.

libs/vkd3d/device.c
doitsujin: done.

libs/vkd3d/state.c
```diff
@@ -4366,12 +4382,30 @@ static HRESULT d3d12_pipeline_state_init_static_pipeline(struct d3d12_pipeline_s
          * be exceedingly rare. */
         can_compile_pipeline_early =
                 (desc->primitive_topology_type != D3D12_PRIMITIVE_TOPOLOGY_TYPE_PATCH || graphics->patch_vertex_count != 0 ||
-                 state->device->device_info.extended_dynamic_state2_features.extendedDynamicState2PatchControlPoints) &&
-                 desc->primitive_topology_type != D3D12_PRIMITIVE_TOPOLOGY_TYPE_UNDEFINED;
+                 state->device->device_info.extended_dynamic_state2_features.extendedDynamicState2PatchControlPoints);
```
HansKristian-Work, line 4385: I think there's a flaw here if patch_vertex_count == 0 and dynamic patch controls are not supported. We're the in a situation where we cannot compile pre-raster state. I think we can just promote 0 to 1 here. 0 will only happen if there are no vertex inputs, and those shaders tend to just report 1 anyways. That would neatly bypass all this jank.

HansKristian-Work (changes requested):
Smol issue

HansKristian-Work (approved):
LGTM, thanks.

Issue #1437, opened 2023-02-19
A Plague Tale: Requiem (2022) - flickering on decals, shadows and parts of the terrain


adolfintel:
As the title says, I'm testing A Plague Tale Requiem using VKD3D (current master, built from source) and I noticed that some decals, some shadows and some parts of the terrain have flickering squares on them.

System info:
* GPU: AMD 6900XT
* Driver: Mesa 23.1 (git, built from source)
* Wine version: Wine-GE-Proton 7-37
* OS: Manjaro, KDE (X11), Kernel 6.1.12

Configuration:
* Game: max settings with RT disabled (doesn't work), version 1.4 from GOG
* OS: RADV_PERFTEST=rt,gpl

I tried changing every setting that I could think of and I also tried tinkering with the application workarounds in device.c (tried every single flag) but nothing had an effect. The flickering is not present on Windows.

The problem can be observed about 10 minutes into the game when you're in the castle and a guy gets killed: the blood flickers, a shadow on the wall on the left also flickers, and the same thing can be observed in several other locations throughout the game.

Nothing interesting in the logs.

Here's a video of what it looks like on a decal on a door in chapter 1:
[aptr flickering.webm](https://user-images.githubusercontent.com/16307689/219970087-13d39938-11da-4087-b6cc-a5fdf496720d.webm)

mbriar:
Did you check if it's a regression in either vkd3d-proton or radv? Might also be worth trying with amdvlk, if it runs the game.

adolfintel:
Did some more tests.
With vkd3d-proton 2.8 the flickering is present, 2.7 crashes on startup.
Surprisingly, AMDVLK runs the game (with terrible performance) but the flickering is present there as well (and on a side note, turning on RT caused a GPU hang)

K0bin:
This also happens on Nvidia GPUs with vkd3d-Proton. It only happens at that one spot in the game, so I never bothered to go back and see if I can reproduce it on Windows.

adolfintel:
I just discovered that the problem is much more noticeable if you lower the resolution and resolution scale. At 720p with the lowest internal resolution the issue can be observed on many objects in that area, not just the terrain and decals.

@K0bin I noticed the problem in other chapters as well, are you sure you didn't have them too?

Blisto91:
Not an issue with the native amd drivers (the door decals at least)

Save game: 
[...]

PR #1440, opened 2023-02-22
Implement rasterizer ordered views

HansKristian-Work:
Fixes https://github.com/HansKristian-Work/vkd3d-proton/issues/1437.

Passes tests on NV and ANV. No AMD driver supports VK_EXT_fragment_shader_interlock yet.


doitsujin:
Given that the most coarse lock we handle is per pixel, do we need to disable VRS for ROV pipelines?

doitsujin (approved):

PR #1441, opened 2023-02-23
Pipeline refactor stuff

doitsujin:
As requested.


libs/vkd3d/vkd3d_private.h
```diff
@@ -1690,6 +1690,26 @@ struct vkd3d_shader_debug_ring_spec_constants
...
+    VkVertexInputBindingDivisorDescriptionEXT vi_divisors[D3D12_IA_VERTEX_INPUT_RESOURCE_SLOT_COUNT];
+    VkPipelineVertexInputDivisorStateCreateInfoEXT vi_divisor_info;
+
+    VkVertexInputAttributeDescription vi_attributes[D3D12_VS_INPUT_REGISTER_COUNT];
+    VkVertexInputBindingDescription vi_bindings[D3D12_IA_VERTEX_INPUT_RESOURCE_SLOT_COUNT];
+    VkPipelineVertexInputStateCreateInfo vi_info;
+
+    VkPipelineInputAssemblyStateCreateInfo ia_info;
+
+    VkDynamicState dy_states[VKD3D_MAX_VERTEX_INPUT_DYNAMIC_STATES];
```
HansKristian-Work, line 1709: To make sure I understand it right, this subset of dynamic state will only take effect when these stages are compiled as GPL, right?

doitsujin: yes, we're not using these fields at all when creating a monolithic pipeline or a shader pipeline library.

HansKristian-Work (approved):
LGTM

PR #1445, opened 2023-03-01
vkd3d-shader: Implement tessellation factor limiting.

HansKristian-Work:
Implement for DXBC for now.


mbriar:
Hey, little drive-by comment, but i also noticed the somewhat unexpected low gpu-bound perf in this game, and it seems like the Volumetric Cloud Quality setting on Very High is what really destroys performance, but also on native AMD d3d12, just not to the same extend. It's a little hard to say just how much the performance improves due to the 60 fps limiter, but going from Very High to Standard gets me from 40 fps to 60 fps while also cutting GPU power draw in half on a 6700xt. For reference, windows is around 55 fps while also completely maxing out the gpu on Very High.  
I assume you probably tested low settings on Deck, but I'll still ask if limiting the tess factor still has such a big impact with Volumetric Cloud Quality set to low?

HansKristian-Work:
> I assume you probably tested low settings on Deck, but I'll still ask if limiting the tess factor still has such a big impact with Volumetric Cloud Quality set to low?

This is with "Best Performance" global setting, and terrain takes > 10 % of frame time, and geometry rendering in general is > 50% of frame time ...

HansKristian-Work:
> but also on native AMD d3d12, just not to the same extend

I tested this with maxed out settings on Windows vs Linux as well, and perf was on-par basically.

mbriar:
Looking at the sky with Volumetric Cloud Quality setting on Very High was quite a bit faster for me on windows, as described above, but all in all terrible on both for what was being rendered..

notpentadactyl:
> Looking at the sky with Volumetric Cloud Quality setting on Very High was quite a bit faster for me on windows, as described above, but all in all terrible on both for what was being rendered..

I didn't look at Windows performance, but on Linux Nvidia (3080) at 4k, max settings I get 29-34 fps (100% GPU load) looking at the sky.

With Volumetric Cloud Quality on Low/Standard: 60fps (65% GPU load) looking at the sky.

doitsujin (approved):

PR #1449, opened 2023-03-03
Bump Vulkan API version to 1.3

doitsujin:
Allows us to get rid of a bunch of KHR extensions and feature checks, since a lot of the features are mandatory in Vulkan 1.3.

`VK_EXT_shader_viewport_index_layer` was technically promoted to 1.2, but unfortunately we need to keep the extension around for now since the SPIR-V caps differ between the extension and core, and we'd need to bump vkd3d-shader to SPIR-V 1.5 in order to fix this.

Overall, this didn't allow for nearly as much of a cleanup as I was hoping, but it's better than nothing.

There is some duplicated info in `vkd3d_vulkan_info`, but cleaning that up should probably be a separate PR.


libs/vkd3d/device.c
```diff
@@ -1261,11 +1261,10 @@ static void vkd3d_physical_device_info_init(struct vkd3d_physical_device_info *i
     info->features2.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_FEATURES_2;
     info->properties2.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_PROPERTIES_2;
 
-    info->subgroup_properties.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_SUBGROUP_PROPERTIES;
-    vk_prepend_struct(&info->properties2, &info->subgroup_properties);
-
-    info->maintenance3_properties.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_MAINTENANCE_3_PROPERTIES;
-    vk_prepend_struct(&info->properties2, &info->maintenance3_properties);
+    info->vulkan_1_1_features.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_VULKAN_1_1_FEATURES;
```
HansKristian-Work, line 1230: Funny enough, the 1_1 struct was added in Core 1.2. Could bump core requirement first then add this.

libs/vkd3d/device.c
```diff
@@ -2174,6 +2184,15 @@ static HRESULT vkd3d_init_device_caps(struct d3d12_device *device,
     physical_device_info->vulkan_1_1_features.protectedMemory = VK_FALSE;
     physical_device_info->vulkan_1_1_features.samplerYcbcrConversion = VK_FALSE;
 
+    physical_device_info->vulkan_1_2_features.shaderInputAttachmentArrayDynamicIndexing = VK_FALSE;
+    physical_device_info->vulkan_1_2_features.shaderInputAttachmentArrayNonUniformIndexing = VK_FALSE;
+    physical_device_info->vulkan_1_2_features.bufferDeviceAddressCaptureReplay = VK_FALSE;
+    physical_device_info->vulkan_1_2_features.bufferDeviceAddressMultiDevice = VK_FALSE;
+    physical_device_info->vulkan_1_2_features.vulkanMemoryModelAvailabilityVisibilityChains = VK_FALSE;
```
HansKristian-Work, line 1997: We're not using vulkan memory model either yet. There's a non-zero risk it can affect some driver's compiler.

8-bit storage is also unused.

libs/vkd3d/device.c
```diff
@@ -2174,6 +2184,15 @@ static HRESULT vkd3d_init_device_caps(struct d3d12_device *device,
...
+
+    physical_device_info->vulkan_1_3_features.descriptorBindingInlineUniformBlockUpdateAfterBind = VK_FALSE;
```
HansKristian-Work, line 1999: - We're not using inline UBO anymore.
- Not using privateData. Might affect winevulkan.
- Not using sync2.
- Not using ASTC.

HansKristian-Work (changes requested):
Comments. Also needs README update.

libs/vkd3d/device.c
HansKristian-Work: There are misc scenarios where unrelated features affect compilers globally (e.g. on NV there was a bug where enabling mesh shader extension affected PSO keys for some reason), or fragment shading rate affected all PSOs in RADV even if the PSO didn't use it.

I think the main argument is consistency though, since we're disabling features we're not using here to begin with, might as well do it consistently.

libs/vkd3d/device.c
doitsujin: Not sure if it's useful to disable sync2, it shouldn't affect anything and we'll likely need it at some point in the future.

HansKristian-Work (approved):
LGTM, thanks!

PR #1454, opened 2023-03-07
Added NO_UPLOAD_HVV for Forza Horizon 5

adolfintel:
FH5 has severe performance drops (visible in the benchmark) with rebar enabled, and also freezes frequently ingame. With this workaround the game performs as expected (apart from the occasional hiccups) and no longer freezes.

Tested on 6900xt, latest master, Mesa 23.1


ishitatsuyuki:
You might be getting hit by resizable BAR bugs specific to certain motherboard vendors. Please check the issue below and see if you're affected.

https://gitlab.freedesktop.org/drm/amd/-/issues/1864

If that issue is the reason, the only solution is to disable the buggy resizable BAR support in your BIOS.

adolfintel:
@ishitatsuyuki looks like you were right, that issue pointed me in the right direction

PR #1459, opened 2023-03-08
Embedded mutable descriptors

HansKristian-Work:
This introduces the idea of "embedded" mutable descriptors.

In this descriptor model, we can embed all information we need for
a given descriptor within the descriptor buffer payload itself.

This requires that the descriptor buffer implementation of the driver
supports some key properties:

- Size >= 32 (we need to reserve 5 LSBs to signal log2 offset to planar
  metadata in case that exists)
- Texel buffer and SSBO can be placed side by side in the mutable
  payload (these are usually 16 bytes each)
- Descriptor buffers can be bound at SSBO descriptor size alignment.
  We are clever here and bind the same descriptor buffer at an offset
  to achieve the packed texel buffer + SSBO descriptor.
- RTAS SRVs is implemented by viewing the descriptor buffer as an SSBO
  directly and shader applies a stride to load the BDA based on
  sizeof(mutable descriptor).
- UAV counters are implemented by repurposing the texel buffer.
  Could have used BDA, but it is somewhat less robust against app bugs
  when we alias texel buffer descriptor with it. We already had code
  paths in place to deal with the pre-BDA world, so this is fine.
- If Size is very large, like on RADV (64 byte) we can take
  advantage of that by embedding the metadata inside the mutable
  descriptor as well. This mostly just serves to save on CPU memory
  and avoids having to check for CPU -> CPU descriptor copies.

This allows us to massively optimize descriptor copies to the point
where the descriptor copy itself is an unrolled memcpy.

To further make this work well, we need to introduce vtable variants.
This improves performance quite a lot for descriptor copies, and is
overall cleaner.

In the descriptor copy micro-benchmark that copies one descriptor at a
time under ideal cache situations,
I observe 12.5 ns -> 2.5 ns per descriptor.

With the vtable variant system in place, it should be possible to add
special optimizations for implementations which do not hit this
optimization path.

### DIFF
--- a/libs/vkd3d/vkd3d_private.h
+++ b/libs/vkd3d/vkd3d_private.h
@@ -1120,6 +1120,8 @@ struct vkd3d_descriptor_binding
 
 #define VKD3D_RESOURCE_DESC_INCREMENT_LOG2 5
 #define VKD3D_RESOURCE_DESC_INCREMENT (1u << VKD3D_RESOURCE_DESC_INCREMENT_LOG2)
+#define VKD3D_RESOURCE_EMBEDDED_METADATA_OFFSET_LOG2_MASK 31
+#define VKD3D_RESOURCE_EMBEDDED_RESOURCE_HEAP_MASK (1ull << 32)
 
 /* Arrange data so that it can pack as tightly as possible.
  * When we copy descriptors, we must copy both structures.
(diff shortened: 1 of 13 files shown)


libs/vkd3d/vkd3d_private.h
```diff
@@ -1409,9 +1409,35 @@ static inline void d3d12_desc_copy_embedded_resource_payload(uint8_t *dst, const
 {
     /* Add special unrolled paths for the usual suspects. */
     if (size == 64)
+    {
+#ifdef __SSE2__
```
doitsujin, line 1415: Not sure if `d3d12_desc_copy_embedded_resource_payload` really adds that much value as an abstraction at this point, should we maybe have something like this in a common header instead and call those directly?
```c
static inline memcpy_nontemporal_64b(void* dst, const void* src) {
#ifdef __SSE2__
    ...
#else
    memcpy(dst, src, 64);
#endif
}
```

Maybe derp an always_inline there as well.

The current code kind of relies too much on the compiler being smart about inlining for my taste, which *probably* happens to work well right now but is by no means guaranteed (I've seen GCC not inline more trivial functions than this for no apparent reason, and if that happens we'll end up getting useless branching for a constant parameter).

HansKristian-Work: Sure, I'll refactor this so it does not require inlining + opts to figure it out.

HansKristian-Work: Rewrote this part now.

doitsujin (approved):
Looks good with the changes.

PR #1465, opened 2023-03-15
Implement ExecuteIndirect multi-dispatch indirect with state changes

HansKristian-Work:
Works around Halo Infinite using even more esoteric D3D12 features.


doitsujin (approved):
Not a fan of this being behind yet another config flag, but given that this is such a niche use case I think this is fine for now and we can think of a more permanent solution if this ever becomes more common.

PR #1468, opened 2023-03-16
idl: Get rid of INLINE_WRAPPER macros.

HansKristian-Work:
They break builds quite a lot since upstream widl and mingw64-tools widl don't agree on how this is supposed to work.
If we need the macros, we have to macro away static which breaks other things in the process.

The robust workaround is to roll our own wrappers for the functions where it's needed.

### DIFF
--- a/demos/demo.h
+++ b/demos/demo.h
@@ -21,7 +21,6 @@
 #endif
 
 #include <vkd3d_windows.h>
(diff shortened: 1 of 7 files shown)


HansKristian-Work:
Abandoning, will try a better way ...

Issue #1389, opened 2023-01-24
DX12 stop working on nvidia gpu with latest proton experimental and 525 driver


an9949an:
After recent proton updates I have black screen in all dx12 games (tested in cyberpunk, deep rock galactic, little nightmares 2). Game works (i hear the sound, even see in nvidia-settings that GPU busy and vram used) but no image. The same situation with Proton Ge 7.44+ .
DirectX11 games work without issues, older proton versions also work as expected.

The problem can be fixed by adding launch options: `VKD3D_DISABLE_EXTENSIONS=VK_KHR_present_id,VK_KHR_present_wait %command%`.

Probably this is the [commit](https://github.com/HansKristian-Work/vkd3d-proton/commit/d2c876211f1e4bd7d4177cc8f0102aabd2b7402d) which produce such behavior.

Here is the [issue](https://github.com/ValveSoftware/Proton/issues/6472) on proton repo with detailed sysinfo and discussion.

mbriar:
You should update to the latest nvidia driver release in the 525 series, which is 525.85.05. As you can tell from the commit you linked, earlier releases in the 525 series had bugs with present_wait, but those should be fixed now.

an9949an:
@mbriar thank you for the answer!
Looks like 525.78.01 is the latest driver shipped with fedora. So, maybe better to add some more precise version check instead just disabling workaround for the whole 525?

mbriar:
I assume the workaround was removed outright because it didn't seem necessary to keep around a workaround in the code that only affects obsolete versions of a single driver series. You can use `VKD3D_DISABLE_EXTENSIONS=VK_KHR_present_wait` as environment variable in the meantime until an update is available for your distro.

an9949an:
@mbriar of course I can use this launch options locally. The main reason of creating this issue here is that looks like removing this workaround for the whole 525 driver series is the wrong decision that breaks compatibility. If driver fixed since 525.78 - workaround should be removed if driver version higher than 525.78 as well. Am I wrong?

mbriar:
The workaround was completely removed and there is no driver version check for this at all anymore. There is also no reason for anyone to stay on an old and buggy point release of the 525 series, so keeping a workaround for that seems unnecessary when there are multiple was to fix it (either updating the driver, or using an older version of vkd3d-proton or the env var for the time being.)

Leopard1907:
[...]

PR #1477, opened 2023-03-20
vkd3d: Disable KHR_present_wait on NV drivers again.

HansKristian-Work:
Various users are reporting vr -3 errors on swapchain creation and there is no fix or known workaround.


doitsujin (approved):
Good to go if user feedback is positive.

mbriar:
@adamnv see https://github.com/HansKristian-Work/vkd3d-proton/issues/1389, the affected people have the issue with every game and even the vkd3d-proton gears.exe demo application.

I've also seen reports about The Outer Worlds: Spacer's Choice Edition and Resident Evil 4 Chainsaw Demo from people on single GPU, non-optimus systems.

adamnv:
So it's everything, eh?  At least it's an easy repro then I guess.  I'll make sure the right folk know!

doitsujin (approved):

PR #1490, opened 2023-03-25
vkd3d: Add option to curb memory PSO cache bloat.

HansKristian-Work:


doitsujin (approved):

K0bin:
Can ANV even run the game in the first place?

renatopereyra:
I don't have an Intel card to test. I just thought that RADV and ANV would be affected similarly here and I didn't see a reason to explicitly leave ANV out.

mbriar:
Games like this that require feature level 12_0 don't run on anv anyways due to missing features and need manual feature level overrides to even try to launch them.

renatopereyra:
ANV will support this feature at level at some point and then hit this bug in all likelihood since it shares much of the shader cache code with RADV. I thought it would save time overall to do both of them together for things that are known to be common between them.

Blisto91:
I don't know much about ANV specifically bit at least for RADV this is intended to be a temporary work around until work in RADV has made a better solution. (afaiu)

PR #1496, opened 2023-03-29
vkd3d: Add PSO cache fix to Uncharted: Legacy of Thieves Collection

runar-work:
Reduces cache bloat somewhat.


HansKristian-Work (approved):
LGTM. If more games show up where this is a problem before Mesa has a more robust fix in place, we'll have to come up with something more general, but these two games represent the "worst case" so far.

PR #1504, opened 2023-04-03
Synchronization2 rework

doitsujin:
Uses sync2 entry points everywhere, and also makes use of some of the new stage and access flags where it makes sense. Also, since `TOP_OF_PIPE` and `BOTTOM_OF_PIPE` are deprecated, those were largely replaced with `ALL_COMMANDS` (and `NONE` in the odd edge case where that makes sense).

Note that this doesn't touch any of the breadcrumb and WBI stuff. This is because we cannot query function pointers for `vkCmdWriteBufferMarker2AMD` and `vkGetQueueCheckpointData2NV` without (redundantly) enabling the `KHR_synchronization2` extension. This shouldn't be an issue since the relevant pipeline stages are all supersets of the sync2 pipeline stages that we use.


libs/vkd3d/memory.c
```diff
@@ -429,18 +432,6 @@ static HRESULT vkd3d_memory_transfer_queue_flush_locked(struct vkd3d_memory_tran
     }
 
     /* Stall future submissions on other queues until the clear has finished */
-    memset(&timeline_info, 0, sizeof(timeline_info));
```
HansKristian-Work, line 432: This breaks 11on12 I think? Would D3D11 land know how to wait on these semaphores?

libs/vkd3d/command.c
```diff
@@ -3395,6 +3400,9 @@ static void d3d12_command_list_clear_attachment_pass(struct d3d12_command_list *
             barrier->subresourceRange.baseArrayLayer = view->info.texture.layer_idx;
             barrier->subresourceRange.layerCount = view->info.texture.layer_count;
 
+            if (clear_op)
+              barrier->srcAccessMask = access;
```
HansKristian-Work, line 3400: indent

HansKristian-Work (changes requested):
Comments

HansKristian-Work:
Missing a commit to remove sync1 entry points perhaps?

libs/vkd3d/memory.c
doitsujin: what's the concern here exactly? we still submit waits to all queues later in the function and stall when locking the queue.

HansKristian-Work: Right, the waiters get flushed on any submission that is not VKD3D_SUBMISSION_WAIT, so the DRAIN command will end up waiting anyways, is that the assertion?

doitsujin: yes, that's the idea. FWIW, our [11on12 test](https://github.com/doitsujin/dxvk-tests/blob/main/d3d11/d3d11_on_12.cpp) which is highly sensitive to this and exposed the memory initialization issue in the first place still works reliably.

libs/vkd3d/command.c
```diff
@@ -2927,7 +2927,7 @@ static void d3d12_command_list_decay_optimal_dsv_resource(struct d3d12_command_l
     barrier.dstStageMask = current_layout_is_shader_visible ? VK_PIPELINE_STAGE_2_ALL_COMMANDS_BIT
             : (VK_PIPELINE_STAGE_2_EARLY_FRAGMENT_TESTS_BIT | VK_PIPELINE_STAGE_2_LATE_FRAGMENT_TESTS_BIT);
     barrier.dstAccessMask = VK_ACCESS_2_DEPTH_STENCIL_ATTACHMENT_READ_BIT |
-            (current_layout_is_shader_visible ? VK_ACCESS_2_SHADER_READ_BIT : 0);
+            (current_layout_is_shader_visible ? VK_ACCESS_2_SHADER_SAMPLED_READ_BIT : 0);
```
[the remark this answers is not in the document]
doitsujin: do we have any idea how recently that is?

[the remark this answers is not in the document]
HansKristian-Work: https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/21271 Last month, so would break on Deck I think.

libs/vkd3d/memory.c
HansKristian-Work: Ok, seems fine then.

libs/vkd3d/command.c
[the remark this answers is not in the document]
doitsujin: Dropped the commit for now, we can probably reintroduce it later.

libs/vkd3d/command.c
```diff
@@ -10918,22 +10982,30 @@ static void d3d12_command_list_resolve_binary_occlusion_queries(struct d3d12_com
...
     d3d12_command_list_invalidate_root_parameters(list, &list->compute_bindings, true);
 
-    vk_barrier.sType = VK_STRUCTURE_TYPE_MEMORY_BARRIER;
-    vk_barrier.pNext = NULL;
-    vk_barrier.dstAccessMask = VK_ACCESS_SHADER_WRITE_BIT;
-    /* If there are any overlapping copy writes, handle them here since we're doing a transfer barrier anyways. */
-    vk_barrier.srcAccessMask = list->tracked_copy_buffer_count ? VK_ACCESS_TRANSFER_WRITE_BIT : 0;
+    memset(&dep_info, 0, sizeof(dep_info));
+    dep_info.sType = VK_STRUCTURE_TYPE_DEPENDENCY_INFO;
+    dep_info.memoryBarrierCount = 0;
```
HansKristian-Work, line 10993: 0?

doitsujin: Should have been 1, fixed.

HansKristian-Work (approved):
LGTM

Issue #1511, opened 2023-04-14
Reconsider use of ms_abi for native builds


expipiplus1:
Neither the official dx headers [1] nor dxvk [2] set `__stdcall`, however vkd3d-proton does [3]. 

This wouldn't necessarily be a problem if one sticks only to using vkd3d-proton, however it does become quite tricky to correctly use vkd3d-proton along with dxvk's dxgi in the same project.

Would it be acceptable to use the default abi for native builds, or make this configurable at build time? 

[1]: https://github.com/microsoft/DirectX-Headers/blob/54d2f42bf6f6c4f7dd21e379b350c676bf60e38b/include/wsl/stubs/basetsd.h#L215
[2]: https://github.com/doitsujin/dxvk/blob/master/include/native/windows/windows_base.h#L190
[3]: https://github.com/HansKristian-Work/vkd3d-proton/blob/master/include/vkd3d_windows.h#L175

HansKristian-Work:
It's a vestigal thing from when vkd3d was implemented in native code rather than PE, so if you know how to modify the widl invocation to avoid ms_abi, feel free to propose a PR.

PR #1513, opened 2023-04-14
Remove use of ms_abi for native builds

expipiplus1:
This brings the native headers in line with the official DirectX headers and dxvk

### DIFF
--- a/include/vkd3d_d3d12.idl
+++ b/include/vkd3d_d3d12.idl
@@ -22,6 +22,10 @@ import "vkd3d_dxgibase.idl";
 
 import "vkd3d_d3dcommon.idl";
 
+cpp_quote("#pragma push_macro(\"WIDL_EXPLICIT_AGGREGATE_RETURNS\")")
+cpp_quote("#undef WIDL_EXPLICIT_AGGREGATE_RETURNS")
+cpp_quote("#define WIDL_EXPLICIT_AGGREGATE_RETURNS")
+
 cpp_quote("#ifndef _D3D12_CONSTANTS")
 cpp_quote("#define _D3D12_CONSTANTS")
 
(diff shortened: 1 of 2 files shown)


expipiplus1:
Closes https://github.com/HansKristian-Work/vkd3d-proton/issues/1511

expipiplus1:
I still need to check if this requires defining `WIDL_EXPLICIT_AGGREGATE_RETURNS`

HansKristian-Work:
The vtables are implemented assuming the explicit aggregate returns fwiw, so I don't think we should mess around with that. Just changing the calling convention should be good enough I think.

HansKristian-Work:
Also, do you consent to this contribution being relicensed to MIT?

expipiplus1:
> Also, do you consent to this contribution being relicensed to MIT?

Sure, that's fine

expipiplus1:
> The vtables are implemented assuming the explicit aggregate returns fwiw, so I don't think we should mess around with that. Just changing the calling convention should be good enough I think. The last comment suggets the PR isn't ready for review, so I'll wait until notified.

Ah, perhaps this is an orthogonal issue caused by me using WIDL 8.0 then... As one would expect with how the vtables are implemented, it is necessary to define `WIDL_EXPLICIT_AGGREGATE_RETURNS` before including `vkd3d_dx12.h`.

I'm happy for this to be merged as is, and sort out this aggregate return issue elsewhere.

For quick reference, this is what WIDL 8.0 generates:

```C++
ID3D12Heap : public ID3D12Pageable
{
#ifdef WIDL_EXPLICIT_AGGREGATE_RETURNS
    virtual D3D12_HEAP_DESC* STDMETHODCALLTYPE GetDesc(
        D3D12_HEAP_DESC *__ret) = 0;
    D3D12_HEAP_DESC STDMETHODCALLTYPE GetDesc(
        )
    {
        D3D12_HEAP_DESC __ret;
        return *GetDesc(&__ret);
    }
#else
    virtual D3D12_HEAP_DESC STDMETHODCALLTYPE GetDesc(
        ) = 0;
#endif
};
```

I think the fix here would be to add `-DWIDL_EXPLICIT_AGGREGATE_RETURNS` to the WIDL invocation.

HansKristian-Work:
```
cpp_quote("#ifndef WIDL_EXPLICIT_AGGREGATE_RETURNS")
cpp_quote("#define WIDL_EXPLICIT_AGGREGATE_RETURNS")
```
That seems fine to me.

expipiplus1:
I implemented it with push/pop_macro so as to not pollute the includer's macro state. I believe that this is supported by gcc, clang and msvc

HansKristian-Work (approved):
LGTM, thanks.

expipiplus1:
Thanks!

Issue #1460, opened 2023-03-10
Resident Evil 4 Chainsaw Demo: CACAO causes striped patterns on some objects


mbriar:
### Software information
Resident Evil 4 Chainsaw Demo, High settings 1440p, Fidelity FX CACAO enabled.

### System information
- GPU: RX 6700 XT 
- Driver: Mesa 23.1.0-devel (git-98cb7e0108)
- Wine version: Proton Hotfix
- VKD3D-Proton version: from Proton Hotfix

Enabling Fidelity FX CACAO causes some striped patterns on some objects that I don't see on windows.

<details>
  <summary>Screenshot linux</summary>
  
![re4_cacao_liux](https://user-images.githubusercontent.com/114156793/224304054-ec1e4282-702a-4e9b-8719-57d2b7834e92.jpg)

</details>

<details>
  <summary>Screenshot windows</summary>
  
![re4_cacao_windows](https://user-images.githubusercontent.com/114156793/224304063-0a8d2740-69d4-4d60-ad0c-e4f6b56a5845.jpg)

</details>

Look at the top of the wooden plank, it's more obvious in motion. To reproduce, start a new game and just run forward a bit until you see the cart with the wooden planks to the left of it.

HansKristian-Work:
Known game bug. Workaround is also known. (Game assumes Wave32 is used when Wave64 is used).

mbriar:
Thanks, I assume the known workaround is nothing users can apply themselves for now? I tried to throw RADV_PERFTEST=pswave32,cswave32 at it and it doesn't seem to help.

HansKristian-Work:
The workaround will likely be to tell shader that wave size is 1, which makes it ignore the broken code path and enter the code path written by AMD.

kodatarule:
Alright this doesn't seem to affect NVIDIA, but there's this random hang just like in RE2 and RE3(it was also reported for RE8 as well) https://github.com/HansKristian-Work/vkd3d-proton/issues/1357
I'm pretty sure it can't be a VRAM issue as I have 24GB of VRAM, I switched to an RTX 3090. Problem is this hang is extremely random and it just freezes up the game.
I managed to get a log of RE3 when I still had the 3070, which I did attach here on this report: https://github.com/ValveSoftware/Proton/issues/3719#issuecomment-1406919590

PR #1523, opened 2023-04-19
Add missing D3D12 constants

expipiplus1:
I sorted the constants by name in the first commit to make any
differences more apparent in the second commit

- Correct the spelling of D3D12_RAYTRACING_INSTANCE_DESCS_BYTE_ALIGNMENT (previously it was D3D12_RAYTRACING_INSTANCE_DESC_BYTE_ALIGNMENT)

- Correct type of D3D12_VIEWPORT_BOUNDS_MIN from UINT to INT


mbriar:
Is this the same as https://github.com/HansKristian-Work/vkd3d-proton/issues/1460 for the Demo? Maybe more shader hashes need to be added to the quirk list..

doitsujin:
what do we need these for exactly?

expipiplus1:
These are useful for compiling D2D12 programs using these constants against the headers in this repo

They were taken from the IDL files in https://github.com/microsoft/DirectX-Headers which are in turn sourced from the DX12 hardware specs

HansKristian-Work:
@doitsujin are there any concerns merging this? Seems okay to me.

expipiplus1:
I added in a missing typedef and a couple of other macros present in the official dx12 headers.

It does seem a little weird to be adding these manually piecemeal, is the API now similar enough to DX12 to encode these headers as some fixup over the Microsoft headers?

HansKristian-Work:
The upstream headers will only ever build on MSVC and generally we add IDL as needed. There hasn't been much use for external projects using our headers so far.

doitsujin (approved):

expipiplus1:
Gotcha, thanks!

PR #1524, opened 2023-04-19
d3d12core: Set the interface pointer returned by GetDebugInterface to NULL.

ivyl:
Minecraft Legends depends on this behavior. I've checked and this is how native D3D12 behaves.

### DIFF
--- a/libs/d3d12core/main.c
+++ b/libs/d3d12core/main.c
@@ -372,6 +372,9 @@ HRESULT STDMETHODCALLTYPE d3d12core_GetDebugInterface(d3d12core_interface *core,
 {
     TRACE("iid %s, debug %p.\n", debugstr_guid(iid), debug);
 
+    if (debug)
+        *debug = NULL;
+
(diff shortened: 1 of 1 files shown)


ivyl:
To run Minecraft Legends you need the following patches on top of Proton 8.0:

https://gist.github.com/ivyl/311613b6ae63113b01505ae3739d34e3

They will make it into bleeding edge once it's updated to use 8.0 codebase.

libs/d3d12core/main.c
```diff
@@ -372,6 +372,8 @@ HRESULT STDMETHODCALLTYPE d3d12core_GetDebugInterface(d3d12core_interface *core,
 {
     TRACE("iid %s, debug %p.\n", debugstr_guid(iid), debug);
 
+    *debug = NULL;
```
HansKristian-Work, line 375: Needs `if (debug)` check I think. Most interfaces allow you to query for success without an out pointer.

PR #1529, opened 2023-04-20
vkd3d: Add RE2 and RE7 to CACAO workaround and remove RE4 demo

runar-work:
Add shader hash used in RE2 and RE7 to list of workaround hashes. Remove RE4 demo as it has a new shader with a fix.


HansKristian-Work (approved):
LGTM

PR #1535, opened 2023-04-21
vkd3d: Implement EXT_swapchain_maintenance1 and replace fence wait with semaphore

HansKristian-Work:
VkFence wait on Mesa is not a noop after all.
There can be very long stalls leading to very strange frame pacing graphs when GPU is rendering slower than refresh rate in windowed modes.

Waiting for semaphores fixes this and with present_wait, there is little reason to not use semaphores when there's a meaningful improvement in jitter.

### DIFF
--- a/libs/d3d12core/main.c
+++ b/libs/d3d12core/main.c
@@ -271,6 +271,12 @@ static HRESULT vkd3d_create_instance_global(struct vkd3d_instance **out_instance
 #endif
     };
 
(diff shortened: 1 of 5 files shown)


HansKristian-Work:
> The semaphore stuff is getting convoluted and hard to reason about though.

ye, it is. Not sure how much better we can do it without timeline WSI though.

doitsujin (approved):

PR #1538, opened 2023-04-25
Extended the wave32 quirk to the whole TLOU application

### LINKED ISSUE #1518: tlou1 god rays can go through buildings
In The Last of Us Part 1 god rays can sometimes be seen going through buildings. This does not happen in the native Windows version, but does happen with vkd3d-proton on Windows.

<details>
  <summary>Screenshot</summary>

image
</details>

### Software information
In The Last of Us Part 1
Ultra preset. Haven't tested others yet.

### System information
- GPU: 7900xtx
- Driver: week old mesa-git
- Wine version: Proton Experimental & BE
- VKD3D-Proton version: Tested master, 2.8 and 2.7. 2.6 wouldn't launch the game. 

### Log files
[steam-1888930.tar.gz](https://github.com/HansKristian-Work/vkd3d-proton/files/11253021/steam-1888930.tar.gz)

### Save game
Just look at the buildings to the left when spawning in
[SAVEFILE00.tar.gz](https://github.com/HansKristian-Work/vkd3d-proton/files/11253095/SAVEFILE00.tar.gz)

Taken from `compatdata/1888930/pfx/drive_c/users/steamuser/Saved Games/The Last of Us Part I/users/<user-id>/savedata/`

adolfintel:
Should finally fix #1518 and the blocky shadows in some areas.

### DIFF
--- a/libs/vkd3d/device.c
+++ b/libs/vkd3d/device.c
@@ -583,12 +583,8 @@ static const struct vkd3d_shader_quirk_info re_quirks = {
 
 /* Subgroup ballots in shadow pre-processing shader assume wave32 despite
  * the shader using 64-wide workgroups. */
-static const struct vkd3d_shader_quirk_hash tlou1_hashes[] = {
-    { 0x4a1afb3a12797c11ull, VKD3D_SHADER_QUIRK_FORCE_MAX_WAVE32 },
-};
-
 static const struct vkd3d_shader_quirk_info tlou1_quirks = {
-    tlou1_hashes, ARRAY_SIZE(tlou1_hashes), 0,
+    NULL, 0, VKD3D_SHADER_QUIRK_FORCE_MAX_WAVE32,
 };
 
 static const struct vkd3d_shader_quirk_meta application_shader_quirks[] = {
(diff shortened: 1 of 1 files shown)


mbriar:
Some questions that don't really apply to this PR, but rather the current state, but I'll ask them here anyways:

- what about the tlou-i-l.exe, which I think is for not AVX2 capable CPUs?
- maybe stupid question, but does this cause any problems for GPUs that don't support wave32, like AMD Vega and older?

adolfintel:
> Some questions that don't really apply to this PR, but rather the current state, but I'll ask them here anyways:
> 
>     * what about the tlou-i-l.exe, which I think is for not AVX2 capable CPUs?

I don't see why it shouldn't work for non-AVX2 capable CPUs, they should use the same shaders. I doubt there are any CPUs old enough to not have AVX2 and able to run this game using vkd3d though.

> 
>     * maybe stupid question, but does this cause any problems for GPUs that don't support wave32, like AMD Vega and older?

It's worth testing, I have an rx 580 here, I'll test it later today.

Blisto91:
For the first point i think mbriar means that it might make sense to include the second exe to also be covered.

mbriar:
> For the first point i think mbriar means that it might make sense to include the second exe to also be covered.

Yes.

HansKristian-Work:
> Finding all the shaders that need this quirk would require playing through the whole game and dumping the shaders in areas where glitches are visible,

That's not what I'm thinking of. Chasing ever-changing shader hashes won't work. I plan on detecting this broken pattern in general in dxil-spirv. I plan on looking into this today.

If you want a blanket workaround for now, I'd suggest RADV_PERFTEST=cswave32.

adolfintel:
Alright, looking forward to your proper fix then :)

By the way, `RADV_PERFTEST=cswave32` doesn't force the group size of 32, it simply enables wave32 for compute shaders, which should already be enabled by default on this series of cards, the glitch is present with this setting.

HansKristian-Work:
> which should already be enabled by default on this series of cards

That's not true.

> the glitch is present with this setting

That doesn't sound right. The workaround only applies to CS.

DadSchoorse:
radv ignores cswave32 for shaders with subgroup operations

adolfintel:
Thanks @DadSchoorse

HansKristian-Work:
https://github.com/HansKristian-Work/vkd3d-proton/pull/1539 should handle it.

PR #1543, opened 2023-04-27
Improve behavior when VRAM is oversubscribed.

adamnv:
The goal is to:

- Enable pageable memory, a.k.a. heap migration (VK_EXT_pageable_device_local_memory)
- Implement explicit D3D12 API residency+prioritization hints on top of VK_EXT_memory_priority
- Closely match the implicit D3D12 prioritization given to resources by default based on usage etc, insofar as publicly documented

**Notes**:
Practically speaking, it's not a panacea for when you're chronically oversubscribed (it may be a better experience... or it may not); the main win is that it recovers a lot more gracefully from transitory oversubscription.

Without VK_EXT_pageable_device_local_memory you can end up with rendertargets (or anything else vital) spilled to a system heap and stay there forever.  That can drive an app off an invisible performance cliff from which it's hard to recover (short of restarting).  With VK_EXT_pageable_device_local_memory and some half-decent priority hints, at least important stuff can make it back to a device-local heap once there's room again and perf gets back to a decent steady-state.

### DIFF
--- a/libs/vkd3d/device.c
+++ b/libs/vkd3d/device.c
@@ -97,6 +97,8 @@ static const struct vkd3d_optional_extension_info optional_device_extensions[] =
     VK_EXTENSION(EXT_IMAGE_SLICED_VIEW_OF_3D, EXT_image_sliced_view_of_3d),
     VK_EXTENSION(EXT_GRAPHICS_PIPELINE_LIBRARY, EXT_graphics_pipeline_library),
     VK_EXTENSION(EXT_FRAGMENT_SHADER_INTERLOCK, EXT_fragment_shader_interlock),
+    VK_EXTENSION(EXT_PAGEABLE_DEVICE_LOCAL_MEMORY, EXT_pageable_device_local_memory),
+    VK_EXTENSION(EXT_MEMORY_PRIORITY, EXT_memory_priority),
     /* AMD extensions */
     VK_EXTENSION(AMD_BUFFER_MARKER, AMD_buffer_marker),
     VK_EXTENSION(AMD_DEVICE_COHERENT_MEMORY, AMD_device_coherent_memory),
@@ -1573,6 +1575,20 @@ static void vkd3d_physical_device_info_init(struct vkd3d_physical_device_info *i
         vk_prepend_struct(&info->features2, &info->fragment_shader_interlock_features);
     }
 
+    if (vulkan_info->EXT_memory_priority)
+    {
+        info->memory_priority_features.sType =
+            VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_MEMORY_PRIORITY_FEATURES_EXT;
+        vk_prepend_struct(&info->features2, &info->memory_priority_features);
+    }
+
+    if (vulkan_info->EXT_pageable_device_local_memory)
+    {
+        info->pageable_device_memory_features.sType =
+            VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_PAGEABLE_DEVICE_LOCAL_MEMORY_FEATURES_EXT;
+        vk_prepend_struct(&info->features2, &info->pageable_device_memory_features);
+    }
+
     VK_CALL(vkGetPhysicalDeviceFeatures2(device->vk_physical_device, &info->features2));
     VK_CALL(vkGetPhysicalDeviceProperties2(device->vk_physical_device, &info->properties2));
 }
@@ -2627,6 +2643,7 @@ static HRESULT d3d12_device_create_scratch_buffer(struct d3d12_device *device, e
         alloc_info.heap_desc.Alignment = D3D12_DEFAULT_RESOURCE_PLACEMENT_ALIGNMENT;
         alloc_info.heap_desc.Flags = D3D12_HEAP_FLAG_ALLOW_ONLY_BUFFERS | D3D12_HEAP_FLAG_CREATE_NOT_ZEROED;
         alloc_info.extra_allocation_flags = VKD3D_ALLOCATION_FLAG_INTERNAL_SCRATCH;
+        alloc_info.vk_memory_priority = vkd3d_convert_to_vk_prio(D3D12_RESIDENCY_PRIORITY_NORMAL);
 
         if (FAILED(hr = vkd3d_allocate_heap_memory(device, &device->memory_allocator,
                 &alloc_info, &scratch->allocation)))
@@ -2644,6 +2661,7 @@ static HRESULT d3d12_device_create_scratch_buffer(struct d3d12_device *device, e
         alloc_info.heap_flags = D3D12_HEAP_FLAG_ALLOW_ONLY_BUFFERS | D3D12_HEAP_FLAG_CREATE_NOT_ZEROED;
         alloc_info.optional_memory_properties = VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT;
         alloc_info.flags = VKD3D_ALLOCATION_FLAG_GLOBAL_BUFFER | VKD3D_ALLOCATION_FLAG_INTERNAL_SCRATCH;
+        alloc_info.vk_memory_priority = vkd3d_convert_to_vk_prio(D3D12_RESIDENCY_PRIORITY_NORMAL);
 
         if (FAILED(hr = vkd3d_allocate_memory(device, &device->memory_allocator,
                 &alloc_info, &scratch->allocation)))
@@ -2662,6 +2680,7 @@ static HRESULT d3d12_device_create_scratch_buffer(struct d3d12_device *device, e
         alloc_info.heap_desc.Alignment = D3D12_DEFAULT_RESOURCE_PLACEMENT_ALIGNMENT;
         alloc_info.heap_desc.Flags = D3D12_HEAP_FLAG_ALLOW_ONLY_BUFFERS | D3D12_HEAP_FLAG_CREATE_NOT_ZEROED;
         alloc_info.extra_allocation_flags = VKD3D_ALLOCATION_FLAG_INTERNAL_SCRATCH;
+        alloc_info.vk_memory_priority = vkd3d_convert_to_vk_prio(D3D12_RESIDENCY_PRIORITY_NORMAL);
 
         if (FAILED(hr = vkd3d_allocate_heap_memory(device, &device->memory_allocator,
                 &alloc_info, &scratch->allocation)))
@@ -5106,17 +5125,135 @@ static HRESULT STDMETHODCALLTYPE d3d12_device_OpenSharedHandleByName(d3d12_devic
 static HRESULT STDMETHODCALLTYPE d3d12_device_MakeResident(d3d12_device_iface *iface,
         UINT object_count, ID3D12Pageable * const *objects)
 {
-    FIXME_ONCE("iface %p, object_count %u, objects %p stub!\n",
+    struct d3d12_device *device = impl_from_ID3D12Device(iface);
+    const struct vkd3d_vk_device_procs *vk_procs = &device->vk_procs;
+
+    TRACE("iface %p, object_count %u, objects %p\n",
             iface, object_count, objects);
 
+    if (device->device_info.pageable_device_memory_features.pageableDeviceLocalMemory)
+    {
+        uint32_t i;
+
+        for (i = 0; i < object_count; i++)
+        {
+            VkDeviceMemory memory = VK_NULL_HANDLE;
+            D3D12_RESIDENCY_PRIORITY priority;
+            ID3D12Resource *resource_iface;
+            ID3D12Heap *heap_iface;
+
+            if (SUCCEEDED(ID3D12Pageable_QueryInterface(objects[i], &IID_ID3D12Heap, (void**)&heap_iface)))
+            {
+                struct d3d12_heap *heap_object = impl_from_ID3D12Heap(heap_iface);
+
+                if (heap_object->priority.allows_dynamic_residency)
+                {
+                    memory = heap_object->allocation.device_allocation.vk_memory;
+                    spinlock_acquire(&heap_object->priority.spinlock);
+                    priority = heap_object->priority.d3d12priority;
+                    heap_object->priority.residency_count++;
+                    spinlock_release(&heap_object->priority.spinlock);
+                }
+
+                ID3D12Heap_Release(heap_iface);
+            }
+            else if (SUCCEEDED(ID3D12Pageable_QueryInterface(objects[i], &IID_ID3D12Resource, (void**)&resource_iface)))
+            {
+                struct d3d12_resource *resource_object = impl_from_ID3D12Resource(resource_iface);
+
+                if (resource_object->priority.allows_dynamic_residency)
+                {
+                    memory = resource_object->mem.device_allocation.vk_memory;
+                    spinlock_acquire(&resource_object->priority.spinlock);
+                    priority = resource_object->priority.d3d12priority;
+                    resource_object->priority.residency_count++;
+                    spinlock_release(&resource_object->priority.spinlock);
+                }
+
+                ID3D12Resource_Release(resource_iface);
+            }
+
+            if (memory)
+            {
+                VK_CALL(vkSetDeviceMemoryPriorityEXT(device->vk_device, memory, vkd3d_convert_to_vk_prio(priority)));
+            }
+        }
+    }
+
     return S_OK;
+}
+
+static HRESULT STDMETHODCALLTYPE d3d12_device_EnqueueMakeResident(d3d12_device_iface *iface,
+        D3D12_RESIDENCY_FLAGS flags, UINT num_objects, ID3D12Pageable *const *objects,
+        ID3D12Fence *fence_to_signal, UINT64 fence_value_to_signal)
+{
+    TRACE("iface %p, flags %#x, num_objects %u, objects %p, fence_to_signal %p, fence_value_to_signal %"PRIu64"\n",
+            iface, flags, num_objects, objects, fence_to_signal, fence_value_to_signal);
+
+    /* note: we ignore flags/D3D12_RESIDENCY_FLAG_DENY_OVERBUDGET; it involves
+       knowing if the app will be made over-budget.  We act as if it won't.  Could perhaps
+       use VK_EXT_memory_budget but don't have an app in-hand that clearly cares. */
+    d3d12_device_MakeResident(iface, num_objects, objects);
+
+    /* we don't block anyway - signal the fence immediately */
+    return ID3D12Fence_Signal(fence_to_signal, fence_value_to_signal);
 }
 
 static HRESULT STDMETHODCALLTYPE d3d12_device_Evict(d3d12_device_iface *iface,
         UINT object_count, ID3D12Pageable * const *objects)
 {
-    FIXME_ONCE("iface %p, object_count %u, objects %p stub!\n",
+    struct d3d12_device *device = impl_from_ID3D12Device(iface);
+    const struct vkd3d_vk_device_procs *vk_procs = &device->vk_procs;
+
+    TRACE("iface %p, object_count %u, objects %p\n",
             iface, object_count, objects);
+
+    if (device->device_info.pageable_device_memory_features.pageableDeviceLocalMemory)
+    {
+        uint32_t i;
+
+        for (i = 0; i < object_count; i++)
+        {
+            VkDeviceMemory memory = VK_NULL_HANDLE;
+            ID3D12Resource *resource_iface;
+            bool now_evicted = false;
+            ID3D12Heap *heap_iface;
+
+            if (SUCCEEDED(ID3D12Pageable_QueryInterface(objects[i], &IID_ID3D12Heap, (void**)&heap_iface)))
+            {
+                struct d3d12_heap *heap_object = impl_from_ID3D12Heap(heap_iface);
+
+                if (heap_object->priority.allows_dynamic_residency)
+                {
+                    memory = heap_object->allocation.device_allocation.vk_memory;
+                    spinlock_acquire(&heap_object->priority.spinlock);
+                    now_evicted = (0 == --heap_object->priority.residency_count);
+                    spinlock_release(&heap_object->priority.spinlock);
+                }
+
+                ID3D12Heap_Release(heap_iface);
+            }
+            else if (SUCCEEDED(ID3D12Pageable_QueryInterface(objects[i], &IID_ID3D12Resource, (void**)&resource_iface)))
+            {
+                struct d3d12_resource *resource_object = impl_from_ID3D12Resource(resource_iface);
+
+                if (resource_object->priority.allows_dynamic_residency)
+                {
+                    memory = resource_object->mem.device_allocation.vk_memory;
+                    spinlock_acquire(&resource_object->priority.spinlock);
+                    now_evicted = (0 == --resource_object->priority.residency_count);
+                    spinlock_release(&resource_object->priority.spinlock);
+                }
+
+                ID3D12Resource_Release(resource_iface);
+            }
+
+            if (memory && now_evicted)
+            {
+                VK_CALL(vkSetDeviceMemoryPriorityEXT(device->vk_device, memory, 0.0f));
+            }
+        }
+    }
 
     return S_OK;
 }
@@ -5410,8 +5547,64 @@ fail:
 static HRESULT STDMETHODCALLTYPE d3d12_device_SetResidencyPriority(d3d12_device_iface *iface,
         UINT object_count, ID3D12Pageable *const *objects, const D3D12_RESIDENCY_PRIORITY *priorities)
 {
-    FIXME_ONCE("iface %p, object_count %u, objects %p, priorities %p stub!\n",
+    struct d3d12_device *device = impl_from_ID3D12Device(iface);
+
+    TRACE("iface %p, object_count %u, objects %p, priorities %p\n",
             iface, object_count, objects, priorities);
+
+    if (device->device_info.pageable_device_memory_features.pageableDeviceLocalMemory)
+    {
+        const struct vkd3d_vk_device_procs *vk_procs = &device->vk_procs;
+        uint32_t i;
+
+        for (i = 0; i < object_count; i++)
+        {
+            D3D12_RESIDENCY_PRIORITY priority = priorities[i];
+            VkDeviceMemory memory = VK_NULL_HANDLE;
+            ID3D12Resource *resource_iface;
+            ID3D12Heap *heap_iface;
+
+            if (SUCCEEDED(ID3D12Pageable_QueryInterface(objects[i], &IID_ID3D12Heap, (void**)&heap_iface)))
+            {
+                struct d3d12_heap *heap_object = impl_from_ID3D12Heap(heap_iface);
+
+                if (heap_object->priority.allows_dynamic_residency)
+                {
+                    spinlock_acquire(&heap_object->priority.spinlock);
+                    heap_object->priority.d3d12priority = priority;
+                    if (heap_object->priority.residency_count)
+                    {
+                        memory = heap_object->allocation.device_allocation.vk_memory;
+                    }
+                    spinlock_release(&heap_object->priority.spinlock);
+                }
+
+                ID3D12Heap_Release(heap_iface);
+            }
+            else if (SUCCEEDED(ID3D12Pageable_QueryInterface(objects[i], &IID_ID3D12Resource, (void**)&resource_iface)))
+            {
+                struct d3d12_resource *resource_object = impl_from_ID3D12Resource(resource_iface);
+
+                if (resource_object->priority.allows_dynamic_residency)
+                {
+                    spinlock_acquire(&resource_object->priority.spinlock);
+                    resource_object->priority.d3d12priority = priority;
+                    if (resource_object->priority.residency_count)
+                    {
+                        memory = resource_object->mem.device_allocation.vk_memory;
+                    }
+                    spinlock_release(&resource_object->priority.spinlock);
+                }
+
+                ID3D12Resource_Release(resource_iface);
+            }
+
+            if (memory)
+            {
+                VK_CALL(vkSetDeviceMemoryPriorityEXT(device->vk_device, memory, vkd3d_convert_to_vk_prio(priority)));
+            }
+        }
+    }
 
     return S_OK;
 }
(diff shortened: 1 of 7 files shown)


libs/vkd3d/utils.c
```diff
@@ -647,6 +648,34 @@ VkFormat vkd3d_get_vk_format(DXGI_FORMAT format)
...
+    if (d3d12prio <= D3D12_RESIDENCY_PRIORITY_NORMAL)
+        result = 0.5f * ((double)d3d12prio) / D3D12_RESIDENCY_PRIORITY_NORMAL;
+    else
+        result = 0.5f + 0.5f * ((double)d3d12prio - D3D12_RESIDENCY_PRIORITY_NORMAL) /
+            (UINT32_MAX - D3D12_RESIDENCY_PRIORITY_NORMAL);
+
+    /* Note: The conversion from a UINT32 d3d priority to a float32 vk priority
+       loses around 9 of the 16 lower-order bits which encode size-based subranking,
+       and thus most of the practical benefit except for massive heaps/resources.
+       Numerous workarounds for this are feasible. */
```
adamnv, line 673: I could go into detail or provide such a workaround, but it was all getting a bit over-engineered-looking given the lack of a specific title where it was a measurable and repeatable benefit.

libs/vkd3d/device.c
```diff
@@ -1572,6 +1573,17 @@ static void vkd3d_physical_device_info_init(struct vkd3d_physical_device_info *i
                 VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_FRAGMENT_SHADER_INTERLOCK_FEATURES_EXT;
         vk_prepend_struct(&info->features2, &info->fragment_shader_interlock_features);
     }
+    
+    if (vulkan_info->EXT_pageable_device_local_memory)
+    {
+        info->memory_priority_features.sType =
```
HansKristian-Work, line 1579: Is this EXT enabled?

include/vkd3d.h
```diff
@@ -165,6 +165,9 @@ HRESULT vkd3d_create_root_signature_deserializer(const void *data, SIZE_T data_s
 
 VkFormat vkd3d_get_vk_format(DXGI_FORMAT format);
 
+uint32_t vkd3d_get_priority_adjust(VkDeviceSize size);
```
HansKristian-Work, line 168: The appropriate place to add internal functions is vkd3d_private.h. vkd3d.h is mostly vestigal at this point.

libs/vkd3d/device.c
```diff
@@ -5106,17 +5121,109 @@ static HRESULT STDMETHODCALLTYPE d3d12_device_OpenSharedHandleByName(d3d12_devic
...
+                priority = heap_object->priority;
+                heap_object->evicted = false;
+
+                ID3D12Heap_Release(heap_iface);
+            }
+            else if (SUCCEEDED(ID3D12DeviceChild_QueryInterface(object, &IID_ID3D12Resource, (void**)&resource_iface)))
+            {
+                struct d3d12_resource *resource_object = impl_from_ID3D12Resource(resource_iface);
+
+                memory = (resource_object->flags & VKD3D_RESOURCE_COMMITTED) ?
```
HansKristian-Work, line 5153: This is not enough. We need to make sure that the allocation is not sub-allocated. If it's suballocated, memory priority cannot be dynamic.

libs/vkd3d/device.c
```diff
@@ -5106,17 +5121,109 @@ static HRESULT STDMETHODCALLTYPE d3d12_device_OpenSharedHandleByName(d3d12_devic
...
+            ID3D12DeviceChild *object = (ID3D12DeviceChild*)objects[i];
+            VkDeviceMemory memory = VK_NULL_HANDLE;
+            D3D12_RESIDENCY_PRIORITY priority;
+
+            if (SUCCEEDED(ID3D12DeviceChild_QueryInterface(object, &IID_ID3D12Heap, (void**)&heap_iface)))
+            {
+                struct d3d12_heap *heap_object = impl_from_ID3D12Heap(heap_iface);
+
+                memory = heap_object->allocation.device_allocation.vk_memory;
+                priority = heap_object->priority;
```
HansKristian-Work, line 5144: Same concern here, a very small heap without DENY_BUFFERS that is smaller than 2MB will be suballocated and should be exempt from memprio. In practice, it shouldn't happen however.

libs/vkd3d/device.c
```diff
@@ -5106,17 +5121,109 @@ static HRESULT STDMETHODCALLTYPE d3d12_device_OpenSharedHandleByName(d3d12_devic
...
+
+    if (device->vk_info.EXT_pageable_device_local_memory)
+    {
+        uint32_t i;
+        
+        for (i = 0; i < object_count; i++)
+        {
+            ID3D12Heap *heap_iface;
+            ID3D12Resource *resource_iface;
+            ID3D12DeviceChild *object = (ID3D12DeviceChild*)objects[i];
```
HansKristian-Work, line 5135: Why not just QueryInterface from the ID3D12Pageable?

libs/vkd3d/device.c
```diff
@@ -5106,17 +5121,109 @@ static HRESULT STDMETHODCALLTYPE d3d12_device_OpenSharedHandleByName(d3d12_devic
...
+            VkDeviceMemory memory = VK_NULL_HANDLE;
+
+            if (SUCCEEDED(ID3D12DeviceChild_QueryInterface(object, &IID_ID3D12Heap, (void**)&heap_iface)))
+            {
+                struct d3d12_heap *heap_object = impl_from_ID3D12Heap(heap_iface);
+
+                memory = heap_object->allocation.device_allocation.vk_memory;
```
HansKristian-Work, line 5205: Same concerns here as for MakeResident.

libs/vkd3d/device.c
```diff
@@ -5106,17 +5121,109 @@ static HRESULT STDMETHODCALLTYPE d3d12_device_OpenSharedHandleByName(d3d12_devic
...
+                ID3D12Heap_Release(heap_iface);
+            }
+            else if (SUCCEEDED(ID3D12DeviceChild_QueryInterface(object, &IID_ID3D12Resource, (void**)&resource_iface)))
+            {
+                struct d3d12_resource *resource_object = impl_from_ID3D12Resource(resource_iface);
+
+                memory = (resource_object->flags & VKD3D_RESOURCE_COMMITTED) ?
+                    resource_object->mem.device_allocation.vk_memory : VK_NULL_HANDLE;
+                priority = resource_object->priority;
+                resource_object->evicted = false;
```
HansKristian-Work, line 5156: Is this a potential race? I'm not sure if it's valid to call MakeResident concurrently on the same resource.

libs/vkd3d/memory.c
```diff
@@ -1057,6 +1057,7 @@ static HRESULT vkd3d_memory_allocation_init(struct vkd3d_memory_allocation *allo
 {
     const struct vkd3d_vk_device_procs *vk_procs = &device->vk_procs;
     VkMemoryRequirements memory_requirements;
+    VkMemoryPriorityAllocateInfoEXT priority_info;
```
HansKristian-Work, line 1060: :christmas_tree: ordering

libs/vkd3d/utils.c
```diff
@@ -647,6 +648,34 @@ VkFormat vkd3d_get_vk_format(DXGI_FORMAT format)
...
+
+    /* 0.0f is reserved for explicitly evicted resources */
+    return fmax(fmin(result, 1.f), 0.01f);
```
HansKristian-Work, line 676: You probably want fmaxf here. But it's probably best to just use the min/max macros. Avoids potential hassle in build dealing with -lm.

libs/vkd3d/memory.c
```diff
@@ -1156,6 +1157,15 @@ static HRESULT vkd3d_memory_allocation_init(struct vkd3d_memory_allocation *allo
         }
     }
 
+    if (device->vk_info.EXT_pageable_device_local_memory &&
```
HansKristian-Work, line 1160: Isn't this gated by VK_EXT_memory_priority, not pageable?

HansKristian-Work: It seems useful to allow this behavior on implementations which only support static memory_priority.

libs/vkd3d/utils.c
```diff
@@ -647,6 +648,34 @@ VkFormat vkd3d_get_vk_format(DXGI_FORMAT format)
...
+
+/* map from 32-bit d3d prio to float (0..1) vk prio.*/
+float vkd3d_convert_to_vk_prio(D3D12_RESIDENCY_PRIORITY d3d12prio)
+{
+    float result;
+
+    /* align D3D12_RESIDENCY_PRIORITY_NORMAL (the default d3d12 prio) with
+       0.5 (the default vk prio) so neither kind wins without explicit prio */
+    if (d3d12prio <= D3D12_RESIDENCY_PRIORITY_NORMAL)
+        result = 0.5f * ((double)d3d12prio) / D3D12_RESIDENCY_PRIORITY_NORMAL;
```
HansKristian-Work, line 665: Weird mixing of float and double.

libs/vkd3d/utils.c
```diff
@@ -647,6 +648,34 @@ VkFormat vkd3d_get_vk_format(DXGI_FORMAT format)
...
+    else
+        result = 0.5f + 0.5f * ((double)d3d12prio - D3D12_RESIDENCY_PRIORITY_NORMAL) /
+            (UINT32_MAX - D3D12_RESIDENCY_PRIORITY_NORMAL);
```
HansKristian-Work, line 668: It might be appropriate to clamp the input priority to D3D12_RESIDENCY_PRIORITY_MINIMUM and MAXIMUM to recover some range.

libs/vkd3d/utils.c
```diff
@@ -647,6 +648,34 @@ VkFormat vkd3d_get_vk_format(DXGI_FORMAT format)
     return VK_FORMAT_UNDEFINED;
 }
 
+/* get some size-based low bits for prioritization in the same way as d3d12 */
+inline uint32_t vkd3d_get_priority_adjust(VkDeviceSize size)
+{
+    return min((size / (10 * 1048576)), 0xFFFFUL);
```
HansKristian-Work, line 654: This is magical. Is the 10 MiB bin-size arbitrary, or is this actually documented?

libs/vkd3d/utils.c
HansKristian-Work: vkd3d_get_priority_adjust just steps by 1 ULP every 10 MiB. This will be lost in noise when doing float divisions here. Float based algorithm should take that into account perhaps to add a special priority band just for that perhaps?

I.e.:

MINIMUM to NORMAL -> linear map from 0.1f to 0.5f.
NORMAL to HIGH -> linear map from 0.5f to 0.75f.
HIGH to HIGH + 0x10000 -> linear map from 0.75f to 0.8f
HIGH to MAXIMUM -> linear map from 0.8f to 1.0f.

libs/vkd3d/vkd3d_private.h
```diff
@@ -638,13 +639,15 @@ struct vkd3d_allocate_memory_info
...
     VkMemoryPropertyFlags optional_memory_properties;
+    float vk_memory_priority;
 };
 
 struct vkd3d_allocate_heap_memory_info
 {
     D3D12_HEAP_DESC heap_desc;
     void *host_ptr;
     uint32_t extra_allocation_flags;
+    float vk_memory_priority; // a heap allocation can actually be a dedicated allocation under the hood (xxx: I think that's true?), in which case we can still fulfil a memory priority
```
HansKristian-Work, line 650: C89 style comments.

HansKristian-Work: And yes, a ID3D12Heap can map 1:1 to a VkDeviceMemory, but only if we don't need suballocation.

libs/vkd3d/device.c
HansKristian-Work: It seems useful to have a VKD3D_RESOURCE flag for whether or not it supports RESIDENCY management. It should kick in if:

- The underlying allocation is not suballocated (there should be a chunk member of vkd3d_device_allocation).
- The resource or heap is DEVICE_LOCAL.
- The EXT is supported.

This can be assigned at resource and heap creation.

libs/vkd3d/device.c
HansKristian-Work: Also worth noting that docs say:

```
MakeResident is ref-counted, such that [Evict](https://learn.microsoft.com/en-us/windows/desktop/api/d3d12/nf-d3d12-id3d12device-evict) must be called the same amount of times as MakeResident before Evict takes effect. Objects that support residency are made resident during creation, so a single Evict call will actually evict the object.
```

So the appropriate implementation should probably use atomic ref-count and priority is atomic relaxed exchange, where initial ref-count is 1 or 0 depending on the heap flag.

HansKristian-Work: Alternatively, to stay sane, a simple lock per-object around the residency processing is fine too.

libs/vkd3d/resource.c
```diff
@@ -2819,6 +2825,20 @@ static HRESULT d3d12_resource_create(struct d3d12_device *device, uint32_t flags
             FIXME_ONCE("ReadFromSubresource may be buggy on host-visible images with ALLOW_SIMULTANEOUS_ACCESS | ALLOW_UNORDERED_ACCESS.\n");
     }
 
+    if ((flags & VKD3D_RESOURCE_COMMITTED) &&
+        (device->vk_info.EXT_pageable_device_local_memory) &&
```
HansKristian-Work, line 2829: Initial priority only depends on EXT_memory_priority.

HansKristian-Work: Also, we tend to check for the feature being enabled, not the EXT since that's the "technically correct" thing to do.

HansKristian-Work: Should this consider D3D12_HEAP_FLAG_CREATE_NOT_RESIDENT? NOT_RESIDENT being lowered to 0.0f priority should only happen when pageable_device_local_memory is supported of course.

HansKristian-Work (changes requested):

libs/vkd3d/utils.c
adamnv: Is it the use of double you object to or using 0.5f instead of 0.5?  Yeah at least the latter is weird - can fix.

libs/vkd3d/utils.c
adamnv: > D3D12_RESIDENCY_PRIORITY_MINIMUM and MAXIMUM
On the one hand it's not an error for an app to submit priorities outside the range, which made me reluctant, but on the other hand that wouldn't be common or smart - so I think that's a good idea.

> HIGH to HIGH + 0x10000 -> linear map from 0.75f to 0.8f

It's a good idea and now that you mention it I'd go further - again although an app can use the full range, the HIGH to HIGH+0x10000 range is particularly special, so I think it should get the lion's share of the 0.5f to 1.0f range to preserve resolution.

libs/vkd3d/resource.c
adamnv: > Should this consider D3D12_HEAP_FLAG_CREATE_NOT_RESIDENT

I believe that's handled elsewhere (at a lower level), I'll check this path actually hits it.

libs/vkd3d/device.c
adamnv: Dunno, will look into it :grinning:

libs/vkd3d/memory.c
adamnv: To be clear, you'd like to support priority-at-allocation-time for the case where only memory_priority is supported?  Seems reasonable.

libs/vkd3d/utils.c
HansKristian-Work: The math should be done in float I think. The end-result is float either way.

adamnv:
Thanks a lot for the review - hope to address it over the coming days. :+1:

libs/vkd3d/device.c
adamnv: I've now added a spinlock around the residency fields.  Although the actual vkSetDeviceMemoryPriorityEXT() isn't covered by the lock, I think I've convinced myself that it should be (in which case that spinlock should really be a mutex). :(

adamnv: Residency-counting is in, now.

libs/vkd3d/memory.c
adamnv: Not familiar with this - what does it mean?  Thanks.

HansKristian-Work: the length of each line is sorted from longest to shortest.

HansKristian-Work:
Doesn't seem like all issues have been addresses, so assume it's still WIP, when it's ready for review again, hit the refresh button under "Reviewers".

adamnv:
Yup, still working through it.  Will refresh when I think all the feedback is addressed - thanks.

libs/vkd3d/resource.c
adamnv: > Also, we tend to check for the feature being enabled, not the EXT since that's the "technically correct" thing to do.

Do you mean check for it in VkPhysicalDeviceFeatures2 or something else?  Thanks.

adamnv:
Still working through it. :)

libs/vkd3d/device.c
adamnv: Addressed this centrally at 9f15cb280f4e86f45fe4aa338774d57b9a0d3529 - please let me know if this wasn't what you had in mind.  Thanks.

adamnv:
Reckon it's ready for (re)review.  Cheers.

libs/vkd3d/device.c
adamnv: FWIW I think this is happening quite a lot (Last Of Us & Spider-Man, but approximately never in UE4 titles).  Not a problem as such, just wanted to check expectations.

adamnv:
FYI I added some tracing spew in history for anyone trying to trace/debug prioritization behavior.  Revert 86ba442098c6658bd8f5cddd7c90f186abfef797 to re-enable the spew.

adamnv:
Games often end up calling SetResidencyPriority() on a heap or resource (mostly heap) that doesn't actually qualify according to the `allows_dynamic_residency` flag I cache in the heap/resource's `priority_info`:

```
heap->priority.allows_dynamic_residency = 
device->device_info.pageable_device_memory_features.pageableDeviceLocalMemory &&
heap->allocation.chunk == NULL /* not suballocated */ &&       
device->memory_properties.memoryTypes[heap->allocation.device_allocation.vk_memory_type].propertyFlags & VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT;
```

I think the reasons are a mix of:
* The heuristics in vkd3d_allocate_memory() which can force a suballocation
* The heuristics in d3d12_resource_create_committed() which can turn a committed allocation into a suballocation
* Apps genuinely asking to set priority on non-device-local or non-committed allocations because YOLO

... but I'd appreciate eyes to double-check.

libs/vkd3d/device.c
```diff
@@ -5106,17 +5125,126 @@ static HRESULT STDMETHODCALLTYPE d3d12_device_OpenSharedHandleByName(d3d12_devic
...
+            D3D12_RESIDENCY_PRIORITY priority;
+            ID3D12Resource *resource_iface;
+            ID3D12Heap *heap_iface;
+
+            if (SUCCEEDED(ID3D12Pageable_QueryInterface(objects[i], &IID_ID3D12Heap, (void**)&heap_iface)))
+            {
+                struct d3d12_heap *heap_object = impl_from_ID3D12Heap(heap_iface);
+
+                if (!heap_object->priority.allows_dynamic_residency)
+                    continue;
```
adamnv, line 5147: d'oh, I added leaks with these early-out.  Fix incoming.

libs/vkd3d/heap.c
```diff
@@ -277,13 +281,36 @@ static HRESULT d3d12_heap_init(struct d3d12_heap *heap, struct d3d12_device *dev
...
             &device->memory_allocator, &alloc_info, &heap->allocation)))
     {
         vkd3d_private_store_destroy(&heap->private_store);
         return hr;
     }
 
+    heap->priority.allows_dynamic_residency = 
+        device->device_info.pageable_device_memory_features.pageableDeviceLocalMemory &&
+        heap->allocation.chunk == NULL /* not suballocated */ &&
+        device->memory_properties.memoryTypes[heap->allocation.device_allocation.vk_memory_type].propertyFlags & VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT;
```
HansKristian-Work, line 312: Parens around bit-wise and.

libs/vkd3d/resource.c
```diff
@@ -17,6 +17,7 @@
  * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
  */
 
+#include "vulkan/vulkan_core.h"
```
HansKristian-Work, line 20: Stray include.

libs/vkd3d/heap.c
```diff
@@ -277,13 +281,36 @@ static HRESULT d3d12_heap_init(struct d3d12_heap *heap, struct d3d12_device *dev
...
 
+    if (device->vk_info.EXT_pageable_device_local_memory)
+    {
+        /* this clause isn't trying to reproduce some precise d3d12 behavior,
+           though it's hinted in the public docs that a similar prioritization
+           is done there... and it seems like a good idea anyway. :) */
+        if (heap->desc.Flags & D3D12_HEAP_FLAG_DENY_NON_RT_DS_TEXTURES)
+        {
+            uint32_t adjust = vkd3d_get_priority_adjust(heap->desc.SizeInBytes);
+            heap->priority.d3d12priority = D3D12_RESIDENCY_PRIORITY_HIGH | adjust;
```
HansKristian-Work, line 292: If only mem-prio is supported, it should be included here I think.

libs/vkd3d/resource.c
```diff
@@ -3033,6 +3057,11 @@ HRESULT d3d12_resource_create_committed(struct d3d12_device *device, const D3D12
         object->res.va = object->mem.resource.va;
     }
 
+    object->priority.allows_dynamic_residency = 
+        device->device_info.pageable_device_memory_features.pageableDeviceLocalMemory &&
+        object->mem.chunk == NULL /* not suballocated */ &&
+        device->memory_properties.memoryTypes[object->mem.device_allocation.vk_memory_type].propertyFlags & VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT;
```
HansKristian-Work, line 3063: parens

libs/vkd3d/utils.c
```diff
@@ -647,6 +647,56 @@ VkFormat vkd3d_get_vk_format(DXGI_FORMAT format)
...
+/* Get some size-based low bits for memory prioritization in the same
+   way as d3d12; d3d12 bumps certain resource priorities to
+   D3D12_RESIDENCY_PRIORITY_HIGH + size-based bits (10MB resolution)
+   see: https://learn.microsoft.com/en-us/windows/win32/direct3d12/residency#default-priority-algorithm */
+uint32_t vkd3d_get_priority_adjust(VkDeviceSize size)
+{
+    return min((size / (10 * 1048576)), 0xFFFFUL);
+}
+
+float vkd3d_lerp_u32_to_float(uint32_t uval, uint32_t ustart, uint32_t uend, float fstart, float fend)
```
HansKristian-Work, line 659: Should be static.

libs/vkd3d/utils.c
```diff
@@ -647,6 +647,56 @@ VkFormat vkd3d_get_vk_format(DXGI_FORMAT format)
...
+{
+    float a;
+
+    if (uval <= ustart) return fstart;
+    else if (uval >= uend) return fend;
+    
+    a = (uval - ustart) / (float)(uend - ustart);
```
HansKristian-Work, line 666: Casting both to float seems good to me.

libs/vkd3d/utils.c
```diff
@@ -647,6 +647,56 @@ VkFormat vkd3d_get_vk_format(DXGI_FORMAT format)
...
+    if (d3d12prio <= D3D12_RESIDENCY_PRIORITY_NORMAL)
+        result = vkd3d_lerp_u32_to_float(d3d12prio,
+            D3D12_RESIDENCY_PRIORITY_MINIMUM, D3D12_RESIDENCY_PRIORITY_NORMAL,
+            0.001f, 0.500f);
+    else if (d3d12prio <= D3D12_RESIDENCY_PRIORITY_HIGH)
+        result = vkd3d_lerp_u32_to_float(d3d12prio,
+            D3D12_RESIDENCY_PRIORITY_NORMAL, D3D12_RESIDENCY_PRIORITY_HIGH,
+            0.500f, 0.520f);
+    else
+        result = vkd3d_lerp_u32_to_float(d3d12prio,
```
HansKristian-Work, line 686: Missing the [HIGH, HIGH + 0x10000] band here I think?

libs/vkd3d/resource.c
```diff
@@ -2819,6 +2828,18 @@ static HRESULT d3d12_resource_create(struct d3d12_device *device, uint32_t flags
...
+    if ((flags & VKD3D_RESOURCE_COMMITTED) &&
+        device->device_info.memory_priority_features.memoryPriority &&
+        (desc->Flags & high_priority_resource_flags))
+    {
+        size_t resource_size = d3d12_resource_is_texture(object) ?
+            vkd3d_compute_resource_layouts_from_desc(device, &object->desc, NULL) :
+            object->desc.Width;
+        uint32_t adjust = vkd3d_get_priority_adjust(resource_size);
+
+        object->priority.d3d12priority = D3D12_RESIDENCY_PRIORITY_HIGH | adjust;
```
HansKristian-Work, line 2840: I don't see where HEAP_FLAG_NOT_RESIDENT is handled for committed resources.

HansKristian-Work (changes requested):

libs/vkd3d/utils.c
adamnv: Missed the most important band in my excitement.  Fixed.

libs/vkd3d/resource.c
adamnv: Thought I double-checked but yes!  Fixed now.

adamnv:
All done.  Again.  I think!

libs/vkd3d/resource.c
```diff
@@ -2819,6 +2827,21 @@ static HRESULT d3d12_resource_create(struct d3d12_device *device, uint32_t flags
...
+
+        if (object->heap_flags & D3D12_HEAP_FLAG_CREATE_NOT_RESIDENT)
```
HansKristian-Work, line 2841: Residency must not be set to 0 unless pageable is supported. Otherwise, it would be locked to 0 prio.

libs/vkd3d/heap.c
```diff
@@ -277,13 +281,36 @@ static HRESULT d3d12_heap_init(struct d3d12_heap *heap, struct d3d12_device *dev
...
+        }
+
+        if (heap->desc.Flags & D3D12_HEAP_FLAG_CREATE_NOT_RESIDENT)
```
HansKristian-Work, line 295: Should only be honored when pageable is supported.

libs/vkd3d/utils.c
```diff
@@ -647,6 +647,59 @@ VkFormat vkd3d_get_vk_format(DXGI_FORMAT format)
...
+       0.5 (the default vk prio) so neither kind wins without explicit prio */
+    if (d3d12prio <= D3D12_RESIDENCY_PRIORITY_NORMAL)
+        result = vkd3d_lerp_u32_to_float(d3d12prio,
+            0, D3D12_RESIDENCY_PRIORITY_NORMAL,
+            0.001f, 0.500f);
+    else if (d3d12prio <= D3D12_RESIDENCY_PRIORITY_HIGH)
+        result = vkd3d_lerp_u32_to_float(d3d12prio,
+            D3D12_RESIDENCY_PRIORITY_NORMAL, D3D12_RESIDENCY_PRIORITY_HIGH,
+            0.500f, 0.700f);
+    else if (d3d12prio <= D3D12_RESIDENCY_PRIORITY_HIGH+0xFFFF)
```
HansKristian-Work, line 685: `D3D12_RESIDENCY_PRIORITY_HIGH + 0xffff`

libs/vkd3d/utils.c
```diff
@@ -647,6 +647,59 @@ VkFormat vkd3d_get_vk_format(DXGI_FORMAT format)
...
+
+/* map from 32-bit d3d prio to float (0..1) vk prio. */
+float vkd3d_convert_to_vk_prio(D3D12_RESIDENCY_PRIORITY d3d12prio)
+{
+    float result;
+
+    /* align D3D12_RESIDENCY_PRIORITY_NORMAL (the default d3d12 prio) with
+       0.5 (the default vk prio) so neither kind wins without explicit prio */
+    if (d3d12prio <= D3D12_RESIDENCY_PRIORITY_NORMAL)
+        result = vkd3d_lerp_u32_to_float(d3d12prio,
```
HansKristian-Work, line 678: Nit, multi-line branch should be in braces.

libs/vkd3d/device.c
```diff
@@ -5410,8 +5538,61 @@ static HRESULT STDMETHODCALLTYPE d3d12_device_SetEventOnMultipleFenceCompletion(
 static HRESULT STDMETHODCALLTYPE d3d12_device_SetResidencyPriority(d3d12_device_iface *iface,
         UINT object_count, ID3D12Pageable *const *objects, const D3D12_RESIDENCY_PRIORITY *priorities)
 {
-    FIXME_ONCE("iface %p, object_count %u, objects %p, priorities %p stub!\n",
```
HansKristian-Work, line 5413: FIXME must be replaced with TRACE here.

libs/vkd3d/device.c
```diff
@@ -5106,17 +5125,126 @@ static HRESULT STDMETHODCALLTYPE d3d12_device_OpenSharedHandleByName(d3d12_devic
...
+    d3d12_device_MakeResident(iface, num_objects, objects);
+
+    /* we don't block anyway - signal the fence immediately */
+    return ID3D12Fence_Signal(fence_to_signal, fence_value_to_signal);
+}
+
 static HRESULT STDMETHODCALLTYPE d3d12_device_Evict(d3d12_device_iface *iface,
         UINT object_count, ID3D12Pageable * const *objects)
 {
-    FIXME_ONCE("iface %p, object_count %u, objects %p stub!\n",
```
HansKristian-Work, line 5118: Replace with TRACE

libs/vkd3d/device.c
```diff
@@ -5106,17 +5125,126 @@ static HRESULT STDMETHODCALLTYPE d3d12_device_OpenSharedHandleByName(d3d12_devic
 static HRESULT STDMETHODCALLTYPE d3d12_device_MakeResident(d3d12_device_iface *iface,
         UINT object_count, ID3D12Pageable * const *objects)
 {
-    FIXME_ONCE("iface %p, object_count %u, objects %p stub!\n",
```
HansKristian-Work, line 5109: Replace with TRACE.

HansKristian-Work (changes requested):
Two bugs and some minor nits.

Please squash all commits into one with intended commit message.

HansKristian-Work (approved):
LGTM, thanks!

PR #1544, opened 2023-05-01
Add debug ring support to normal ExecuteIndirect

HansKristian-Work:
When debugging execute indirect faults, we need to know the input arguments to have any hope chasing things down.


HansKristian-Work:
Obsoleted.

PR #1546, opened 2023-05-02
tests: add a multi-threaded test runner

zmike:
# Background

the test suite here is a bit minimal and runs itself synchronously, which is problematic for cases where:
* a test runs itself for any amount of time
* a test crashes

in the former case, all 511 other threads of the user's powerful supercomputer are sitting idle instead of being put to work

in the latter case, the entire test process stops and no further test results can be gathered without either starting a new run that filters out the crashing test(s) or fixing the crash

this test runner solves both issues by running one test case per cpu and then tracking the failures such that all cores are always in use and any crashes don't affect the rest of the tests

# Performance
obviously I gotta talk about perf, so here's some rough data:

|Num CPU threads|Driver|Normal runtime (seconds)|Runner runtime (seconds)|%change|
|----------------|-------|----|--------|-------|
|8|ANV|86|33|62%|
|64|RADV|46|8|83%|

# Sample outputs
```
Running 313 vkd3d-proton D3D12 tests...
FAILED test_object_interface
FAILED test_shader_instructions_dxil
107 / 313 complete (2 failures)...
FAILED test_shader_instructions
203 / 313 complete (3 failures)...
FAILED test_pipeline_library
FAILED test_undefined_read_typed_buffer_as_untyped_dxbc
FAILED test_undefined_read_typed_buffer_as_untyped_dxil
237 / 313 complete (6 failures)...
FAILED test_advanced_cbv_layout
301 / 313 complete (7 failures)...
310 / 313 complete (7 failures)...
***********************
Finished in 32s!
7 FAILURES: (run with 'VKD3D_TEST_FILTER=<name> build/tests/d3d12')
test_object_interface
test_shader_instructions_dxil
test_shader_instructions
test_pipeline_library
test_undefined_read_typed_buffer_as_untyped_dxbc
test_undefined_read_typed_buffer_as_untyped_dxil
test_advanced_cbv_layout
```

```
Running 313 vkd3d-proton D3D12 tests...
255 / 313 complete (0 failures)...
***********************
Finished in 8s!
ALL PASSED!
```

### DIFF
--- a/tests/test-runner.sh
+++ b/tests/test-runner.sh
@@ -1,0 +1,141 @@
+#!/usr/bin/env bash
+
+# arg variable init
+output_dir=
+run_stress=
+args=()
+# by default run one job per cpu thread
+nr_cpus=$(grep -w processor /proc/cpuinfo|wc -l)
+while [[ $# -gt 0 ]]; do
+	case $1 in
+	-o|--output-dir)
+		# this outputs the logs for each test to a separate file
+		output_dir="$2"
+		shift
+		shift
+		;;
+	-s|--run-stress-tests)
+		# stress tests take a long time to run and are disabled by default
+		run_stress=1
+		shift
+		;;
+	-j|--jobs)
+		# override the default job count
+		nr_cpus=$2
+		shift
+		shift
+		;;
+	-h|--help)
+		echo "./test-runner.sh [-o|--output-dir logfile_dir] [-s|--run-stress-tests] [-j|--jobs N] path/to/tests/d3d12"
+		exit 1
+		;;
+	-*|--*)
+		echo "unknown arg $1"
+		;;
+	*)
+		args+=("$1")
+		shift
+		;;
+	esac
+done
+
+set -- "${args[@]}"
+d3d12_bin="$1"
+
+if [[ -z $d3d12_bin || ! -f "$d3d12_bin" ]] ; then
+	echo "Must specify path to valid d3d12 test exe!"
+	exit 1
+fi
+
+tests=()
+if [[ -z $run_stress ]] ; then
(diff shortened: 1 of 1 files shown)


test-runner.sh
```diff
@@ -0,0 +1,141 @@
...
+set -- "${args[@]}"
+d3d12_bin="$1"
+
+if [[ -z $d3d12_bin || ! -f "$d3d12_bin" ]] ; then
+	echo "Must specify path to valid d3d12 test exe!"
+	exit 1
+fi
+
+tests=()
+if [[ -z $run_stress ]] ; then
```
HansKristian-Work, line 51: This is such a cute way to get which tests to iterate over.

test-runner.sh
```diff
@@ -0,0 +1,141 @@
...
+		last_notify=$cur_time
+	fi
+done
+
+
+# nice summary output at the end
+echo "***********************"
+echo "Finished in ${SECONDS}s!"
+
+if [[ ${#fails[@]} != 0 ]] then
```
HansKristian-Work, line 134:
```
Running 313 vkd3d-proton D3D12 tests...
174 / 313 complete (0 failures)...
285 / 313 complete (0 failures)...
***********************
Finished in 12s!
./test-runner.sh: line 134: syntax error near unexpected token `then'
./test-runner.sh: line 134: `if [[ ${#fails[@]} != 0 ]] then'
```

test-runner.sh
HansKristian-Work, line 1: Would probably make more sense to move this script into tests/.

HansKristian-Work (changes requested):
Seems to be a script bug.

test-runner.sh
zmike: oops typo

HansKristian-Work (approved):
LGTM.

PR #1550, opened 2023-05-09
Fix root signature feature check

### LINKED ISSUE #1548: Improvement of ID3D12Device::CheckFeatureSupport for D3D12_FEATURE_ROOT_SIGNATURE
For now, m_device->CheckFeatureSupport(D3D12_FEATURE_ROOT_SIGNATURE, &featureData, sizeof(featureData))) always outputs nothing in featureData (D3D12_FEATURE_DATA_ROOT_SIGNATURE). For other features, I wonder if CheckFeatureSupport should be necessary. However, it seems CheckFeatureSupport(D3D12_FEATURE_ROOT_SIGNATURE) for the highest root signature version can be commonly used in DX12 apps for root signature compatibility (as DX12 has a versioned root signature creation method), and thus I think vkd3d-proton should at least put D3D_ROOT_SIGNATURE_VERSION_1_0 into D3D12_FEATURE_DATA_ROOT_SIGNATURE::HighestVersion.

Although I can make a workaround manually by:
D3D12_FEATURE_DATA_ROOT_SIGNATURE featureData = {};

if (FAILED(m_device->CheckFeatureSupport(D3D12_FEATURE_ROOT_SIGNATURE,
		&featureData, sizeof(featureData))) || !featureData.HighestVersion)
		return D3D_ROOT_SIGNATURE_VERSION_1_0;

I think it would be better to handle it by vkd3d-proton. Thanks.

### Software information
Name of the game, settings used etc.

### System information
- GPU: AMD RX6800 and NVIDIA RTX 3080
- Driver: both latest
- Wine version: NA
- VKD3D-Proton version: latest staging

### Log files
NA. I made a breakpoint to check the output value from CheckFeatureSupport.

doitsujin:
Turns out that the runtime will return `E_INVALIDARG` when the app passes in an unrecognized or unsupported root signature version, which makes this whole function a bit... weird.

Fixes #1548.

### DIFF
--- a/libs/vkd3d/device.c
+++ b/libs/vkd3d/device.c
@@ -3652,6 +3652,12 @@ static HRESULT STDMETHODCALLTYPE d3d12_device_CheckFeatureSupport(d3d12_device_i
                 return E_INVALIDARG;
             }
 
+            if (!data->HighestVersion || data->HighestVersion > D3D_ROOT_SIGNATURE_VERSION_1_1)
+            {
+                WARN("Unrecognized root signature version %#x.\n", data->HighestVersion);
+                return E_INVALIDARG;
+            }
+
             TRACE("Root signature requested %#x.\n", data->HighestVersion);
             data->HighestVersion = min(data->HighestVersion, D3D_ROOT_SIGNATURE_VERSION_1_1);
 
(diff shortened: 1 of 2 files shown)


HansKristian-Work (approved):
LGTM, I suppose we'll need to bump that check 1.2 when we add that.

PR #1561, opened 2023-05-18
test_denorm_behavior: early breakout from loop if the compute pipeline couldn't be created

shadeslayer:
Some additional null checks


HansKristian-Work:
This is not an acceptable fix. The proper fix would be in dxil-spirv.

PR #1582, opened 2023-05-31
d3d12: Add D3D12PIXEventsReplaceBlock and D3D12PIXGetThreadInfo stub …

nqsxw:
…to fix failing to load DirectML.dll

If we use vkd3d-proton in an application with DirectML enabled. The program will error out Entry Point Not Found.

The procedure entry point D3D12PIXEventsReplaceBlock/D3D12PIXGetThreadInfo could not be located in the dynamic link library C:\Windows\SYSTEM32\DirectML.dll

And the DirectML.dll will be unloaded.

### DIFF
--- a/libs/d3d12/main.c
+++ b/libs/d3d12/main.c
@@ -206,3 +206,15 @@ HRESULT WINAPI DLLEXPORT D3D12GetInterface(REFCLSID rcslid, REFIID iid, void **d
         return E_NOINTERFACE;
     return IVKD3DCoreInterface_GetInterface(core, rcslid, iid, debug);
 }
+
+HRESULT WINAPI DLLEXPORT D3D12PIXEventsReplaceBlock()
+{
+    FIXME("stub!\n");
+    return E_NOTIMPL;
+}
+
+HRESULT WINAPI DLLEXPORT D3D12PIXGetThreadInfo()
+{
+    FIXME("stub!\n");
+    return E_NOTIMPL;
+}
--- a/libs/d3d12/d3d12.def
+++ b/libs/d3d12/d3d12.def
@@ -9,3 +9,6 @@ EXPORTS
     D3D12EnableExperimentalFeatures
     D3D12SerializeRootSignature
     D3D12SerializeVersionedRootSignature
+
+    D3D12PIXEventsReplaceBlock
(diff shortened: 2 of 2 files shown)


HansKristian-Work:
This is very questionable. Does the native d3d12.dll runtime export these symbols?

nqsxw:
> This is very questionable. Does the native d3d12.dll runtime export these symbols?

Yeah. It should be. You can try to use GetProcAddress to get that symbol from the system d3d12.dll.

And I found the MinGW runtime also exports the symbols. This link [d3d12.def](https://third-party-mirror.googlesource.com/mingw-w64/+/refs/heads/master/mingw-w64-crt/lib-common/d3d12.def) should be helpful.

libs/d3d12/main.c
```diff
@@ -206,3 +206,13 @@ HRESULT WINAPI DLLEXPORT D3D12GetInterface(REFCLSID rcslid, REFIID iid, void **d
         return E_NOINTERFACE;
     return IVKD3DCoreInterface_GetInterface(core, rcslid, iid, debug);
 }
+
+void WINAPI DLLEXPORT D3D12PIXEventsReplaceBlock()
```
HansKristian-Work, line 210: At least the return types should be correct here and return E_NOTIMPL. I doubt these functions return void.

doitsujin: Similarly, printing a FIXME probably screws with some registers so all these should probably do is return.... something.

nqsxw: Yes. I approve. I've updated the codes to return the E_NOTIMPL. Could you help review them?

libs/d3d12/main.c
```diff
@@ -206,3 +206,15 @@ HRESULT WINAPI DLLEXPORT D3D12GetInterface(REFCLSID rcslid, REFIID iid, void **d
         return E_NOINTERFACE;
     return IVKD3DCoreInterface_GetInterface(core, rcslid, iid, debug);
 }
+
+HRESULT WINAPI DLLEXPORT D3D12PIXEventsReplaceBlock()
```
HansKristian-Work, line 210: Does it really take no arguments? () != (void) in C fwiw. Proper return types is important for WINAPI calling convention.

PR #1591, opened 2023-06-09
Experimentally implement SM 6.7

HansKristian-Work:
For now, this needs to be enabled as an opt-in, because we cannot fully implement it with Vulkan as-is without relying on magic driver behavior, but fortunately the features happen to work as-is on at least RADV and NV.

This adds:

- MSAA UAV
  - Trivial Vulkan catch-up
- QuadAny/All
  - This is more awkward, but based on testing and docs, all lanes must be active, so a Quad shuffle is sufficient, and can be optimized by compilers.
- WaveOpsIncludeHelperLanes
  - This is technically default Vulkan behavior, but spec allows for helper lanes to flicker out for non-quad ops, but generally this works as expected. To be 100% compliant, we need a new extension to guarantee this behavior.
- SampleCmpLevel
  - Trivial catch-up
- RawGather
  - It is actually just GatherRed on UINT textures. Fortunately, the image views must be R16/R32/R32G32_UINT, so this is just syntactic sugar, except for 64-bit, which is just 2 gathers with bitcasts.
- Integer sampling
  - Straight Vulkan catch-up.
  - Border color on stencil is awkward. I've verified that AMD and NV differ in behavior in native drivers.
- Dynamic texture offsets on non-gather instructions
  - There is no feature for this in Vulkan yet, but it just "happens to work" on both RADV and NV at least. To be 100% compliant, we need a new extension to guarantee this behavior.


HansKristian-Work:
@Blisto91 @runar-work For testing, this PR significantly changes shader outputs in:

- F1 2021
- F1 2022
- Farming Simulator 2022
- Halo Infinite
- Monster Hunter Rise (and Sunbreak)
- RE2
- RE8
- Capcom 2nd Stadium
- CP77
- Ghost and Goblins Resurrected
- Jedi Survivor
- RE4
- SF6

runar-work:
On NV:

- Cyberpunk: Had some random crashes and issues with Xid 31, but reproduced with plain PE
- F1 2021: Xid 69 when trying to start a race, but again reproduced with plain PE
- RE2, RE8: tested for about 15-20 minutes each without issues
- RE4: Xid 109 and some rendering glitches, but both are familiar problems

On RADV:

- F1 2021: GPU hangs after 5-20 minutes, reproduced with plain BE
- Halo Infinite: Some weird LOD glitch or something which makes the geometry extremely low-poly, same with plain BE
- MHR, Ghosts 'n Goblins Resurrected, Jedi Survivor, RE4: Tested for 20 minutes without issues
- Cyberpunk: framerate dropped massively and system turned unresponsive after about an hour of playing (when the ambulance shows up in the first mission), no sign of GPU hang in dmesg, and wasn't able to reproduce this again in the same location after 20 minutes of trying

Blisto91:
Checked the remaining games on RADV. F1 2022, Farming Simulator 2022, RE2, RE8, Capcom 2nd Stadium and SF6 tested for around 20 min without anything out of the ordinary.
Also poked at Cyberpunk for a bit, but didn't encounter anything like Runars issue so far.

libs/vkd3d/resource.c
```diff
@@ -2215,6 +2215,13 @@ HRESULT d3d12_resource_validate_desc(const D3D12_RESOURCE_DESC1 *desc, struct d3
             }
             /* Fall through. */
         case D3D12_RESOURCE_DIMENSION_TEXTURE2D:
+            if (desc->SampleDesc.Count > 1 &&
+                    !(desc->Flags & (D3D12_RESOURCE_FLAG_ALLOW_DEPTH_STENCIL | D3D12_RESOURCE_FLAG_ALLOW_RENDER_TARGET)))
+            {
+                WARN("Multi-sampled textures must be created with render target or depth attachment usage.\n");
+                return E_INVALIDARG;
+            }
+            /* Fall through */
```
doitsujin, line 2224: Theis whole sample count thing seems weird, we're implicitly allowing multisampled 1D + 3D textures for some reason. Is this intended?

We might also want to error out if ALLOW_UAV is set with >1 sample and the device feature is not supported.

HansKristian-Work: Right, that was not intended.

HansKristian-Work: Fixed.

doitsujin (approved):

PR #1592, opened 2023-06-09
Add tar.gz as an option archive for the release

crramirez:
I am using a tool (CrossOver) that doesn't recognize the tar.zst archive type. 

Another alternative archive type can be added?


K0bin:
Why does Crossover need to support the archive type? Why do you use Crossover in general instead of Proton?

crramirez:
I am adding an installer for a game that needs the d3d12.dll in this repo. Because the built-in is not enough. 

I prefer using Crossover for games outside Steam like the battle net ones. I am a collaborator with CodeWeavers.

I use Crossover since 2006.

csutcliff:
> I am a collaborator with CodeWeavers.
Sounds like you should collaborate with them to implement zstd archive support

crramirez:
Yes I raised the flag there, it involves changing a python library or using another one. But probably the change won't arrive until CrossOver 23.

The change here is easier, if it only involves what I've done. Also, I noticed that gz produces a smaller file than zst.

PR #1593, opened 2023-06-12
Bump API versions for (static) samplers.

HansKristian-Work:
To test SM 6.7 which is in flight, we need to be able to test uint border colors.

This requires CreateSampler2 from ID3D12Device11 as well as Root Signature 1.2 to test the static variant of that.


libs/vkd3d/resource.c
```diff
@@ -5899,7 +5902,7 @@ void d3d12_desc_create_sampler_embedded(vkd3d_cpu_descriptor_va_t desc_va,
 }
 
 void d3d12_desc_create_sampler(vkd3d_cpu_descriptor_va_t desc_va,
-        struct d3d12_device *device, const D3D12_SAMPLER_DESC *desc)
+        struct d3d12_device *device, const D3D12_SAMPLER_DESC2 *desc)
```
doitsujin, line 5910: Should probably put the FIXME about ignoring flags here too?

libs/vkd3d/vkd3d_main.c
```diff
@@ -344,18 +344,7 @@ static ULONG STDMETHODCALLTYPE d3d12_versioned_root_signature_deserializer_Relea
...
-            return VKD3D_ROOT_SIGNATURE_VERSION_1_0;
-        case D3D_ROOT_SIGNATURE_VERSION_1_1:
-            return VKD3D_ROOT_SIGNATURE_VERSION_1_1;
-        case D3D_ROOT_SIGNATURE_VERSION_1_2:
-            return VKD3D_ROOT_SIGNATURE_VERSION_1_2;
-        default:
-            WARN("Unknown root signature version %#x.\n", version);
-            return 0;
-    }
+    return (enum vkd3d_root_signature_version)version;
```
doitsujin, line 347: why do we even have our own enum for this at this point?

HansKristian-Work: I'm not sure why we declare internal structs which are clearly supposed to be direct aliases with the d3d ones. It might be a holdover from when vkd3d-shader was a standalone .so thing ...

HansKristian-Work: We can probably just refactor this away later, but I'd rather not do that yak shave right now.

libs/vkd3d/resource.c
HansKristian-Work: Fixed.

doitsujin (approved):

PR #1595, opened 2023-06-14
vkd3d: Protect access to the view map with the spinlock.

giomasce:
Just spotted a little oversight. I didn't test it, except that it builds.


libs/vkd3d/resource.c
```diff
@@ -4971,6 +4971,8 @@ static struct vkd3d_view *vkd3d_create_texture_srv_view(struct d3d12_device *dev
     {
         key.u.texture.miplevel_clamp = floor(key.u.texture.miplevel_clamp);
 
+        rw_spinlock_acquire_read(&resource->view_map.spinlock);
```
HansKristian-Work, line 4974: Better to acquire_read, assign hash_map_find to bool, then release.

HansKristian-Work:
Tried booting up Halo Infinite, and it ran fine, so the smoke test passes.

PR #1610, opened 2023-06-29
Implement Enhanced Barrier's Resource Format Casting - Phase 3/3

HansKristian-Work:
Adds implementation. It's mostly plumbing, but changes some minor things:

- Add strict format feature validation on resource creation.
- Change UAV Uint clear to deal with format cast list.
- Plumb through the format list everywhere it's needed.


libs/vkd3d/resource.c
```diff
@@ -2263,6 +2339,95 @@ HRESULT d3d12_resource_validate_desc(const D3D12_RESOURCE_DESC1 *desc, struct d3
...
+                    return E_INVALIDARG;
+        }
+        else
+        {
+            const struct vkd3d_vk_device_procs *vk_procs = &device->vk_procs;
+            VkFormatFeatureFlags required_flags = 0;
+            VkFormatFeatureFlags total_flags = 0;
+            VkFormatProperties format_properties;
+
+            format = vkd3d_format_from_d3d12_resource_desc(device, desc, 0);
```
HansKristian-Work, line 2359: We might want to make this general, where we validate all resource descs, but it will mean constructing a temporary format cast list and iterating over that for the non-castable path.

tests/d3d12_enhanced_barriers.c
```diff
@@ -259,6 +259,260 @@ void test_enhanced_barrier_castable_formats_validation(void)
...
+        DXGI_FORMAT tex_format;
+        DXGI_FORMAT dsv_format;
+        DXGI_FORMAT srv_format;
+        float clear_value;
+        uint32_t reference_value;
+        unsigned int pso_index;
+    } tests[] = {
+        { DXGI_FORMAT_D32_FLOAT, DXGI_FORMAT_D32_FLOAT, DXGI_FORMAT_R32_FLOAT, 1.0f, 0x3f800000, PSO_INDEX_FLOAT },
+        { DXGI_FORMAT_D32_FLOAT, DXGI_FORMAT_D32_FLOAT, DXGI_FORMAT_R32_FLOAT, 0.5f, 0x3f000000, PSO_INDEX_FLOAT },
+        { DXGI_FORMAT_D32_FLOAT, DXGI_FORMAT_D32_FLOAT, DXGI_FORMAT_R32_FLOAT, 0.0f, 0, PSO_INDEX_FLOAT },
```
HansKristian-Work, line 297: Should probably add a test for R32_FLOAT as base format with D32_FLOAT as castable. That might break our current impl ...

libs/vkd3d/resource.c
```diff
@@ -244,8 +244,72 @@ static unsigned int max_miplevel_count(const D3D12_RESOURCE_DESC1 *desc)
...
+
+    if (list.format_count < 2)
+        return false;
+
+    *vk_flags |= VK_IMAGE_CREATE_MUTABLE_FORMAT_BIT;
+
+    /* If we overflow the format list for whatever reason,
+     * just fall back to plain MUTABLE_FORMAT_BIT. */
+    if (list.format_count > ARRAY_SIZE(list.vk_formats))
+        return false;
```
HansKristian-Work, line 305: Hmmmm, returning early here could screw over Clear UAV Uint.

libs/vkd3d/utils.c
```diff
@@ -391,13 +391,24 @@ void vkd3d_format_compatibility_list_add_format(struct vkd3d_format_compatibilit
...
-        assert(list->format_count < ARRAY_SIZE(list->vk_formats));
-        list->vk_formats[list->format_count++] = vk_format;
+        if (list->format_count < ARRAY_SIZE(list->vk_formats))
+        {
+            list->vk_formats[list->format_count++] = vk_format;
+        }
+        else
+        {
+            /* Mark overflow. Caller can fall back to plain MUTABLE_FORMAT_BIT without format list. */
+            list->format_count = VKD3D_FORMAT_COMPAT_LIST_ALL_FORMATS;
```
doitsujin, line 410: Not sure I like this design - might just want to `return false` on overflow and print a warning or something, I feel like that's safer and easier to reason about without having to be aware of the format_count quirk.

HansKristian-Work: Would it make more sense to use formatCount == 0 as the sentinel?

HansKristian-Work: Updated.

doitsujin (approved):

PR #1611, opened 2023-06-29
vkd3d: Implement negative viewport height feature.

null77:
This mirrors a Vulkan feature and doesn't require any emulation.

@HansKristian-Work LMK if you'd prefer I de-duplicate the test code.


libs/vkd3d/device.c
```diff
@@ -6949,6 +6946,15 @@ static void d3d12_device_caps_init_feature_options11(struct d3d12_device *device
             device->device_info.properties2.properties.limits.minStorageBufferOffsetAlignment <= 16;
 }
 
+static void d3d12_device_caps_init_feature_options13(struct d3d12_device *device)
+{
+    D3D12_FEATURE_DATA_D3D12_OPTIONS13 *options13 = &device->d3d12_caps.options13;
+
+    /* Negative viewports become core in Vulkan 1.2. */
+    assert(VKD3D_MIN_API_VERSION >= VK_API_VERSION_1_2);
```
HansKristian-Work, line 6954: This assert is misleading. We always required negative viewports. Non-negative viewports in D3D12 become negative viewports in Vulkan. Negative viewport in D3D12 will become non-negative I think.

tests/d3d12_command.c
```diff
@@ -490,6 +490,163 @@ void test_fractional_viewports(void)
...
+            context.root_signature, desc.rt_format, &vs, &ps, &input_layout);
+
+    vb = create_upload_buffer(context.device, sizeof(quad), quad);
+
+    vbv.BufferLocation = ID3D12Resource_GetGPUVirtualAddress(vb);
+    vbv.StrideInBytes = sizeof(*quad);
+    vbv.SizeInBytes = sizeof(quad);
+
+    set_viewport(&viewport, 0, context.render_target_desc.Height,
+            context.render_target_desc.Width, -(float)context.render_target_desc.Height, 0.0f, 1.0f);
```
HansKristian-Work, line 609: I think a viewport test like this should test 4 draw calls targeting 4 quadrants to make sure the intended X / Y offsets end up where we expect.

tests/d3d12_command.c
```diff
@@ -490,6 +490,163 @@ void test_fractional_viewports(void)
...
+
+    ID3D12GraphicsCommandList_ClearRenderTargetView(command_list, context.rtv, white, 0, NULL);
+
+    ID3D12GraphicsCommandList_OMSetRenderTargets(command_list, 1, &context.rtv, false, NULL);
+    ID3D12GraphicsCommandList_SetGraphicsRootSignature(command_list, context.root_signature);
+    ID3D12GraphicsCommandList_SetPipelineState(command_list, context.pipeline_state);
+    ID3D12GraphicsCommandList_IASetPrimitiveTopology(command_list, D3D_PRIMITIVE_TOPOLOGY_TRIANGLESTRIP);
+    ID3D12GraphicsCommandList_IASetVertexBuffers(command_list, 0, 1, &vbv);
+    ID3D12GraphicsCommandList_RSSetViewports(command_list, 1, &viewport);
```
HansKristian-Work, line 618: Were no changes to Vulkan code required here? Maybe the math just happens to work out.

tests/d3d12_command.c
```diff
@@ -490,6 +490,163 @@ void test_fractional_viewports(void)
...
+    desc.rt_format = DXGI_FORMAT_R32G32B32A32_FLOAT;
+    desc.no_root_signature = true;
+    if (!init_test_context(&context, &desc))
+        return;
+    command_list = context.list;
+    queue = context.queue;
+    device = context.device;
+
+    memset(&options13, 0, sizeof(options13));
+    hr = ID3D12Device_CheckFeatureSupport(device, D3D12_FEATURE_D3D12_OPTIONS13, &options13, sizeof(options13));
```
HansKristian-Work, line 583: I think this would require AgilitySDK. Did you figure out how to run tests on Windows with that?

HansKristian-Work (changes requested):
Smol nits.

libs/vkd3d/device.c
null77: Right right. Removed assert.

tests/d3d12_command.c
null77: Didn't seem to be. 🤞

tests/d3d12_command.c
null77: Yep, I tested with the Agility SDK.

tests/d3d12_command.c
null77: Sounds good, done. PTAL

tests/d3d12_command.c
```diff
@@ -490,6 +490,197 @@ void test_fractional_viewports(void)
...
+    };
+    static const struct quad
+    {
+        struct vec4 position;
+        struct vec2 texcoord;
+    }
+    quads[4][4] =
+    {
+        {
+            {{0.0f,  1.0f, 1.0f, 1.0f}, {0.1f, 0.1f}},
```
HansKristian-Work, line 568: There is flat TEXCOORD input here to PS, so it does not verify that Y-flip actually happens.

tests/d3d12_command.c
```diff
@@ -490,6 +490,197 @@ void test_fractional_viewports(void)
...
+        0x00000001, 0x0100003e,
+    };
+    static const D3D12_SHADER_BYTECODE vs = {vs_code, sizeof(vs_code)};
+    static const DWORD ps_code[] =
+    {
+#if 0
+        float4 main(float4 position : SV_Position,
+                float2 texcoord : TEXCOORD) : SV_Target
+        {
+            return float4(position.xyz, texcoord.x);
```
HansKristian-Work, line 541: Should be texcoord.y here I think, since that will be "flipped" when viewport is flipped.

tests/d3d12_command.c
```diff
@@ -490,6 +490,197 @@ void test_fractional_viewports(void)
...
+    ID3D12GraphicsCommandList_OMSetRenderTargets(command_list, 1, &context.rtv, false, NULL);
+    ID3D12GraphicsCommandList_SetGraphicsRootSignature(command_list, context.root_signature);
+    ID3D12GraphicsCommandList_SetPipelineState(command_list, context.pipeline_state);
+    ID3D12GraphicsCommandList_IASetPrimitiveTopology(command_list, D3D_PRIMITIVE_TOPOLOGY_TRIANGLESTRIP);
+    ID3D12GraphicsCommandList_RSSetViewports(command_list, 1, &viewport);
+    ID3D12GraphicsCommandList_RSSetScissorRects(command_list, 1, &context.scissor_rect);
+
+    for (i = 0; i < 4; i++)
+    {
+        ID3D12GraphicsCommandList_IASetVertexBuffers(command_list, 0, 1, &vbvs[i]);
```
HansKristian-Work, line 645: I meant that we need 4 different RSSetViewports here rather than using the VBO to do the same thing.

tests/d3d12_command.c
null77: Changed to a variable TEXCOORD Y.

tests/d3d12_command.c
null77: Right, changed to output y.

tests/d3d12_command.c
null77: Done, PTAL.

HansKristian-Work (approved):
Passes on both AMD and NV. LGTM.

PR #1614, opened 2023-06-30
Add robustness to interface queries.

HansKristian-Work:


libs/vkd3d/utils.c
```diff
@@ -808,6 +808,13 @@ HRESULT return_interface(void *iface, REFIID iface_iid,
     IUnknown *unknown = iface;
     HRESULT hr;
 
+    /* Don't check IID here, we always have to return S_FALSE. */
+    if (!object)
+    {
+        IUnknown_Release(unknown);
```
Joshua-Ashton, line 814: Why release? We only release normally because we actually called QueryInterface which increments...?

PR #1620, opened 2023-07-04
vkd3d: Mask embedded metadata VA.

HansKristian-Work:
Fixes unaligned memory access issue in packed metadata. No functional issue on x86, but still very dumb to rely on.


doitsujin (approved):

ishitatsuyuki (approved):
Tests passing on RADV with no alignment warnings from UBSan.

PR #1626, opened 2023-07-06
vkd3d: Close memory handle in d3d12_device_CreateSharedHandle().

gofman:
vkGetMemoryWin32HandleKHR with VK_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_WIN32_BIT returns a duplicated handle which needs to be closed by the caller (otherwise leaking shared resource memory).

### DIFF
--- a/libs/vkd3d/device.c
+++ b/libs/vkd3d/device.c
@@ -5022,6 +5022,7 @@ static HRESULT STDMETHODCALLTYPE d3d12_device_CreateSharedHandle(d3d12_device_if
                 if (!vkd3d_set_shared_metadata(*handle, &metadata, sizeof(metadata)))
                     ERR("Failed to set metadata for shared resource, importing created handle will fail.\n");
             }
+            CloseHandle(*handle);
         }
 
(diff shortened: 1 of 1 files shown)


libs/vkd3d/device.c
```diff
@@ -5022,6 +5022,7 @@ static HRESULT STDMETHODCALLTYPE d3d12_device_CreateSharedHandle(d3d12_device_if
                 if (!vkd3d_set_shared_metadata(*handle, &metadata, sizeof(metadata)))
                     ERR("Failed to set metadata for shared resource, importing created handle will fail.\n");
             }
+            CloseHandle(*handle);
```
HansKristian-Work, line 5025: This looks strange. This is in CreateSharedHandle, and we're supposed to return a proper handle to the caller here, right?

gofman: Oh yeah, sorry for the noise. Some tests would help to avoid such stupid mistakes, I think we will get there with shared resources.

PR #1627, opened 2023-07-06
Implement Enhanced Barrier format casting - Phase 2

HansKristian-Work:
Adds tests.

- RGBA8 / R32 casting.
- Use new style of DSV casting. Verify rendered result.
- Extensive checks for resource creation validation. Found lots of weird and wonderful behavior ...
- Verify weird things you can do with resource casting on buffers (cast to DXGI_FORMAT_UNKNOWN :frog:).
- To validate resource casting more exhaustively, I ran it through Vulkan validation to make sure there were no validation errors regarding missing formats in the VkImageFormatList.

"Features" that were left out from testing:
- Apparently you can cast between unrelated BC formats. Validation does not complain. This is not technically allowed in Vulkan since BC formats live in their own format class, but it's one of those things that will "just work" I'd imagine if any application is crazy enough to rely on that. (It's also a meaningless usecase, unlike BC <-> UINT which is now supported in D3D12).
- Apparently validation does not complain if you try to cast D32 to R32_UINT and even non-sense like R16G16_UINT. I proved some failures when doing this on AMD Windows, so it's likely unintended.

We have to use the new enhanced barrier creation APIs here, so using legacy barriers with that interface is mildly questionable. We have to go via COMMON layout to do this. The D3D12 validation layers are very crash prone here if you don't do it exactly like this.

### DIFF
--- a/tests/d3d12_enhanced_barriers.c
+++ b/tests/d3d12_enhanced_barriers.c
@@ -1,0 +1,318 @@
+/*
+ * Copyright 2023 Hans-Kristian Arntzen for Valve Corporation
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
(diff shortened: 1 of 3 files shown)


doitsujin (approved):
Looks like this is another well-designed and well-tested feature in the D3D12 Agility SDK...

PR #1633, opened 2023-07-19
Add no_upload_hvv workaround to Halo Infinite

Etaash-mathamsetty:
Seems to be causing long loading times and poor performance for some people, I guess we can work around it for now.

### DIFF
--- a/libs/vkd3d/device.c
+++ b/libs/vkd3d/device.c
@@ -465,11 +465,13 @@ static const struct vkd3d_instance_application_meta application_override[] = {
      * This works okay with zerovram on first game boot, but not later, since this memory is guaranteed to be recycled.
      * Game also relies on indirectly modifying CBV root descriptors, which means we are forced to rely on RAW_VA_CBV.
      * It also relies on multi-dispatch indirect with state updates which is ... ye.
-     * Need another config flag to workaround that as well. */
+     * Need another config flag to workaround that as well.
+     * Poor loading times and performance with ReBar on some devices.
+     */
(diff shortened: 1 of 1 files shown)


jodaco:
Can confirm that this takes loading time from about a minute or so down to about 10 seconds.

libs/vkd3d/device.c
```diff
@@ -465,11 +465,13 @@ static const struct vkd3d_instance_application_meta application_override[] = {
      * This works okay with zerovram on first game boot, but not later, since this memory is guaranteed to be recycled.
      * Game also relies on indirectly modifying CBV root descriptors, which means we are forced to rely on RAW_VA_CBV.
      * It also relies on multi-dispatch indirect with state updates which is ... ye.
-     * Need another config flag to workaround that as well. */
+     * Need another config flag to workaround that as well.
+     * Poor loading times and performance with ReBar on some devices.
```
MadLittleMods, line 469: @Etaash-mathamsetty Heya, just poking as I noticed this was fixed in-game and went looking for what fixed it. Thanks for the addressing :heart: 

I'm curious how you narrowed down long load times in Halo Infinite to this fix? Any good tips to debug? Is there discussion/context around this problem somewhere?

Etaash-mathamsetty: it was found by accident

Blisto91: It had a similar effect in AoW4 and was tried and tested a bunch for this game too when the issue was being looked in to.

PR #1640, opened 2023-07-26
Specify + Implement ID3D12GraphicsCommandListExt1 with CuBIN smem Support

liam-middlebrook:
Specify `ID3D12GraphicsCommandListExt1` which contains a new function for creating CuBINs with smem usage. This function mirrors the api of the NvAPI function `NvAPI_D3D12_CreateCubinComputeShaderEx`.

This functionality is needed for CuBINs that use smem.

### DIFF
--- a/include/vkd3d_command_list_vkd3d_ext.idl
+++ b/include/vkd3d_command_list_vkd3d_ext.idl
@@ -30,3 +30,13 @@ interface ID3D12GraphicsCommandListExt : IUnknown
    HRESULT LaunchCubinShader(D3D12_CUBIN_DATA_HANDLE *handle, UINT32 block_x, UINT32 block_y, UINT32 block_z, const void *params, UINT32 param_size);
 }
 
+[
+    uuid(d53b0028-afb4-4b65-a4f1-7b0daaa65b4f),
+    object,
+    local,
+    pointer_default(unique)
+]
+interface ID3D12GraphicsCommandListExt1 : ID3D12GraphicsCommandListExt
+{
+   HRESULT LaunchCubinShaderEx(D3D12_CUBIN_DATA_HANDLE *handle, UINT32 block_x, UINT32 block_y, UINT32 block_z, UINT32 smem_size, const void *params, UINT32 param_size, const void *raw_params, UINT32 raw_params_count);
+}
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -4642,7 +4642,7 @@ static void d3d12_command_list_track_query_heap(struct d3d12_command_list *list,
     }
 }
 
-extern ULONG STDMETHODCALLTYPE d3d12_command_list_vkd3d_ext_AddRef(ID3D12GraphicsCommandListExt *iface);
+extern ULONG STDMETHODCALLTYPE d3d12_command_list_vkd3d_ext_AddRef(d3d12_command_list_vkd3d_ext_iface *iface);
 
 HRESULT STDMETHODCALLTYPE d3d12_command_list_QueryInterface(d3d12_command_list_iface *iface,
         REFIID iid, void **object)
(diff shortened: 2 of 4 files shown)


include/vkd3d_device_vkd3d_ext.idl
```diff
@@ -35,6 +35,17 @@ interface ID3D12DeviceExt : IUnknown
     HRESULT CaptureUAVInfo(D3D12_UAV_INFO *uav_info);
 }
 
+[
+    uuid(13228264-a243-42dd-b9e3-12dc8856d5f9),
+    object,
+    local,
+    pointer_default(unique)
+]
+interface ID3D12DeviceExt2 : ID3D12DeviceExt
```
Saancreed, line 44: Nitpicking, but I believe Microsoft's naming convention for D3D interfaces goes `ID3D12Thing` → `ID3D12Thing1` → `ID3D12Thing2` → … so this should be `ID3D12DeviceExt1` instead if we are to stay true to it.

liam-middlebrook: Thanks for the catch. Addressed.

Joshua-Ashton:
Is it worth us just exposing pParams right here even though NVAPI does not need it yet?

I would like to avoid another iteration on this interface.

include/vkd3d_vk_includes.h
```diff
@@ -51,6 +51,7 @@ typedef struct D3D12_CUBIN_DATA_HANDLE
     UINT32 blockX;
     UINT32 blockY;
     UINT32 blockZ;
+    UINT32 smemSize;
```
HansKristian-Work, line 54: This is an ABI break and needs to be a new struct ala `D3D12_CUBIN_DATA_HANDLE1`

libs/vkd3d/device_vkd3d_ext.c
```diff
@@ -129,6 +130,34 @@ static HRESULT STDMETHODCALLTYPE d3d12_device_vkd3d_ext_CreateCubinComputeShader
...
+                                                                  block_z,
+                                                                  0, /* smem_size */
+                                                                  shader_name,
+                                                                  out_handle);
+}
+
+static HRESULT STDMETHODCALLTYPE d3d12_device_vkd3d_ext_CreateCubinComputeShaderEx(ID3D12DeviceExt1 *iface, const void *cubin_data,
+       UINT32 cubin_size, UINT32 block_x, UINT32 block_y, UINT32 block_z, UINT32 smem_size, const char *shader_name, D3D12_CUBIN_DATA_HANDLE **out_handle)
+{
+    return d3d12_device_vkd3d_ext_CreateCubinComputeShader_common(iface,
```
HansKristian-Work, line 150: Lots of pointer type warnings here with the ID3D12DeviceExt interface.

liam-middlebrook: Whoops, have this resolved locally with a pre-commit. Looks like the original changes never really made use of the `d3d12_device_vkd3d_ext_iface` typedef that they added.

include/vkd3d_vk_includes.h
liam-middlebrook: This got me thinking, an alternative way of doing this would be to add a new function like `LaunchCubinShaderWithSmem` and do `ID3D12GraphicsCommandListExt1` instead of extending the device further.

This would better fit with `VK_NVX_binary_import`, as well as the `NvAPI_D3D12_LaunchCuKernelChain` and `NvAPI_D3D12_LaunchCuKernelChainEx` functions. It seems like that's probably a better goal to hit here (especially given @Joshua-Ashton's comment on wanting to limit future interface iterations). Just extending the command-list to take in smem + pParams should help us avoid future changes around the launch interface.

Does that seem like a reasonable move to you? I think with this ABI breakage we would need to bump the command-list interface anyways, so I prefer this new alternative to what I had previously planned.

HansKristian-Work: Extending command list interface is fine too.

liam-middlebrook: I've implemented that. The new `LaunchCubinShaderEx` function introduced should also cover @Joshua-Ashton's initial request to plumb `pParams` to avoid further iterations on this interface.

HansKristian-Work (approved):
LGTM. Trivial extension.

PR #1643, opened 2023-08-05
Implement DirectStorage metacommand using VK_NV_memory_decompression

doitsujin:
Fixes GPU hangs in Ratchet & Clank that are caused by the Nvidia driver miscompiling the fallback compute shader for GPU decompression. Compared to a working GLSL version of that fallback shader, using `VK_NV_memory_decompression` also yields better throughput (by around 10-15% on my mobile RTX 2060).

### Data layouts
Compressed data is organized into multiple tile streams (which will roughly correspond to one larger GPU resource or subresource), with the so-called control buffer storing stream metadata as follows:
```c
struct stream_t {
  uint32_t src_offset; /* relative to the start of the input buffer */
  uint32_t dst_offset; /* relative to the start of the output buffer */
};

struct control_buffer_t {
  uint32_t stream_count;
  stream_t streams[stream_count];
};
```

Each stream then starts with a metadata header of the following format:

```c
struct tile_stream_t {
  uint16_t unknown; /* always 0x01fe? */
  uint16_t tile_count;
  uint32_t frog; /* stores the uncompressed size of the last tile in bit 2 onwards */
  uint32_t last_tile_compressed_size;
  uint32_t tile_offsets[tile_count - 1]; /* relative to the end of the metadata section, for tile 1 onward */
};
```

This is in contrast to `VK_NV_memory_decompression`, which expects a flat array of the following structure *for each tile*:
```c
struct region_t {
  uint64_t src_va;
  uint64_t dst_va;
  uint64_t compressed_size;
  uint64_t decompressed_size;
  uint64_t method;
};
```

In order to resolve this mismatch, we dispatch compute shaders to parse each tile stream header and compute the appropriate address ranges. Tiles other than the last one in each stream always have a decompressed size of 64k.

The elephant in the room is that we have no way of knowing the number of tiles to decompress in advance, only the stream count and buffer addresses. In order to resolve this, we basically have two options:
1. Request a fixed-size scratch buffer large enough to hold memory region data for the maximum number of tiles in one stream (65535), and run one iteration of the decompression command per stream.
2. Request a fixed-size scratch buffer and write the entire memory region array to it, and just hope it won't overflow.

Option 1, while technically more correct, is very slow in practice and only allows for a throughput of ~1.2 GB/s in the bulk load demo on my mobile RTX 2060, compared to ~5.0 GB/s when using Option 2. Therefore, this implementation uses the latter approach, allowing for up to 65536 tiles per call, which corresponds to a maximum output data size of 4 GiB per call while the scratch buffer is roughly 2.5 MB in size.

The assumption that applications will not attempt to decode more than 4GiB of data at once seems reasonable for a number of reasons:
- All output data will be written to the same ID3D12Resource since there is no real BDA support in D3D12, and stream descriptions use offsets into that buffer.
- When streaming in textures, output data must further be copied to actual image resources, which means that a very large output buffer is essentially wasted memory.
- DirectStorage defaults to a staging buffer size of 32 MiB.

### Undocumented APIs
While D3D12 exposes a limited amount of information about meta commands, they are not publicly documented, and this implementation is mostly based on observed behaviour from invoking the command on AMD's Windows driver, as well as testing interactions with DirectStorage itself by running the [BulkLoadDemo](https://github.com/microsoft/DirectStorage/blob/main/Samples/BulkLoadDemo/README.md) on top of vkd3d-proton.

As such, the `D3D12_FEATURE_DATA_QUERY_META_COMMAND` part remains a mystery. The API consists of two opaque blobs, and I was only able to work out the meanig of some of the fields based on DirectStorage behaviour. This is a problem for the decompression test: Nvidia's Windows driver will set one of the unknown fields to 1, and attempting to run the test will cause a crash. It is not clear how applications are expected to respond to that particular bit.

I did not include a test for querying `D3D12_FEATURE_DATA_QUERY_META_COMMAND` since behaviour on Windows drivers varies greatly; there is no consistency at all except for a small set of known-good inputs. When passing invalid inputs, AMD will often return errors and Nvidia will often crash, but Nvidia also accepts some inputs that cause errors on AMD. It does not seem overly useful to test behaviour there, especially as long as we do not understand most of the data structure.

### Fallback shader
I also considered adding a custom fallback shader for systems that do not support `VK_NV_memory_decompression`, and testing showed a roughly 15% increase in throughput on the Steam Deck compared to the built-in shader from DirectStorage, while performance on my desktop 6900XT is unaffected. Unfortunately, we cannot ship this for the time being since the GDeflate reference implementation it is based on is licensed under Apache 2.0, so I removed it again.


libs/vkd3d/meta.c
```diff
@@ -1413,6 +1413,46 @@ void vkd3d_execute_indirect_ops_cleanup(struct vkd3d_execute_indirect_ops *meta_
     pthread_mutex_destroy(&meta_indirect_ops->mutex);
 }
 
+static HRESULT vkd3d_dstorage_ops_init(struct vkd3d_dstorage_ops *dstorage_ops, struct d3d12_device *device)
+{
+    VkPushConstantRange push_range;
```
HansKristian-Work, line 1418: The shaders require Int64. We should only create these meta shaders if memoryDecompression is supported.

libs/vkd3d/meta_commands.c
```diff
@@ -20,6 +20,121 @@
...
+    REFGUID command_id;
+    LPCWSTR name;
+    D3D12_GRAPHICS_STATES init_dirty_states;
+    D3D12_GRAPHICS_STATES exec_dirty_states;
+    unsigned int parameter_count;
+    const struct d3d12_meta_command_parameter_info *parameters;
+    d3d12_meta_command_create_proc create_proc;
+};
+
+static const struct d3d12_meta_command_parameter_info d3d12_meta_command_dstorage_parameter_infos[] =
```
HansKristian-Work, line 68: How was this determined? By enumerating parameters from a native implementation?

libs/vkd3d/meta_commands.c
```diff
@@ -20,6 +20,121 @@
...
+                offsetof(struct d3d12_meta_command_dstorage_exec_args, status_buffer_va) } },
+    { D3D12_META_COMMAND_PARAMETER_STAGE_EXECUTION,
+        { u"StatusBufferSize", D3D12_META_COMMAND_PARAMETER_TYPE_UINT64,
+                D3D12_META_COMMAND_PARAMETER_FLAG_INPUT, 0,
+                offsetof(struct d3d12_meta_command_dstorage_exec_args, status_buffer_size) } },
+};
+
+const struct d3d12_meta_command_info d3d12_meta_command_infos[] =
+{
+  { &IID_META_COMMAND_DSTORAGE, u"DirectStorage", 0,
```
HansKristian-Work, line 135: Odd indent

libs/vkd3d/meta_commands.c
```diff
@@ -20,6 +20,121 @@
 
 #include "vkd3d_private.h"
 
+const GUID IID_META_COMMAND_DSTORAGE = {0x1bddd090,0xc47e,0x459c,{0x8f,0x81,0x42,0xc9,0xf9,0x7a,0x53,0x08}};
```
HansKristian-Work, line 23: How was this determined?

libs/vkd3d/vkd3d_private.h
```diff
@@ -4522,6 +4522,24 @@ static inline struct d3d12_state_object *impl_from_ID3D12StateObject(ID3D12State
...
+#define VKD3D_DSTORAGE_MAX_TILE_COUNT (0x10000u)
+
+struct d3d12_meta_command_dstorage_scratch_header
+{
+    /* Workgroup count for fallback shader */
+    VkDispatchIndirectCommand region_count;
+    /* Padding to ensure 8-byte alignment of region array */
+    uint32_t padding;
+    /* Region array to be interpreted by either vkCmdDecompressMemoryNV
+     * or the fallback shader */
```
HansKristian-Work, line 4558: There is no fallback shader is there?

libs/vkd3d/meta_commands.c
```diff
@@ -22,6 +22,9 @@
 
 const GUID IID_META_COMMAND_DSTORAGE = {0x1bddd090,0xc47e,0x459c,{0x8f,0x81,0x42,0xc9,0xf9,0x7a,0x53,0x08}};
 
+static HRESULT d3d12_meta_command_create_dstorage(struct d3d12_meta_command* meta_command,
```
HansKristian-Work, line 25: `*` placement of meta_Command and parameter_data.

libs/vkd3d/meta_commands.c
```diff
@@ -132,9 +135,102 @@ const struct d3d12_meta_command_info d3d12_meta_command_infos[] =
...
+            return &d3d12_meta_command_infos[i];
+    }
+
+    return NULL;
+}
+
+static bool vkd3d_check_meta_command_support(struct d3d12_device *device, REFGUID command_id)
+{
+    if (!memcmp(command_id, &IID_META_COMMAND_DSTORAGE, sizeof(*command_id)))
+        return device->device_info.memory_decompression_features.memoryDecompression;
```
HansKristian-Work, line 157: For completeness, should check for SM 6.0 (subgroup) and Int64.

libs/vkd3d/meta_commands.c
```diff
@@ -267,6 +363,183 @@ CONST_VTBL struct ID3D12MetaCommandVtbl d3d12_meta_command_vtbl =
...
+    struct d3d12_meta_command *object;
+    HRESULT hr;
+
+    if (!vkd3d_check_meta_command_support(device, guid))
+    {
+        FIXME("Unsupported meta command %s.\n", debugstr_guid(guid));
+        return E_INVALIDARG;
+    }
+
+    if (!(object = vkd3d_malloc(sizeof(*object))))
```
HansKristian-Work, line 514: calloc?

libs/vkd3d/device.c
```diff
@@ -4174,6 +4174,48 @@ static HRESULT STDMETHODCALLTYPE d3d12_device_CheckFeatureSupport(d3d12_device_i
...
+
+            if ((data->QueryInputDataSizeInBytes && !data->pQueryInputData) ||
+                    (data->QueryOutputDataSizeInBytes && !data->pQueryOutputData))
+                return E_INVALIDARG;
+
+            if (!memcmp(&data->CommandId, &IID_META_COMMAND_DSTORAGE, sizeof(data->CommandId)))
+            {
+                /* It is not clear what any of these parameters do, but as of right now
+                 * it appears that only a single combination of parameters is supported. */
+                const struct d3d12_meta_command_dstorage_query_in_args* in_args = data->pQueryInputData;
```
HansKristian-Work, line 4189: `*`

libs/vkd3d/device.c
```diff
@@ -4174,6 +4174,48 @@ static HRESULT STDMETHODCALLTYPE d3d12_device_CheckFeatureSupport(d3d12_device_i
...
+                if (in_args->unknown2 == 1)
+                {
+                    out_args->unknown0 = 1;
+                    /* Limit stream count to something reasonable. Given that we have a hard limit
+                     * on the number of tiles we can process in one call, we should make it unlikely
+                     * for applications to hit that upper limit even if they try. */
+                    out_args->max_stream_count = min(256u, in_args->stream_count);
+                    /* Reserve data for the indirect tile count, then the memory region for
+                     * each tile, and finally a compute workgroup count for each stream. */
+                    out_args->scratch_size = sizeof(struct d3d12_meta_command_dstorage_scratch_header) +
```
HansKristian-Work, line 4212: Is there any required alignment for scratch area? Seems like we require 8-byte alignment in the impl.

tests/d3d12_dstorage.c
```diff
@@ -0,0 +1,354 @@
...
+
+    for (i = 0; i < count && !supports_meta_command; i++)
+        supports_meta_command = !memcmp(&meta_command_descs[i].Id, &IID_META_COMMAND_DSTORAGE, sizeof(meta_command_descs[i].Id));
+
+    free(meta_command_descs);
+
+    if (!supports_meta_command)
+    {
+        skip("DirectStorage meta command not supported.\n");
+        return;
```
HansKristian-Work, line 138: Missing release.

libs/vkd3d/meta_commands.c
doitsujin: Yes, these match the output of EnumerateMetaCommandParameters.

tests/d3d12_dstorage.c
```diff
@@ -0,0 +1,354 @@
...
+    memset(query_in, 0, sizeof(query_in));
+    query_in[0] = 0x100001u; /* up to 16 streams */
+    query_in[2] = 1u;
+
+    hr = ID3D12Device5_CheckFeatureSupport(device5, D3D12_FEATURE_QUERY_META_COMMAND, &query, sizeof(query));
+
+    if (FAILED(hr))
+    {
+        skip("DirectStorage meta command not supported.\n");
+        refcount = ID3D12Device5_Release(device5);
```
HansKristian-Work, line 172: Need to release plain device as well.

libs/vkd3d/meta_commands.c
HansKristian-Work: Ok.

HansKristian-Work (changes requested):
Overall, it seems fine. Mostly a bunch of nits.

libs/vkd3d/device.c
doitsujin: I don't know if there are any alignment guarantees, we're getting a fresh resource from DStorage though.

HansKristian-Work: Ok.

vkushwaha-nv:
>I also considered adding a custom fallback shader for systems that do not support VK_NV_memory_decompression
Have you looked at using https://github.com/KhronosGroup/Vulkan-ExtensionLayer/tree/main/layers/decompression

HansKristian-Work:
> > I also considered adding a custom fallback shader for systems that do not support VK_NV_memory_decompression
> > Have you looked at using https://github.com/KhronosGroup/Vulkan-ExtensionLayer/tree/main/layers/decompression

Not legal for us to use, it's Apache 2.0, which is incompatible with LGPL.

adamnv:
Thanks for doing this!  I was also poking around at how to implement this metacommand, but it was looking like a PITA and you just did it. :grinning:   Ratchet & Clank thank you!  And presumably a bunch of future titles!

PR #1647, opened 2023-08-08
Workaround games which forget to set SetRootDescriptorTable

HansKristian-Work:
Gets rid of the "active" mask. It's vestigial and useless and only serves to push garbage data from stack to GPU.

Should fix spurious GPU hangs in SF6, which hits this scenario.


doitsujin (approved):
Looks simple enough.

Issue #1636, opened 2023-07-23
Resident Evil 4 Remake: Reflection glitches in some areas


mbriar:
In some areas in RE4 Remake, specifically at the end of chapter 10, in the area where you use the liquid nitrogen to fight the monster, the light reflected from some glass surfaces is glitching out. This does not happen with native d3d12 on windows.
Turning off Anti-Aliasing makes the glitch much more severe, but it still happens with AA set to TAA+FXAA.

Here is a savegame to reach the relevant spot ([path to put it](https://www.pcgamingwiki.com/wiki/Resident_Evil_4_(2023)#Save_game_data_location)), load the save in chapter 10, second slot after the autosave: https://mega.nz/file/WyJXwIpI#K4cY3R6IjM5OpsKbia_ZV7uo-CiU-rSPX-3swUl9Gww

Here is a video that shows how to reach the spot where the glitch happens from the save location and what it looks like: https://mega.nz/file/O2o2AIqQ#Qi7DPNqAfSZlAfMoEIyD2uuDjPV3sb9Fo6wIr5liSec

Same spot on windows running on the same hardware and settings that doesn't show the glitch: https://mega.nz/file/CvIilajD#ddxhbg-f-XxvNpnxBKZQZGGtu61SMnjVMUVnvOIAz5Y

Vulkan renderdoc capture of the glitch: https://mega.nz/file/eyxjjTCB#R-9hW12bw4I_i1Mxd_ZmaxY8-BSd0BzKakyT9gmkICI

### Software information
Resident Evil 4 Remake, most settings max, Anti-Aliasing off

### System information
- GPU: 6700XT
- Driver: Mesa main and AMDVLK
- Wine version: Proton Experimental 
- VKD3D-Proton version: db608963e43416a

### Log files
[re4_vkd3d_log.zip](https://github.com/HansKristian-Work/vkd3d-proton/files/12139657/re4_vkd3d_log.zip)

runar-work:
Great that you managed to capture this. I'm unable to replay the capture with my RX 7600 since I'm missing an extension (VK_KHR_fragment_shader_barycentric), and save game transfer doesn't work either. However, I happened to find a location in the very beginning of chapter 2 where a light fixture causes the same issue when the light is reflected in surrounding puddles (also on the other side of walls). Looks like some compute shader shenanigans.

mbriar:
[...]

PR #1648, opened 2023-08-09
FP16 cast shenanigans.

HansKristian-Work:
Add more detailed tests around FP32 <-> FP16 conversion behavior. Uncovered some interesting nits:

- Truncating FP32 to FP16 can be RTZ always (AMD) or RTE always (NV). Truncating 100k returns +inf on NV for example.
- Casting FP32 -> FP16 -> FP32 must perform the cast. That does not happen on Vulkan drivers. Work around this by forcing a QuantizeToFP16 if this pattern is observed. Only test suite and Metro Exodus EE hits this in some trivial places.

Attempts to workaround a shader flicker glitch in RE4 where we found a case where a value > 100k is truncated to FP16 into +inf, which then causes issues. Using FP32 for this shader seemed to work fine locally. Just attempt disabling FP16 globally for this game, since a heuristic didn't work out.


mbriar:
Hey, I realize this is still a draft, so apologies for jumping the gun here, but this does not fix #1636 for me on a 6700XT with RADV. (don't know if this is even intended to fix that issue, but might as well mention it since I tried now..)

runar-work:
It appears there's more than one cause of this glitch. This PR fixes the blinking issue with the ceiling light at the beginning of chapter 2 on a 5700 XT. Now that I tested a bit more, it does not fix the blinking in the eyes of the NPC in the first cutscene in chapter 1. I've been unable to capture that case, however.

I've been working my way through the chapters while testing for GPU hangs, so it shouldn't be too long before I reach the spot @mbriar described in the original issue. There are a few other instances of blinking glitches I've encountered, so I'll have to see what else I can capture after this fix.

doitsujin (approved):

PR #1649, opened 2023-08-10
dxil-spirv: Update submodule.

HansKristian-Work:
Another pass over strict waveops.

Optimizes away useless WaveReadLaneFirst instructions that read provably wave uniform values.
Use LSB/Shuffle method instead of branch + phi(undef, broadcast) style. Just as fast on AMD and more robust. Should workaround some issues on NV, especially in CP77 and clouds as observed by @runar-work.

Some games affected by patches:

- F1 2022
- Halo Infinite
- CP77
- Misc RE engine (Monster Hunter Rise, RE2, RE8, RE4, SF6)


runar-work:
For the record, NV driver I used was 535.98.
	
I also tested the games for 15-20 minutes with RX 7600 and Mesa 23.1.5.

No issues in F1 22, Cyberpunk, MHR, RE2, RE8, RE4 or SF6, except for the usual game freeze on exit with RE2/RE4/SF6.

Halo has an issue where one CPU core gets stuck at 100% and the game stops rendering, but that happened with an older build as well. Also forgot I had this problem on NV once in the beginning, but wasn't able to reproduce it in the time I tested.

PR #1651, opened 2023-08-11
Batch raytracing acceleration structure builds

doitsujin:
Significantly improves Raytracing performance on RADV in a number of games, especially Hitman 3.

Handling WriteBufferImmediate is a bit janky as always, since we want to maintain correct command ordering with `BOTTOM` markers.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -4876,6 +4876,7 @@ static HRESULT STDMETHODCALLTYPE d3d12_command_list_Close(d3d12_command_list_ifa
 
     d3d12_command_list_end_current_render_pass(list, false);
     d3d12_command_list_end_transfer_batch(list);
+    d3d12_command_list_flush_rtas_batch(list);
 
     if (list->predicate_enabled)
         VK_CALL(vkCmdEndConditionalRenderingEXT(list->vk_command_buffer));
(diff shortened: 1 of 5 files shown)


libs/vkd3d/command.c
```diff
@@ -8769,6 +8777,10 @@ static void STDMETHODCALLTYPE d3d12_command_list_ResourceBarrier(d3d12_command_l
                  * If we don't know the resource, we must assume a global UAV transition
                  * which also includes RTAS. */
                 state_mask = 0;
+
+                /* Always flush pending RTAS builds since we might be synchronizing a scratch buffer */
+                d3d12_command_list_flush_rtas_batch(list);
```
HansKristian-Work, line 8782: Need to consider aliasing barrier as well I think.

libs/vkd3d/command.c
```diff
@@ -12504,10 +12518,19 @@ static void STDMETHODCALLTYPE d3d12_command_list_BuildRaytracingAccelerationStru
         return;
     }
 
+    /* Do not batch TLAS and BLAS builds into the same command, since doing so
+     * is disallowed if there are data dependencies between the builds. This
+     * happens in Cyberpunk 2077, which does not emit appropriate UAV barriers. */
+    if (rtas_batch->build_info_count && rtas_batch->build_type != desc->Inputs.Type)
+        d3d12_command_list_flush_rtas_batch(list);
```
HansKristian-Work, line 12525: If there are no appropriate barriers, should be add them here? Seems like app is doing:
BOTTOM -> BOTTOM -> BOTTOM -> no barrier -> TOP -> TOP -> TOP?

HansKristian-Work (changes requested):
Seems okay. Mostly some nits.

libs/vkd3d/command.c
doitsujin: Can do, I omitted them because things just happen to work™ but then again this is only expected to happen once/frame anyway, so it won't hurt performance.

HansKristian-Work: Might randomly fix issues on NV perhaps.

HansKristian-Work (approved):
LGTM!

PR #1659, opened 2023-08-17
vkd3d-shader: Mask DXBC code that is statically unreachable.

### LINKED ISSUE #1657: Xenia: Shader validation errors
When running the Xenia emulator (I specifically tried RDR1) on vkd3d-Proton, validation errors complain about a handful of shaders. It uses DXBC.

```
./31c2c452708da23c.spv
error: line 4219: InBoundsAccessChain must appear in a block
  %3510 = OpInBoundsAccessChain %_ptr_Function_float %r13 %uint_1

./76d532a2d81dc53e.spv
error: line 4224: InBoundsAccessChain must appear in a block
  %3514 = OpInBoundsAccessChain %_ptr_Function_float %r14 %uint_1
  
./874ee0cdb053f772.spv
error: line 4214: InBoundsAccessChain must appear in a block
  %3509 = OpInBoundsAccessChain %_ptr_Function_float %r14 %uint_1

./c859d7317d49e6b2.spv
error: line 9671: InBoundsAccessChain must appear in a block
  %7972 = OpInBoundsAccessChain %_ptr_Function_float %r4 %uint_1
```

I've dumped the shaders and also disassembled both the DXBC and SPIR-V of the first one.
All of them exibit the same validation error, so hopefully it's the same issue for all of them.
[xenia shaders.zip](https://github.com/HansKristian-Work/vkd3d-proton/files/12362450/xenia.shaders.zip)

HansKristian-Work:
Xenia emits questionable DXBC code that is dead and does not exist at the appropriate scopes. Continue -> Break inside a switch for example ...

Fix #1657.

### DIFF
--- a/libs/vkd3d-shader/spirv.c
+++ b/libs/vkd3d-shader/spirv.c
@@ -11141,9 +11141,36 @@ static bool is_dcl_instruction(enum VKD3D_SHADER_INSTRUCTION_HANDLER handler_idx
             || handler_idx == VKD3DSIH_HS_DECLS;
 }
 
+static bool is_scope_generating_control_flow_instruction(enum VKD3D_SHADER_INSTRUCTION_HANDLER handler)
+{
+    /* Ignore continue, break, ret and any unconditional branches which don't logically generate new labels on their own.
+     * In some Xenia shaders, there is questionable back-to-back continue + continue, followed by break.
+     * If we've terminated a block, just ignore it.
+     * Technically, if there are if/endif pairs after a continue or break, it might be more
+     * correct to track shadow control flow scopes and ignore code until we pop out of the shadow stack,
+     * but this is silly and should only be considered if actually needed. */
+    switch (handler)
+    {
+        case VKD3DSIH_CASE:
+        case VKD3DSIH_DEFAULT:
+        case VKD3DSIH_ELSE:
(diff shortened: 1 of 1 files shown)


libs/vkd3d-shader/spirv.c
```diff
@@ -11141,9 +11141,36 @@ static bool is_dcl_instruction(enum VKD3D_SHADER_INSTRUCTION_HANDLER handler_idx
...
+    {
+        case VKD3DSIH_CASE:
+        case VKD3DSIH_DEFAULT:
+        case VKD3DSIH_ELSE:
+        case VKD3DSIH_ENDIF:
+        case VKD3DSIH_ENDLOOP:
+        case VKD3DSIH_ENDSWITCH:
+        case VKD3DSIH_IF:
+        case VKD3DSIH_LOOP:
+        case VKD3DSIH_SWITCH:
```
doitsujin, line 11162: I think this needs all the HS phases as well?

Technically `LABEL` too but we don't support that at the moment.

HansKristian-Work: Don't think HS phases are needed. If we are in this situation we are considered masked until the switch/loop situation is popped. HS phases don't generate their own logical scopes per-se I think.

doitsujin (approved):

PR #1663, opened 2023-08-21
vkd3d: Check for query pool when deduplicating query ranges.

doitsujin:


libs/vkd3d/command.c
```diff
@@ -5043,10 +5043,10 @@ static void d3d12_command_list_read_query_range(struct d3d12_command_list *list,
      * in at most one range. */
     while (lo < hi)
     {
-        if (pos < list->query_ranges_count)
-        {
-            range = &list->query_ranges[pos];
+        range = &list->query_ranges[pos];
```
HansKristian-Work, line 5046: Dereferencing out of bounds like this seems questionable to me. I've had issues in the past where compiler assumed things based on that.

doitsujin: Shouldn't this be equivalent to `list->query_ranges + pos`? Guess we can write it like that explicitly but the actual dereference won't happen until after the range check anyways.

HansKristian-Work (approved):
LGTM

libs/vkd3d/command.c
HansKristian-Work: technically, it'd be a dereference, then addrof, and I think compiler is allowed to assume a dereferenced pointer to point to valid memory, so reorder could happen, but not 100% sure here, but the proposed fix should be fine.

PR #1674, opened 2023-08-30
Use VK_KHR_maintenance5 features if available

doitsujin:
Mostly A8_UNORM, but also index buffer robustness. Tested on RADV git as well as 23.1.6.


HansKristian-Work (approved):
LGTM, thanks!

PR #1676, opened 2023-08-31
Implement Enhanced Barriers - Phase 2

HansKristian-Work:
This primarily focuses on global / buffer barriers.

Mainly, we're concerned with translating stages and access masks here.


doitsujin (approved):
Bit weird to have stuff added and then immediately refactored / moved around, but code itself looks good.

PR #1694, opened 2023-09-07
Add special game perf workaround for Starfield and other DGC junkies

HansKristian-Work:
**NOTE: The meaning of this PR, potential perf impact and the problem it tries to solve is being grossly misrepresented on other sites.**

The goal of this refactor is to optimize for cases where games (Starfield in particular) uses advanced ExecuteIndirect in very inefficient ways.

- Indirect count, but indirect count ends up being 0.
- Non-indirect count, but none of the active draws inside the multi-draw indirect actually result in a draw. Multiple back-to-back ExecuteIndirects are called like this, causing big bubbles on the GPU.

In RADV, a special optimization was added which uses indirectCount in DGC as a predicate when possible. This lets us skip over the prepare cmdbuffer as well as the INDIRECT_BUFFER execution which is about 10x faster than spawning a small CS, adding extra sync and then executing a NOP-ed out indirect buffer.

We can take advantage of this by doing our own prologue which scans through the ExecuteIndirect buffer in the scenario where non-indirect count is used. Using indirect count is not slower than direct count at all.

To make this efficient, this PR refactors the command buffer emit system so that instead of having one init command buffer and one "real" command buffer, we have N sequences of this pattern. This allows us to split a command stream when observing an INDIRECT_ARGUMENT barrier, and we can batch up any patch CS easily. For example, given a D3D12 command stream like:

- CS to generate indirect arguments
- ResourceBarrier(UAV -> INDIRECT_ARGUMENT)
- ExecuteIndirect (arg0, command_count = 16)
- ExecuteIndirect (arg1, command_count = 16)
- ExecuteIndirect (arg2, command_count = 16)
- ExecuteIndirect (arg3, command_count = 16)

we can now transform this into:

iteration[0].vk_cmd_buffer:
- CS to generate
- UAV -> INDIRECT_ARGUMENT
- Begin new sequence

iteration[1].vk_init_buffer:
- Scan arg0 (emit count 0 if all empty draws)
- Scan arg1 (emit count 0 if all empty draws)
- Scan arg2 (emit count 0 if all empty draws)
- Scan arg3 (emit count 0 if all empty draws)
- CS -> INDIRECT_ARGUMENT

iteration[1].vk_cmd_buffer:
- ExecuteIndirect(arg0, count0) (over 10x faster throughput if we can predicate out work)
- ExecuteIndirect(arg0, count0)
- ExecuteIndirect(arg0, count0)
- ExecuteIndirect(arg0, count0)

This kind of command reordering can be extended later to whatever use case we have in mind, but reordering indirect work is the main important case I think. 

As a heuristic, splitting command buffers isn't ideal, so we only consider a split if a device ever created a fancy execute indirect command signature and existing content outside Halo and Starfield should not observe any difference in behavior here.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -11985,6 +11985,8 @@ static void d3d12_command_list_execute_indirect_state_template_dgc(
 {
     const struct vkd3d_vk_device_procs *vk_procs = &list->device->vk_procs;
     const VkPhysicalDeviceDeviceGeneratedCommandsPropertiesNV *props;
+    VkConditionalRenderingBeginInfoEXT conditional_begin_info;
+    struct vkd3d_scratch_allocation predication_allocation;
     struct vkd3d_scratch_allocation preprocess_allocation;
     struct vkd3d_scratch_allocation stream_allocation;
     struct vkd3d_scratch_allocation count_allocation;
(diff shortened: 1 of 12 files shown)


HansKristian-Work:
Seeing some very random glitches in Starfield with this PR unfortunately, so needs more time in the oven ...

jams3223:
It's sad that developers can't take their time to properly code their games due to their greed for money, and people like you, Kristian, have to fix their mistakes. We're so lucky to have you, even though I didn't really like the game, but I appreciate your efforts. You're a gift from God to the community, and I look up to your work as a young programmer.

HansKristian-Work:
Found the issue, was my own bug after all ;_; forgot to update indirect_meta after a split so we ended up missing some barriers.

Silvrav:
This is excellent news, thank you HansKristian. When is this being planned to be released to experimental?

Ruzihm:
> Found the issue, was my own bug after all ;_; forgot to update indirect_meta after a split so we ended up missing some barriers.

Is this to say the issue mentioned in the body of this pr and attributed to the devs mistakes does not actually exist, or is this about the very random glitches mentioned in a prior comment? Trying to understand how this situation is developing.

Blisto91:
The PR introduced visual bugs which the cause for was then found and fixed.

Also try to keep the posting in the PR to a minimum.

HansKristian-Work:
To be clear, the gains expected here are very minute. Single percent range to pop some final bubbles that Mesa didn't clean up on its own. The real gains come from recent Mesa patches on main.

HansKristian-Work:
Hid some comments. Github PRs is not the place for drive-by comments.

PrincessAkira:
> The PR introduced visual bugs for which the cause was then found and fixed.
> 
> Also try to keep the posting in the PR to a minimum.

I can still notice some flickers of textures.
I can rule out this is happening due to Mesa.
On request, i can provide a video.

Using this Branch with mesa from AUR and normal proton.
RADV is in use aswell.

Blisto91:
> Using this Branch with mesa from AUR and normal proton.

Then it is unrelated to the PR. You should open a new issue if you suspect it is a vkd3d-proton issue.

libs/vkd3d/command.c
```diff
@@ -1998,6 +2000,16 @@ static void d3d12_command_list_begin_new_sequence(struct d3d12_command_list *lis
     d3d12_command_list_invalidate_all_state(list);
     /* Extra special consideration since we're starting a fresh command buffer. */
     list->descriptor_heap.buffers.heap_dirty = true;
+    d3d12_command_list_debug_mark_label(list, "Split", 0.0f, 0.0f, 0.0f, 1.0f);
+}
+
+static void d3d12_command_list_consider_new_sequence(struct d3d12_command_list *list)
+{
+    if (list->cmd.vk_command_buffer == list->cmd.vk_init_commands_post_indirect_barrier &&
+            vkd3d_atomic_uint32_load_explicit(&list->device->device_has_dgc_templates, vkd3d_memory_order_relaxed))
```
doitsujin, line 2024: For correctness, we also need to make sure that there are no scoped queries active (think it's only `PIPELINE_STATISTICS` at this point, which isn't super common anyway). We currently don't have code for that, but just adding a counter to the command list for active query tracking should do.

Blisto91:
@PrincessAkira I checked out the scope text flicker, but haven't found this PR to affect it. I get the same both with regular Experimental and with vkd3d-proton 2.9. Also checked native game on Windows and it's the same there for me.

PrincessAkira:
> @PrincessAkira I checked out the scope text flicker, but haven't found this PR to affect it. I get the same both with regular Experimental and with vkd3d-proton 2.9. Also checked native game on Windows and it's the same there for me.

Weird.
Ive only observed this with this.
I can double-check again if u want though

libs/vkd3d/command.c
HansKristian-Work: That's what https://github.com/HansKristian-Work/vkd3d-proton/pull/1694/commits/ded0fc4df308d2fbc67e46ac67c2d87160c18d2c does.

doitsujin: missed that somehow, derp.

doitsujin (approved):

PR #1707, opened 2023-09-19
dxil-spirv: Update dxil-spirv.

HansKristian-Work:
Fixes hang in shader compilation on AMD with Mortal Kombat 1.

This game seems to require AGS shaders to compile successfully despite AGS not being reported as supported :<


HansKristian-Work:
@runar-work @Blisto91 This also affects AMD in F1 2021, so worth giving that a smoketest.

PR #1713, opened 2023-09-26
Implement support for multiple conflicting global root signatures in DXR

HansKristian-Work:
Battlefield V requires this particular weird feature apparently :|

DXR allows multiple global root signatures per RTPSO,
with the restriction that given a bound root-sig you can only execute
shaders which are compatible.

The solution here is to create one unique VkPipeline per unique global
root signature. When tracing rays, we bind the pipeline that corresponds
to the bound root signature.

The refactor here is somewhat substantial, but the core gist of it
is that we first enumerate all unique root signatures in our
associations data structures. Then we attempt to compile one pipeline
per variant.

The implementation here is somewhat simplified, since we only expect
more than one variant in extremely rare circumstances.
Basic gist is that we test every entry point / hit group / collection
and only include things which are compatible for that given variant.

For collections there is only one variant that might be compatible.
For entry point exports, they must be associated with a global root signature,
so it will be compiled into one variant.
Hit groups are included if all their references are part of the variant.

Exports and entry points now receive two new indices:
- Pipeline variant index
- Group / module index within that variant

This is necessary since we can no longer assume a 1:1 relation between
export index / group index and entry point / stage index.


doitsujin (approved):
bit of a chungus pr, but lgtm.

PR #1721, opened 2023-10-04
Implement various Vulkan compat features

doitsujin:
Does not include line rasterization stuff yet since that is more tricky, and currently unsupported anyway.


tests/d3d12_pso.c
```diff
@@ -2703,3 +2703,442 @@ void test_topology_triangle_fan(void)
...
+
+    push_args.depth = 0.5f;
+    push_args.slope = 0.0f;
+    ID3D12GraphicsCommandList9_SetGraphicsRoot32BitConstants(command_list9, 0, sizeof(push_args) / sizeof(uint32_t), &push_args, 0);
+
+    ID3D12GraphicsCommandList9_BeginQuery(command_list9, query_heap, D3D12_QUERY_TYPE_BINARY_OCCLUSION, 7);
+    ID3D12GraphicsCommandList9_DrawInstanced(command_list9, 3, 1, 0, 0);
+    ID3D12GraphicsCommandList9_EndQuery(command_list9, query_heap, D3D12_QUERY_TYPE_BINARY_OCCLUSION, 7);
+
+#if 0
```
HansKristian-Work, line 3091: Should this test be noped out if both drivers are robust against it?

doitsujin: I only did it because it is technically wildly out of spec, but can re-enable.

libs/vkd3d/state.c
```diff
@@ -4562,6 +4566,13 @@ static HRESULT d3d12_pipeline_state_init_graphics_create_info(struct d3d12_pipel
         goto fail;
     }
 
+    /* Tests show that D3D12 drivers behave as if D3D12_PIPELINE_STATE_FLAG_DYNAMIC_DEPTH_BIAS
+     * was always set, howevert doing that would invalidate existing pipeline caches, so avoid
```
HansKristian-Work, line 4619: howevert typo

libs/vkd3d/state.c
```diff
@@ -4562,6 +4566,13 @@ static HRESULT d3d12_pipeline_state_init_graphics_create_info(struct d3d12_pipel
...
+     * this until proven necessary. */
+    if ((desc->flags & D3D12_PIPELINE_STATE_FLAG_DYNAMIC_DEPTH_BIAS) &&
+            (graphics->ds_desc.depthTestEnable))
```
HansKristian-Work, line 4622: Depth bias affects FragCoord.z I think, which exists without depth test I think?

subprojects/dxil-spirv
HansKristian-Work, line 1: Was this intentional?

libs/vkd3d/vkd3d_private.h
```diff
@@ -1964,7 +1965,9 @@ struct d3d12_graphics_pipeline_state
     VkPipelineRasterizationDepthClipStateCreateInfoEXT rs_depth_clip_info;
     VkPipelineRasterizationStateStreamCreateInfoEXT rs_stream_info;
 
-    uint32_t dynamic_state_flags; /* vkd3d_dynamic_state_flag */
+    /* vkd3d_dynamic_state_flag */
+    uint32_t explicit_dynamic_states;
```
HansKristian-Work, line 1969: Is explicit_dynamic_states really necessary? Seems like we could just look at pipeline_dynamic_states?

libs/vkd3d/cache.c
```diff
@@ -2623,8 +2623,7 @@ void vkd3d_pipeline_cache_compat_from_state_desc(struct vkd3d_pipeline_cache_com
         HF32(desc->rasterizer_state.DepthBiasClamp);
         HF32(desc->rasterizer_state.SlopeScaledDepthBias);
         H32(desc->rasterizer_state.DepthClipEnable);
-        H32(desc->rasterizer_state.MultisampleEnable);
```
HansKristian-Work, line 2626: Should these be removed?

HansKristian-Work: Or was MultisampleEnable purely line rendering related?

HansKristian-Work (changes requested):
Smol nits mostly, lgtm otherwise.

libs/vkd3d/cache.c
doitsujin: These are entirely removed from RASTERIZER_DESC2 and only affect line rasterization.

libs/vkd3d/state.c
doitsujin: As per discord discussion, this will need some additional testing, the way we currently deal with DSV with no depth test might be incorrect.

libs/vkd3d/vkd3d_private.h
doitsujin: Having a separate `explicit_dynamic_states` seems more robust to me, `pipeline_dynamic_states` is only written at the time the pipeline gets compiled and may or may not depend on things like GPL support.

libs/vkd3d/state.c
doitsujin: fixed.

HansKristian-Work (approved):
LGTM

PR #1741, opened 2023-10-15
Implement line rasterization features

doitsujin:
This appears to be a complete mess on Windows drivers, Alpha lines are broken on AMD (both the old and the new way to enable those), wide lines aren't really all that wide on NV.

This implementation religiously adheres to publicly available documentation instead, and behaviour is consistent between NV and RADV.


HansKristian-Work (approved):
LGTM

PR #1745, opened 2023-10-20
Enable DGC preprocess on RADV as well.

HansKristian-Work:


HansKristian-Work:
Tested against https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/26172

~1% in Starfield
~15% (!) in Halo Infinite

doitsujin (approved):

PR #1758, opened 2023-10-30
vkd3d: Add workaround for broken RTAS barrier in Witcher 3.

HansKristian-Work:


HansKristian-Work:
Does not seem to be the actual root cause.

PR #1774, opened 2023-11-08
vkd3d: Disable NV_dgcc on 545 drivers.

runar-work:
Workaround for glitches and hangs in Starfield introduced with DGCC-enabled 545-drivers.


doitsujin (approved):
pain

HansKristian-Work:
It never ends. Just waiting for updated commit since the email is wrong.

PR #1776, opened 2023-11-10
Implement DXR local -> external dependencies

HansKristian-Work:
Forgot about this cursed RTPSO flag. Warhammer: Darktide uses this.

Works now on RADV (but Xid 109 on NV, but probably not our bug).


doitsujin (approved):
Should be good to go if tested enough.

PR #1782, opened 2023-11-17
vkd3d: Add internal refcounts to command allocator.

HansKristian-Work:
Workaround game bugs in Windjammer 2 where game is releasing command allocator while there are pending submissions in flight.

This is obviously broken, but workaround it by repurposing the pending submission counter into private refcounts instead.


doitsujin (approved):
*sigh*

PR #1790, opened 2023-11-24
Implement stencil copies without stencil export

doitsujin:
Unblocks R8->stencil copies on NV. We'll need the same trickery for stencil resolves anyway, so might as well support this as well.


HansKristian-Work (approved):
image

PR #1806, opened 2023-12-08
Always enable storage image usage for non-RTV 2D images

doitsujin:
Needed for multisample resolve in some cases, but this is scary and needs testing on a wide range of hardware.

Builds on #1805.

### DIFF
--- a/libs/vkd3d/resource.c
+++ b/libs/vkd3d/resource.c
@@ -774,6 +774,16 @@ static HRESULT vkd3d_get_image_create_info(struct d3d12_device *device,
     if (!(desc->Flags & D3D12_RESOURCE_FLAG_DENY_SHADER_RESOURCE) || desc->SampleDesc.Count > 1)
         image_info->usage |= VK_IMAGE_USAGE_SAMPLED_BIT;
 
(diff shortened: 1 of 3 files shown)


HansKristian-Work:
Asked an NV dev and they suggested this won't change anything for them.

libs/vkd3d/resource.c
```diff
@@ -774,6 +774,16 @@ static HRESULT vkd3d_get_image_create_info(struct d3d12_device *device,
     if (!(desc->Flags & D3D12_RESOURCE_FLAG_DENY_SHADER_RESOURCE) || desc->SampleDesc.Count > 1)
         image_info->usage |= VK_IMAGE_USAGE_SAMPLED_BIT;
 
+    /* Add storage image usage if the image cannot be rendered to, but may
+     * be used as a destination image for multisample resolves. */
+    if (desc->Dimension == D3D12_RESOURCE_DIMENSION_TEXTURE2D && desc->SampleDesc.Count == 1 &&
```
HansKristian-Work, line 779: It should be safe to land this if we check for AMD or NVIDIA for now. I'm still concerned about e.g. Intel/Turnip.

PR #1807, opened 2023-12-13
Fix various problems with the resinfo instruction

doitsujin:
Addresses the following issues:
- `precise` flags can be set on this instruction, but this would lead to the uint path not working properly.
- if the queried mip level is out of bounds, the returned width, height and array size is specified to be 0. AMD Windows drivers do not seem to implement this, but Nvidia passes the relevant test.
- `_rcpfloat` was not supported. FXC never seems to emit this, but I've seen this in the wild with DXVK, probably from custom DXBC emitters.

The first issue was encountered in Avatar Frontiers of Pandora.

### DIFF
--- a/libs/vkd3d-shader/spirv.c
+++ b/libs/vkd3d-shader/spirv.c
@@ -10739,13 +10739,15 @@ static void vkd3d_dxbc_compiler_emit_bufinfo(struct vkd3d_dxbc_compiler *compile
 static void vkd3d_dxbc_compiler_emit_resinfo(struct vkd3d_dxbc_compiler *compiler,
         const struct vkd3d_shader_instruction *instruction)
 {
+    uint32_t type_id, lod_id, val_id, one_id, rcp_id, cond_id, rcp_type_id, miplevel_count_id;
     struct vkd3d_spirv_builder *builder = &compiler->spirv_builder;
     const struct vkd3d_shader_dst_param *dst = instruction->dst;
     const struct vkd3d_shader_src_param *src = instruction->src;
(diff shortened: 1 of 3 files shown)


libs/vkd3d-shader/spirv.c
```diff
@@ -10779,16 +10789,41 @@ static void vkd3d_dxbc_compiler_emit_resinfo(struct vkd3d_dxbc_compiler *compile
...
+                indices[i] = i;
+
+            rcp_id = vkd3d_spirv_build_op_vector_shuffle(builder, rcp_type_id,
+                    val_id, val_id, indices, size_component_count);
+
+            one_id = vkd3d_dxbc_compiler_get_constant_float_vector(compiler, 1.0f, size_component_count);
+            rcp_id = vkd3d_spirv_build_op_fdiv(builder, rcp_type_id, one_id, rcp_id);
+
+            for (i = 0; i < VKD3D_VEC4_SIZE; i++)
+                indices[i] = (i < size_component_count ? VKD3D_VEC4_SIZE : 0) + i;
```
HansKristian-Work, line 10822: Is val_id always 4 components?

doitsujin: Yes, the code above all of this sets it up that way. Reasoning being that the instruction is defined to return `vec4(w, h, d, mips)`, where `h` or `d` might also be the array size depending on the resource type, or may be zeroed if it's not meaningful (e.g. `d` for non-array Texture2D).

HansKristian-Work (approved):

PR #1815, opened 2023-12-18
build: Replace stdcall-alias with kill-at, remove __declspec(dllexport)

ishitatsuyuki:
This mirrors changes at DXVK [1], [2].

The intent in [1] was to prevent 32-bit export function names from containing @Ordinals at the end.

This also turns out to be useful for compatibility with LLVM-based toolchain, in particular when using LLD, because LLD does not recognize --add-stdcall-alias but does recognize --kill-at.

As in [1], we use the system-provided d3d12 dependency instead of the
implib we generated, because MinGW LD insists on emitting implib with
ordinal suffixes (LLD does the correct thing here).

Another part of the change is to remove __declspec(dllexport). The .def
files we use specify the linkage already, so the attributes are obsolete.
This fixes LLD warnings like:

    ld.lld: warning: duplicate /export option: D3D12CreateDevice

There is another LLD compatibility issue here: while MSVC and MinGW will
give priority to the .def definitions, older versions of LLD would end up
honoring the attribute instead, resulting in default ordinals to be
exported instead of the specified one [3]. Not specifying dllexport in the
source fixes this.

This probably needs more testing. I'm primarily doing this since I use the LLVM toolchain for development ([like this cross file](https://github.com/ishitatsuyuki/dxvk/blob/clang/build-clang-win64.txt)), and I don't have 32-bit D3D12 games at hand to test. I did test that the 64-bit binaries produced by Clang/LLD runs correctly, unlike before where ordinals were messed up.

cc @Joshua-Ashton 

[1]: https://github.com/doitsujin/dxvk/commit/a62117cd13618f98e5c0f0da27de9193cbc6dae2
[2]: https://github.com/doitsujin/dxvk/commit/904d3e6c9014bff7aa0843630b009c51879a1d06
[3]: https://github.com/llvm/llvm-project/issues/62329

### DIFF
--- a/libs/d3d12/main.c
+++ b/libs/d3d12/main.c
@@ -40,17 +40,11 @@
 #include "vkd3d_string.h"
 #include "vkd3d_platform.h"
 
-/* We need to specify the __declspec(dllexport) attribute
- * on MinGW because otherwise the stdcall aliases/fixups
- * don't get exported.
- */
-#if defined(_MSC_VER)
-  #define DLLEXPORT
-#elif defined(__MINGW32__)
-  #define DLLEXPORT __declspec(dllexport)
+#if defined(__WINE__) || !defined(_WIN32)
+#define DLLEXPORT __attribute__((visibility("default")))
+#include <dlfcn.h>
 #else
-  #define DLLEXPORT __attribute__((visibility("default")))
-  #include <dlfcn.h>
+#define DLLEXPORT
 #endif
 
 static pthread_once_t library_once = PTHREAD_ONCE_INIT;
--- a/libs/d3d12core/main.c
+++ b/libs/d3d12core/main.c
@@ -34,17 +34,11 @@
 
 #include "debug.h"
 
-/* We need to specify the __declspec(dllexport) attribute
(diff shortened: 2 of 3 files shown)


ishitatsuyuki:
Apparently tests were affected by the .lib/.a thing... I'll fix this later.

Joshua-Ashton (commented):
Once you fix the tests linking to lib_d3d12 which is redefined to be d3d12_dep (why?), looks fine.

I think it should probably go away if you just nuke the lib_d3d12 = d3d12_dep thing on Win32 builds.

ishitatsuyuki:
Sigh, MinGW LD is just broken (LLD emitted artifacts work fine). I've updated to use the same approach as DXVK (link to system D3D12). There seems to be a couple direct reference to D3D12 functions in the tests.

HansKristian-Work:
Is this PR stalled? Needs to be rebased at least.

ishitatsuyuki:
OK, I will rebase.

ishitatsuyuki:
Done.

PR #1823, opened 2023-12-31
d3d12: Pick the adapter chosen by the app when duplicate LUIDs are present

### LINKED ISSUE #1809: Wrong GPU being selected on hybrid graphics system
I am on a laptop that has both integrated graphics on the Intel CPU and a discrete NVIDIA GPU. When games are launched on Proton, they should be selecting the dGPU to run on, but when on D3D12 they consistently only use the integrated graphics.
`__NV_PRIME_RENDER_OFFLOAD` does not solve the issue, unfortunately. The discrete GPU still shows up at 0% usage in my system monitor and the game will run at less than 10 FPS.
This only occurs with D3D12 - when I run a game that uses Vulkan or D3D11, the NVIDIA GPU gets picked, as expected.

### Software information
I have tested this with Satisfactory (D3D12 vs. Vulkan rendering backends), Deep Rock Galactic (D3D12 vs. D3D11 rendering backends, and Forza Horizon 5 (only D3D12 and just won't run).

### System information
- GPU: Mesa Intel® Graphics and NVIDIA RTX 5000 Ada Generation Laptop GPU
- Driver: 4.18.1 and 545.29.06 respectively
- Wine version: Proton Experimental, but it also happens on 8.0.4
- VKD3D-Proton version: 2.11

### Log files
[steam-548430.log](https://github.com/HansKristian-Work/vkd3d-proton/files/13667371/steam-548430.log)

TechnoPorg:
Add a heuristic when initializing a d3d12 physical device to take the discrete GPU if available, and fall back to the integrated GPU. This mirrors the device selection code in vkd3d_select_physical_device. Previously, the first available device was selected without checking for others, which would lead to the wrong device being selected on hybrid graphics systems.

Fixes #1809

### DIFF
--- a/libs/d3d12core/main.c
+++ b/libs/d3d12core/main.c
@@ -199,7 +199,9 @@ static VkPhysicalDevice d3d12_find_physical_device(struct vkd3d_instance *instan
         if (id_properties.deviceLUIDValid && !memcmp(id_properties.deviceLUID, &adapter_desc->AdapterLuid, VK_LUID_SIZE))
         {
             vk_physical_device = vk_physical_devices[i];
-            break;
+            /* Workaround for systems where multiple devices have the same LUID */
+            if (properties2.properties.deviceID == adapter_desc->DeviceId && properties2.properties.vendorID == adapter_desc->VendorId)
+                break;
         }
     }
 


K0bin:
Are there 2 GPUs with the same LUID or why is this necessary?

TechnoPorg:
Yes, both my integrated graphics and discrete GPU seem to have the same LUID. Here are the results of some debug logging I did with the following code:
```c
        if (id_properties.deviceLUIDValid && !memcmp(id_properties.deviceLUID, &adapter_desc->AdapterLuid, VK_LUID_SIZE))
        {
            TRACE("GPU %u has LUID %u\n", vk_physical_devices[i], id_properties.deviceLUID);
            if(properties2.properties.deviceType == VK_PHYSICAL_DEVICE_TYPE_DISCRETE_GPU && !dgpu_device)
                dgpu_device = vk_physical_devices[i];
            if(properties2.properties.deviceType == VK_PHYSICAL_DEVICE_TYPE_INTEGRATED_GPU && !igpu_device)
                igpu_device = vk_physical_devices[i];
        }
    }
    TRACE(dgpu_device ? "There is a dgpu\n" : "There is no dgpu\n");
    if(dgpu_device) TRACE("The dgpu is %u\n", dgpu_device);
    TRACE(igpu_device ? "There is an igpu\n" : "There is no igpu\n");
    if(igpu_device) TRACE("The igpu is %u\n", igpu_device);
```
```
0134:trace:d3d12_find_physical_device: Matching adapters by LUIDs.
0134:trace:d3d12_find_physical_device: Device name: Intel(R) Graphics (RPL-S).
0134:trace:d3d12_find_physical_device: GPU 247451976 has LUID 12182896
0134:trace:d3d12_find_physical_device: Device name: NVIDIA RTX 5000 Ada Generation Laptop GPU.
0134:trace:d3d12_find_physical_device: GPU 247451992 has LUID 12182896
0134:trace:d3d12_find_physical_device: Device name: llvmpipe (LLVM 17.0.6, 256 bits).
0134:trace:d3d12_find_physical_device: There is a dgpu
0134:trace:d3d12_find_physical_device: The dgpu is 247451992
0134:trace:d3d12_find_physical_device: There is an igpu
0134:trace:d3d12_find_physical_device: The igpu is 247451976
0134:trace:d3d12_find_physical_device: Using device: NVIDIA RTX 5000 Ada Generation Laptop GPU, 0x10de:0x2730.
```

Etaash-mathamsetty:
> Are there 2 GPUs with the same LUID or why is this necessary?
> 
> Otherwise it will only ever set either `igpu` or `dgpu`, never both.

I think 2 GPUs should not have the same LUID and this should be fixed dxvk side

Etaash-mathamsetty:
I would probably make a integer called score and make a function to calculate the score based on the device properties, then everytime you see a bigger score set vk_physical_device to that value. 

Not sure if this is the best way, but I think it's a cleaner way

TechnoPorg:
Such a score would probably have to be based on available graphics memory (see [this StackExchange answer](https://gamedev.stackexchange.com/questions/124738/how-to-select-the-most-powerful-vkdevice#124751)). I'm not sure in what scenario a more extensible way to select a graphics device would be necessary, though, because one can already manually choose with an environment variable, and to my knowledge, there are no mainstream gaming devices with more than 2 discrete/integrated GPUs, or where someone is gaming on a software rasterizer :laughing:. This implementation is the same way the device is selected in `device.c`, which is why I chose to use it. To be clear, I'm not trying to be defensive or aggressive, just explaining my rationale.

> Otherwise it will only ever set either igpu or dgpu, never both.

I'm not familiar with how adapter LUIDs work - could you explain why this is the case?

To be honest, I'm not sure why this particular device selection function is needed at all, since it's only called in one place, in `main.c`, and later on in that chain of calls, `vkd3d_select_physical_device` in `device.c` is called to do the same thing. This is why it's good that more skilled graphics programmers than me are the ones writing most of this code!

K0bin:
> I would probably make a integer called score and make a function to calculate the score based on the device properties, then everytime you see a bigger score set vk_physical_device to that value.

That seems overkill.

Etaash-mathamsetty:
> > I would probably make a integer called score and make a function to calculate the score based on the device properties, then everytime you see a bigger score set vk_physical_device to that value.
> 
> That seems overkill.

in this case it would be a simple function score of 2 with dgpu and score of 1 with igpu, but yeah it definitely is overkill lol

doitsujin:
For Windows builds we could really just call into DXGI and query whatever the first adapter is, then find the corresponding Vulkan adapter by LUID. This also has the advantage that it would be consistent with DXVK's behaviour, I don't really a reason to reinvent the wheel here.

With the holiday season now well and truly over I'll look into this in the coming days.

libs/d3d12core/main.c
```diff
@@ -198,11 +201,17 @@ static VkPhysicalDevice d3d12_find_physical_device(struct vkd3d_instance *instan
 
         if (id_properties.deviceLUIDValid && !memcmp(id_properties.deviceLUID, &adapter_desc->AdapterLuid, VK_LUID_SIZE))
         {
-            vk_physical_device = vk_physical_devices[i];
-            break;
+            if(properties2.properties.deviceType == VK_PHYSICAL_DEVICE_TYPE_DISCRETE_GPU && !dgpu_device)
+                dgpu_device = vk_physical_devices[i];
+            if(properties2.properties.deviceType == VK_PHYSICAL_DEVICE_TYPE_INTEGRATED_GPU && !igpu_device)
+                igpu_device = vk_physical_devices[i];
```
doitsujin, line 207: I'd rather just do something like

```c
vk_physical_device = vk_physical_devices[i];

if (properties2.properties.deviceID == adapter_desc->deviceId &&
    properties2.properties.vendorID == adapter_desc->vendorId)
  break;
```

doitsujin: See comment. Multiple adapters sharing the same LUID shouldn't be happening in the first place, but instead of adding some weird heuristic we should make an effort to actually match the device that the app wants to use by checking the remaining adapter properties.

doitsujin (changes requested):

libs/d3d12core/main.c
Etaash-mathamsetty: I was thinking about this, but would it work when dxgi "fakes" a different vendor id on nvidia?

TechnoPorg:
Updated as per your feedback @doitsujin. I hadn't realized that this was just selecting the device selected by the app, so it makes a lot more sense what's going on here now!

libs/d3d12core/main.c
doitsujin: That's a problem, but at least there are ways to work around that from the Proton/DXVK side.

doitsujin: I guess we could also look at the device name as a last resort. That can technically be faked too, but no part in the stack is going to do that by default and DXVK just forwards the raw Vulkan device name anyway.

HansKristian-Work:
superseded, closing

PR #1824, opened 2024-01-05
d3d12: Be more robust with LUID-based adapter selection.

### LINKED ISSUE #1809: Wrong GPU being selected on hybrid graphics system

doitsujin:
Like the original PR (#1823), this should fix situations where multiple Vulkan adapters get assigned the same LUID, but with comparing device names this should also be robust against DXVK-side vendor ID overrides, which are still sometimes needed.

This is a situation that generally *shouldn't* happen and is probably some sort of winex11 issue in combination with systems that have multiple GPUs that can all output to the same display in one way or another.

Pinging @Etaash-mathamsetty , @TechnoPorg

Supersedes #1823.
Fixes #1809.

### DIFF
--- a/libs/d3d12core/main.c
+++ b/libs/d3d12core/main.c
@@ -148,9 +148,10 @@ static VkPhysicalDevice d3d12_find_physical_device(struct vkd3d_instance *instan
     VkPhysicalDeviceProperties2 properties2;
     VkPhysicalDevice *vk_physical_devices;
     VkInstance vk_instance;
-    unsigned int i;
+    unsigned int i, j;
     uint32_t count;
     VkResult vr;
+    bool match;
 
     vk_instance = vkd3d_instance_get_vk_instance(instance);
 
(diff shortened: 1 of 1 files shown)


libs/d3d12core/main.c
```diff
@@ -198,8 +199,33 @@ static VkPhysicalDevice d3d12_find_physical_device(struct vkd3d_instance *instan
...
+                {
+                    /* For simplicity, assume that adapter names are all ASCII characters */
+                    match = true;
+
+                    for (j = 0; j < ARRAY_SIZE(adapter_desc->Description); j++)
+                    {
+                        WCHAR a = (WCHAR)properties2.properties.deviceName[j];
+                        WCHAR b = adapter_desc->Description[j];
+
+                        if (!(match = (a == b)) || !a || !b)
```
HansKristian-Work, line 221: This seems a bit iffy if string lengths don't match?

HansKristian-Work: Might be better to add a wide string compare util to our string code?

doitsujin: If string lengths don't match, then there's a point where one character is 0 but not the other, so a != b -> match = false. The last part wouldn't even be evaluated in that case.

HansKristian-Work: Right ...

HansKristian-Work (approved):

TechnoPorg:
Can confirm that this fixes the bug for me. Thanks a lot!

PR #1827, opened 2024-01-08
Implement render passes

doitsujin:
Title, basically.

The implementation doesn't really do anything special, it essentially behaves as if OMSetRenderTargets was used instead, with discards, clears and resolves being performed as necessary. As a result, we keep exposing `RENDER_PASS_TIER_0`, which is equivalent to runtime emulation.

There's a few things tests don't cover:
- Resolve parameter validation. There doesn't seem to be any, passing the wrong resource or passing null pointers for the subresource array can easily result in a crash.
- `NO_ACCESS` behaviour for individual depth/stencil aspects. If e.g. depth is written and the stencil aspect is `NO_ACCESS`, it is still bound and can even be written in practice, but doing so is obviously invalid.


libs/vkd3d/command.c
```diff
@@ -14588,13 +14640,35 @@ static void STDMETHODCALLTYPE d3d12_command_list_BeginRenderPass(d3d12_command_l
...
 {
-    FIXME("iface %p, rt_count %u, render_targets %p, depth_stencil %p, flags %#x stub!\n",
+    struct d3d12_command_list *list = impl_from_ID3D12GraphicsCommandList(iface);
+
+    TRACE("iface %p, rt_count %u, render_targets %p, depth_stencil %p, flags %#x.\n",
             iface, rt_count, render_targets, depth_stencil, flags);
+
+    if (list->is_inside_render_pass)
+    {
+        d3d12_command_list_mark_as_invalid(list, "BeginRenderPass called outside a render pass.\n");
```
HansKristian-Work, line 15022: inside a render pass?

libs/vkd3d/command.c
```diff
@@ -3482,9 +3482,9 @@ static void d3d12_command_list_debug_mark_end_region(struct d3d12_command_list *
         VK_CALL(vkCmdEndDebugUtilsLabelEXT(list->cmd.vk_command_buffer));
 }
 
-static void d3d12_command_list_clear_attachment_pass(struct d3d12_command_list *list, struct d3d12_resource *resource,
+static void d3d12_command_list_load_attachment(struct d3d12_command_list *list, struct d3d12_resource *resource,
         struct vkd3d_view *view, VkImageAspectFlags clear_aspects, const VkClearValue *clear_value, UINT rect_count,
-        const D3D12_RECT *rects, bool is_bound)
```
HansKristian-Work, line 3487: Was is_bound always false? Seems like it from the diff at least.

libs/vkd3d/command.c
```diff
@@ -14669,11 +14692,327 @@ static void STDMETHODCALLTYPE d3d12_command_list_SetProtectedResourceSession(d3d
...
+            access == D3D12_RENDER_PASS_BEGINNING_ACCESS_TYPE_DISCARD ||
+            access == D3D12_RENDER_PASS_BEGINNING_ACCESS_TYPE_PRESERVE ||
+            access == D3D12_RENDER_PASS_BEGINNING_ACCESS_TYPE_PRESERVE_LOCAL_RENDER)
+        return true;
+
+    if (access == D3D12_RENDER_PASS_BEGINNING_ACCESS_TYPE_PRESERVE_LOCAL_SRV)
+    {
+        bind_read_only_dsv_flags = 0u;
+
+        if (aspect == VK_IMAGE_ASPECT_DEPTH_BIT)
```
HansKristian-Work, line 14995: Is only one aspect supported here?

libs/vkd3d/command.c
```diff
@@ -14684,8 +15023,104 @@ static void STDMETHODCALLTYPE d3d12_command_list_BeginRenderPass(d3d12_command_l
...
+            if ((rtv_desc = d3d12_rtv_desc_from_cpu_handle(rt->cpuDescriptor)) && rtv_desc->resource)
+            {
+                list->rtvs[rt_index] = *rtv_desc;
+
+                VKD3D_BREADCRUMB_AUX64(rtv_desc->view->cookie);
+                VKD3D_BREADCRUMB_AUX32(i);
+                VKD3D_BREADCRUMB_TAG("RTV bind");
+
+                if (!(flags & D3D12_RENDER_PASS_FLAG_RESUMING_PASS))
+                    d3d12_command_list_load_render_pass_rtv(list, &list->rtvs[rt_index], rt);
```
HansKristian-Work, line 15059: I think we're missing image initialization tracking here. If CLEAR op is used, we need to do similar things that OMSetRenderTargets does and set d3d12_command_list_track_resource_usage(false) if appropriate.

If we don't do anything, e.g. partial CLEAR or LOAD, we'll hit the init in d3d12_command_list_emit_render_pass_transition.

libs/vkd3d/command.c
```diff
@@ -14669,11 +14692,327 @@ static void STDMETHODCALLTYPE d3d12_command_list_SetProtectedResourceSession(d3d
...
+                break;
+            }
+        }
+
+        if (!merged_regions)
+        {
+            vkd3d_array_reserve((void **)&list->rtv_resolve_regions, &list->rtv_resolve_region_size,
+                    list->rtv_resolve_region_count + 1, sizeof(*list->rtv_resolve_regions));
+
+            list->rtv_resolve_regions[resolve->region_index + (resolve->region_count++)] = region;
```
HansKristian-Work, line 14974: Should we assert here that resolve->region_index + resolve->region_count == list->rtv_resolve_region_count? Seems like it should be true, but it's slightly confusing to not index by the list->rtv_resolve_region_count here.

HansKristian-Work (changes requested):
Comments.

Overall the strategy seems alright, at least for IMR. It's a TIER_0 implementation either way.

libs/vkd3d/device.c
```diff
@@ -7566,9 +7566,7 @@ static void d3d12_device_caps_init_feature_options17(struct d3d12_device *device
 static void d3d12_device_caps_init_feature_options18(struct d3d12_device *device)
 {
     D3D12_FEATURE_DATA_D3D12_OPTIONS18 *options18 = &device->d3d12_caps.options18;
-
-    /* We don't currently support render passes at all */
-    options18->RenderPassesValid = FALSE;
+    options18->RenderPassesValid = TRUE;
```
HansKristian-Work, line 7569: Don't we have to expose RenderPassTier0 as well?

libs/vkd3d/command.c
doitsujin: Yes, there was only ever one call to this with is_bound being false, so I removed it.

libs/vkd3d/command.c
doitsujin: Yes, the parameter type is `VkImageAspectFlagBits` for that reason (rather than `Flags`).

libs/vkd3d/device.c
doitsujin: We already do that (`options5`). There is no "not supported" tier because of runtime emulation.

libs/vkd3d/command.c
doitsujin: I can change this to be more clear.

libs/vkd3d/command.c
doitsujin: derp

libs/vkd3d/command.c
doitsujin: Right, put the relevant calls in `d3d12_command_list_load_render_pass_rtv` for simplicity.

libs/vkd3d/command.c
```diff
@@ -5419,6 +5421,26 @@ static void d3d12_command_list_init_default_descriptor_buffers(struct d3d12_comm
...
 
+static void d3d12_command_list_reset_rtv_resolves(struct d3d12_command_list *list)
+{
+    unsigned int i;
+
+    for (i = 0; i < list->rtv_resolve_count; i++)
+    {
+        struct d3d12_rtv_resolve *resolve = &list->rtv_resolves[i];
+
+        resolve->src_resource = NULL;
```
HansKristian-Work, line 5432: Any reason to clear out memory like this?

libs/vkd3d/command.c
```diff
@@ -14669,11 +14692,337 @@ static void STDMETHODCALLTYPE d3d12_command_list_SetProtectedResourceSession(d3d
...
+
+static void d3d12_command_list_load_render_pass_rtv(struct d3d12_command_list *list,
+        struct d3d12_rtv_desc *rtv_info, const D3D12_RENDER_PASS_RENDER_TARGET_DESC *rt)
+{
+    VkAttachmentLoadOp load_op = vk_load_op_from_d3d12(rt->BeginningAccess.Type);
+    VkClearValue clear_value;
+
+    vk_clear_color_value_from_d3d12(&clear_value.color, &rt->BeginningAccess.Clear.ClearValue);
+
+    d3d12_command_list_track_resource_usage(list, rtv_info->resource,
```
HansKristian-Work, line 14729: The RTV must cover the full resource to be able to signal false here. There should be a helper vkd3d_rtv_and_aspects_fully_cover_resource that's used by d3d12_command_list_clear_attachment.

HansKristian-Work (changes requested):
Some nits.

libs/vkd3d/command.c
doitsujin: This used to store an array locally but I guess this can just be left alone now.

doitsujin: Removed this bit, the array does not need to be initialized at all anymore.

libs/vkd3d/command.c
doitsujin: urg, I hate this. Anyway, should be fixed.

HansKristian-Work (approved):
LGTM

PR #1828, opened 2024-01-09
Disable alpha to coverage when fragment shader exports sample mask

doitsujin:
Not immediately relevant for any game, but we've encountered this issue in D3D11, and D3D12 behaves the same.

Also proves that sample mask export still affects rendering with non-multisamples render targets.


libs/vkd3d-shader/spirv.c
```diff
@@ -11813,6 +11827,56 @@ void vkd3d_shader_extract_feature_meta(struct vkd3d_shader_code *code)
                 meta |= VKD3D_SHADER_META_FLAG_EMITS_TRIANGLES;
 
         }
+        else if ((op == SpvOpDecorate && count == 4) ||
+                (op == SpvOpMemberDecorate && count == 5))
```
HansKristian-Work, line 11831: Member decorate? SpvBuiltInSampleMask cannot be placed in IO block.

libs/vkd3d-shader/spirv.c
```diff
@@ -11813,6 +11827,56 @@ void vkd3d_shader_extract_feature_meta(struct vkd3d_shader_code *code)
...
+
+            if (storage_class == SpvStorageClassOutput || storage_class == SpvStorageClassInput)
+            {
+                var_id = spirv[offset + 2];
+
+                for (i = 0; i < tracked_builtin_count; i++)
+                {
+                    const struct vkd3d_tracked_builtin *entry = &tracked_builtins[i];
+
+                    if (entry->var_id != var_id)
```
HansKristian-Work, line 11865: Is it written like this to potentially look for more than builtin later?

libs/vkd3d-shader/spirv.c
HansKristian-Work: Is the intention to be more extensible here?

HansKristian-Work (changes requested):
Nits/questions

libs/vkd3d-shader/spirv.c
doitsujin: Pretty much, yes.

libs/vkd3d-shader/spirv.c
doitsujin: The idea here was to introduce a mechanism which can be easily extended in the future. I think there's value to that, e.g. if we ever upgrade one of the shader compilers to SPIR-V 1.6 and things like viewport/layer export change, and potential future changes.

HansKristian-Work (approved):
LGTM

PR #1836, opened 2024-01-16
RFC: Add application to compile batches of HLSL shaders

doitsujin:
As discussed on Discord, this could be used to make writing tests less painful by allowing us to simply run the exe within a directory that contains HLSL shaders and spits out plain C headers. Draft for now to gauge whether this is going in the right direction.

Currently, this works as follows:
- For DXIL, `dxcompiler.dll` must be in the search path (either next to the exe or in the wine prefix). As for headers, I added a new IDL with the necessary interfaces, but it does not cover the entire DXC API.
- For DXBC, we require `d3dcompiler_47.dll` as the latest version. I haven't added a custom IDL for the d3dcompiler APIs since MinGW ships the necessary headers anyway, but I can add a minimal IDL here as well if we want to go that route.
- Input directory defaults to CWD, can be changed via command line. This currently does not scan directories recursively.
- Output directory defaults to `headers` within the source directory.
- Input files must named something like `frog.vs_6_0.vs_5_0.hlsl`, with support for up to one DXIL target and up to one DXBC target (of course, specifying only one also works). The output file name would be `frog.h`, and names of the corresponding `D3D12_SHADER_BYTECODE` constants are `frog_dxil` and `frog_dxbc`.
- The entry point **must** be called `main`. When porting our existing shaders, this would to be adjusted in some places.
- Debug and reflection info are currently always stripped out of the binaries to keep the size small.
- In terms of naming, we assume ASCII characters for everything excluding directory paths. This is kind of necessary anyway since C wouldn't be overly happy about having 🐸 as a variable name, but also simplifies string conversions in various places because *of course* the shader compiler APIs are inconsistent about requiring wide strings vs utf-8 strings.

All of this could be changed or expanded if necessary (e.g. to allow for changing entry point names, or supporting multiple entry points per file), but for now I went with a fairly minimal approach. I haven't really looked into whether special options are necessary when compiling raytracing/workgraph shaders or shaders with embedded root signatures.

### DIFF
--- a/programs/vkd3d-hlsl-build/dxil_library.cpp
+++ b/programs/vkd3d-hlsl-build/dxil_library.cpp
@@ -1,0 +1,202 @@
+/*
+ * Copyright 2024 Philip Rebohle for Valve Software
+ *
+ * This library is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 2.1 of the License, or (at your option) any later version.
+ *
+ * This library is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
+ */
+#define INITGUID
+
+#include <array>
+#include <cstring>
+#include <iostream>
+
+#include "dxil_library.h"
+
+dxc_library dxc_library::instance;
+
+
+dxc_library::dxc_library()
+{
+    m_dxcompiler = LoadLibraryA("dxcompiler.dll");
+
+    if (!m_dxcompiler)
+    {
+        std::wcerr << "Failed to load dxcompiler.dll." << std::endl;
+        return;
+    }
+
+    m_dxc_create_instance = reinterpret_cast<DxcCreateInstanceProc>(
+        GetProcAddress(m_dxcompiler, "DxcCreateInstance"));
+
+    if (!m_dxc_create_instance)
+    {
+        std::wcerr << "DxcCreateInstance not found in dxcompiler.dll." << std::endl;
+        FreeLibrary(m_dxcompiler);
+        m_dxcompiler = nullptr;
+    }
+}
+
+
+dxc_library::~dxc_library()
+{
+    FreeLibrary(m_dxcompiler);
+}
+
+
+bool dxc_library::compile(const std::vector<char> &code, LPCWSTR source_name, LPCWSTR target,
+        LPCWSTR entry_point, const std::vector<std::wstring> &arguments, std::vector<char> &binary) const
+{
+    if (!m_dxcompiler)
+        return false;
+
+    /* These are not thread-safe, just create new instances each time */
+    IDxcUtils *utils = nullptr;
+    HRESULT hr = m_dxc_create_instance(CLSID_DxcUtils, IID_PPV_ARGS(&utils));
+
+    if (FAILED(hr))
+    {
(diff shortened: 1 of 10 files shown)


programs/vkd3d-hlsl-build/dxil_library.cpp
```diff
@@ -0,0 +1,199 @@
...
+            {
+                binary.resize(binary_blob->GetBufferSize());
+                std::memcpy(binary.data(), binary_blob->GetBufferPointer(), binary.size());
+            }
+        }
+
+        if (binary_blob)
+            binary_blob->Release();
+        else
+            std::wcerr << "Failed to retrieve shader binary." << std::endl;
```
HansKristian-Work, line 161: wcerr?

programs/vkd3d-hlsl-build/dxil_library.cpp
```diff
@@ -0,0 +1,199 @@
...
+        utils->Release();
+        compiler->Release();
+        return false;
+    }
+
+    /* Set up parameters that minimize the generated binary size */
+    IDxcCompilerArgs *args = nullptr;
+
+    std::array<LPCWSTR, 2> arguments = {
+        L"-Qstrip_debug",
```
HansKristian-Work, line 98: u"" in case we port this to the Linux .so version of this at some unknown point?

programs/vkd3d-hlsl-build/main.cpp
```diff
@@ -0,0 +1,478 @@
...
+                break;
+            }
+
+            if (profile_count == profiles.size())
+            {
+                std::wcerr << in_file << ": Too many profiles." << std::endl;
+                break;
+            }
+
+            if (!profiles.at(profile_count).parse(extension.c_str()))
```
HansKristian-Work, line 167: The critical part here is that we need a way to pass down custom command line options. E.g. some of the shaders need -enable-16bit-types, ftz denorm modes, etc.

I'm not too picky which direction we choose. SPIRV-Cross and dxil-spirv use a style where we have:

foo.dothisthing.dothatthing.frag

and the python script looks for pre-determined patterns of `.dothisthing.` and appends CLI arguments as necessary.

Another idea would be to make this completely generic where we can tag with e.g. `frog.native-types.cs_6_2.hlsl` and CLI can look up the file `native-types.txt` or something where the command line is written. This tool seems useful enough in general that it might be worth going the generic route.

HansKristian-Work (commented):
Overall, I like this approach, it should work very well for our needs.

programs/vkd3d-hlsl-build/dxil_library.cpp
doitsujin: This produces compiler errors as casting char16_t* to wchar_t* implicitly does not work, this would need more ugly casts.

I'm not sure how useful it is to think about porting right now given that linux dxc does not produce signed binaries and we do not have a good enough alternative to d3dcompiler right now; if this ever becomes a concern we might as well revisit it then.

HansKristian-Work: Fair enough.

programs/vkd3d-hlsl-build/dxil_library.cpp
doitsujin: We kind of need to use the `w` variants here since a lot of the strings we get are UTF-16 (including `std::filesystem::path`), and AFAIK on Windows this *needs* to be consistent throughout the application so we can't mix and match wcerr and cerr etc.

HansKristian-Work: But we're passing an ASCII string to wcerr. Is that supposed to handle the conversion for us?

doitsujin: It does at least appear to work whereas the other way around (feeding wchar_t* into cerr) does not.

programs/vkd3d-hlsl-build/main.cpp
doitsujin: I changed this to interpret everything that isn't a shader profile as a command line option, e.g. `frog.vs_6_0.denorm ftz.hlsl` would be equivalent to passing `-denorm ftz` to DXC.

PR #1841, opened 2024-01-23
Fix view creation with denorm MinLodClamp

doitsujin:
Atlas Fallen sets ResourceMinLODClamp to binary 0x1 for most texture views, so we end up skipping view creation a lot of the time and rendering appears completely black.

The correct behaviour is a bit obscure since at least AMD native does not match what the D3D11 functional spec suggest in terms of rounding, but the test passes with this implementation.

### DIFF
--- a/tests/d3d12_descriptors.c
+++ b/tests/d3d12_descriptors.c
@@ -4397,6 +4397,7 @@ void test_view_min_lod(void)
     {
         const D3D12_SHADER_BYTECODE *ps;
         int most_detailed_mip;
+        int mip_count;
         float test_lod;
         float min_lod;
         unsigned int expected_color;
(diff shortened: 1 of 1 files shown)


libs/vkd3d/resource.c
```diff
@@ -4450,7 +4451,9 @@ bool vkd3d_create_texture_view(struct d3d12_device *device, const struct vkd3d_t
     /* If the clamp is defined such that it would only access mip levels
      * outside the view range, don't make a view and use a NULL descriptor.
      * The clamp is absolute, and not affected by the baseMipLevel. */
-    if (vkd3d_float_to_fixed_24_8(desc->miplevel_clamp) <= vkd3d_float_to_fixed_24_8(desc->miplevel_idx + desc->miplevel_count - 1))
+    miplevel_clamp_fixed = vkd3d_float_to_fixed_24_8(desc->miplevel_clamp);
```
HansKristian-Work, line 4454: Why vkd3d-proton code here? Seems like something should have been squashed.

HansKristian-Work (changes requested):
Just nits.

HansKristian-Work (approved):

PR #1843, opened 2024-01-24
Handle global RS mismatch better

HansKristian-Work:
Warhammer 40k: Darktide relies on this to ... work when NVAPI is enabled. Native drivers seem more or less robust against this, so we need to be more robust too, sigh ...


include/private/hashmap.h
```diff
@@ -292,4 +292,15 @@ static inline uint64_t hash_fnv1_iterate_string(uint64_t h, const char *str)
     return h;
 }
 
+static inline uint64_t hash_fnv1_iterate_data(uint64_t h, const void *ptr_, size_t size)
+{
+    const uint8_t *ptr = ptr_;
+    size_t i;
+
+    for (i = 0; i < size; i++)
+        h = hash_fnv1_iterate_u8(h, ptr[i]);
```
doitsujin, line 301: Given that we use this to hash blobs, might be worth doing this per-dword rather than per-byte and then just gather the last couple of bytes by hand if the size isn't aligned to 4?

libs/vkd3d-shader/vkd3d_shader_main.c
```diff
@@ -980,3 +980,29 @@ int vkd3d_shader_parse_root_signature_v_1_2_from_raw_payload(const struct vkd3d_
...
+        hash = hash_fnv1_iterate_u32(hash, desc->parameters[i].parameter_type);
+        if (desc->parameters[i].parameter_type == VKD3D_ROOT_PARAMETER_TYPE_32BIT_CONSTANTS)
+            hash = hash_fnv1_iterate_u32(hash, desc->parameters[i].constants.value_count);
+        else
+            hash = hash_fnv1_iterate_u32(hash, 0);
+    }
+
+    /* static sampler desc are tightly packed in memory. */
+    hash = hash_fnv1_iterate_data(hash, desc->static_samplers,
+            desc->static_sampler_count * sizeof(*desc->static_samplers));
```
doitsujin, line 1005: `static_sampler_desc` still contains register space+number information, is it intended to include those?

include/private/hashmap.h
HansKristian-Work: There's also alignment of the pointer to consider.

libs/vkd3d-shader/vkd3d_shader_main.c
HansKristian-Work: That is true ... I suppose we need more intrusive hashing.

include/private/hashmap.h
HansKristian-Work: Removed function.

libs/vkd3d-shader/vkd3d_shader_main.c
HansKristian-Work: Fixed.

doitsujin (approved):

PR #1850, opened 2024-01-28
vkd3d: Falsely report RADV doesn't implicitly clear

Etaash-mathamsetty:
radv_zero_vram=true is causing performance problems for many people after some kernel commits, so just workaround the problem for now. (To be clear: at this point in time this is not intended to get merged, only meant for testing purposes)


ishitatsuyuki:
The issue is tracked at https://gitlab.freedesktop.org/mesa/mesa/-/issues/9189, hopefully the kernel patchset will be upstream soon.

PR #1855, opened 2024-01-29
Basic workgraph tests

HansKristian-Work:
This needs SDK 613 headers and latest DXC version.

These are basic "hello world" tests. There is no data passing between nodes here.

WARP has some bugs with multiple GPU input nodes.
AMD driver hangs GPU with CPU inputs + 0 stride. Most likely they screw up memory allocation requirements for that CPU data.

Otherwise, the tests passed on RX 7600 with the adrenalin driver now.


oscarbg:
Do you want/need testing on a NV Ada card?

HansKristian-Work:
Tests pass as-is on RTX 4070.

doitsujin:
Looks like there are some unused variable warnings from the new tests, other than that LGTM.

HansKristian-Work:
Also fixed the annoying so_strides warning nit while I was at it.

PR #1883, opened 2024-02-07
Implement fallback for compute shader derivatives

### LINKED ISSUE #1866: Any UE5 apps shipped as D3D12/SM6 fail to start on Windows
UE5 engine fails with generic RHI message
`DirectX 12 is not supported on your system. Try running without the -dx12 or -d3d12 command line argument.`
See:  https://github.com/EpicGames/UnrealEngine/blob/072300df18a94f18077ca20a14224b5d99fee872/Engine/Source/Runtime/RHI/Private/Windows/WindowsDynamicRHI.cpp#L869

### Software information
Any UE5 game that ships in configuration D3D12 RHI/SM6: Talos Principle 2, Only Up, The Isle,. etc

### System information
- GPU: GTX 1070
- Driver: 545.xx
- Wine version: --- 
- Windows version: 10 21h2
- VKD3D-Proton version:  2.11.1
- Vulak API version: 1.3.260

### Log files
[dxvk.log](https://github.com/HansKristian-Work/vkd3d-proton/files/14125421/dxvk.log)

HansKristian-Work:
Allows NV Pascal to expose SM 6.6, albeit with somewhat questionable conformance :) UE5 seems to require SM 6.6 to even boot now. Fix #1866.


doitsujin (approved):
Looks good, I'd be fine with not locking the fallback behind an Nvidia driver check too if we print a warning or something, but I can also kind of see why we're doing it.

HansKristian-Work:
> Looks good, I'd be fine with not locking the fallback behind an Nvidia driver check too if we print a warning or something, but I can also kind of see why we're doing it.

I don't want other drivers to start thinking they can skirt around not implementing the feature basically.

PR #1885, opened 2024-02-09
Use quad control / reconvergence features if available

doitsujin:
And enable SM 6.7. Still keeping the `EXPERIMENTAL_FEATURES` check due to driver support being more on the limited side of things right now.


include/vkd3d_shader.h
```diff
@@ -370,6 +370,8 @@ enum vkd3d_shader_target_extension
     VKD3D_SHADER_TARGET_EXTENSION_SUPPORT_FP64_DENORM_PRESERVE,
     VKD3D_SHADER_TARGET_EXTENSION_SUPPORT_SUBGROUP_PARTITIONED_NV,
     VKD3D_SHADER_TARGET_EXTENSION_COMPUTE_SHADER_DERIVATIVES_NV,
+    VKD3D_SHADER_TARGET_EXTENSION_MAXIMAL_RECONVERGENCE,
```
HansKristian-Work, line 373: Quad control extension depends on maximal reconvergence, so we should merge the extension to avoid the awkward inner loop.

HansKristian-Work (changes requested):
Smol nit.

HansKristian-Work (approved):
LGTM

PR #1907, opened 2024-02-19
vkd3d: R32_FLOAT supports both COLOR and DEPTH but is currently only reporting COLOR

dave-juicelabs:
A game engine is checking R32_FLOAT for DEPTH but finding it is not supported even though it is.

### DIFF
--- a/libs/vkd3d/device.c
+++ b/libs/vkd3d/device.c
@@ -3724,8 +3724,11 @@ static HRESULT d3d12_device_get_format_support(struct d3d12_device *device, D3D1
     const struct vkd3d_vk_device_procs *vk_procs = &device->vk_procs;
     VkPhysicalDeviceImageFormatInfo2 format_info;
     VkImageFormatProperties2 format_properties;
+    const struct vkd3d_format *depth_format;
+    VkFormatFeatureFlags2 buffer_features;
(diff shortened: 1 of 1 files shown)


libs/vkd3d/device.c
```diff
@@ -3740,8 +3743,10 @@ static HRESULT d3d12_device_get_format_support(struct d3d12_device *device, D3D1
         return S_OK;
     }
 
-    if (!(format = vkd3d_get_format(device, data->Format, false)))
-        format = vkd3d_get_format(device, data->Format, true);
+    format = vkd3d_get_format(device, data->Format, false);
```
HansKristian-Work, line 3746: Could you add some basic test coverage to prove this?

dave-juicelabs:
After looking in to this further, I fear if this PR goes it that it will cause unintended issues considering the format for R32_FLOAT changes when it is COLOR vs DEPTH and the features supported do not match. Specifically, it appears Vulkans R32_FLOAT does not support the required features for D3D12_FORMAT_SUPPORT1_SHADER_SAMPLE_COMPARISON while Vulkans D32_FLOAT does. Knowing that the resulting crash without this PR is most certainly a bug in the game engine, I will work with them to get it resolved.

PR #1934, opened 2024-03-07
vkd3d: Remove Cyberpunk 2077 shader quirks

pixelcluster:
The bug has been fixed game-side. Yay!

### DIFF
--- a/libs/vkd3d/device.c
+++ b/libs/vkd3d/device.c
@@ -681,16 +681,6 @@ static const struct vkd3d_shader_quirk_info witcher3_quirks = {
     witcher3_hashes, ARRAY_SIZE(witcher3_hashes), 0,
 };
 
-static const struct vkd3d_shader_quirk_hash cp77_hashes[] = {
-    /* Shader accesses descriptor heap out of bounds spuriously, causing GPU hang on RADV. */
-    { 0x55540466536c9e11, VKD3D_SHADER_QUIRK_DESCRIPTOR_HEAP_ROBUSTNESS },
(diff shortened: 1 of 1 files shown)


Joshua-Ashton:
Is it worth removing these? If they have new shaders, then this should only trigger on the older versions?

pixelcluster:
Yeah, I don't think the quirks get triggered anymore. But do we really want to keep essentially dead workarounds for outdated versions of the game? I don't think we expect any users to run game versions 2.0-2.11 anymore (the issue started appearing in 2.0 and has been fixed in 2.12 IIRC).

Some people might still use the 1.x branch from before Phantom Liberty released because 2.0 raised system requirements, but the issue doesn't appear there either.

Joshua-Ashton:
Yeah, mainly just thinking of modders/speedrunners given that game has a big scene there.

shelterx:
My thought is just why not just add a comment on what versions those workarounds are for and saying it's fixed from version x.x. instead.

HansKristian-Work:
> Yeah, mainly just thinking of modders/speedrunners given that game has a big scene there.

Is there any evidence speedrunners are targeting >= 2.0 && <= 2.11? Why would they play with DXR on? And if they lock to specific versions, any reason why they cannot use specific vkd3d-protons versions as well?

doitsujin:
Agree that we should remove the workaround if this has been fixed in game.

HansKristian-Work (approved):

HansKristian-Work:
Ok, let's do that then.

PR #1939, opened 2024-03-12
Use dynamic rasterization samples for multisampled PSOs

### LINKED ISSUE #1839: World of Warcraft Classic Season of Discovery - DX12 Anti-Aliasing Crash
World of Warcraft Season of Discovery (this is the only version I really play, so this may be present on other versions, I am just unaware) crashes for me specifically when I receive an honorable kill from player vs. player combat. I have not experienced any other crashes besides this, which happens 100% of the time when triggered from an honorable kill.

This crash does not occur for me when anti-aliasing is turned off while on DX12.
This crash does not occur for me when DX11 is used instead of DX12. I can use full anti-aliasing on DX11 without crashing while attempting to trigger the crash.

### Software information
World of Warcraft Classic Era, specifically Season of Discovery if that matters. 

Graphics Settings:
DirectX12, Anti-Aliasing - Advanced, Image-Based Techniques - CMAA, Multisample Techniques - 'Color 8x / Depth 8x', Multisample Alpha-Test - True

### System information
- GPU: NVIDIA RTX 3080TI
- Driver: 535.146.02
- Wine version: wine-ge-8-25-x86_64
- VKD3D-Proton version: v2.11.1

### Log files
attached 
[wowcrash.txt](https://github.com/HansKristian-Work/vkd3d-proton/files/14016798/wowcrash.txt)

doitsujin:
Fixes #1839 (hopefully).

### DIFF
--- a/libs/vkd3d/device.c
+++ b/libs/vkd3d/device.c
@@ -94,6 +94,7 @@ static const struct vkd3d_optional_extension_info optional_device_extensions[] =
     VK_EXTENSION(EXT_TRANSFORM_FEEDBACK, EXT_transform_feedback),
     VK_EXTENSION(EXT_VERTEX_ATTRIBUTE_DIVISOR, EXT_vertex_attribute_divisor),
     VK_EXTENSION(EXT_EXTENDED_DYNAMIC_STATE_2, EXT_extended_dynamic_state2),
+    VK_EXTENSION(EXT_EXTENDED_DYNAMIC_STATE_3, EXT_extended_dynamic_state3),
     VK_EXTENSION(EXT_EXTERNAL_MEMORY_HOST, EXT_external_memory_host),
     VK_EXTENSION(EXT_SHADER_IMAGE_ATOMIC_INT64, EXT_shader_image_atomic_int64),
     VK_EXTENSION(EXT_MESH_SHADER, EXT_mesh_shader),
@@ -1596,6 +1597,12 @@ static void vkd3d_physical_device_info_init(struct vkd3d_physical_device_info *i
         vk_prepend_struct(&info->features2, &info->extended_dynamic_state2_features);
     }
 
+    if (vulkan_info->EXT_extended_dynamic_state3)
+    {
+        info->extended_dynamic_state3_features.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_EXTENDED_DYNAMIC_STATE_3_FEATURES_EXT;
+        vk_prepend_struct(&info->features2, &info->extended_dynamic_state3_features);
+    }
+
     if (vulkan_info->EXT_external_memory_host)
     {
         info->external_memory_host_properties.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_EXTERNAL_MEMORY_HOST_PROPERTIES_EXT;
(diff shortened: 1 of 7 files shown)


libs/vkd3d/state.c
```diff
@@ -3959,6 +3970,12 @@ uint32_t d3d12_graphics_pipeline_state_get_dynamic_state_flags(struct d3d12_pipe
     if (graphics->index_buffer_strip_cut_value && !is_mesh_pipeline)
         dynamic_state_flags |= VKD3D_DYNAMIC_STATE_PRIMITIVE_RESTART;
 
+    /* Enable dynamic sample count for multisampled pipelines so we can work around
+     * bugs where the app may render to a single sampled render target. */
+    if ((!key || !key->rasterization_samples) && d3d12_graphics_pipeline_needs_dynamic_rasterization_samples(&state->graphics) &&
```
HansKristian-Work, line 3975: This seems somewhat strange. I would expect `!key || key->rasterization_samples`? If rasterization_samples is 0, i.e. no RTs, rasterization samples is static.

libs/vkd3d/vkd3d_private.h
```diff
@@ -2185,6 +2186,7 @@ struct vkd3d_pipeline_key
 {
     D3D12_PRIMITIVE_TOPOLOGY topology;
     VkFormat dsv_format;
+    VkSampleCountFlags rasterization_samples;
```
HansKristian-Work, line 2189: FlagBits?

libs/vkd3d/device.c
```diff
@@ -1596,6 +1597,12 @@ static void vkd3d_physical_device_info_init(struct vkd3d_physical_device_info *i
         vk_prepend_struct(&info->features2, &info->extended_dynamic_state2_features);
     }
 
+    if (vulkan_info->EXT_extended_dynamic_state3)
+    {
+        info->extended_dynamic_state3_features.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_EXTENDED_DYNAMIC_STATE_3_FEATURES_EXT;
+        vk_prepend_struct(&info->features2, &info->extended_dynamic_state3_features);
```
HansKristian-Work, line 1603: Should we explicitly disable all the features we don't use? Not sure if drivers would do anything differently based on those feature bits, but might good to be explicit about which features we actually use. We do similarly for EDS2.

HansKristian-Work (changes requested):
Need a CHANGELOG update, but other than nits looks solid.

libs/vkd3d/vkd3d_private.h
doitsujin: Fixed.

libs/vkd3d/device.c
doitsujin: Fair point, disabled all features now that aren't used by vkd3d-proton or dxvk.

libs/vkd3d/state.c
doitsujin: Removed this check. It was entirely redundant, `key->rasterization_samples` is 0 iff EDS3 rasterization samples is supported, see `get_or_create_pipeline` - the reasoning here is that we'd otherwise create redundant pipelines if we hit the fallback path for any other reason (DSV format memes without GPL).

There was a related issue though, `get_pipeline` wouldn't check for `dyn_state->rasterization_samples == 0` and instead we'd create a redundant fallback pipeline with the same properties as the main one. Fixed that too.

HansKristian-Work (approved):
LGTM

RepH3x:
This definitely worked for me, the OP from #1839. Thanks so much for the fix!

PR #1951, opened 2024-03-20
vkd3d: Ensure monotonic frame ID order for Reflex.

HansKristian-Work:
Some applications use app IDs in "random" order which completely breaks the existing implementation.

### DIFF
--- a/include/vkd3d_vk_includes.h
+++ b/include/vkd3d_vk_includes.h
@@ -68,28 +68,31 @@ typedef struct D3D12_UAV_INFO
     UINT64 gpuVASize;  
 } D3D12_UAV_INFO;
 
+typedef struct D3D12_FRAME_REPORT
+{
+    UINT64 frameID;
+    UINT64 inputSampleTime;
+    UINT64 simStartTime;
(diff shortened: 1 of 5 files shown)


adamnv:
I gave this a spin.  It successfully avoids the slow swapchain-thrashing that was previously reproducible in Spider-Man Miles Morales, Ratchet & Clank, and RoboCop Rogue City.

HansKristian-Work:
@esullivan-nvidia This exposes a bug on NV driver. When IDs aren't one after the other, the algorithm seems to break down. This is especially the case in GetLatencyTimingsNV where the outputs are complete bogus when using stride != 1.

HansKristian-Work:
@esullivan-nvidia any updates?

esullivan-nvidia:
I have submitted the fix for this issue internally. It should ship in the next developer beta driver. I will let you know when that happens.

HansKristian-Work:
Seems to work on latest beta driver, yes.

esullivan-nvidia:
LGTM, thanks!

doitsujin (approved):

PR #1961, opened 2024-04-08
Implement an adaptive internal latency handle.

HansKristian-Work:
Optimize for smoother frame pace rather than strict adherence to the
present wait internal. The scenario where this can greatly help
is on FRR display where GPU is rendering slower than refresh rate.
In this situation, we can observe that DXGI present blocks spuriously
for a full frame interval, which can cause starvation issues, since
CPU might become too late to submit more work to GPU in time,
which is especially devastating on Deck due to its power management
scheme.


HansKristian-Work:
Obsolete.

PR #1964, opened 2024-04-15
vkd3d: Enable vulkan extensions required for OpenVR.

yshui:


HansKristian-Work:
Is there a reason for us to query this from the registry? Can't we just enable the relevant extensions? Which extension does VR even need beyond what we enable in the first place?

HansKristian-Work:
If you only need to enable extensions and not enable device level features as well, I think the proper place to do so is in d3d12/main.c where you can give optional extensions to instance/device creation. See where we enable surface_maint1 and swapchain_maint1. The code also needs to be ifdeffed behind WIN32.

gofman:
Strictly speaking, we do not know in Proton what are the required extensions for VR, whether they are the same for every GPU or when they are going to be changed by Steam VR implementation. We query that from VR. So far we were avoiding hardcoding that and not sure if doing so would be a good idea.

yshui:
@HansKristian-Work for instance extensions i think it's possible to move it to `d3d12core/main.c`, but for device extensions I need to know the physical device first

HansKristian-Work:
> we do not know in Proton what are the required extensions for VR

Most extensions require a Feature struct to be used though. Has that not been a problem before?

libs/vkd3d/device.c
```diff
@@ -379,6 +379,184 @@ static bool vkd3d_disable_nvx_extensions(struct d3d12_device *device, const char
     return disabled;
 }
 
+static BOOL vkd3d_wait_vr_key(HKEY vr_key)
+{
+    DWORD type, value, wait_status, size;
+    LSTATUS status;
+    HANDLE event;
+    DWORD max_retry = 10; /* Each value query and each timeout counts as a try. */
```
HansKristian-Work, line 387: Why is a timeout needed?

libs/vkd3d/device.c
```diff
@@ -379,6 +379,184 @@ static bool vkd3d_disable_nvx_extensions(struct d3d12_device *device, const char
...
+    }
+
+    CloseHandle(event);
+    return value == 1;
+}
+
+static char *vkd3d_get_openvr_instance_extensions()
+{
+    HKEY vr_key;
+    LSTATUS status = RegOpenKeyA(HKEY_CURRENT_USER, "Software\\Wine\\VR", &vr_key);
```
HansKristian-Work, line 442: Is this key only set in registry when VR is intended to be used?

yshui: IIUC this key is always set.

libs/vkd3d/device.c
```diff
@@ -379,6 +379,184 @@ static bool vkd3d_disable_nvx_extensions(struct d3d12_device *device, const char
...
+    if (type != REG_DWORD)
+    {
+        ERR("OpenVR: unexpected value type %d", type);
+        return false;
+    }
+
+    if (value)
+        return value == 1;
+
+    event = CreateEventA(NULL, FALSE, FALSE, NULL);
```
HansKristian-Work, line 404: error handling?

libs/vkd3d/device.c
HansKristian-Work: Ok. Need inverse :christmas_tree: ordering though.

libs/vkd3d/device.c
```diff
@@ -379,6 +379,184 @@ static bool vkd3d_disable_nvx_extensions(struct d3d12_device *device, const char
...
+            break;
+        }
+        if (value)
+            break;
+        max_retry--;
+
+        while ((wait_status = WaitForSingleObject(event, 1000)) == WAIT_TIMEOUT && max_retry)
+        {
+            WARN("VR state wait timeout (retries left %u)", max_retry);
+            max_retry--;
```
HansKristian-Work, line 425: continue after this?

yshui:
> Most extensions require a Feature struct to be used though

OpenVR doesn't have a mechanism for telling the application what feature struct to pass. In practice none of the extensions they use need a feature struct.

libs/vkd3d/device.c
```diff
@@ -379,6 +379,184 @@ static bool vkd3d_disable_nvx_extensions(struct d3d12_device *device, const char
...
+        if ((status = RegQueryValueExA(vr_key, "state", NULL, &type, (BYTE *)&value, &size)))
+        {
+            ERR("OpenVR: could not query value, status %d", status);
+            break;
+        }
+        if (value)
+            break;
+        max_retry--;
+
+        while ((wait_status = WaitForSingleObject(event, 1000)) == WAIT_TIMEOUT && max_retry)
```
HansKristian-Work, line 422: Is there a risk that we wait for a full second if steam.exe sets the registry entry after RegQuery but before RegNotifyChangeKeyValue?

libs/vkd3d/device.c
```diff
@@ -379,6 +379,184 @@ static bool vkd3d_disable_nvx_extensions(struct d3d12_device *device, const char
...
+    RegCloseKey(vr_key);
+    return openvr_device_extensions;
+}
+
+static uint32_t vkd3d_parse_openvr_extension_list(char *extension_str, const char ***extension_list)
+{
+    char *cursor = extension_str;
+    uint32_t count = 0;
+
+    if (extension_list) *extension_list = NULL;
```
HansKristian-Work, line 536: Code on next line for single-line ifs.

libs/vkd3d/device.c
```diff
@@ -2931,6 +3129,13 @@ static HRESULT vkd3d_create_vk_device(struct d3d12_device *device,
         return hr;
     }
 
+    openvr_extensions_str = vkd3d_get_openvr_device_extensions(device);
+    openvr_extension_count = vkd3d_parse_openvr_extension_list(openvr_extensions_str, &openvr_extensions);
+    TRACE("OpenVR device extensions:");
+    for (i = 0; i < openvr_extension_count; i++) {
```
HansKristian-Work, line 3135: No braces needed.

libs/vkd3d/device.c
```diff
@@ -2949,6 +3155,8 @@ static HRESULT vkd3d_create_vk_device(struct d3d12_device *device,
     device_info.ppEnabledExtensionNames = extensions;
     device_info.pEnabledFeatures = &device->device_info.features2.features;
     vkd3d_free(user_extension_supported);
+    vkd3d_free(openvr_extensions);
```
HansKristian-Work, line 3158: This seems problematic. vkd3d_enable_extensions does not seem to strdup the extensions.

HansKristian-Work:
> I need to know the physical device first

I think you'll know that from the DXGI adapter in D3D12CreateDevice, right?

Joshua-Ashton:
Why are we using the registry? Can't we just ask OpenVR itself?

Joshua-Ashton:
https://github.com/doitsujin/dxvk/blob/master/src/dxvk/dxvk_openvr.cpp#L207

We do both in DXVK, but idk why? I don't see what the point of the vr key stuff is...

yshui:
@Joshua-Ashton I assume there must be a reason for this :thinking: @gofman do you know?

libs/vkd3d/device.c
HansKristian-Work: Yes.

yshui: Why do we need that?

gofman: I am afraid 3 seconds timeout might be a bit risky? Is there any reason not to take what dxvk has? I think that was working for some years now without any issues.

gofman: Yes, it may take some time on first run at least (if it is installed and configured of course. The initialization in steam.exe helper is done asynchronously not to delay starting up everything which everything has a lot of things to do before first d3d device is created.

For the years of that code being in dxvk I am not aware of a single instance of hang due to that. There is an error logging during wait so it would be immediately clear if it is hanging waiting for VR [registry] initialization.

gofman: In any case, I doubt 3 or 5 sec timeout solves anything here. If in theory we will be hitting that timeout on each [other] run and continuing after wait this is not a solution anyway, such delays on each device creation are unacceptable anyway.

libs/d3d12core/main.c
```diff
@@ -53,6 +54,197 @@ static void *vulkan_module = NULL;
...
+            break;
+        }
+        if (value)
+            break;
+        max_retry--;
+
+        while ((wait_status = WaitForSingleObject(event, 1000)) == WAIT_TIMEOUT && max_retry)
+        {
+            WARN("VR state wait timeout (retries left %u)", max_retry);
+            max_retry--;
```
HansKristian-Work, line 107: Need to continue; after this line, otherwise a misleading ERR is reported.

libs/d3d12core/main.c
```diff
@@ -53,6 +54,197 @@ static void *vulkan_module = NULL;
...
+{
+    char *cursor = extension_str;
+    uint32_t count = 0;
+
+    if (!cursor)
+        return 0;
+
+    while (*cursor)
+    {
+        char *next = strchr(cursor, L' ');
```
HansKristian-Work, line 224: wide char?

libs/d3d12core/main.c
```diff
@@ -53,6 +54,197 @@ static void *vulkan_module = NULL;
...
+        cursor = next + 1;
+    }
+
+    if (!extension_list)
+        return count;
+
+    cursor = extension_str;
+    for (uint32_t i = 0; i < count; i++)
+    {
+        char *next = strchr(cursor, L' ');
```
HansKristian-Work, line 237: wide char?

libs/d3d12core/main.c
```diff
@@ -53,6 +54,197 @@ static void *vulkan_module = NULL;
...
+{
+    HKEY vr_key;
+
+    LSTATUS status = RegOpenKeyA(HKEY_CURRENT_USER, "Software\\Wine\\VR", &vr_key);
+    char *openvr_instance_extensions = NULL;
+    DWORD len = 0, type;
+
+    if (status != ERROR_SUCCESS)
+    {
+        WARN("Failed to open VR registry key, status %d.\n", status);
```
HansKristian-Work, line 131: We expect this path to be taken if vkd3d-proton is running outside of Steam, right?

HansKristian-Work (changes requested):
Some nits, but overall approach seems reasonable.

libs/d3d12core/main.c
yshui: I think it's better to exclude `WAIT_TIMEOUT` from the it below, if that what you meant?

libs/d3d12core/main.c
yshui: Yeah

libs/d3d12core/main.c
HansKristian-Work: Or that. It should not print error on timeout.

HansKristian-Work (approved):
LGTM

yshui:
@HansKristian-Work Thank you so much!

PR #1972, opened 2024-04-26
vkd3d: Introduce interface for creating d3d12 resources from VkImages

yshui:
Motivation for adding this is to support OpenXR extension XR_KHR_D3D12_enable.

The way OpenXR is designed, the application gets an array of swapchain images from the runtime, and renders into them. (Unlike OpenVR, with which the application will submit textures to be displayed in the VR environment). This means we need to convert the VkImages we get on the unix side to ID3D12Resources on the PE side.

* * *

I hope I am doing this correctly. I tried the `hello_xr` demo on D3D12 which works, and I didn't see any warnings from the validation layer.

### DIFF
--- a/libs/vkd3d/device.c
+++ b/libs/vkd3d/device.c
@@ -3401,7 +3401,8 @@ HRESULT STDMETHODCALLTYPE d3d12_device_QueryInterface(d3d12_device_iface *iface,
         return S_OK;
     }
 
-    if (IsEqualGUID(riid, &IID_ID3D12DeviceExt))
+    if (IsEqualGUID(riid, &IID_ID3D12DeviceExt)
+            || IsEqualGUID(riid, &IID_ID3D12DeviceExt1))
     {
         struct d3d12_device *device = impl_from_ID3D12Device(iface);
         d3d12_device_vkd3d_ext_AddRef(&device->ID3D12DeviceExt_iface);
@@ -8563,7 +8564,7 @@ static void d3d12_device_replace_vtable(struct d3d12_device *device)
     }
 }
 
-extern CONST_VTBL struct ID3D12DeviceExtVtbl d3d12_device_vkd3d_ext_vtbl;
+extern CONST_VTBL struct ID3D12DeviceExt1Vtbl d3d12_device_vkd3d_ext_vtbl;
 extern CONST_VTBL struct ID3D12DXVKInteropDeviceVtbl d3d12_dxvk_interop_device_vtbl;
 extern CONST_VTBL struct ID3DLowLatencyDeviceVtbl d3d_low_latency_device_vtbl;
 
--- a/libs/vkd3d/resource.c
+++ b/libs/vkd3d/resource.c
@@ -3405,6 +3405,37 @@ static void d3d12_resource_tag_debug_name(struct d3d12_resource *resource,
         vkd3d_set_vk_object_name(device, (uint64_t)resource->res.vk_buffer, VK_OBJECT_TYPE_BUFFER, name_buffer);
 }
 
+HRESULT d3d12_resource_create_committed_borrowed(struct d3d12_device *device, const D3D12_RESOURCE_DESC1 *desc,
+        UINT64 vk_handle, struct d3d12_resource **resource)
+{
+    D3D12_HEAP_PROPERTIES heap_props;
+    struct d3d12_resource *object;
+    HRESULT hr;
+    if (desc->Dimension != D3D12_RESOURCE_DIMENSION_TEXTURE2D)
+    {
+        FIXME("Only creation of Texture2D resources are currently supported.\n");
+        return E_NOTIMPL;
+    }
+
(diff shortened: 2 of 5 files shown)


libs/vkd3d/device.c
```diff
@@ -3401,10 +3401,11 @@ HRESULT STDMETHODCALLTYPE d3d12_device_QueryInterface(d3d12_device_iface *iface,
         return S_OK;
     }
 
-    if (IsEqualGUID(riid, &IID_ID3D12DeviceExt))
+    if (IsEqualGUID(riid, &IID_ID3D12DeviceExt)
+            || IsEqualGUID(riid, &IID_ID3D12DeviceExt1))
     {
         struct d3d12_device *device = impl_from_ID3D12Device(iface);
-        d3d12_device_vkd3d_ext_AddRef(&device->ID3D12DeviceExt_iface);
+        ID3D12DeviceExt1_AddRef(&device->ID3D12DeviceExt_iface);
```
HansKristian-Work, line 3408: Why drop the explicit AddRef call?

libs/vkd3d/resource.c
```diff
@@ -3405,6 +3405,37 @@ static void d3d12_resource_tag_debug_name(struct d3d12_resource *resource,
...
+HRESULT d3d12_resource_create_committed_borrowed(struct d3d12_device *device, const D3D12_RESOURCE_DESC1 *desc,
+        UINT64 vk_handle, struct d3d12_resource **resource)
+{
+    D3D12_HEAP_PROPERTIES heap_props;
+    struct d3d12_resource *object;
+    HRESULT hr;
+    if (desc->Dimension != D3D12_RESOURCE_DIMENSION_TEXTURE2D)
+    {
+        FIXME("Only creation of Texture2D resources are currently supported.\n");
+        return VKD3D_ERROR_NOT_IMPLEMENTED;
```
HansKristian-Work, line 3417: E_NOTIMPL.

libs/vkd3d/resource.c
```diff
@@ -3405,6 +3405,37 @@ static void d3d12_resource_tag_debug_name(struct d3d12_resource *resource,
...
+
+    hr = d3d12_resource_create(device, VKD3D_RESOURCE_COMMITTED | VKD3D_RESOURCE_EXTERNAL,
+            desc, &heap_props, D3D12_HEAP_FLAG_SHARED, D3D12_RESOURCE_STATE_COMMON, NULL, 0, NULL, &object);
+    if (FAILED(hr))
+        return hr;
+
+    object->res.vk_image = (VkImage)vk_handle;
+    if (!object->desc.MipLevels)
+        object->desc.MipLevels = max_miplevel_count(desc);
+    object->initial_layout_transition = 1;
```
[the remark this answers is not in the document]
HansKristian-Work: Also need to get common_layout. Maybe common_layout should be passed down as a uint32_t vk_common_image_layout.

See end of vkd3d_get_image_create_info.

libs/vkd3d/resource.c
```diff
@@ -3405,6 +3405,37 @@ static void d3d12_resource_tag_debug_name(struct d3d12_resource *resource,
...
+        return E_NOTIMPL;
+    }
+
+    heap_props.Type = D3D12_HEAP_TYPE_DEFAULT;
+    heap_props.CPUPageProperty = D3D12_CPU_PAGE_PROPERTY_UNKNOWN;
+    heap_props.MemoryPoolPreference = D3D12_MEMORY_POOL_UNKNOWN;
+    heap_props.CreationNodeMask = 0;
+    heap_props.VisibleNodeMask = 0;
+
+    hr = d3d12_resource_create(device, VKD3D_RESOURCE_COMMITTED | VKD3D_RESOURCE_EXTERNAL,
```
HansKristian-Work, line 3426: I think you might have to set the PLACED flag instead of COMMITTED. That way Discard and Clears will do proper transitions from UNDEFINED.

yshui: thanks, i didn't know this. btw, should I also change the function name to `CreatePlacedFromBorrowedVulkanHandle`?

HansKristian-Work: CreateResourceFromBorrowedHandle makes more sense.

gofman:
Not sure if that is related here, but it looks like we are also missing queue index from ID3D12DXVKInteropDevice::GetVulkanQueueInfo which we need to pass to OpenXR (and we have that from IDXGIVkInteropDevice2 from dxvk / d3d11). I am not sure if that should be considered as an independent change and PR, if not maybe worth adding at once some way?

HansKristian-Work:
> ID3D12DXVKInteropDevice::GetVulkanQueueInfo which we need to pass to OpenXR (and we have that from IDXGIVkInteropDevice2 from dxvk / d3d11)

If DXVK added further revisions of that interface, there's no problem adding that to vkd3d-proton too. I don't mind it being in same PR as long as it's in separate commits as usual.

HansKristian-Work (approved):
LGTM

PR #1975, opened 2024-04-30
AC: Mirage workaround for missing barriers

HansKristian-Work:


doitsujin (approved):

Saancreed:
@HansKristian-Work Fyi Ubisoft games nowadays ship with two separate binaries, one for users that purchased the game and another one for subscribers of some Ubisoft+ thingy. In case of Mirage, the former is indeed `ACMirage.exe` but the latter is `ACMirage_plus.exe` so it might be a good idea to add the second name to the list of patterns as well.

```
$ ls -1 *.exe
ACMirage.exe
ACMirage_plus.exe
```

Also seen previously in a3d21494f7cc74c6b3763e87b59e94a40e98c589 (but back then it was AC Valhalla).

PR #1976, opened 2024-04-30
HFW workaround for Mesa 23 on Deck

HansKristian-Work:
Very ugly workaround, but we kind need to consider driver-specific CFG workarounds in this case, otherwise, we cannot reasonably move dxil-spirv forward ... :(


doitsujin (approved):
pain, but i guess we can at least get rid of this once people can be expected to run a fixed RADV.

PR #1979, opened 2024-05-02
More interesting workgraph tests

HansKristian-Work:


HansKristian-Work:
Passes on NV and WARP. Several compiler crashes on AMD, but no validation errors in sight.

doitsujin (approved):
Probably good to merge tests now.

PR #1981, opened 2024-05-03
Release swapchain command queue reference after swapchain decref

esullivan-nvidia:
This prevents possibly destroying the d3d12 device while releasing the frame_latency_fence if the device ref count hits zero.


HansKristian-Work (approved):
Nice find

PR #1983, opened 2024-05-04
Add config flag to disable SSBO alignment workaround

emryd:
Add `no_ssbo_alignment` config flag to disable SSBO alignment workaround.

This fixes the crashes in Horizon: Forbidden West in the Death's Door area which appear for some users, tested on a GeForce GTX 1080. See https://github.com/ValveSoftware/Proton/issues/7590#issuecomment-2020331165 for similar crashes reports.

Here is a debug log when such crashes occur:
```
6769.539:011c:01fc:info:vkd3d-proton:vkd3d_pipeline_library_disk_thread_main: Flushing disk cache (wakeup counter since last flush = 319). It seems like application has stopped creating new PSOs for the time being.
```


HansKristian-Work:
I don't buy outright that this is a fix. This will also completely cripple performance, since no descriptor buffer and offset buffers for SSBO which is *devastatingly bad*

PR #1985, opened 2024-05-08
vkd3d: Add an interface for getting all information about a VkQueue

yshui:
OpenXR's vulkan extension expects a queue index instead of a VkQueue handle directly, so we need to expose this information from vkd3d-proton to support that.


HansKristian-Work (approved):
LGTM

PR #1987, opened 2024-05-10
vkd3d: Support indexed draws with NULL index buffer.

doitsujin:
Brought up by @rlocatti-nv . `KHR_maintenance6` doesn't seem overly interesting for us, but it does allow us to handle this edge case properly. Tested on both RADV and Nvidia.


HansKristian-Work (approved):
LGTM

PR #1991, opened 2024-05-10
vkd3d: Use proper barrier stage for XFB pause/resume.

HansKristian-Work:


doitsujin (approved):

doitsujin:
Checked the spec, looks like this indeed got changed so that the counter read no longer happens at `DRAW_INDIRECT`.

HansKristian-Work:
Seems like this was fixed in sync2, ye.

PR #2000, opened 2024-05-28
tests: Add test for mismatched render target sizes.

doitsujin:
Confirms that while this is three miles deep into UB territory, our implementation is valid.

Behaviour for pixels not included in any of the bound render targets differs between implementations. It seems like AMD just uses the first bound RTV as a reference point, while Nvidia's behaviour matches the current vkd3d-proton implementation and WARP just does its own thing yet again.


HansKristian-Work (approved):
LGTM

PR #2004, opened 2024-05-29
Support sRGB compute resolve

### LINKED ISSUE #2001: Arma Reforger: map is invisible when using "hardware antialiasing" option
### Software information
"Hardware antialiasing" option causes the map to not draw in Arma Reforger. Disabling this option makes the map visible again
Here's a screenshot with "Hardware antialiasing" set to "x2":
Снимок экрана_20240528_201128
Here's a screenshot with "Hardware antialiasing" disabled:
Снимок экрана_20240528_201117

### System information
- GPU: RX 6700 XT 
- Driver: Mesa 24.1.0-arch1.1 RADV
- Wine version: GE-Proton9-5
- VKD3D-Proton version: 2.12.0+

### Log files
[steam-1874880.log](https://github.com/HansKristian-Work/vkd3d-proton/files/15473536/steam-1874880.log)

doitsujin:
Should fix #2001 for real now. We're hitting the compute path for the map resolve because the destination image is typeless and also not renderable.

Draft for now since I want to test something w.r.t. the format parameter.


libs/vkd3d/command.c
```diff
@@ -8605,19 +8605,16 @@ static VkResolveModeFlagBits vk_resolve_mode_from_d3d12(D3D12_RESOLVE_MODE mode)
 static const struct vkd3d_format *d3d12_command_list_get_resolve_format(struct d3d12_command_list *list,
         const struct d3d12_resource *dst_resource, const struct d3d12_resource *src_resource, DXGI_FORMAT format)
 {
+    if (!format)
+      return NULL;
```
HansKristian-Work, line 8609: 4 spaces?

libs/vkd3d/command.c
```diff
@@ -8605,19 +8605,16 @@ static VkResolveModeFlagBits vk_resolve_mode_from_d3d12(D3D12_RESOLVE_MODE mode)
...
+
     /* Depth-stencil formats require special care since the app will need to
      * pass in the stencil plane as subresource 1, but may also use a format
      * for which we only have one aspect flag set. Just ignore the explicit
      * format in that case so that we always know the full aspect mask. */
-    if (dst_resource->format->vk_aspect_mask & (VK_IMAGE_ASPECT_DEPTH_BIT | VK_IMAGE_ASPECT_STENCIL_BIT))
-        return dst_resource->format;
-
-    if (dst_resource->format->type != VKD3D_FORMAT_TYPE_TYPELESS)
+    if (dst_resource->format->vk_aspect_mask == (VK_IMAGE_ASPECT_DEPTH_BIT | VK_IMAGE_ASPECT_STENCIL_BIT))
```
HansKristian-Work, line 8615: Why did bitwise-and change to equal here?

libs/vkd3d/command.c
```diff
@@ -8645,18 +8649,21 @@ enum vkd3d_resolve_image_path d3d12_command_list_select_resolve_path(struct d3d1
...
-        /* Prefer the direct path even for typeless formats if possible */
-        if (vkd3d_format->vk_format == dst_resource->format->vk_format &&
-                vkd3d_format->vk_format == src_resource->format->vk_format)
-            path = VKD3D_RESOLVE_IMAGE_PATH_DIRECT;
-        else
-            path = VKD3D_RESOLVE_IMAGE_PATH_RENDER_PASS_ATTACHMENT;
+        path = VKD3D_RESOLVE_IMAGE_PATH_RENDER_PASS_ATTACHMENT;
+
+        /* Some drivers ignore the view format for resolve attachments */
+        if (list->device->device_info.vulkan_1_2_properties.driverID == VK_DRIVER_ID_NVIDIA_PROPRIETARY ||
```
HansKristian-Work, line 8658: Should gate this behind vkd3d_config driver workaround check?

tests/d3d12_copy.c
```diff
@@ -1684,6 +1684,229 @@ void test_multisample_resolve_formats(void)
...
+    {
+        vkd3d_test_set_context("Test %u", i);
+
+        memset(&resource_desc, 0, sizeof(resource_desc));
+        resource_desc.Dimension = D3D12_RESOURCE_DIMENSION_TEXTURE2D;
+        resource_desc.Width = 4;
+        resource_desc.Height = 4;
+        resource_desc.DepthOrArraySize = 1;
+        resource_desc.MipLevels = 1;
+        resource_desc.SampleDesc.Count = 2;
```
HansKristian-Work, line 1870: is MSAA 2x guaranteed to be supported? Would also work to use 4x MSAA and write sample_index / 2u in shader I think.

libs/vkd3d/command.c
doitsujin: Doesn't technically change anything, but the whole `TYPELESS` thing mentioned in the comment only ever matters for images that have both aspects set.

HansKristian-Work: ok.

tests/d3d12_copy.c
doitsujin: *technically* it's not required, but I also can't think of a single implementation that doesn't support 2x here. Can change it though.

libs/vkd3d/command.c
doitsujin: fixed

tests/d3d12_copy.c
HansKristian-Work: Should be trivial to use 4x here, so I'd rather we do that.

doitsujin: ok, updated.

HansKristian-Work (approved):
LGTM

PR #2009, opened 2024-06-05
Expose Shader Model 6.8

doitsujin:
Mostly just tests and wave size memes.

Tests verified to pass both on WARP and native Nvidia, not sure how to convince native AMD to expose SM6.8.

Draft untl we land the dxil-spirv stuff.


include/vkd3d_shader.h
```diff
@@ -86,10 +86,12 @@ struct vkd3d_shader_meta
     vkd3d_shader_hash_t hash;
     unsigned int cs_workgroup_size[3]; /* Only contains valid data if uses_subgroup_size is true. */
     unsigned int patch_vertex_count; /* Relevant for HS. May be 0, in which case the patch vertex count is not known. */
-    unsigned int cs_required_wave_size; /* If non-zero, force a specific CS subgroup size. */
+    unsigned int cs_wave_size_min; /* If non-zero, minimum or required subgroup size. */
```
HansKristian-Work, line 89: Given how this will bloat our caches, we should pack this in uint8_t x 3 (+1 pad) I think. MaxWaveSize fits fine in 8-bit.

libs/vkd3d/device.c
```diff
@@ -8230,6 +8231,25 @@ static void d3d12_device_caps_init_shader_model(struct d3d12_device *device)
...
+         *   - Completely replaces WaveSize at the DXIL level
+         *   - Specifies a range of subgroup sizes that the shader can operate
+         *     with, as well as a preferred subgroup size.
+         * - Expanded Comparison Sampling
+         *   - Basically Grad/Bias operands for Dref instructions
+         * - Work graphs
+         *   - uhh...
+         */
+        if (device->d3d12_caps.max_shader_model == D3D_SHADER_MODEL_6_7 &&
+                (vkd3d_config_flags & VKD3D_CONFIG_FLAG_ENABLE_EXPERIMENTAL_FEATURES))
```
HansKristian-Work: Is there a reason to put this behind experimental flag? We can enable SM 6.7 unconditionally now. We just lack AdvancedTexturing ops. This commit should probably be moved last so it cannot break bisects.

libs/vkd3d/state.c
```diff
@@ -2536,16 +2538,25 @@ static HRESULT vkd3d_setup_shader_stage(struct d3d12_pipeline_state *state, stru
...
+                subgroup_size_alignment = spirv_code->meta.cs_wave_size_preferred;
+                override_subgroup_size = true;
+            }
+            else if (spirv_code->meta.cs_wave_size_min && (
+                    spirv_code->meta.cs_wave_size_min > device->device_info.vulkan_1_3_properties.minSubgroupSize ||
+                    spirv_code->meta.cs_wave_size_max < device->device_info.vulkan_1_3_properties.maxSubgroupSize))
+            {
+                /* We generally want to let the driver decide on a subgroup size, but if the device supports
+                 * subgroup sizes outside the range required by the shader, we need to specify it manually.
+                 * For now, prefer large subgroups in order to hit wave64 on AMD. */
```
HansKristian-Work, line 2556: I would add a vendor check for Intel here and opt for smallest possible. Intel implements large wave in a slow way.

libs/vkd3d/device.c
```diff
@@ -8302,8 +8302,7 @@ static void d3d12_device_caps_init_shader_model(struct d3d12_device *device)
          * - Work graphs
          *   - uhh...
          */
-        if (device->d3d12_caps.max_shader_model == D3D_SHADER_MODEL_6_7 &&
-                (vkd3d_config_flags & VKD3D_CONFIG_FLAG_ENABLE_EXPERIMENTAL_FEATURES))
+        if (device->d3d12_caps.max_shader_model == D3D_SHADER_MODEL_6_7)
```
HansKristian-Work, line 8305: Seems weird to not just fuse the previous commit with this one.

HansKristian-Work (changes requested):

libs/vkd3d/device.c
doitsujin: done

include/vkd3d_shader.h
doitsujin: Fair, I had considered this but then didn't because the rest also uses larger types than necessary.

libs/vkd3d/state.c
doitsujin: Done.

HansKristian-Work (approved):
LGTM!

PR #2014, opened 2024-06-11
Implement frame statistics

doitsujin:
Implements necessary functionality for `IDXGISwapChain::GetFrameStatistics` to return somewhat meaningful data. Without it, we pretend that a present operation completes immediately, which is bad in case the app uses this for frame pacing purposes, and worse if it makes assumptions on whether or not it is safe to free GPU resources. There are a couple of games using this (albeit for unknown purposes), most notably Nixxes ports.

TL;DR is that the API quite literally counts successful presents, and we happen to have an internal timeline doing exactly that in a robust manner already so we can just expose that to the app. Some refactoring was needed since this requires us to always have the waiter thread active, even if present wait is not used for a given present.

Additionally, this adds a frame rate limiter. The main motivation for this interface on the DXVK side of things was to support games that disable their own built-in limiter when running in full-screen with a 60 Hz display mode, but we can do the same here. This is less important for D3D12, but works in games that allow changing the refresh rate, e.g. Horizon Zero Dawn. If this functionality isn't desired, I can drop the commit.

Currently untested on environments where present wait is unsupported. There's no rush to land this either, it's probably a good idea to tackle this after 2.13 and do some broad testing on whether it affects (or breaks) any existing games.

### DIFF
--- a/libs/vkd3d/swapchain.c
+++ b/libs/vkd3d/swapchain.c
@@ -118,11 +118,9 @@ struct dxgi_vk_swap_chain
         uint64_t present_id;
         bool present_id_valid;
 
-        /* Atomically updated after a PRESENT queue command has processed.
-         * We don't care about wrap around.
-         * We just care about equality check so we can atomically check if all outstanding present events have completed on CPU timeline.
-         * This is used to implement occlusion check. */
-        uint32_t present_count;
+        /* Atomically updated after a PRESENT queue command has processed. Used to atomically check if
+         * all outstanding present events have completed on CPU timeline. */
+        UINT64 present_count;
 
         /* For blits. Use simple VkFences since we have to use binary semaphores with WSI release anyways.
          * We don't need to wait on these fences on main thread. */
@@ -182,7 +180,7 @@ struct dxgi_vk_swap_chain
         struct d3d12_resource *backbuffers[DXGI_MAX_SWAP_CHAIN_BUFFERS];
         VkImageView vk_image_views[DXGI_MAX_SWAP_CHAIN_BUFFERS];
         uint64_t blit_count;
-        uint32_t present_count;
+        uint64_t present_count;
         UINT index;
 
         DXGI_COLOR_SPACE_TYPE dxgi_color_space_type;
@@ -853,7 +851,7 @@ static void dxgi_vk_swap_chain_set_hdr_metadata(struct dxgi_vk_swap_chain *chain
 
 static bool dxgi_vk_swap_chain_present_task_is_idle(struct dxgi_vk_swap_chain *chain)
 {
-    uint32_t presented_count = vkd3d_atomic_uint32_load_explicit(&chain->present.present_count, vkd3d_memory_order_acquire);
+    uint64_t presented_count = vkd3d_atomic_uint64_load_explicit(&chain->present.present_count, vkd3d_memory_order_acquire);
     return presented_count == chain->user.present_count;
 }
 
@@ -1162,8 +1160,12 @@ static HRESULT STDMETHODCALLTYPE dxgi_vk_swap_chain_SetHDRMetaData(IDXGIVkSwapCh
 
 static void STDMETHODCALLTYPE dxgi_vk_swap_chain_GetLastPresentCount(IDXGIVkSwapChain2 *iface, UINT64 *present_count)
 {
-    /* Doing nothing here is fine for now, DXVK will fill in dummy values. */
-    FIXME_ONCE("iface %p, present_count %p stub!\n", iface, present_count);
+    struct dxgi_vk_swap_chain *chain = impl_from_IDXGIVkSwapChain(iface);
+
+    TRACE("iface %p, present_count %p.\n", iface, present_count);
+
+    /* Represents the number of times Present has been called successfully */
+    *present_count = chain->user.present_count;
 }
 
 static void STDMETHODCALLTYPE dxgi_vk_swap_chain_GetFrameStatistics(IDXGIVkSwapChain2 *iface,
@@ -2474,7 +2476,7 @@ static void dxgi_vk_swap_chain_present_callback(void *chain_)
 {
     const struct dxgi_vk_swap_chain_present_request *next_request;
     struct dxgi_vk_swap_chain *chain = chain_;
-    uint32_t next_present_count;
+    uint64_t next_present_count;
     uint32_t present_count;
     uint32_t i;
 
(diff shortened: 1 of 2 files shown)


libs/vkd3d/swapchain.c
```diff
@@ -122,7 +122,7 @@ struct dxgi_vk_swap_chain
          * We don't care about wrap around.
          * We just care about equality check so we can atomically check if all outstanding present events have completed on CPU timeline.
          * This is used to implement occlusion check. */
-        uint32_t present_count;
+        uint64_t present_count;
```
HansKristian-Work, line 142: Probably need VKD3D_ALIGN(8) for 32-bit. Also, the comment above is now misleading w.r.t. not caring about wraparound.

libs/vkd3d/swapchain.c
```diff
@@ -853,7 +853,7 @@ static void dxgi_vk_swap_chain_set_hdr_metadata(struct dxgi_vk_swap_chain *chain
 
 static bool dxgi_vk_swap_chain_present_task_is_idle(struct dxgi_vk_swap_chain *chain)
 {
-    uint32_t presented_count = vkd3d_atomic_uint32_load_explicit(&chain->present.present_count, vkd3d_memory_order_acquire);
+    uint32_t presented_count = vkd3d_atomic_uint64_load_explicit(&chain->present.present_count, vkd3d_memory_order_acquire);
```
HansKristian-Work, line 896: Need uint64_t here since user.present count is also 64-bit now.

libs/vkd3d/swapchain.c
```diff
@@ -2571,26 +2576,35 @@ static void *dxgi_vk_swap_chain_wait_worker(void *chain_)
...
+                        entry.id, UINT64_MAX));
+            }
+            vkd3d_queue_timeline_trace_complete_present_wait(timeline_trace, cookie);
+
+            if (entry.begin_frame_time_ns)
+                end_frame_time_ns = vkd3d_get_current_time_ns();
+        }
+        else
+        {
+            d3d12_fence_set_event_on_completion(impl_from_ID3D12Fence1(chain->present.frame_latency_fence),
```
HansKristian-Work, line 2770: We could probably just wait on our own blit semaphore here instead of round-tripping through fence waiter threads and submission thread?

libs/vkd3d/swapchain.c
```diff
@@ -2552,16 +2539,34 @@ static void dxgi_vk_swap_chain_present_callback(void *chain_)
...
+#ifdef _WIN32
+    /* DXGI returns the raw QPC value */
+    LARGE_INTEGER li;
+    QueryPerformanceCounter(&li);
+    time = li.QuadPart;
+#else
+    time = vkd3d_get_current_time_ns();
+#endif
+
+    pthread_mutex_lock(&chain->frame_statistics.lock);
```
HansKristian-Work, line 2620: This seems like a good candidate for a spin-lock, but I don't have a strong opinion.

HansKristian-Work (changes requested):
Overall approach seems reasonable. Some nits though.

libs/vkd3d/swapchain.c
doitsujin: Sounds reasonable.

libs/vkd3d/swapchain.c
doitsujin: I changed this to `UINT64` since that is naturally aligned.

libs/vkd3d/swapchain.c
doitsujin: derp, fixed

libs/vkd3d/swapchain.c
doitsujin: Looking at this again we should no longer need the old frame latency fence at all and the complete semaphore is guaranteed to get signaled with the correct timeline anyway, so yeah, good point.

libs/vkd3d/swapchain.c
```diff
@@ -2701,6 +2888,76 @@ static HRESULT dxgi_vk_swap_chain_init_low_latency(struct dxgi_vk_swap_chain *ch
...
+}
+
+static HRESULT dxgi_vk_swap_chain_init_frame_rate_limiter(struct dxgi_vk_swap_chain *chain)
+{
+    double target_frame_rate;
+    char env[16];
+
+    pthread_mutex_init(&chain->frame_rate_limit.lock, NULL);
+
+    if (vkd3d_get_env_var("VKD3D_FRAME_RATE", env, sizeof(env)) ||
```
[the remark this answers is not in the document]
doitsujin: As mentioned in the other thread, I don't see a reason to overcomplicate things here.

HansKristian-Work (changes requested):
Potential sync issue, would be good to refactor out the frame limiter commit to its own PR after this one, since it's a bit more complicated and the other commits are looking pretty good on their own.

libs/vkd3d/swapchain.c
```diff
@@ -313,17 +319,6 @@ static void dxgi_vk_swap_chain_drain_queue(struct dxgi_vk_swap_chain *chain)
         if (chain->present.vk_acquire_semaphore[i] && chain->present.acquire_semaphore_signalled[i])
             dxgi_vk_swap_chain_wait_acquire_semaphore(chain, chain->present.vk_acquire_semaphore[i], false);
 
-    /* Ensures that all pending ReleaseSemaphore() calls are also made.
-     * This happens on the fence waiter queues, so it's not enough to call vkQueueWaitIdle to be 100% sure.
-     * The fence waiter thread processes requests in-order,
-     * so if we observe that an EVENT has been signalled,
-     * we know all pending semaphore signals have happened as well. */
-
-    chain->present.frame_latency_count += 1;
```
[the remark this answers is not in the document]
doitsujin: added a semaphore wait to fix this.

HansKristian-Work (approved):
LGTM

phaserblast:
This is a much-needed feature, so good work. Can you please just confirm here for good that **DXVK_FRAME_RATE** will be the common variable checked across both DXVK and vkd3d moving forward?

Blisto91:
It also sets the frame rate cap for vkd3d-proton yes. If both are set then `VKD3D_FRAME_RATE` takes precedence.

phaserblast:
> It also sets the frame rate cap for vkd3d-proton yes. If both are set then `VKD3D_FRAME_RATE` takes precedence.

> Note that this is one-way only. DXVK itself doesn't recognize `VKD3D_FRAME_RATE`, but VKD3D does recognize `DXVK_FRAME_RATE`, and, as Blisto said, `VKD3D_FRAME_RATE` takes precedence.

This is great! Thanks for the clarification.

PR #2017, opened 2024-06-17
Fix multiple issues and inconsistencies in log message formats

Saancreed:
:broom:, but I dumped most of it into a single commit so let me know if you want this stuff to be split differently.

* Adds missing newlines
* Adds some dots where appropriate
* Removes some stray spaces, in particular at the end of message just before newline and when there was more than one space in the middle of format string
* Changes some specifiers to better match argument types
* Fixes some conversion specifiers missing their arguments or arguments missing their conversion specifiers
* Adds extra arguments to log messages in two places where it would make sense to log them

Two places I'm not sure about are

* `ERR("DSV: %p\n", graphics->dsv_format);` where I don't know if there's any value in logging the pointer value and maybe we should log `graphics->dsv_format->dxgi_format` or `graphics->dsv_format->vk_format` here instead (or both?)
* `RT_TRACE(" RDAT subobject: %s (type #%x).\n", debugstr_w(data->subobjects[j].name), data->subobjects[j].kind);` where the comment suggests that the strings might actually be ASCII but IIUC `name` is actually already post-conversion to `WCHAR`.

There's also another commit with two `NULL` checks added (for `extension_count`) where I found it weird that they are not checked before dereferencing. Finally, last commit attempts to ensure leading zeros are always printed when formatting the build commit message.

### DIFF
--- a/libs/vkd3d/device.c
+++ b/libs/vkd3d/device.c
@@ -3064,7 +3064,7 @@ static void d3d12_add_device_singleton(struct d3d12_device *device, LUID luid)
 
     if (!(e = vkd3d_malloc(sizeof(*e))))
     {
-        ERR("Failed to register device singleton for adapter.");
+        ERR("Failed to register device singleton for adapter.\n");
         return;
     }
     e->adapter_luid = luid;
@@ -3175,7 +3175,7 @@ HRESULT d3d12_device_get_scratch_buffer(struct d3d12_device *device, enum vkd3d_
 
     if (min_size > pool->block_size)
     {
-        FIXME("Requesting scratch buffer kind %u larger than limit (%"PRIu64" > %u). Expect bad performance.\n",
+        FIXME("Requesting scratch buffer kind %u larger than limit (%"PRIu64" > %"PRIu64"). Expect bad performance.\n",
                 kind, min_size, pool->block_size);
         return d3d12_device_create_scratch_buffer(device, kind, min_size, memory_types, scratch);
     }
@@ -4223,7 +4223,7 @@ static HRESULT STDMETHODCALLTYPE d3d12_device_CheckFeatureSupport(d3d12_device_i
             }
 
             data->Support = D3D12_PROTECTED_RESOURCE_SESSION_SUPPORT_FLAG_NONE;
-            TRACE("Protected resource session support %#x.", data->Support);
+            TRACE("Protected resource session support %#x.\n", data->Support);
             return S_OK;
         }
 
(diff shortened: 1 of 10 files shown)


libs/vkd3d/state.c
```diff
@@ -5548,7 +5548,7 @@ static void d3d12_pipeline_state_log_graphics_state(const struct d3d12_pipeline_
 
     if (graphics->dsv_format)
     {
-        ERR("DSV: %u\n", graphics->dsv_format);
+        ERR("DSV: %p\n", graphics->dsv_format);
```
HansKristian-Work, line 5551: Should be graphics->dsv_format->dxgi_format, and format should be "DSV: #%x".

libs/vkd3d/device.c
```diff
@@ -1029,7 +1029,7 @@ static HRESULT vkd3d_instance_init(struct vkd3d_instance *instance,
     application_info.engineVersion = vkd3d_get_vk_version();
     application_info.apiVersion = loader_version;
 
-    INFO("vkd3d-proton - build: %"PRIx64".\n", vkd3d_build);
+    INFO(vkd3d_build < (1ull << 60) ? "vkd3d-proton - build: %015"PRIx64".\n" : "vkd3d-proton - build: %016"PRIx64".\n", vkd3d_build);
```
HansKristian-Work, line 1032: This feels kinda goofy. Why doesn't just a plain %016"PRIx64" work as is?

HansKristian-Work (changes requested):
Thanks for the cleanup, just have a few nits.

libs/vkd3d/device.c
Saancreed: Ye, sorry about that. It's because after my previous change in https://github.com/HansKristian-Work/vkd3d-proton/commit/61b13cc2786cdeeb96232f0ba7e2dd8aa789e012, `vkd3d_build` will be either `0xfffffffffffffff` on clean trees or `0xfffffffffffffff0` on dirty ones and afaict always printing with `%016` would prepend a misleading `0` to the former case.

libs/vkd3d/state.c
Saancreed: Is `#%x` an actual format specifier I'm just unaware of or is it just supposed to be unmodified unsigned-int-as-hex with `#` prepended in front if it? I believe `%#x` would be a proper format (unsigned-int-as-hex-with-0x-in-front) but I see in a _lot_ of places `#%x` being used so I'll go with that one for now, though I can't guess the intent here.

libs/vkd3d/device.c
HansKristian-Work: Right, but if the hash is legitimately less than 1ull << 60, this will be funny. I think we need to fix the meson build to mark dirtyness a different way then.

HansKristian-Work (changes requested):
The two previous commits are fine as-is. Can you remove the last commit and start a new PR so we can hash it out there?

Saancreed:
> Can you remove the last commit and start a new PR so we can hash it out there?

Done, I'll submit the other PR in a moment.

PR #2020, opened 2024-06-19
Add vkd3d_counted_by

Joshua-Ashton:
Probably not too helpful right now, but probably good to do going forward.

Added it to the flexi-array members I found that had a member associated with their size, and checked the size was set before their use.


HansKristian-Work:
Seems fine, but I'll merge after tagging 2.13.

HansKristian-Work (approved):
Trivial enough

PR #2024, opened 2024-06-24
Implement frame rate limiter

doitsujin:
Fallout from #2014. Implementation matches that of DXVK.

The `SetTargetFrameRate` function is called with:
- A **positive** value if there's a DXVK app profile (or user config). Hopefully won't ever be relevant for us since this is only really used to work around broken games that don't understand the concept of high-refresh monitors and just break at >60 FPS.
- A **negative** value if the app is running in fullscreen mode (i.e. app calls `SetFullscreenState(TRUE)`) and presents with SyncInterval > 0. The target frame rate is equal to the refresh rate of the selected display mode, and the intention is to only engage the frame rate limiter if we can't do a mode switch and display at a higher rate. Again *mostly* relevant for older games, but also makes things like the "fullscreen" mode and refresh rate options in Horizon Zero Dawn work as intended.
- A value of 0.0 otherwise, in which case there is no limit.

The algorithm itself is fairly simple and intends to roughly emulate the behaviour of running on a lower refresh rate display:
- Let `Δt = 1 / |target_frame_rate|` and `t_f` be the time stamp at which `vkWaitForPresentKHR` completes.
- Each frame has a window `[t0..t1)` during which we want the frame to be presented, with `t1 = t0 + Δt`.
- if `t_f < t1`, the frame is on time and we advance the next frame's window by setting `t0_next = t1`.
- if also `t_f < t0`, sleep until `t0`.
- if `t_f >= t1`, the current frame was delayed and we restart the frame window by setting `t0_next = t_f + Δt`.

Sleeping needs to be reasonably accurate to not mess up frame pacing, which is why we intentionally ask to be woken up a few milliseconds early and busy-wait for the remainder of the duration time. The downside here is power consumption, but the only way to really improve on that would be to do the SpecialK thing of abusing the `monitorx` / `mwaitx` instruction on Ryzen CPUs, which isn't trivial since they require cpuid checks and operate in units of TSC ticks.

### DIFF
--- a/libs/vkd3d/swapchain.c
+++ b/libs/vkd3d/swapchain.c
@@ -78,6 +78,23 @@ struct present_wait_entry
     uint64_t begin_frame_time_ns;
 };
 
+#if defined(_WIN32)
+typedef UINT (WINAPI *PFN_NtDelayExecution) (BOOL, LARGE_INTEGER*);
+typedef UINT (WINAPI *PFN_NtQueryTimerResolution) (ULONG*, ULONG*, ULONG*);
+typedef UINT (WINAPI *PFN_NtSetTimerResolution) (ULONG, BOOL, ULONG*);
+#endif
+
+struct platform_sleep_state
+{
+#if defined(_WIN32)
+    PFN_NtDelayExecution NtDelayExecution;
+    PFN_NtQueryTimerResolution NtQueryTimerResolution;
+    PFN_NtSetTimerResolution NtSetTimerResolution;
+#endif
+
+    uint64_t sleep_threshold_ns;
+};
+
 struct dxgi_vk_swap_chain
 {
     IDXGIVkSwapChain2 IDXGIVkSwapChain_iface;
@@ -194,6 +211,20 @@ struct dxgi_vk_swap_chain
         uint64_t count;
         uint64_t time;
     } frame_statistics;
+
+    struct
+    {
+        pthread_mutex_t lock;
+
+        bool enable;
+        bool has_user_override;
+        uint64_t target_interval_ns;
+        uint64_t next_deadline_ns;
+        uint64_t frame_timestamps[16];
+        uint32_t frame_window_index;
+
+        struct platform_sleep_state sleep_state;
+    } frame_rate_limit;
 
     struct
     {
@@ -385,6 +416,8 @@ static void dxgi_vk_swap_chain_cleanup(struct dxgi_vk_swap_chain *chain)
     pthread_mutex_destroy(&chain->wait_thread.lock);
     pthread_cond_destroy(&chain->wait_thread.cond);
     vkd3d_free(chain->wait_thread.wait_queue);
+
+    pthread_mutex_destroy(&chain->frame_rate_limit.lock);
 
     vkd3d_native_sync_handle_destroy(chain->frame_latency_event);
     vkd3d_native_sync_handle_destroy(chain->frame_latency_event_internal);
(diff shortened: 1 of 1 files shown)


libs/vkd3d/swapchain.c
```diff
@@ -2526,6 +2596,111 @@ static void dxgi_vk_swap_chain_update_frame_statistics(struct dxgi_vk_swap_chain
...
+            chain->frame_rate_limit.frame_window_index += 1;
+
+            window_total_time_ns = current_time_ns - window_start_time_ns;
+
+            /* Enable the frame rate limiter if frames are being delivered faster than expected.
+             * Discard the first window worth of data since it may be inaccurate, especially if
+             * we do not hit the present_wait path. */
+            if (chain->frame_rate_limit.frame_window_index >= 2 * window_size)
+            {
+                chain->frame_rate_limit.enable = 103 * window_total_time_ns < 100 * window_size * chain->frame_rate_limit.target_interval_ns;
```
HansKristian-Work, line 2657: Is this meant to kick in if there is more than 3% error compared to target rate?

doitsujin: basically, yes.

libs/vkd3d/swapchain.c
```diff
@@ -2526,6 +2596,111 @@ static void dxgi_vk_swap_chain_update_frame_statistics(struct dxgi_vk_swap_chain
...
+                }
+            }
+        }
+
+        if (chain->frame_rate_limit.enable)
+        {
+            next_deadline_ns = chain->frame_rate_limit.next_deadline_ns;
+
+            if (chain->frame_rate_limit.next_deadline_ns + chain->frame_rate_limit.target_interval_ns > current_time_ns)
+                chain->frame_rate_limit.next_deadline_ns += chain->frame_rate_limit.target_interval_ns;
```
HansKristian-Work, line 2673: This path is triggered if frame is delivered within [t0, t1] region?

libs/vkd3d/swapchain.c
```diff
@@ -2526,6 +2596,111 @@ static void dxgi_vk_swap_chain_update_frame_statistics(struct dxgi_vk_swap_chain
...
+            else
+                chain->frame_rate_limit.next_deadline_ns = current_time_ns + chain->frame_rate_limit.target_interval_ns;
```
HansKristian-Work, line 2675: And this is the path for when frame is delivered late?
Would probably be good to comment this in more detail, since it's a bit hard to follow otherwise.

libs/vkd3d/swapchain.c
```diff
@@ -2526,6 +2596,111 @@ static void dxgi_vk_swap_chain_update_frame_statistics(struct dxgi_vk_swap_chain
...
+
+    pthread_mutex_unlock(&chain->frame_rate_limit.lock);
+
+    if (next_deadline_ns <= current_time_ns)
+        return;
+
+    /* Busy-wait for the last couple of milliseconds for accuracy, only use the
+     * platform's sleep function for durations longer than the given threshold. */
+    sleep_duration_ns = next_deadline_ns - current_time_ns;
+    sleep_threshold_ns = sleep_state->sleep_threshold_ns + sleep_duration_ns / 6;
```
HansKristian-Work, line 2687: Why the divide by 6?

libs/vkd3d/swapchain.c
HansKristian-Work: any particular reason 3% was chosen? Should probably make that a define.

HansKristian-Work (changes requested):
I grok the implementation and seems okay overall, but got some nits.

libs/vkd3d/swapchain.c
doitsujin: Fair, can add some comments.

libs/vkd3d/swapchain.c
```diff
@@ -2526,6 +2596,111 @@ static void dxgi_vk_swap_chain_update_frame_statistics(struct dxgi_vk_swap_chain
...
+            window_total_time_ns = current_time_ns - window_start_time_ns;
+
+            /* Enable the frame rate limiter if frames are being delivered faster than expected.
+             * Discard the first window worth of data since it may be inaccurate, especially if
+             * we do not hit the present_wait path. */
+            if (chain->frame_rate_limit.frame_window_index >= 2 * window_size)
+            {
+                chain->frame_rate_limit.enable = 103 * window_total_time_ns < 100 * window_size * chain->frame_rate_limit.target_interval_ns;
+
+                if (chain->frame_rate_limit.enable)
```
HansKristian-Work, line 2659: Do we want to engage this if vsync is off? Or will DXVK set target frame rate to 0.0 then?

doitsujin: DXVK handles the vsync off case.

libs/vkd3d/swapchain.c
doitsujin: The idea here is to just very conservatively estimate some sort of error on platform sleep to increase the likelyhood of waking up no later than `t0`. Essentially we expect it to be no worse than `a + b * duration`, where `a` is a function of the timer interval (extremely relevant on Windows) and `b` is just some arbitrary scaling factor that happens to work well in practice.

HansKristian-Work (approved):
LGTM (also very cursed)

PR #2028, opened 2024-06-26
Planar formats, part 2: Handle subsampled planes and plane view aspects

doitsujin:
This is a fairly invasive refactor, so I'm doing this as a separate PR.

The goal here is to unify all the different ways in which we compute the size of a subresource to make it a) more consistent and b) correct w.r.t. subsampled planes. Currently it's all over the place and largely broken.

Another change is that we now store the aspect flag that is used to create a view in the view info. This is important in cases where we need to reconstruct the exact subresource of the image that the view covers, since it can be something like `PLANE_1_BIT` even if the format's aspect mask is `COLOR_BIT`.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -3033,7 +3033,8 @@ static int d3d12_command_list_find_attachment_view(struct d3d12_command_list *li
         if (!list->rendering_info.dsv.imageView)
             return -1;
 
-        if (dsv->info.texture.miplevel_idx == subresource->mipLevel &&
+        if (dsv->info.texture.aspect_mask == subresource->aspectMask &&
+                dsv->info.texture.miplevel_idx == subresource->mipLevel &&
                 dsv->info.texture.layer_idx == subresource->baseArrayLayer &&
                 dsv->info.texture.layer_count == subresource->layerCount)
             return D3D12_SIMULTANEOUS_RENDER_TARGET_COUNT;
@@ -3050,7 +3051,8 @@ static int d3d12_command_list_find_attachment_view(struct d3d12_command_list *li
             if (list->rtvs[i].resource != resource)
                 continue;
 
-            if (rtv->info.texture.miplevel_idx == subresource->mipLevel &&
+            if (rtv->info.texture.aspect_mask == subresource->aspectMask &&
+                    rtv->info.texture.miplevel_idx == subresource->mipLevel &&
                     rtv->info.texture.layer_idx == subresource->baseArrayLayer &&
                     rtv->info.texture.layer_count == subresource->layerCount)
                 return i;
@@ -4152,7 +4154,7 @@ static void d3d12_command_list_emit_render_pass_transition(struct d3d12_command_
 
             if (rtv->resource->flags & VKD3D_RESOURCE_LINEAR_STAGING_COPY)
             {
-                vk_subresource_layers.aspectMask = rtv->format->vk_aspect_mask;
+                vk_subresource_layers.aspectMask = rtv->view->info.texture.aspect_mask;
                 vk_subresource_layers.mipLevel = rtv->view->info.texture.miplevel_idx;
                 vk_subresource_layers.baseArrayLayer = rtv->view->info.texture.layer_idx;
                 vk_subresource_layers.layerCount = rtv->view->info.texture.layer_count;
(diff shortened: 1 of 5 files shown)


libs/vkd3d/command.c
```diff
@@ -11268,7 +11270,7 @@ static bool vkd3d_rtv_and_aspects_fully_cover_resource(const struct d3d12_resour
         const struct vkd3d_view *view, VkImageAspectFlags clear_aspects)
 {
     /* Check that we're clearing all aspects. */
-    return view->format->vk_aspect_mask == clear_aspects &&
+    return resource->format->vk_aspect_mask == clear_aspects &&
```
HansKristian-Work, line 11269: Is this correct? If we have a planar image like NV12, and take a view of plane 0, we should never consider that fully-covered. I would expect to see view->format still being used here.

libs/vkd3d/resource.c
```diff
@@ -4769,6 +4769,7 @@ bool vkd3d_create_texture_view(struct d3d12_device *device, const struct vkd3d_t
     object->vk_image_view = vk_view;
     object->format = format;
     object->info.texture.vk_view_type = desc->view_type;
+    object->info.texture.aspect_mask = desc->aspect_mask;
```
HansKristian-Work, line 4784: Do we remember to hash the aspect mask anywhere?

libs/vkd3d/command.c
```diff
@@ -11568,18 +11568,21 @@ static void d3d12_command_list_clear_uav(struct d3d12_command_list *list,
...
         }
 
-        miplevel_idx = 0;
         layer_count = 1;
         pipeline = vkd3d_meta_get_clear_buffer_uav_pipeline(&list->device->meta_ops,
                 !args->has_view || args->u.view->format->type == VKD3D_FORMAT_TYPE_UINT,
                 !args->has_view);
         workgroup_size = vkd3d_meta_get_clear_buffer_uav_workgroup_size();
+
+        view_extent.width = resource->desc.Width;
```
HansKristian-Work, line 11569: This seems redundant. In line 11594, full_rect.right is overwritten based on byte_count. Seems slightly misleading to set view_extent.width to anything other than 0 here with a comment saying  it'll be overriden. Another idea might be to pull that branch into here. sampler_feedback_clear is only for images anyway.

libs/vkd3d/command.c
```diff
@@ -4101,16 +4100,11 @@ static void d3d12_command_list_flush_subresource_updates(struct d3d12_command_li
         const struct vkd3d_subresource_tracking *entry = &list->subresource_tracking[i];
 
         /* Entries will only ever have one aspect set, so this is fine */
-        plane_idx = d3d12_plane_index_from_vk_aspect((VkImageAspectFlagBits)entry->subresource.aspectMask);
```
HansKristian-Work, line 4098: The comment above is misleading now when these lines are removed.

HansKristian-Work (changes requested):
Minor nits, but overall very good refactor.

libs/vkd3d/command.c
doitsujin: This is actually precisely *why* this changed - the view format would have `COLOR`, but `clear_aspects` would also be `COLOR` whereas the resource aspects would be `PLANE_0 | PLANE_1`.

HansKristian-Work: But wouldn't the view aspects be PLANE_0 or PLANE_1 in that case? Taking a COLOR view of planar is not legal for render or sampling (without Ycbcr samplers).

HansKristian-Work: Ah ... Derp. Yes, now it makes sense.

HansKristian-Work: And to be sure, for depth-stencil, this won't cause any issues since it's not possible to view a subset of the aspects there?

libs/vkd3d/resource.c
HansKristian-Work: Only case I can think of is when creating 3 different R8_UNORM views of Y, Cb, and Cr of 420.

doitsujin: Right, need to do that.

libs/vkd3d/command.c
doitsujin: FWIW if I understand the purpose of this function correctly, we're checking here if we're discarding the *entire resource*, so even if we *could* create a depth-only view this should still be correct.

HansKristian-Work: Right.

libs/vkd3d/command.c
doitsujin: Cleaned this up a bit.

HansKristian-Work (approved):
LGTM

PR #2030, opened 2024-06-27
tests: stop skipping Task+Mesh DGC tests with RADV

hakzsam:
RADV is passing them now.

For reference https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/29935


doitsujin (approved):
Can merge once the RADV MR lands.

PR #2034, opened 2024-07-01
Validate that resolve format is actually supported before we try creating views.

### LINKED ISSUE #2032: Invalid Vulkan usage in test_multisample_resolve_strongly_typed
`$ VKD3D_TEST_FILTER=test_multisample_resolve_strongly_typed ./build/tests/d3d12`

```
test_multisample_resolve_strongly_typed:2029:Test 2: Todo: Got hr 0, expected E_INVALIDARG.
VUID-VkImageViewCreateInfo-image-01762(ERROR / SPEC): msgNum: 1695986743 - Validation Error: [ VUID-VkImageViewCreateInfo-image-01762 ] Object 0: handle = 0x8fd57200000000e5, type = VK_OBJECT_TYPE_IMAGE; | MessageID = 0x6516b437 | vkCreateImageView(): pCreateInfo->format VK_FORMAT_R16_UNORM is different from VkImage 0x8fd57200000000e5[] format (VK_FORMAT_R16_SFLOAT). Formats MUST be IDENTICAL unless VK_IMAGE_CREATE_MUTABLE_FORMAT BIT was set on image creation. The Vulkan spec states: If image was not created with the VK_IMAGE_CREATE_MUTABLE_FORMAT_BIT flag, or if the format of the image is a multi-planar format and if subresourceRange.aspectMask is VK_IMAGE_ASPECT_COLOR_BIT, format must be identical to the format used to create image (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-VkImageViewCreateInfo-image-01762)
    Objects: 1
        [0] 0x8fd57200000000e5, type: 10, name: NULL
d3d12: ../src/vulkan/runtime/vk_image.c:475: vk_image_view_init: Assertion `image_view->format == image->format' failed.
Aborted (core dumped)
```

The test currently asserts on Mesa/RADV.

HansKristian-Work:
Fix #2032.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -8618,6 +8618,30 @@ static const struct vkd3d_format *d3d12_command_list_get_resolve_format(struct d
     return vkd3d_format_from_d3d12_resource_desc(list->device, &dst_resource->desc, format);
 }
 
+static bool d3d12_resource_view_format_is_compatible(
+        const struct d3d12_resource *resource, const struct vkd3d_format *format)
+{
+    const struct vkd3d_format_compatibility_list *compat;
+    unsigned int i;
+
+    if (resource->format->vk_format == format->vk_format)
+        return true;
+
+    compat = &resource->format_compatibility_list;
+
+    /* Full mutable, we can cast to whatever we want. */
+    if (compat->format_count == 0)
+        return true;
+
+    for (i = 0; i < compat->format_count; i++)
+    {
+        if (compat->vk_formats[i] == format->vk_format)
+            return true;
+    }
+
+    return false;
+}
+
 enum vkd3d_resolve_image_path d3d12_command_list_select_resolve_path(struct d3d12_command_list *list,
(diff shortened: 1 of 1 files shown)


libs/vkd3d/command.c
```diff
@@ -8977,6 +8977,18 @@ static void d3d12_command_list_execute_resolve(struct d3d12_command_list *list,
         src_view_desc.image_usage = VK_IMAGE_USAGE_SAMPLED_BIT;
         src_view_desc.allowed_swizzle = true;
 
+        if (!d3d12_resource_view_format_is_compatible(dst_resource, dst_view_desc.format))
```
doitsujin, line 8980: `d3d12_command_list_select_resolve_path` is kind of intended to have checks like these, should probably just return `VKD3D_RESOLVE_IMAGE_PATH_UNSUPPORTED` if we end up with an edge case we can't support.

HansKristian-Work: Right ... I'll fix it.

PR #2053, opened 2024-07-31
Extend `ID3D12DXVKInteropDevice` into `ID3D12DXVKInteropDevice1` to add `VkCommandBuffer` interop

liam-middlebrook:
Add functionality for grabbing the `VkCommandBuffer` which is internal to a `ID3D12CommandList` object. This allows for applications to do interop using DX12 synchronization primitives but enqueue work onto queues which are otherwise not exposed through traditional DX12 APIs.

This will be used as a basis for `dxvk-nvapi` to create a DX12 offering of the `VK_NV_optical_flow` functionality. Tagging @jp7677 and @Saancreed from that project as any feedback received here will be relevant to a follow-up pull request I will be making there.

### Interface Naming
Should this be called `ID3D12DXVKInteropDevice1` ? It has a high overlap in functionality with `ID3D12DXVKInteropDevice` so it seemed appropriate for it to inherit the methods. Saancreed suggested via Discord that it may make sense to rename `ID3D12DXVKInteropDevice` to `ID3D12VKInteropDevice` (keeping the same GUID, thus preserving ABI).

### Scope of VkCommandBuffer retrieval
I chose to keep a simpler implementation where we can only retrieve the internal `VkCommandBuffer` from command lists which are reset with an interop-enabled `ID3D12CommandAllocator`. We could also in theory define some more fine-grained synchronization contracts so that the internal command buffer could be retrieved from non-interop command lists.

I don't see a pressing use-case for that to be needed right now, but I can think of plenty reasons why it may be attractive though. One may be to move the VK calls for `CubinComputeShader*` functionality back into dxvk-nvapi, where it can just record dispatches into the command-buffer directly rather than relying on vkd3d-proton to be updated in order to support new functions in that space. I'm unclear on what the maintenance burden would be for supporting this kind of thing (I know that vkd3d-proton does commandlist-internal work tracking and this would definitely complicate that if it had to account for a mix of internally+externally recorded work).

### DIFF
--- a/include/vkd3d_device_vkd3d_ext.idl
+++ b/include/vkd3d_device_vkd3d_ext.idl
@@ -56,6 +56,21 @@ interface ID3D12DXVKInteropDevice : IUnknown
 }
 
 [
+    uuid(902d8115-59eb-4406-9518-fe00f991ee65),
+    object,
+    local,
+    pointer_default(unique)
+]
+interface ID3D12DXVKInteropDevice1 : ID3D12DXVKInteropDevice
+{
+    HRESULT GetVulkanResourceInfo1(ID3D12Resource *resource, UINT64 *vk_handle, UINT64 *buffer_offset, VkFormat *format);
+    HRESULT CreateInteropCommandQueue(const D3D12_COMMAND_QUEUE_DESC *pDesc, UINT32 vk_queue_family_index, ID3D12CommandQueue **ppQueue);
+    HRESULT CreateInteropCommandAllocator(D3D12_COMMAND_LIST_TYPE type, UINT32 vk_queue_family_index, ID3D12CommandAllocator **ppAllocator);
+    HRESULT BeginVkCommandBufferInterop(ID3D12CommandList *pCmdList, VkCommandBuffer *pCommandBuffer);
+    HRESULT EndVkCommandBufferInterop(ID3D12CommandList *pCmdList);
+}
+
+[
     uuid(f3112584-41f9-348d-a59b-00b7e1d285d6),
     object,
     local,
--- a/libs/vkd3d/breadcrumbs.c
+++ b/libs/vkd3d/breadcrumbs.c
@@ -599,21 +599,24 @@ void vkd3d_breadcrumb_tracer_report_device_lost(struct vkd3d_breadcrumb_tracer *
     if (device->vk_info.NV_device_diagnostic_checkpoints)
     {
         /* vkGetQueueCheckpointDataNV does not require us to synchronize access to the queue. */
-        queue_family_info = d3d12_device_get_vkd3d_queue_family(device, D3D12_COMMAND_LIST_TYPE_DIRECT);
+        queue_family_info = d3d12_device_get_vkd3d_queue_family(device, D3D12_COMMAND_LIST_TYPE_DIRECT, VK_QUEUE_FAMILY_IGNORED);
+
         for (i = 0; i < queue_family_info->queue_count; i++)
         {
             vk_queue = queue_family_info->queues[i]->vk_queue;
             vkd3d_breadcrumb_tracer_report_queue_nv(tracer, device, vk_queue);
         }
 
-        queue_family_info = d3d12_device_get_vkd3d_queue_family(device, D3D12_COMMAND_LIST_TYPE_COMPUTE);
+        queue_family_info = d3d12_device_get_vkd3d_queue_family(device, D3D12_COMMAND_LIST_TYPE_COMPUTE, VK_QUEUE_FAMILY_IGNORED);
+
         for (i = 0; i < queue_family_info->queue_count; i++)
         {
             vk_queue = queue_family_info->queues[i]->vk_queue;
             vkd3d_breadcrumb_tracer_report_queue_nv(tracer, device, vk_queue);
         }
 
-        queue_family_info = d3d12_device_get_vkd3d_queue_family(device, D3D12_COMMAND_LIST_TYPE_COPY);
+        queue_family_info = d3d12_device_get_vkd3d_queue_family(device, D3D12_COMMAND_LIST_TYPE_COPY, VK_QUEUE_FAMILY_IGNORED);
+
         for (i = 0; i < queue_family_info->queue_count; i++)
         {
             vk_queue = queue_family_info->queues[i]->vk_queue;
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -2629,19 +2629,33 @@ static CONST_VTBL struct ID3D12CommandAllocatorVtbl d3d12_command_allocator_vtbl
 };
 
 struct vkd3d_queue_family_info *d3d12_device_get_vkd3d_queue_family(struct d3d12_device *device,
-        D3D12_COMMAND_LIST_TYPE type)
+        D3D12_COMMAND_LIST_TYPE type,
+        uint32_t vk_family_index)
 {
-    switch (type)
+    if (vk_family_index == VK_QUEUE_FAMILY_IGNORED)
     {
-        case D3D12_COMMAND_LIST_TYPE_DIRECT:
-            return device->queue_families[VKD3D_QUEUE_FAMILY_GRAPHICS];
-        case D3D12_COMMAND_LIST_TYPE_COMPUTE:
-            return device->queue_families[VKD3D_QUEUE_FAMILY_COMPUTE];
-        case D3D12_COMMAND_LIST_TYPE_COPY:
-            return device->queue_families[VKD3D_QUEUE_FAMILY_TRANSFER];
-        default:
-            FIXME("Unhandled command list type %#x.\n", type);
-            return device->queue_families[VKD3D_QUEUE_FAMILY_GRAPHICS];
+        switch (type)
+        {
+            case D3D12_COMMAND_LIST_TYPE_DIRECT:
+                return device->queue_families[VKD3D_QUEUE_FAMILY_GRAPHICS];
+            case D3D12_COMMAND_LIST_TYPE_COMPUTE:
+                return device->queue_families[VKD3D_QUEUE_FAMILY_COMPUTE];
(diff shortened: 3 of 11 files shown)


libs/vkd3d/device_vkd3d_ext.c
```diff
@@ -406,12 +406,22 @@ static HRESULT STDMETHODCALLTYPE d3d12_dxvk_interop_device_GetVulkanResourceInfo
...
         *buffer_offset = 0;
     }
 
     return S_OK;
 }
 
+static HRESULT STDMETHODCALLTYPE d3d12_dxvk_interop_device_GetVulkanResourceInfo(d3d12_dxvk_interop_device_iface *iface,
+        ID3D12Resource *resource, UINT64 *vk_handle, UINT64 *buffer_offset)
+{
+	return d3d12_dxvk_interop_device_GetVulkanResourceInfo1(iface, resource, vk_handle, buffer_offset, NULL);
```
HansKristian-Work, line 422: Bad intent.

libs/vkd3d/device_vkd3d_ext.c
```diff
@@ -434,7 +444,84 @@ static HRESULT STDMETHODCALLTYPE d3d12_dxvk_interop_device_UnlockCommandQueue(d3
...
+    struct d3d12_command_list *cmd_list;
+    cmd_list = d3d12_command_list_from_iface(pCmdList);
+
+    if (!cmd_list)
+        return E_INVALIDARG;
+
+    /* Only allow this on command-lists associated to allocators with
+     * vk_interop set. Otherwise barriers would be a lot more complex with
+     * this external command-list API. */
+    if (!cmd_list->allocator->vk_interop)
```
HansKristian-Work, line 516: I guess using that bool to validate this is fine ...

HansKristian-Work (changes requested):
The headers for commits should be vkd3d: not command:.

As for the name of the interface, I don't really care if it's ID3D12VKInterop or not, that's more up to @doitsujin.

HansKristian-Work:
> I'm unclear on what the maintenance burden would be for supporting this kind of thing (I know that vkd3d-proton does commandlist-internal work tracking and this would definitely complicate that if it had to account for a mix of internally+externally recorded work).

I think this could work if there's a "begin interop" where we effectively invalidate all state in the command list (d3d12_invalidate_all_state), and flush any pending tracking. That basically means doing work that Close command list does.

libs/vkd3d/command.c
```diff
@@ -2763,13 +2781,17 @@ static HRESULT d3d12_command_allocator_init(struct d3d12_command_allocator *allo
 
     allocator->current_command_list = NULL;
 
+    allocator->vk_interop = vk_interop;
```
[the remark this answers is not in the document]
liam-middlebrook: The thought here was to specifically note whether or not a command allocator was going to be used for interop. So that the interop functions wouldn't be used with non-interop enabled allocators. This extra validation step isn't strictly necessary, so I'm happy to get rid of it if you prefer.

libs/vkd3d/command.c
```diff
@@ -2785,7 +2807,40 @@ HRESULT d3d12_command_allocator_create(struct d3d12_device *device,
...
+    TRACE("Created command allocator %p.\n", object);
+
+    *allocator = object;
+
+    return S_OK;
+}
+
+HRESULT d3d12_command_allocator_create_vk(struct d3d12_device *device,
+        uint32_t vk_family_index,
+        D3D12_COMMAND_LIST_TYPE type,
```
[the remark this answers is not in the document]
liam-middlebrook: Thanks, that makes sense. That simplifies the change a lot as well.

libs/vkd3d/command.c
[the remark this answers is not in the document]
liam-middlebrook: On further thought (and reading your comments below) this is unnecessary. I think a model described below (with command list state invalidation) makes more sense, and is a bit more forward-thinking.

tests/d3d12_dxvk_interop_device.c
```diff
@@ -0,0 +1,157 @@
...
+        ok(hr == S_OK, "ID3D12Device_CreateCommandList failed %#x.\n", hr);
+        destroy_test_context(&context);
+        return;
+    }
+
+    if (FAILED(ID3D12DXVKInteropDevice1_BeginVkCommandBufferInterop(interop_device,
+        (ID3D12CommandList*)command_list,
+        &vkCmdBuf)))
+    {
+        ok(hr == S_OK, "ID3D12DXVKInteropDevice1_GetVulkanCmdHandles failed %#x.\n", hr);
```
Saancreed, line 104: Shouldn't this now be something like the following?

```suggestion
    if (FAILED(hr = ID3D12DXVKInteropDevice1_BeginVkCommandBufferInterop(interop_device,
        (ID3D12CommandList*)command_list,
        &vkCmdBuf)))
    {
        ok(hr == S_OK, "ID3D12DXVKInteropDevice1_BeginVkCommandBufferInterop failed %#x.\n", hr);
```

libs/vkd3d/device_vkd3d_ext.c
```diff
@@ -434,7 +446,61 @@ static HRESULT STDMETHODCALLTYPE d3d12_dxvk_interop_device_UnlockCommandQueue(d3
...
+static HRESULT STDMETHODCALLTYPE d3d12_dxvk_interop_device_BeginVkCommandBufferInterop(d3d12_dxvk_interop_device_iface *iface,
+        ID3D12CommandList *pCmdList, VkCommandBuffer *cmdBuf)
+{
+    struct d3d12_command_list *cmd_list;
+    cmd_list = d3d12_command_list_from_iface(pCmdList);
+
+    if (!cmd_list)
+        return E_INVALIDARG;
+
+    d3d12_command_list_invalidate_all_state(cmd_list);
```
HansKristian-Work, line 496: This isn't quite enough to make the interop work, but it's a start. We can probably flesh that out after this lands. Just add a FIXME comment here for now.

tests/d3d12_dxvk_interop_device.c
```diff
@@ -0,0 +1,157 @@
...
+
+    if (FAILED(hr = ID3D12Device_CreateCommandList(device, 0, D3D12_COMMAND_LIST_TYPE_DIRECT,
+            allocator, NULL, &IID_ID3D12GraphicsCommandList, (void **)&command_list)))
+    {
+        ok(hr == S_OK, "ID3D12Device_CreateCommandList failed %#x.\n", hr);
+        destroy_test_context(&context);
+        return;
+    }
+
+    if (FAILED(ID3D12DXVKInteropDevice1_BeginVkCommandBufferInterop(interop_device,
```
HansKristian-Work, line 100: hr is not set here.

tests/d3d12_dxvk_interop_device.c
```diff
@@ -0,0 +1,157 @@
...
+                D3D12_RESOURCE_STATE_COPY_DEST);
+    upload_buffer_data(buffer, 0, sizeof(data_values), data_values, queue, command_list);
+    reset_command_list(command_list, allocator);
+
+
+    // Export it to VK and use VK functions to manipulate it
+    ID3D12DXVKInteropDevice1_GetVulkanResourceInfo1(interop_device,
+        buffer,
+        (uint64_t*)&vkBuffer,
+        &offset,
```
HansKristian-Work, line 124: This needs to be UINT64 to get the 8 byte alignment on 32-bit. Both the handle and offset it seems.

libs/vkd3d/breadcrumbs.c
```diff
@@ -599,21 +599,30 @@ void vkd3d_breadcrumb_tracer_report_device_lost(struct vkd3d_breadcrumb_tracer *
     if (device->vk_info.NV_device_diagnostic_checkpoints)
     {
         /* vkGetQueueCheckpointDataNV does not require us to synchronize access to the queue. */
-        queue_family_info = d3d12_device_get_vkd3d_queue_family(device, D3D12_COMMAND_LIST_TYPE_DIRECT);
+        queue_family_info = d3d12_device_get_vkd3d_queue_family(device,
+                                D3D12_COMMAND_LIST_TYPE_DIRECT,
```
HansKristian-Work, line 603: Just keep this on one line, the indents here are a bit goofy. Ditto the similar code below.

include/vkd3d_device_vkd3d_ext.idl
```diff
@@ -55,6 +55,20 @@ interface ID3D12DXVKInteropDevice : IUnknown
...
+    object,
+    local,
+    pointer_default(unique)
+]
+interface ID3D12DXVKInteropDevice1 : ID3D12DXVKInteropDevice
+{
+    HRESULT GetVulkanResourceInfo1(ID3D12Resource *resource, UINT64 *vk_handle, UINT64 *buffer_offset, VkFormat *format);
+    HRESULT CreateInteropCommandQueue(const D3D12_COMMAND_QUEUE_DESC *pDesc, UINT32 vk_queue_family_index, ID3D12CommandQueue **ppQueue);
+    HRESULT CreateInteropCommandAllocator(D3D12_COMMAND_LIST_TYPE type, UINT32 vk_queue_family_index, ID3D12CommandAllocator **ppAllocator);
+    HRESULT BeginVkCommandBufferInterop(ID3D12CommandList *pCmdList, VkCommandBuffer *pCommandBuffer);
```
HansKristian-Work, line 69: I think we need an EndVkCommandBufferInterop as well, just in case we need to suspect certain things in the future like conditional rendering and queries. For now, just implement it as a noop.

HansKristian-Work (changes requested):
The overall approach is reasonable I think and is not that invasive. Mostly nits now.

HansKristian-Work (changes requested):
This LGTM now, but needs to change the `command:` to `vkd3d:` on the commits as requested earlier.

HansKristian-Work (approved):
LGTM. Passes test suite on both NV and RADV.

PR #2056, opened 2024-08-05
Implement depth bias control

doitsujin:
Fairly basic, includes a test that fails w/o the extension on AMD but passes with.

Like DXVK we also enable `exactDepthBias` if supported by the implementation since this matches AMD/Nvidia/WARP D3D behaviour, but the test does not explicitly check this. I can add stricter validation if desired. Without this, Nvidia behaves differently with `UNORM` DSV formats, no difference on RADV.

For the unknown DSV case, we can just abuse dynamic depth bias to set up the correct depth bias representation. As a reminder, pipelines with and without dynamic depth bias behave the same way on native D3D12, and any dynamic state is overwritten with PSO state at the time the pipeline is bound, so this should be correct. Since depth bias is part of pre-raster state, we cannot use GPL functionality for this.

For the no DSV case, there's essentially nothing we can do to match native behaviour. We already have a test for this, the only reliable workaround would be to bind some sort of fake D32 DSV which isn't really an option.

Unfortunately this is going to invalidate a bunch of fossizlize dbs (at least pipelines with depth bias enabled); don't think we can do an awful lot about that short of relying on RADV implementation details and just not passing the struct unless we use an emulated format. I can add that check if desired.


tests/d3d12_depth_stencil.c
```diff
@@ -2409,3 +2409,176 @@ void test_depth_bias_behaviour(void)
...
+        bool expect_pass;
+    }
+    tests[] =
+    {
+        { DXGI_FORMAT_D16_UNORM,            0x3c802000, 0x3c800000, 0, false },
+        { DXGI_FORMAT_D16_UNORM,            0x3c802000, 0x3c800000, 1, true  },
+        { DXGI_FORMAT_D16_UNORM,            0x3c820000, 0x3c800000, 1, false },
+
+        { DXGI_FORMAT_D24_UNORM_S8_UINT,    0x3c800020, 0x3c800000, 0, false },
+        { DXGI_FORMAT_D24_UNORM_S8_UINT,    0x3c800020, 0x3c800000, 1, true  },
```
HansKristian-Work, line 2450: Isn't this slightly above 1.0? How does that work?

HansKristian-Work: To properly test D24_UNORM behavior on emulated format, shouldn't we use smaller FP values here?

HansKristian-Work:
> Unfortunately this is going to invalidate a bunch of fossizlize dbs (at least pipelines with depth bias enabled); don't think we can do an awful lot about that short of relying on RADV implementation details and just not passing the struct unless we use an emulated format. I can add that check if desired.

Depth-bias is a fairly limited number of pipelines, so I don't mind. It'd be different if we invalidated all pipelines.

libs/vkd3d/vkd3d_private.h
```diff
@@ -5569,6 +5570,21 @@ static inline unsigned int d3d12_resource_get_sub_resource_count(const struct d3
...
+static inline void vkd3d_get_depth_bias_representation(VkDepthBiasRepresentationInfoEXT *info,
+        const struct d3d12_device *device, const struct vkd3d_format *dsv_format)
+{
+    memset(info, 0, sizeof(*info));
+    info->sType = VK_STRUCTURE_TYPE_DEPTH_BIAS_REPRESENTATION_INFO_EXT;
+    info->depthBiasRepresentation = VK_DEPTH_BIAS_REPRESENTATION_LEAST_REPRESENTABLE_VALUE_FORMAT_EXT;
+    info->depthBiasExact = device->device_info.depth_bias_control_features.depthBiasExact;
+
+    /* Checking only strongly typed formats should work here since we take the
+     * format from the DSV or PSO desc, where typelesss formats are not allowed */
```
HansKristian-Work, line 5582: yessssss (typo on typelesss).

HansKristian-Work (changes requested):
Implementation looks solid overall. Some nits.

tests/d3d12_depth_stencil.c
doitsujin: The numbers are all around 1/64 actually. Will add a comment.

libs/vkd3d/vkd3d_private.h
doitsujin: fixed.

tests/d3d12_depth_stencil.c
HansKristian-Work: Ah, right. I looked at the wrong thing in the FP32 calculator :frog:

HansKristian-Work (approved):
LGTM

PR #2059, opened 2024-08-08
Rework check_multisample_quality_levels

doitsujin:
Fixes some oversights and misleading warnings. We *generally* got the correct return values regardless (minus sparse), but the tiling mode inconsistencies have the potential to become a real problem.

I changed the helper function to use `vkGetPhysicalDeviceImageFormatProperties2` to make it more readable.

I also did some further testing on Windows to check if there's any correlation between `SampleCount > 1` being supported and `D3D12_FORMAT_SUPPORT1_MULTISAMPLE_*` being exposed, and it's all over the place and different across implementations, so our implementation should be more or less fine as-is. The only thing we're really getting wrong in `check_format_support` right now is typeless formats.


HansKristian-Work (approved):
LGTM

Issue #2071, opened 2024-08-16
vkd3d seems to be ignoring maxTexelBufferElements in some games, leading to visual corruption


pzanoni-intel:
Hello

Original bug report: https://gitlab.freedesktop.org/mesa/mesa/-/issues/9963

On Intel graphics with the Mesa drivers we're observing that some games such as Marvel's Spider-Man Remastered and Assassin's Creed: Valhalla seem to be emitting `vkGetDescriptorEXT()` calls that ignore `VUID-VkDescriptorGetInfoEXT-type-09427`, which relates to `VkPhysicalDeviceLimits::maxTexelBufferElements`.

https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkDescriptorGetInfoEXT.html#VUID-VkDescriptorGetInfoEXT-type-09427

Since `vkGetDescriptorEXT()` doesn't have a way to return errors, I tried to change anv to completely disable VK_EXT_descriptor_buffer. With this, we have the same kind of problem, but now with `vkCreateBufferView()` calls violating VkBufferViewCreateInfo-range-00930, which is basically the same thing.

https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkBufferViewCreateInfo.html#VUID-VkBufferViewCreateInfo-range-00930

Now, `vkCreateBufferView()` is able to return errors, but if I make it return `VK_ERROR_VALIDATION_FAILED_EXT` on the relevant cases then Spider-Man GPU hangs on the main menu. But an interesting thing is that, when I do this, the white parts of his uniform (the spider in his chest and the back of his hands), which previously showed rendering corruption, are now completely black. So this change really seems to have an effect on the current visual corruption.

Please notice: in Release builds the anv driver doesn't do any checking related to this at all, but as a result we end up programming the hardware with wrong values. In debug builds we catch this with an assertion deep in our call stack:

https://gitlab.freedesktop.org/mesa/mesa/-/blob/main/src/intel/isl/isl_surface_state.c?ref_type=heads#L997

... as a result, the driver ignores the errors. On the other hand, both games show some rendering corruption. Still, we shouldn't be issuing invalid Vulkan commands.

Is there something that can be done about this issue in the vkd3d-proton layer?

Thanks in advance!

doitsujin:
What kind of numbers are we talking about here? D3D12 requirements are at least 2^27 texels regardless of format.

[...]

PR #2074, opened 2024-08-21
tests: Test very large texel buffer views.

doitsujin:
Fallout from #2071.

This essentially tests whether we can safely create and use texel buffer views up to 2^29 texels for smaller formats, which is roughly the upper limit of what the games from the issue are relying on. Passes on AMD and Nvidia as well as WARP, no idea how this behaves on Intel.


HansKristian-Work (approved):
Passes on d3d12 Intel iGPU too.

PR #2075, opened 2024-08-21
Async compute present - attempt to squeeze a little perf out of AMD

HansKristian-Work:


HansKristian-Work:
Obsolete. Didn't really fix FSR3 since it relies on graphics work too.

PR #2076, opened 2024-08-21
Implement staggered submissions

doitsujin:
Not sure if we should do this first or #2075 first, rebasing in either direction isn't entirely trivial but not *that* bad either (I have a branch of this on top of 2075 somewhere).


libs/vkd3d/command.c
```diff
@@ -18059,6 +18071,63 @@ static VkResult d3d12_command_queue_submit_split_locked(struct d3d12_device *dev
...
+
+    command_queue->last_submission_timeline_value = 0;
+}
+
+static bool d3d12_command_queue_needs_staggered_submissions_locked(struct d3d12_command_queue *command_queue)
+{
+    static const uint64_t inactive_threshold_ns = 1000000000ull; /* 1s */
+    uint64_t current_time_ns;
+
+    if (command_queue->vkd3d_queue->virtual_queue_count == 1)
```
HansKristian-Work, line 18100: I wonder if a less risky implementation is to only consider a queue high priority if it has a swapchain associated with it. It's the only scenario I can think of where we actually want to go to this length.

HansKristian-Work: The current impl will also trigger for COMPUTE. I guess it will only trigger for cases where COMPUTE + COPY queues is more than 4. But then again, staggering work in async compute / copy is certainly less bad than direct queue.

libs/vkd3d/command.c
```diff
@@ -18643,6 +18765,7 @@ static void *d3d12_command_queue_submission_worker_main(void *userdata)
 
         case VKD3D_SUBMISSION_WAIT:
             VKD3D_REGION_BEGIN(queue_wait);
+            d3d12_command_queue_wait_staggered_submission(queue);
```
HansKristian-Work, line 18768: As mentioned on Discord, I think if we're staggered, we should wait on the fence on CPU.

libs/vkd3d/command.c
```diff
@@ -18059,6 +18071,63 @@ static VkResult d3d12_command_queue_submit_split_locked(struct d3d12_device *dev
...
+    VkResult vr;
+
+    if (!command_queue->last_submission_timeline_value)
+        return;
+
+    memset(&wait_info, 0, sizeof(wait_info));
+    wait_info.sType = VK_STRUCTURE_TYPE_SEMAPHORE_WAIT_INFO;
+    wait_info.semaphoreCount = 1;
+    wait_info.pSemaphores = &command_queue->vkd3d_queue->submission_timeline;
+    wait_info.pValues = &command_queue->last_submission_timeline_value;
```
HansKristian-Work, line 18087: Did we test if a full wait-for-idle is really necessary?

libs/vkd3d/command.c
```diff
@@ -18123,84 +18165,164 @@ static void d3d12_command_queue_execute(struct d3d12_command_queue *command_queu
...
+
+                for (i = 0; i < num_submits; i++)
+                    submit_desc[i].pNext = &latency_submit_present_info;
+            }
+
+            if (low_latency_swapchain)
+                dxgi_vk_swap_chain_decref(low_latency_swapchain);
+        }
+
+        if (split_submissions)
```
HansKristian-Work, line 18292: Should not hit this path if submit_count > 1.

libs/vkd3d/command.c
doitsujin: I actually did experiment with that last night and it did work, yes.

libs/vkd3d/command.c
doitsujin: is this fine even if we have the transition command buffer in there? I'm not sure what the RADV problem we're working around there is exactly.

HansKristian-Work: Ah, right ... Well, maybe just leave it as-is. It's there to workaround async compute DGCC issues on RADV where if there's more than one command buffer in the submit bad things happen.

libs/vkd3d/command.c
doitsujin: FWIW I added logging so we can easily see if we hit this path when we don't want to; given that queue priorities don't seem to be used much anyway this probably won't happen very often anyway.

libs/vkd3d/command.c
```diff
@@ -17606,11 +17608,42 @@ static CONST_VTBL struct ID3D12CommandQueueVtbl d3d12_command_queue_vtbl =
...
+    for (i = 0; i < command_queue->vkd3d_queue->command_queue_count; i++)
+    {
+        struct d3d12_command_queue *q = command_queue->vkd3d_queue->command_queues[i];
+
+        if (q == command_queue)
+            continue;
+
+        /* If any other virtual queue is actively doing submissions, resolve waits
+         * on the CPU in order to avoid delays caused by false dependencies. */
+        queue_submit_time_ns = vkd3d_atomic_uint64_load_explicit(&q->last_submission_time_ns, vkd3d_memory_order_relaxed);
```
HansKristian-Work, line 17630: Need explicit alignment for last_submission_time_ns on 32-bit.

libs/vkd3d/command.c
```diff
@@ -2678,16 +2680,40 @@ struct vkd3d_queue *d3d12_device_allocate_vkd3d_queue(struct vkd3d_queue_family_
...
     return queue;
 }
 
-void d3d12_device_unmap_vkd3d_queue(struct vkd3d_queue *queue)
+void d3d12_device_unmap_vkd3d_queue(struct vkd3d_queue *queue, struct d3d12_command_queue *command_queue)
 {
+    size_t i;
+
     pthread_mutex_lock(&queue->mutex);
     queue->virtual_queue_count--;
```
HansKristian-Work, line 2704: Do we need both virtual_queue_count and command_queue_count?

HansKristian-Work: It might make sense for the async compute present path since that branch is creating a virtual queue without the associated d3d12 command queue, so maybe it makes sense to keep the distinction?

doitsujin: We can allocate internal queues as well in which case there would be no d3d12 command queue, so unless we want to deal with random null pointers in the array this is kind of needed.

HansKristian-Work: Very well.

libs/vkd3d/command.c
doitsujin: made it `UINT64`.

libs/vkd3d/command.c
```diff
@@ -18168,6 +18168,9 @@ static bool d3d12_command_queue_needs_staggered_submissions_locked(struct d3d12_
     if (command_queue->vkd3d_queue->command_queue_count == 1)
         return false;
 
+    if (vkd3d_config_flags & VKD3D_CONFIG_FLAG_NO_STAGGERED_SUBMIT)
```
HansKristian-Work, line 18171: I think we'll want to flip this into opt-in until the regression in Wukong is solved?

doitsujin: done.

HansKristian-Work (approved):
LGTM

PR #2077, opened 2024-08-22
Workaround RADV 24.1.6 bug causing vertex explosions.

HansKristian-Work:


doitsujin (approved):
Ah well, at least we should be able to remove this once it's fixed and shipping in stable mesa.

PR #2079, opened 2024-08-22
Enable reporting of format warnings in CI builds and fix some of them

Saancreed:
~Draft for now because there are still some dragons hiding in the land of tests and programs. I'll try to get to them tomorrow.~


Saancreed:
Okay, I think that's it. `-Wno-format-truncation` is there for warnings generated by calls to `snprintf` to potentially-too-small buffers. Cases related to shader overrides can be silenced by doing `if (snprintf(…) >= ARRAY_SIZE(…)) { ERR(…); return false; }` but cases related to cache path are more tricky so I just silenced these.

There are still two `format not a string literal, argument types not checked [-Wformat-nonliteral]` reported in `vkd3d-rs-parse` but I don't see a way to resolve this cleanly so maybe the correct way to go would be to add `-Wno-format-nonliteral` as well.

HansKristian-Work (approved):

PR #2081, opened 2024-08-23
Add GPU cost heuristic for staggered submissions

doitsujin:
Doesn't fix all our problems with staggered submissions, but batches a bunch of submissions (command buffers just doing a single clear / copy or literally just doing barriers without any GPU work). This helps FF16 a bit as well.

Basic idea here is to assume that any draw / dispatch is expensive enough to warrant a dedicated submission, but copies, clears etc typically aren't unless there's a whole bunch of them.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -2018,6 +2018,7 @@ static HRESULT d3d12_command_allocator_allocate_command_buffer(struct d3d12_comm
     }
 
     list->cmd.iteration_count = 1;
+    list->cmd.estimated_cost = 0;
 
 #ifdef VKD3D_ENABLE_BREADCRUMBS
     if (vkd3d_config_flags & VKD3D_CONFIG_FLAG_BREADCRUMBS)
@@ -2046,6 +2047,10 @@ static void d3d12_command_list_begin_new_sequence(struct d3d12_command_list *lis
 
     if (list->cmd.iteration_count >= VKD3D_MAX_COMMAND_LIST_SEQUENCES)
         return;
+
+    assert(list->cmd.iteration_count);
+    list->cmd.iterations[list->cmd.iteration_count - 1].estimated_cost = list->cmd.estimated_cost;
+    list->cmd.estimated_cost = 0;
 
     iteration = &list->cmd.iterations[list->cmd.iteration_count];
     command_buffer_info.sType = VK_STRUCTURE_TYPE_COMMAND_BUFFER_ALLOCATE_INFO;
@@ -3654,6 +3659,9 @@ static void d3d12_command_list_load_attachment(struct d3d12_command_list *list, 
     VkExtent3D view_extent;
     VkAccessFlags2 access;
     bool clear_op;
+
+    /* This does not get called with LOAD_OP_LOAD */
+    list->cmd.estimated_cost += VKD3D_COMMAND_COST_LOW;
 
     memset(initial_layouts, 0, sizeof(initial_layouts));
     memset(final_layouts, 0, sizeof(final_layouts));
@@ -5390,6 +5398,9 @@ static HRESULT STDMETHODCALLTYPE d3d12_command_list_Close(d3d12_command_list_ifa
         WARN("Failed to end command buffer, vr %d.\n", vr);
         return hresult_from_vk_result(vr);
     }
+
+    assert(list->cmd.iteration_count);
+    list->cmd.iterations[list->cmd.iteration_count - 1].estimated_cost = list->cmd.estimated_cost;
 
     if (list->allocator)
     {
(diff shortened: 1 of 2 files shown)


libs/vkd3d/command.c
```diff
@@ -17334,11 +17391,15 @@ static void STDMETHODCALLTYPE d3d12_command_queue_ExecuteCommandLists(ID3D12Comm
         {
             if (cmd_list->cmd.iterations[iter].vk_init_commands)
             {
+                cmd_cost[j] = VKD3D_COMMAND_COST_HIGH;
```
HansKristian-Work, line 17394: Do we really do significant work in init command lists?

libs/vkd3d/command.c
```diff
@@ -17334,11 +17391,15 @@ static void STDMETHODCALLTYPE d3d12_command_queue_ExecuteCommandLists(ID3D12Comm
...
+
                 buffer = &buffers[j++];
                 buffer->sType = VK_STRUCTURE_TYPE_COMMAND_BUFFER_SUBMIT_INFO;
                 buffer->commandBuffer = cmd_list->cmd.iterations[iter].vk_init_commands;
             }
 
+            cmd_cost[j] = cmd_list->cmd.estimated_cost;
```
HansKristian-Work, line 17401: Shouldn't the cost be per iteration?

libs/vkd3d/command.c
doitsujin: If there's ExecuteIndirect memes involved then apparently yes. Not fusing init commands destroys frame pacing in FF16 at least.

libs/vkd3d/command.c
doitsujin: Technically yes, but it just seems like extra complexity that doesn't really make a difference in terms of decision making right now anyway.

libs/vkd3d/vkd3d_private.h
```diff
@@ -2796,6 +2801,10 @@ struct d3d12_command_list_sequence
     unsigned int active_non_inline_running_queries;
     bool uses_dgc_compute_in_async_compute;
 
+    /* Number of draws, dispatches, copies etc. Used to fuse barrier-only
+     * command buffers for staggered submissions. */
+    uint32_t estimated_cost;
```
HansKristian-Work, line 2806: I think we'll want a cost estimate for the init command list as well. Preprocess generation work can be considered LOW cost for example, but mostly it's rather trivial stuff I think that shouldn't be considered to have cost at all I think?

HansKristian-Work (changes requested):
comments

libs/vkd3d/command.c
HansKristian-Work: The added complexity seems fairly minimal, this could just change to cmd_list->cmd.iterations[iter].estimated_cost instead, so I think it's worth doing it.

libs/vkd3d/command.c
HansKristian-Work: Do we know what happens in the init command lists? I can't think of anything expensive unless it's doing DGC.

libs/vkd3d/command.c
HansKristian-Work: Maybe a simpler solution is to consider the full cost of the ID3D12CommandList (sans first init command buffer), and set costs to 0 for everything except the last VkCommandBuffer we add.

HansKristian-Work: Then we could have a init cost and total cost for normal commands.

HansKristian-Work: Also missing 0 cost in
```
    if (command_queue->vkd3d_queue->barrier_command_buffer)
    {
        /* Append a full GPU barrier between submissions.
         * This command buffer is SIMULTANEOUS_BIT. */
        buffer = &buffers[j++];
        buffer->sType = VK_STRUCTURE_TYPE_COMMAND_BUFFER_SUBMIT_INFO;
        buffer->commandBuffer = command_queue->vkd3d_queue->barrier_command_buffer;
    }
```

libs/vkd3d/command.c
HansKristian-Work: The only things I can find that would use init command buffer is:

- Query pool reset (should be virtually free)
- DGC fixups (should still be cheap)
- Emitting predicated commands

Could there an off-by-one error where the cost metric is wrong and we confuse init buffer for a real command buffer?

libs/vkd3d/command.c
```diff
@@ -17279,6 +17330,12 @@ static void STDMETHODCALLTYPE d3d12_command_queue_ExecuteCommandLists(ID3D12Comm
         return;
     }
 
+    if (!(cmd_cost = vkd3d_calloc(num_command_buffers, sizeof(*cmd_cost))))
```
HansKristian-Work, line 17333: missing free in fail path on line 17375.

doitsujin: fixed.

HansKristian-Work (approved):

PR #2082, opened 2024-08-24
Signal fences on CPU to reduce delays on muxed queues

doitsujin:
Suppose we have two virtual queues `A` and `B` both sharing the same `vkd3d_queue`, and the application does something like this:
```
A->ExecuteCommands(...);
B->ExecuteCommands(...);
A->Signal(fence, ...);
```
then actually signaling the fence may be delayed until work submitted to `B` completes, which can delay both app threads waiting on the fence as well as command submissions to any queues that wait on said fence. To solve this, we instead queue up a CPU-side wait on the last submission performed by the signaling command queue and signal the fence manually.

This also takes the early signal idea from #2075 and queues up a CPU signal in case the virtual queue has no pending submissions.

This expands upon the per-queue submission tracker introduced in 5ac07816f05770401ccc9d27b96e715549f83c79, which now needs to be updated for **all** submissions made from d3d12 scope (waits, sparse binding, 11on12 interop stuff, etc) for correctness.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -364,6 +364,7 @@ static void vkd3d_queue_push_waiters_to_worker_locked(struct vkd3d_queue *vkd3d_
         VkSemaphore timeline, uint64_t value)
 {
     struct vkd3d_queue_timeline_trace_cookie cookie;
+    struct vkd3d_fence_wait_info fence_info;
     HRESULT hr;
     size_t i;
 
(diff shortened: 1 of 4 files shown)


libs/vkd3d/command.c
```diff
@@ -618,12 +619,24 @@ static void vkd3d_wait_for_gpu_timeline_semaphore(struct vkd3d_fence_worker *wor
...
+        if (fence->fence_info.signal_mode == VKD3D_FENCE_SIGNAL_GPU_PHYSICAL)
+        {
+            TRACE("Signaling fence %p physical value %#"PRIx64".\n", local_fence, fence->fence_info.vk_semaphore_value);
+
+            if (FAILED(hr = d3d12_fence_signal(local_fence, worker, fence->fence_info.vk_semaphore_value)))
+                ERR("Failed to signal D3D12 fence, hr %#x.\n", hr);
+        }
+        else if (fence->fence_info.signal_mode == VKD3D_FENCE_SIGNAL_CPU_VIRTUAL)
+        {
+            TRACE("Signaling fence %p virtual value %#"PRIx64".\n", local_fence, fence->fence_info.vk_semaphore_value);
```
HansKristian-Work, line 635: Shouldn't that be fence->fence_info.virtual_value?

doitsujin: derp

HansKristian-Work (approved):
LGTM

PR #2093, opened 2024-08-27
Align suballocated committed textures to the device buffer image granularity

esullivan-nvidia:
This is required to prevent issues with GPUs that have a buffer image granularity that is larger than the memory alignment returned from vkGetImageMemoryRequirements2. Without this alignment linear and optimal tiled resources could alias each other resulting in memory corruption.

### DIFF
--- a/libs/vkd3d/resource.c
+++ b/libs/vkd3d/resource.c
@@ -3619,6 +3619,11 @@ HRESULT d3d12_resource_create_committed(struct d3d12_device *device, const D3D12
             allocate_info.flags = VKD3D_ALLOCATION_FLAG_GLOBAL_BUFFER |
                     VKD3D_ALLOCATION_FLAG_ALLOW_IMAGE_SUBALLOCATION;
 
+            /* If image suballocation is allowed force the memory alignment to respect
+             * the device buffer image granularity to prevent resource aliasing */
+            allocate_info.memory_requirements.alignment = max(allocate_info.memory_requirements.alignment,
+                    device->device_info.properties2.properties.limits.bufferImageGranularity);
+
             /* For suballocations, we only care about being able to clear the memory,
              * not anything else. */
             allocate_info.explicit_global_buffer_usage = VK_BUFFER_USAGE_TRANSFER_DST_BIT;


doitsujin:
What kind of alignment numbers are we talking about here and does this fix anything in particular?

Mostly asking because there are potentially issues with placed resources too and then things will get all sorts of complicated.

esullivan-nvidia:
This was discovered from us running the vkd3d-proton tests in our test automation. It specifically impacts Maxwell GPUs that have a buffer image granularity size of 65536. Which is just enough to have vkd3d-proton report support for D3D12_RESOURCE_HEAP_TIER_2 and enable image suballocation. Two tests that started failing once image suballocation was enabled are test_descriptor_tables and test_copy_descriptors_range_sizes. These tests create all of their resources as committed so they end up getting suballocated with linear resources like constant buffers. This change resolves those issues.

I did consider the placed resource path but didn't notice anything that would cause the same issue seen here. It looks like vkd3d-proton tries to have applications create placed resources with an alignment of D3D12_DEFAULT_RESOURCE_PLACEMENT_ALIGNMENT by reporting that from GetResourceAllocationInfo. So it would seem like an application bug to me if a resource was placed at an offset that was not compatible with that.

HansKristian-Work:
Buffers are always aligned at 64k for placed (4k is banned), which I think would cover this. Small image alignment (4k) is a D3D12 requirement. But I think placed will honor buffer image granularity as-is.

libs/vkd3d/resource.c
```diff
@@ -3619,6 +3619,11 @@ HRESULT d3d12_resource_create_committed(struct d3d12_device *device, const D3D12
             allocate_info.flags = VKD3D_ALLOCATION_FLAG_GLOBAL_BUFFER |
                     VKD3D_ALLOCATION_FLAG_ALLOW_IMAGE_SUBALLOCATION;
 
+            /* If image suballocation is allowed force the memory alignment to respect
+             * the device buffer image granularity to prevent resource aliasing */
+            allocate_info.memory_requirements.alignment = max(allocate_info.memory_requirements.alignment,
```
HansKristian-Work, line 3624: This is confusing. We use explicit_global_buffer_usage below there, which should ensure that suballocation from images and buffer never allocate from the same chunk. Did those tests you mention really end up binding a UBO and image from the same VkDeviceMemory?

HansKristian-Work:
I added logging with this diff:

```
diff --git a/libs/vkd3d/memory.c b/libs/vkd3d/memory.c
index fd4e8671..02b6d4c6 100644
--- a/libs/vkd3d/memory.c
+++ b/libs/vkd3d/memory.c
@@ -1889,7 +1889,14 @@ HRESULT vkd3d_allocate_memory(struct d3d12_device *device, struct vkd3d_memory_a
     }
 
     if (suballocate)
+    {
         hr = vkd3d_suballocate_memory(device, allocator, info, allocation);
+
+        INFO("Heap flag #%x: Suballocated memory %p @ %"PRIx64".\n",
+                info->heap_flags,
+                allocation->device_allocation.vk_memory,
+                allocation->offset);
+    }
     else
         hr = vkd3d_memory_allocation_init(allocation, device, allocator, info);
 
```

I cannot prove any aliasing between suballocated buffers and images, at least on modern NV cards.

esullivan-nvidia:
So when I initially looked at the API log I captured from the test I incorrectly thought the memory chunk global buffer was a resource created by the test. My bad on that one, and sorry for the noise.

It seems like the memory clear that vkd3d-proton does through the global buffer when suballocating an image resource out of a shared memory chunk might be causing this issue. With the image suballocation change it is possible for multiple resources to end up sharing a single page. I suspect the clear for one can end up corrupting the other resources sharing the same page. This is why forcing the image memory alignment to guarantee only a single image resource lives on the page resolves the issue.

HansKristian-Work:
> This is why forcing the image memory alignment to guarantee only a single image resource lives on the page resolves the issue.

Ok, that makes sense.

PR #2095, opened 2024-08-28
Signal rework

doitsujin:
Builds on the more pecise submission tracking and half-reverts the CPU signal thing by removing the timeline semaphore from non-shared fences and just using queue timelines instead to reduce submission spam.

I had some concerns about semaphore lifetimes initially since we can use the timeline semaphore belonging to a `vkd3d_queue` can be used on *any other* `vkd3d_queue` now, but since D3D12 command queues ref the device and vkd3d queues have the same lifetime as the device, this should be a non-issue for now.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -1383,29 +1383,19 @@ CONST_VTBL struct ID3D12Fence1Vtbl d3d12_fence_vtbl =
     d3d12_fence_GetCreationFlags,
 };
 
-static HRESULT d3d12_fence_init_timeline(struct d3d12_fence *fence, struct d3d12_device *device,
-        UINT64 initial_value)
-{
-    fence->virtual_value = initial_value;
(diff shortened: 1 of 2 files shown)


libs/vkd3d/command.c
```diff
@@ -17756,11 +17765,14 @@ static void d3d12_command_queue_wait(struct d3d12_command_queue *command_queue,
...
-    wait_count = d3d12_fence_get_physical_wait_value_locked(fence, value);
+    has_wait = d3d12_fence_get_physical_wait_value_locked(fence, value, &fence_value);
 
     d3d12_fence_unlock(fence);
 
-    assert(fence->timeline_semaphore);
+    if (!has_wait)
+        return;
+
+    assert(fence_value.virtual_value == value);
```
HansKristian-Work, line 17766: Is this true? What if we signal 10 and wait for 9?

HansKristian-Work (approved):
LGTM

PR #2102, opened 2024-08-30
vkd3d: Implement better heuristic to enable frame rate limiter.

doitsujin:
Allows for a bit more leniency w.r.t. swap chain buffering and timing errors, but still triggers early if the delivered frame rate is significantly higher than expected (e.g. 144 vs 60 fps).

Fixes situations where we engage the limiter in Elden Ring despite the game bringing its own 60 FPS lock.


HansKristian-Work (approved):
Seems reasonable to me.

PR #2109, opened 2024-09-03
Null sibling descriptors, attempt to workaround GPU hang on NV in FF XVI.

HansKristian-Work:
Another game bug ...


HansKristian-Work:
This seems to have been a red herring.

PR #2117, opened 2024-09-05
Use KHR_compute_shader_derivatives

doitsujin:
Instead of NV where supported.

Also allows us to expose mesh/task derivatives if supported by the underlying implementation. The MS defivative test is extremely basic and mostly there to verify that the pipeline compiles properly, and passes on WARP as well as our implementation on RADV.

Draft because this does depend on dxil-spirv changes.


HansKristian-Work (commented):
Looks great. Just needs new dxil-spirv commit.

HansKristian-Work (approved):

PR #2127, opened 2024-09-20
vkd3d: Fix wrong format being used for depth bias representation.

doitsujin:
The christmas tree certainly contributed to this one going unnoticed, `format` isn't our dsv format....

Explicitly pass the DXGI format instead to avoid confusion.


anthonymoon:
@doitsujin Christmas tree?

PR #2130, opened 2024-09-23
vkd3d: Refactor device ref-count logic.

HansKristian-Work:
Avoid going through vtable just to do internal addref/release. Also splits up the logging so we can distinguish between public and private ref-counts.


doitsujin (approved):
LGTM

PR #2134, opened 2024-09-25
Deferred device destroy

HansKristian-Work:


HansKristian-Work:
Don't think this went anywhere.

PR #2135, opened 2024-09-26
vkd3d: enable EXT_device_generated_commands

zmike:
this adds new paths to use the EXT version of the functionality when available instead of NV, enabling broader support across drivers/hardware while providing the same functionality


HansKristian-Work:
Superseded.

PR #2139, opened 2024-09-28
Implement GPU upload heap.

K0bin:
God of War Ragnarok appears to use it.

Not 100% sure about the interaction with the existing system that promotes the regular UPLOAD_HEAP to rebar based on a budget.
I decided to go for the following:
- We only report support for the GPU_UPLOAD_HEAP if the existing logic has decided that we have a Rebar budget. Otherwise the same problem of potentially taking VRAM from other resources which need it more applies even if the application itself decides to put it in Rebar.
- When actually allocating memory the budget is ignored if it was the application that decided to put it into Rebar instead of our automatic promotion logic.
- After feedback: Once we allocate memory for either GPU_UPLOAD_HEAP or an equivalent custom heap, the automatically promotion logic gets disabled.
-  Resource creation checks whether GPU_UPLOAD_HEAP is supported when it (or an equivalent custom heap) gets specified.

I hope I'm properly covering all spots that interact with the heap types or custom heap properties.

Memory property flags are based on: https://microsoft.github.io/DirectX-Specs/d3d/D3D12GPUUploadHeaps.html

For tests I mostly copied everything that tested something for HEAP_TYPE_UPLOAD and adapted it to also test HEAP_TYPE_GPU_UPLOAD. I noticed that it was legal to create textures on HEAP_TYPE_GPU_UPLOAD, so I tested that with both render targets and depth stencil textures. I also tried to test the more flexible initial state.

While implementing this, I noticed there were a lot of spots where `is_cpu_accessible_heap()` gets called. As you can see from the diff, I replaced some of those with checks for system memory.

I wasn't sure about all spots where this function got used. So here are the ones that I did **not** change. Please also review those!

- https://github.com/HansKristian-Work/vkd3d-proton/blob/7856e20cccdc090319d485288caf7a6f113c85ec/libs/vkd3d/resource.c#L637-L642
- https://github.com/HansKristian-Work/vkd3d-proton/blob/7856e20cccdc090319d485288caf7a6f113c85ec/libs/vkd3d/memory.c#L1960-L1968
- https://github.com/HansKristian-Work/vkd3d-proton/blob/7856e20cccdc090319d485288caf7a6f113c85ec/libs/vkd3d/resource.c#L208-L212
- https://github.com/HansKristian-Work/vkd3d-proton/blob/7856e20cccdc090319d485288caf7a6f113c85ec/libs/vkd3d/resource.c#L851-L852
- https://github.com/HansKristian-Work/vkd3d-proton/blob/7856e20cccdc090319d485288caf7a6f113c85ec/libs/vkd3d/resource.c#L921-L929
- https://github.com/HansKristian-Work/vkd3d-proton/blob/7856e20cccdc090319d485288caf7a6f113c85ec/libs/vkd3d/resource.c#L3434-L3437

### DIFF
--- a/libs/vkd3d/device.c
+++ b/libs/vkd3d/device.c
@@ -977,6 +977,7 @@ static const struct vkd3d_debug_option vkd3d_config_options[] =
     {"force_dynamic_msaa", VKD3D_CONFIG_FLAG_FORCE_DYNAMIC_MSAA},
     {"instruction_qa_checks", VKD3D_CONFIG_FLAG_INSTRUCTION_QA_CHECKS},
     {"transfer_queue", VKD3D_CONFIG_FLAG_TRANSFER_QUEUE},
+    {"no_gpu_upload_heap", VKD3D_CONFIG_FLAG_NO_GPU_UPLOAD_HEAP},
 };
 
 static void vkd3d_config_flags_init_once(void)
@@ -5719,7 +5720,7 @@ static D3D12_HEAP_PROPERTIES * STDMETHODCALLTYPE d3d12_device_GetCustomHeapPrope
         case D3D12_HEAP_TYPE_DEFAULT:
             heap_properties->CPUPageProperty = D3D12_CPU_PAGE_PROPERTY_NOT_AVAILABLE;
             heap_properties->MemoryPoolPreference = d3d12_device_is_uma(device, NULL)
-                    ?  D3D12_MEMORY_POOL_L0 : D3D12_MEMORY_POOL_L1;
+                    ? D3D12_MEMORY_POOL_L0 : D3D12_MEMORY_POOL_L1;
             break;
 
         case D3D12_HEAP_TYPE_UPLOAD:
@@ -5731,6 +5732,13 @@ static D3D12_HEAP_PROPERTIES * STDMETHODCALLTYPE d3d12_device_GetCustomHeapPrope
         case D3D12_HEAP_TYPE_READBACK:
             heap_properties->CPUPageProperty = D3D12_CPU_PAGE_PROPERTY_WRITE_BACK;
             heap_properties->MemoryPoolPreference = D3D12_MEMORY_POOL_L0;
+            break;
+
+        case D3D12_HEAP_TYPE_GPU_UPLOAD:
+            heap_properties->CPUPageProperty = d3d12_device_is_uma(device, &coherent) && coherent
+                    ? D3D12_CPU_PAGE_PROPERTY_WRITE_BACK : D3D12_CPU_PAGE_PROPERTY_WRITE_COMBINE;
+            heap_properties->MemoryPoolPreference = d3d12_device_is_uma(device, NULL)
+                    ? D3D12_MEMORY_POOL_L0 : D3D12_MEMORY_POOL_L1;
             break;
 
         default:
@@ -8106,7 +8114,7 @@ static void d3d12_device_caps_init_feature_options16(struct d3d12_device *device
     D3D12_FEATURE_DATA_D3D12_OPTIONS16 *options16 = &device->d3d12_caps.options16;
 
     options16->DynamicDepthBiasSupported = TRUE;
-    options16->GPUUploadHeapSupported = FALSE;
+    options16->GPUUploadHeapSupported = device->memory_info.has_gpu_upload_heap;
 }
 
 static void d3d12_device_caps_init_feature_options17(struct d3d12_device *device)
--- a/libs/vkd3d/memory.c
+++ b/libs/vkd3d/memory.c
@@ -642,7 +642,9 @@ static D3D12_HEAP_TYPE vkd3d_normalize_heap_type(const D3D12_HEAP_PROPERTIES *he
         case D3D12_CPU_PAGE_PROPERTY_WRITE_BACK:
             return D3D12_HEAP_TYPE_READBACK;
         case D3D12_CPU_PAGE_PROPERTY_WRITE_COMBINE:
-            return D3D12_HEAP_TYPE_UPLOAD;
+            return heap_properties->MemoryPoolPreference == D3D12_MEMORY_POOL_L1
+                ? D3D12_HEAP_TYPE_GPU_UPLOAD
+                : D3D12_HEAP_TYPE_UPLOAD;
         default:
             break;
     }
@@ -663,6 +665,10 @@ static HRESULT vkd3d_select_memory_flags(struct d3d12_device *device, const D3D1
             *type_flags = device->memory_info.upload_heap_memory_properties;
             break;
 
+        case D3D12_HEAP_TYPE_GPU_UPLOAD:
+            *type_flags = VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT | VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT | VK_MEMORY_PROPERTY_HOST_COHERENT_BIT;
+            break;
+
         case D3D12_HEAP_TYPE_READBACK:
             *type_flags = VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT | VK_MEMORY_PROPERTY_HOST_CACHED_BIT;
             break;
@@ -678,8 +684,11 @@ static HRESULT vkd3d_select_memory_flags(struct d3d12_device *device, const D3D1
                     break;
                 case D3D12_CPU_PAGE_PROPERTY_WRITE_COMBINE:
                     *type_flags = VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT | VK_MEMORY_PROPERTY_HOST_COHERENT_BIT;
+
                     if (vkd3d_config_flags & VKD3D_CONFIG_FLAG_FORCE_HOST_CACHED)
                         *type_flags |= VK_MEMORY_PROPERTY_HOST_CACHED_BIT;
+                    else if (heap_properties->MemoryPoolPreference == D3D12_MEMORY_POOL_L1)
+                        *type_flags |= VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT;
                     break;
                 case D3D12_CPU_PAGE_PROPERTY_NOT_AVAILABLE:
                     *type_flags = VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT;
@@ -1061,6 +1070,9 @@ static void *vkd3d_allocate_write_watch_pointer(const D3D12_HEAP_PROPERTIES *pro
     case D3D12_HEAP_TYPE_UPLOAD:
         protect = PAGE_READWRITE | PAGE_WRITECOMBINE;
         break;
+    case D3D12_HEAP_TYPE_GPU_UPLOAD:
+        protect = PAGE_READWRITE | PAGE_WRITECOMBINE;
+        break;
     case D3D12_HEAP_TYPE_READBACK:
         /* WRITE_WATCH fails for this type in native D3D12,
          * otherwise it would be PAGE_READWRITE. */
@@ -1258,6 +1270,16 @@ static HRESULT vkd3d_memory_allocation_init(struct vkd3d_memory_allocation *allo
         }
     }
 
+    /* Custom heaps have been resolved to regular heap types earlier. */
+    assert(allocation->heap_type != D3D12_HEAP_TYPE_CUSTOM);
+
+    if (allocation->heap_type == D3D12_HEAP_TYPE_GPU_UPLOAD
+        && !device->memory_info.has_gpu_upload_heap)
+    {
+        ERR("Trying to allocate memory on GPU_UPLOAD_HEAP which is not supported on current device.\n");
+        return E_INVALIDARG;
+    }
+
     if (device->device_info.memory_priority_features.memoryPriority &&
         (type_flags & VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT))
     {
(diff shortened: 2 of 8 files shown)


doitsujin:
I think for tests we would want mostly what we already do for all the regular heap stuff:
- verify that creating heaps works, both normal as well as custom heaps. There's a conflicting test somewhere because the combination of heap properties enabling GPU upload heaps is currently banned.
- check API validation around resource creation, regular upload heaps are fairly restricted and you need custom heaps for certain usages such as anything GPU-writable

These can probably be folded into existing tests somehow without having to write too much new code.

HansKristian-Work:
> Another thing we also need to consider is what to do with normal UPLOAD heaps once the app starts using GPU upload heaps; ideally we just stop putting those into ReBAR as soon as the first GPU upload heap is created or something.

Yes, we'll need a device-global atomic or something where if we have observed GPU heap being used, UPLOAD heap does not get DEVICE_LOCAL memory property anymore.

libs/vkd3d/device.c
```diff
@@ -5673,6 +5673,13 @@ static D3D12_HEAP_PROPERTIES * STDMETHODCALLTYPE d3d12_device_GetCustomHeapPrope
             heap_properties->MemoryPoolPreference = D3D12_MEMORY_POOL_L0;
             break;
 
+        case D3D12_HEAP_TYPE_GPU_UPLOAD:
+            heap_properties->CPUPageProperty = d3d12_device_is_uma(device, &coherent) && coherent
+                    ? D3D12_CPU_PAGE_PROPERTY_WRITE_BACK : D3D12_CPU_PAGE_PROPERTY_WRITE_COMBINE;
+            heap_properties->MemoryPoolPreference = d3d12_device_is_uma(device, NULL)
+                    ?  D3D12_MEMORY_POOL_L0 : D3D12_MEMORY_POOL_L1;
```
HansKristian-Work, line 5680: Weird space after `?`

libs/vkd3d/device.c
```diff
@@ -8046,7 +8053,7 @@ static void d3d12_device_caps_init_feature_options16(struct d3d12_device *device
     D3D12_FEATURE_DATA_D3D12_OPTIONS16 *options16 = &device->d3d12_caps.options16;
 
     options16->DynamicDepthBiasSupported = TRUE;
-    options16->GPUUploadHeapSupported = FALSE;
+    options16->GPUUploadHeapSupported = device->memory_info.rebar_budget != 0;
```
HansKristian-Work, line 8056: This is conflicting with

```
    if (heap_properties->MemoryPoolPreference == D3D12_MEMORY_POOL_L1
        && heap_properties->CPUPageProperty == D3D12_CPU_PAGE_PROPERTY_WRITE_COMBINE
        && !(device->memory_info.upload_heap_memory_properties & VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT))
    {
        WARN("Invalid memory pool preference (device does not support rebar).\n");
        return E_INVALIDARG;
    }
```

which uses the upload_heap_memory_properties instead to determine if GPUUploadHeap is supported.

libs/vkd3d/memory.c
```diff
@@ -678,8 +684,14 @@ static HRESULT vkd3d_select_memory_flags(struct d3d12_device *device, const D3D1
                     break;
                 case D3D12_CPU_PAGE_PROPERTY_WRITE_COMBINE:
                     *type_flags = VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT | VK_MEMORY_PROPERTY_HOST_COHERENT_BIT;
+
+                    if (heap_properties->MemoryPoolPreference == D3D12_MEMORY_POOL_L1)
+                        *type_flags |= VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT;
+
                     if (vkd3d_config_flags & VKD3D_CONFIG_FLAG_FORCE_HOST_CACHED)
                         *type_flags |= VK_MEMORY_PROPERTY_HOST_CACHED_BIT;
+                    else if (heap_properties->MemoryPoolPreference == D3D12_MEMORY_POOL_L1)
```
HansKristian-Work, line 701: This looks kinda redundant to me?

K0bin: You're right, I think the top one snuck back in during one of the rebases.

libs/vkd3d/memory.c
```diff
@@ -1258,6 +1273,16 @@ static HRESULT vkd3d_memory_allocation_init(struct vkd3d_memory_allocation *allo
         }
     }
 
+    if (!(device->memory_info.upload_heap_memory_properties & VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT)
```
HansKristian-Work: This branch is impossible to read. Please split this up into somewhat readable booleans.

libs/vkd3d/resource.c
```diff
@@ -17,6 +17,7 @@
  * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
  */
 
+#include "vkd3d.h"
```
HansKristian-Work, line 20: Is this include accidental?

libs/vkd3d/device.c
HansKristian-Work: I see this is inherited from the D3D12_HEAP_TYPE_DEFAULT line above. Might as well fix that while you're at it.

libs/vkd3d/resource.c
K0bin: VSCode keeps adding in random includes >.<

libs/vkd3d/device.c
HansKristian-Work: I think we might want to use vkd3d_memory_topology_is_uma_like() check in vkd3d_memory_info_upload_hvv_memory_properties() somehow.

Perhaps the cleaner solution is to not request DEVICE_LOCAL properties for UMA-like, since device-local and host-visible is kinda the same thing anyway, and shouldn't change anything.

libs/vkd3d/memory.c
```diff
@@ -1273,14 +1281,22 @@ static HRESULT vkd3d_memory_allocation_init(struct vkd3d_memory_allocation *allo
...
     {
-        ERR("Trying to allocate memory on GPU_UPLOAD_HEAP which is not supported on current device.\n");
-        return E_INVALIDARG;
+        if (!(device->memory_info.upload_heap_memory_properties & VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT))
+        {
+            ERR("Trying to allocate memory on GPU_UPLOAD_HEAP which is not supported on current device.\n");
+            return E_INVALIDARG;
+        }
+
+        if (!vkd3d_atomic_uint64_exchange_explicit(&device->memory_info.has_used_gpu_upload_heap, true, vkd3d_memory_order_relaxed)
```
HansKristian-Work, line 1295: Why uint64? The code above is u32.

tests/d3d12_crosstest.h
```diff
@@ -781,4 +781,15 @@ static inline void enable_d3d12_debug_layer(int argc, char **argv)
     }
 }
 
+static inline bool device_supports_gpu_upload_heap(ID3D12Device *device)
+{
+    D3D12_FEATURE_DATA_D3D12_OPTIONS16 options16;
+    HRESULT hr;
+    hr = ID3D12Device_CheckFeatureSupport(device, D3D12_FEATURE_D3D12_OPTIONS16, &options16, sizeof(options16));
+    if (hr != S_OK) {
```
HansKristian-Work, line 789: Brace on next line. Can just drop brace here.

HansKristian-Work (changes requested):
Comments.

libs/vkd3d/memory.c
```diff
@@ -642,7 +642,9 @@ static D3D12_HEAP_TYPE vkd3d_normalize_heap_type(const D3D12_HEAP_PROPERTIES *he
         case D3D12_CPU_PAGE_PROPERTY_WRITE_BACK:
             return D3D12_HEAP_TYPE_READBACK;
         case D3D12_CPU_PAGE_PROPERTY_WRITE_COMBINE:
-            return D3D12_HEAP_TYPE_UPLOAD;
+            return heap_properties->MemoryPoolPreference == D3D12_MEMORY_POOL_L1
+                ? D3D12_HEAP_TYPE_UPLOAD
```
HansKristian-Work, line 646: Shouldn't this be the other way around?

libs/vkd3d/memory.c
```diff
@@ -1280,7 +1305,7 @@ static HRESULT vkd3d_memory_allocation_init(struct vkd3d_memory_allocation *allo
     else
     {
         hr = vkd3d_allocate_device_memory(device, memory_requirements.size, type_flags,
-                type_mask, &flags_info, true, &allocation->device_allocation);
+                type_mask, &flags_info, info->heap_properties.Type == D3D12_HEAP_TYPE_UPLOAD, &allocation->device_allocation);
```
HansKristian-Work, line 1325: I don't understand the rationable behind this line.

K0bin: Ignore the budget if we are allocating memory for GPU UPLOAD HEAP. In case the game only allocates GPU upload heap memory after allocating enough regular UPLOAD HEAP memory to use up all the budget.

libs/vkd3d/vkd3d_private.h
```diff
@@ -4972,6 +4972,20 @@ static inline bool is_cpu_accessible_heap(const D3D12_HEAP_PROPERTIES *propertie
...
 
+static inline bool is_cpu_accessible_system_memory_heap(const D3D12_HEAP_PROPERTIES *properties)
+{
+    if (properties->Type == D3D12_HEAP_TYPE_DEFAULT)
+        return false;
+    if (properties->Type == D3D12_HEAP_TYPE_GPU_UPLOAD)
+        return false;
+    if (properties->Type == D3D12_HEAP_TYPE_CUSTOM)
+    {
+        return properties->CPUPageProperty == D3D12_CPU_PAGE_PROPERTY_NOT_AVAILABLE
```
HansKristian-Work, line 4984: Why == NOT_AVAILABLE? That doesn't make much sense.

libs/vkd3d/resource.c
```diff
@@ -8776,6 +8776,7 @@ HRESULT vkd3d_memory_info_init(struct vkd3d_memory_info *info,
     info->descriptor_heap_memory_properties =
             vkd3d_memory_info_descriptor_heap_memory_properties(&topology, device);
     vkd3d_memory_info_init_budgets(info, &topology, device);
+    info->has_used_gpu_upload_heap = false;
```
HansKristian-Work, line 8779: Feels weird to use bool on a u32.

libs/vkd3d/memory.c
```diff
@@ -1272,15 +1280,24 @@ static HRESULT vkd3d_memory_allocation_init(struct vkd3d_memory_allocation *allo
...
-        ERR("Trying to allocate memory on GPU_UPLOAD_HEAP which is not supported on current device.\n");
-        return E_INVALIDARG;
+        is_rebar_supported = (device->memory_info.upload_heap_memory_properties & VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT) != 0;
+        if (!is_rebar_supported)
+        {
+            ERR("Trying to allocate memory on GPU_UPLOAD_HEAP which is not supported on current device.\n");
+            return E_INVALIDARG;
+        }
+
+        if (!vkd3d_atomic_uint32_exchange_explicit(&device->memory_info.has_used_gpu_upload_heap, true, vkd3d_memory_order_relaxed)
```
HansKristian-Work, line 1296: Just use 1 instead of true, it's a uint32.

libs/vkd3d/memory.c
HansKristian-Work: Right, this is the respect budget field, thought it was something about fallback ...

To make sure, CUSTOM heap with L0 heap will be normalized to UPLOAD here, right?

HansKristian-Work: I think we can assert here that heap property is no longer CUSTOM once we reach this code path.

K0bin: This ignores custom heap, so I changed it to `!is_host_visible_vram`.

HansKristian-Work:
There is a test failure when running on AMD without ReBAR.

```
VKD3D_CONFIG=no_upload_hvv VKD3D_TEST_FILTER=test_map_texture_validation build-native/tests/d3d12
```

`test_map_texture_validation:3680:Test 10: Test failed: Unexpected hr 0.`

HansKristian-Work (commented):
LGTM, but I'd like @doitsujin to have a second pair of eyes on it.

K0bin:
Changed it, so the GPU upload heap is no longer controlled by the config option `no_upload_hvv`. Instead it can now be separately disabled using `no_gpu_upload_heap`.

libs/vkd3d/resource.c
```diff
@@ -8645,34 +8653,18 @@ static VkMemoryPropertyFlags vkd3d_memory_info_descriptor_heap_memory_properties
     }
 }
 
-static VkMemoryPropertyFlags vkd3d_memory_info_upload_hvv_memory_properties(
+static bool vkd3d_memory_info_decide_hvv_usage(
         const struct vkd3d_memory_topology *topology,
         const struct d3d12_device *device)
 {
-    if (vkd3d_config_flags & VKD3D_CONFIG_FLAG_FORCE_HOST_CACHED)
```
HansKristian-Work, line 8652: For FORCE_HOST_CACHED we don't want to expose ReBAR at all. Mostly to avoid RenderDoc captures being unusably slow.

libs/vkd3d/resource.c
```diff
@@ -8691,25 +8683,69 @@ static VkMemoryPropertyFlags vkd3d_memory_info_upload_hvv_memory_properties(
...
+
+
+static bool vkd3d_memory_info_decide_gpu_upload_heap(bool is_hvv_use_allowed)
+{
+    if (!is_hvv_use_allowed)
+        return false;
+
+    if (vkd3d_config_flags & VKD3D_CONFIG_FLAG_FORCE_HOST_CACHED)
+        return false;
+
```
HansKristian-Work, line 8743: Weird newline.

libs/vkd3d/resource.c
HansKristian-Work: Or maybe this is resolved below ...

HansKristian-Work (approved):
LGTM

doitsujin (approved):
LGTM, tests pass on a real non-ReBAR system too.

PR #2152, opened 2024-10-05
vkd3d: Allow exposing GPU_UPLOAD heap on smaller VRAM as well.

HansKristian-Work:
If game explicitly asks for GPU_UPLOAD we should honor it. Also allows our internal scratch allocations for uniform memory to use ReBAR on more modest VRAM configurations as well.

### DIFF
--- a/libs/vkd3d/memory.c
+++ b/libs/vkd3d/memory.c
@@ -1289,8 +1289,9 @@ static HRESULT vkd3d_memory_allocation_init(struct vkd3d_memory_allocation *allo
             return E_INVALIDARG;
         }
 
-        if (vkd3d_atomic_uint32_exchange_explicit(&device->memory_info.has_used_gpu_upload_heap, 1, vkd3d_memory_order_relaxed) != 1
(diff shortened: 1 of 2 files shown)


K0bin:
What prompted this change of opinion? 
I asked about this in the PR.

HansKristian-Work:
> What prompted this change of opinion?

There was an unintended fallout from this change, which is that our own internal scratch allocations would not be able to use ReBAR anymore, but this change makes it so that UPLOAD heap stays sysmem, but if GPU_UPLOAD is explicitly asked for, we should allow application to use it. The main reason we're careful about ReBAR on 8 GB is that every random UPLOAD thing is placed in ReBAR potentially, which bloats things.

K0bin:
The issue with internal allocations makes sense of course. I'm just a bit worried that games aren't smart enough about memory management which will end up hurting us more than Windows.

HansKristian-Work:
> The issue with internal allocations makes sense of course. I'm just a bit worried that games aren't smart enough about memory management which will end up hurting us more than Windows.

We can always re-evaluate if we see GPU_UPLOAD games in the wild that cause issues.

K0bin (approved):

doitsujin (approved):

PR #2156, opened 2024-10-09
Implement ID3DDestructionNotifier

doitsujin:
I lost multiple brain cells writing this code, but the TL;DR is that *every* D3D12 COM object supports this interface, and according to the docs it's supposed to fire any registered callback *before* references to other objects are released, hence the jank around objects that we keep alive internally.

Apparently this is needed by The Forever Winter, although I haven't seen it use the interface myself while testing.

### DIFF
--- a/libs/vkd3d/utils.c
+++ b/libs/vkd3d/utils.c
@@ -1654,3 +1654,158 @@ HRESULT d3d_blob_create(void *buffer, SIZE_T size, struct d3d_blob **blob)
 
     return S_OK;
 }
+
+
+static struct d3d_destruction_notifier *impl_from_ID3DDestructionNotifier(ID3DDestructionNotifier *iface)
+{
+    return CONTAINING_RECORD(iface, struct d3d_destruction_notifier, ID3DDestructionNotifier_iface);
+}
+
+static HRESULT STDMETHODCALLTYPE d3d_destruction_notifier_QueryInterface(ID3DDestructionNotifier *iface, REFIID riid, void **object)
+{
+    struct d3d_destruction_notifier *notifier = impl_from_ID3DDestructionNotifier(iface);
+
+    TRACE("iface %p, riid %s, object %p.\n", iface, debugstr_guid(riid), object);
+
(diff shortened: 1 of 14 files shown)


libs/vkd3d/vkd3d_private.h
```diff
@@ -539,7 +539,8 @@ struct d3d_destruction_notifier
 
 void d3d_destruction_notifier_init(struct d3d_destruction_notifier *notifier, IUnknown *parent);
 void d3d_destruction_notifier_free(struct d3d_destruction_notifier *notifier);
-void d3d_destruction_notifier_notify(struct d3d_destruction_notifier *notifier);
+void d3d_destruction_notifier_notify
+(struct d3d_destruction_notifier *notifier);
```
HansKristian-Work, line 543: This line break seems unintentional?

HansKristian-Work (changes requested):
Hnnnng. Only one trivial nit, otherwise lgtm.

libs/vkd3d/utils.c
```diff
@@ -1654,3 +1654,158 @@ HRESULT d3d_blob_create(void *buffer, SIZE_T size, struct d3d_blob **blob)
...
+    return S_OK;
+}
+
+static CONST_VTBL struct ID3DDestructionNotifierVtbl d3d_destruction_notifier_vtbl =
+{
+    /* IUnknown methods */
+    d3d_destruction_notifier_QueryInterface,
+    d3d_destruction_notifier_AddRef,
+    d3d_destruction_notifier_Release,
+    /* ID3DBlob methods */
```
misyltoad, line 1772: copy paste error

libs/vkd3d/vkd3d_private.h
doitsujin: derp

libs/vkd3d/utils.c
doitsujin: fixed that too

HansKristian-Work (approved):
LGTM

PR #2158, opened 2024-10-11
vkd3d: Workaround buggy API usage patterns with pipeline libraries.

HansKristian-Work:
FF XVI attempts to serialize back to same pointer that unserialized. This is not legal, but it would work if we guaranteed stable serialization, which we don't. Workaround it by detecting overlapping memory and serializing to temporary memory.


doitsujin (approved):
well at least the workaround isn't too crazy

PR #2161, opened 2024-10-15
meta: Finish document for workgraph emulation.

HansKristian-Work:
This is purely non-functional.
Before attempting to land experimental workgraph support, this document should be useful :')
It's probably best to not look at the diff, but the full view since I rewrote the entire thing.
Also serves as a pseudo blog-post.


HansKristian-Work:
Waiving review per discord discussion.

PR #2164, opened 2024-10-16
Implement ExecuteIndirect tier 1.1.

HansKristian-Work:


tests/shaders/command/execute_indirect_tier11_mesh.ms_6_5.hlsl
```diff
@@ -0,0 +1,10 @@
+RWStructuredBuffer<uint> RWBuf : register(u0);
+cbuffer Cbuf : register(b0) { uint DrawID; };
+
+[numthreads(1, 1, 1)]
+[outputtopology("triangle")]
+void main()
+{
+	uint o;
+	InterlockedAdd(RWBuf[DrawID], 1, o);
```
doitsujin, line 9: Unless there's some DXC magic these days, we should probably call `SetMeshOutputCounts(0, 0)` here or AMD native might hang. I ran into this while initially implementing mesh shaders.

doitsujin (commented):
Test nit/question, looks good otherwise if there's no way we compact indirect draws or something.

tests/shaders/command/execute_indirect_tier11_mesh.ms_6_5.hlsl
HansKristian-Work: AMD doesn't expose Tier 1.1 yet, so *shrug* for now I think.

doitsujin: fair.

doitsujin (approved):

PR #2171, opened 2024-10-22
Force precise vertex shaders in Hunt: Showdown

HansKristian-Work:
Should fix glitched graphics due to game bug.


doitsujin (approved):
pain, but at least it builds on an existing workaround i guess

PR #2177, opened 2024-10-29
Improve accounting of tests marked as driver bugs

LingMan:
1. Count tests marked as driver bugs as executed, since they actually do get run.
2. Currently all tests marked as driver bugs get counted as "bug" no matter if they succeed or fail. Count them separately to make it clear how many of them actually fail.

Before:
```
d3d12: 3872234 tests executed (0 failures, 412 successful todo, 88 skipped, 332 todo, 2738 bugs).
```
After:
```
d3d12: 3874972 tests executed (0 failures, 412 successful todo, 88 skipped, 332 todo, 1 bugs, 2737 successful bugs).
```


doitsujin:
"successful bugs" reads a bit weird, maybe "fixed" would be better?

Just a nit though, I do agree that we should only count bugs as actual bugs if the corresponding test fails.

LingMan:
Right, I actually used "fixed" at first. Then I thought that a passing test doesn't say all that much about bugs being fixed. With flaking tests, tests that only fail on some hardware, etc. That made me mirror the `todo`/`successful todo` wording instead.

Happy to use whatever terminology you'd prefer though.

Blisto91:
Can't it just be explicitly named? Something like "not our bugs", "driver bugs" or "invalid bugs" etc. Or does that sounds weird.

R1chterScale:
Forgive me if I'm misunderstanding, but this is bugs that are from native DX12 yes? If so, perhaps "Bugs replicated" might be a preferred term?

LingMan:
No, these are bugs in the underlying Vulkan driver and should ideally be fixed there. Modulo misattributions of course.

PR #2202, opened 2024-11-11
Initialize page tables for sparse resources

doitsujin:
Extends the existing function that allocates and binds the image metadata aspect to also initialize all page mappings to null.

For synchronization purposes, this uses the same approach as memory clears in order to avoid a full `vkQueueWaitIdle`. Resources in our test suite can take ~0.5ms to init, so doing this asynchronously is pretty much required for performance reasons, and it only gets worse on devices that don't have a dedicated sparse queue.

Tested with SotTR and DD2, not seeing perf or other kinds of regressions.


doitsujin:
Test fails on amdvlk now for some reason, investigating.

libs/vkd3d/resource.c
```diff
@@ -3109,26 +3098,38 @@ static HRESULT d3d12_resource_bind_sparse_metadata(struct d3d12_resource *resour
         goto cleanup;
     }
 
-    if ((vr = VK_CALL(vkQueueBindSparse(vk_queue, 1, &bind_info, VK_NULL_HANDLE))) < 0)
+    sparse->init_timeline_value = ++vkd3d_queue->submission_timeline_count;
```
HansKristian-Work, line 3177: This is flawed I think. Sparse commands run completely out of order with normal submissions, so signal ordering is a problem here. I think we need a wait for previous value to be safe.

libs/vkd3d/resource.c
```diff
@@ -2972,123 +2972,199 @@ static HRESULT d3d12_resource_validate_create_info(const D3D12_RESOURCE_DESC1 *d
...
 }
 
-static HRESULT d3d12_resource_bind_sparse_metadata(struct d3d12_resource *resource,
+static HRESULT d3d12_resource_init_page_table(struct d3d12_resource *resource,
         struct d3d12_device *device, struct d3d12_sparse_info *sparse)
 {
     const struct vkd3d_vk_device_procs *vk_procs = &device->vk_procs;
     VkSparseImageMemoryRequirements *sparse_requirements = NULL;
-    VkSparseImageOpaqueMemoryBindInfo opaque_bind;
+    unsigned int  mip_tail_layer_count, standard_mip_count;
```
HansKristian-Work, line 2980: Weird space between int and mip_tail.

libs/vkd3d/resource.c
```diff
@@ -2972,123 +2972,199 @@ static HRESULT d3d12_resource_validate_create_info(const D3D12_RESOURCE_DESC1 *d
...
+    bind_info.sType = VK_STRUCTURE_TYPE_BIND_SPARSE_INFO;
 
-    if (!(sparse_requirements = vkd3d_malloc(sparse_requirement_count * sizeof(*sparse_requirements))))
+    if (d3d12_resource_is_buffer(resource))
     {
-        ERR("Failed to allocate sparse memory requirement array.\n");
-        hr = E_OUTOFMEMORY;
-        goto cleanup;
+        memset(&buffer_bind, 0, sizeof(buffer_bind));
+        buffer_bind.size = resource->desc.Width;
```
HansKristian-Work, line 3010: Do we make sure to round this up to nearest 64 K?

HansKristian-Work:
```
        /* If we attempt to bind sparse buffer with non-64k pages, we crash drivers.
         * Specs seems a bit unclear how non-aligned VkBuffer sizes are supposed to work,
         * so be safe. Pad out sparse buffers to their natural page size. */
        buffer_info.size = align64(buffer_info.size, VKD3D_TILE_SIZE);
```

doitsujin: fixed, wasn't aware that we were padding sparse buffers.

libs/vkd3d/resource.c
doitsujin: I suppose this is only really an issue if we don't have a dedicated sparse binding queue, but added an unconditional wait nonetheless since it shouldn't harm anything in that case either.

doitsujin: Should be fixed now via a dedicated timeline.

libs/vkd3d/resource.c
```diff
@@ -3109,26 +3098,40 @@ static HRESULT d3d12_resource_bind_sparse_metadata(struct d3d12_resource *resour
...
+
+    memset(&bind_info, 0, sizeof(bind_info));
+    bind_info.sType = VK_STRUCTURE_TYPE_BIND_SPARSE_INFO;
+    bind_info.pNext = &timeline_info;
+    bind_info.imageOpaqueBindCount = 1;
+    bind_info.pImageOpaqueBinds = &opaque_bind;
+    bind_info.signalSemaphoreCount = 1;
+    bind_info.pSignalSemaphores = &device->sparse_init_timeline;
+
+    vr = VK_CALL(vkQueueBindSparse(vk_queue, 1, &bind_info, VK_NULL_HANDLE));
```
HansKristian-Work, line 3189: We still need to wait for the previous timeline value, since sparse updates are unordered, and we still risk signaling N + 1 before N is signaled in theory. (Confirmed this with Tobias).

HansKristian-Work (approved):
LGTM.

PR #2211, opened 2024-11-14
Enable float controls for DXBC shaders

doitsujin:
Fairly trivial, should probably be tested with Veilguard to confirm it's actually fixed.

Ignores the FP16 stuff for now since we don't emit that in our DXBC compiler anyway, and the one thing that's interesting - forcing RTZ for f32tof16 - isn't possible to support anyway.


HansKristian-Work (approved):
LGTM. Needs confirmation it fixes Veilguard properly before merging.

PR #2213, opened 2024-11-15
tests: Workaround MSVC build failure.

HansKristian-Work:
Weird that it started failing now though ...


Blisto91:
> What's even weirder is that it doesn't fail in #2214...

Than one ran with `msvc 19.41.34123` whereas the failing one with `msvc 19.42.34433`

Issue #2215, opened 2024-11-16
Cyberpunk 2077 performance is reduced since commit bcfc25c9


Sporif:
Cyberpunk 2077 performance is reduced, especially during holocalls. 

Toggling the MangoHud overlay on during holocalls seems to increase the fps somewhat (but performance is still not at the expected level).

The commit that caused this issue is [bcfc25c9](https://github.com/HansKristian-Work/vkd3d-proton/commit/bcfc25c90aff70a9f6ed2299231d1105fdec4b71).

### Software information
Cyberpunk 2077, run with these [graphics settings](https://imgur.com/a/Duj0e9C). I also have gamescope, ReShade, and MangoHud enabled but the issue still exists with all of them disabled.
 
### System information
- GPU: AMD RX 7900 XTX
- Driver: Mesa 24.2.7
- Wine version: Proton 9.0-3
- VKD3D-Proton version: 2.13.r92.gcf34c0bd
- Distribution: Arch Linux
- Kernel: 6.11.8-1-custom

### Log files
[steam-14489251341945274368.log](https://github.com/user-attachments/files/17787588/steam-14489251341945274368.log)

These are performance logs taken using MangoHud with a holocall initiated in the middle of the run. You can see that the affected commit causes a drop in performance overall, but a severe drop during the holocall.

[cf34c0bd](https://github.com/HansKristian-Work/vkd3d-proton/commit/cf34c0bd47ffcee6aa4a49168ebb47936bbb3f9c): [Cyberpunk2077_2024-11-14_19-35-45.csv](https://github.com/user-attachments/files/17787591/Cyberpunk2077_2024-11-14_19-35-45.csv)
[bcfc25c9](https://github.com/HansKristian-Work/vkd3d-proton/commit/bcfc25c90aff70a9f6ed2299231d1105fdec4b71): [Cyberpunk2077_2024-11-14_19-40-33.csv](https://github.com/user-attachments/files/17787592/Cyberpunk2077_2024-11-14_19-40-33.csv)

doitsujin:
What is holocall?

Is there a way to visualize the CSV files or get any useful numbers out of them? What kind of performance delta are we talking about?

mbriar:
> Is there a way to visualize the CSV files

![cp1](https://github.com/user-attachments/assets/fc1f4899-831c-4ed6-aa82-46b7f0adc07c)

doitsujin:
Need a **fast** way to repro in game then.

[...]

PR #2226, opened 2024-11-25
Rework semaphore waiting

doitsujin:
Workaround for #2215.

This moves fence wait tracking from `vkd3d_queue` to the virtual queues and optimizes away most of the empty submissions that only do a semaphore wait and a semaphore signal, which is what caused the massive performance degradation in Cyberpunk since AMDGPU does not handle this very well.

Additionally, sparse binding serialization now uses the queue's timeline instead of the binary semaphore. At worst, there will now be one extra `vkQueueSubmit` on the first sparse binding call in order to flush waiters, rather than doing a round-trip over the graphics queue each time.

One potential concern is that this may delay fence signals in scenarios where we set `NO_STAGGERED_SUBMIT` in case the app does something like `queue->Wait(...); nothing for quite some time; queue->Signal(...)` with no commands being submitted. The only solution I can think of would involve stalling the signaling queue worker until all pending waits have completed in that case, which is something that we're kind of trying to avoid, but I'm not sure if this is even an issue in practice - Black Myth Wukong doesn't seem to run into this problem with FSR framegen enabled at all, so at least UE5 should be fine.

Most of this code was written on four hours of sleep, so this will need extensive testing.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -18079,6 +18079,8 @@ static void d3d12_command_queue_eliminate_completed_waits(struct d3d12_command_q
 
 static void d3d12_command_queue_gather_wait_semaphores_locked(struct d3d12_command_queue *command_queue, VkSubmitInfo2 *submit_info, uint32_t wait_flags)
 {
+    struct VkSemaphoreSubmitInfo serializing_semaphore;
+
     d3d12_command_queue_add_wait_semaphores(command_queue,
             submit_info->waitSemaphoreInfoCount,
             submit_info->pWaitSemaphoreInfos);
@@ -18092,6 +18094,16 @@ static void d3d12_command_queue_gather_wait_semaphores_locked(struct d3d12_comma
                 command_queue->vkd3d_queue->wait_semaphores);
 
         command_queue->vkd3d_queue->wait_count = 0u;
+    }
+
+    if ((wait_flags & VKD3D_WAIT_SEMAPHORES_SERIALIZING) && command_queue->serializing_semaphore_signaled)
+    {
+        memset(&serializing_semaphore, 0, sizeof(serializing_semaphore));
+        serializing_semaphore.sType = VK_STRUCTURE_TYPE_SEMAPHORE_SUBMIT_INFO;
+        serializing_semaphore.semaphore = command_queue->serializing_semaphore;
+        serializing_semaphore.stageMask = VK_PIPELINE_STAGE_2_ALL_COMMANDS_BIT;
+
+        d3d12_command_queue_add_wait_semaphores(command_queue, 1, &serializing_semaphore);
     }
 
     d3d12_command_queue_eliminate_completed_waits(command_queue);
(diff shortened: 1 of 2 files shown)


libs/vkd3d/vkd3d_private.h
```diff
@@ -3298,6 +3298,10 @@ struct d3d12_command_queue
     struct vkd3d_fence_virtual_wait *wait_fences;
     size_t wait_fences_size;
     size_t wait_fence_count;
+
+    VkSemaphoreSubmitInfo *wait_semaphores;
```
HansKristian-Work, line 3307: When is this freed?

libs/vkd3d/command.c
```diff
@@ -18142,6 +18154,9 @@ static void d3d12_command_queue_flush_waiters(struct d3d12_command_queue *comman
 
         if ((vr = VK_CALL(vkQueueSubmit2(vk_queue, 1, &submit_info, VK_NULL_HANDLE))))
             ERR("Failed to submit semaphore waits, vr %d.\n");
```
HansKristian-Work, line 18156: Missing "vr" here.

libs/vkd3d/command.c
```diff
@@ -18044,6 +17976,88 @@ static bool d3d12_command_queue_needs_cpu_waits_locked(struct d3d12_command_queu
...
+    }
+
+    vkd3d_queue_flush_waiters_locked(command_queue->vkd3d_queue, vk_queue, command_queue, vk_procs);
+
+    command_queue->last_submission_timeline_value = command_queue->vkd3d_queue->submission_timeline_count;
+
+    d3d12_command_queue_push_fence_waits_to_worker(command_queue);
+    d3d12_command_queue_reset_fence_waits(command_queue);
+
+    vkd3d_queue_release(command_queue->vkd3d_queue);
```
HansKristian-Work, line 18164: There should be no need to hold the lock here. It can be released after `command_queue->last_submission_timeline_value` assignment I think?

libs/vkd3d/command.c
```diff
@@ -18037,23 +18042,73 @@ static void d3d12_command_queue_add_wait(struct d3d12_command_queue *command_que
...
+    signal_semaphore.semaphore = command_queue->vkd3d_queue->submission_timeline;
+    signal_semaphore.value = command_queue->last_submission_timeline_value;
 
-    d3d12_command_queue_push_fence_waits_to_worker(command_queue);
-    d3d12_command_queue_reset_fence_waits(command_queue);
+    submit_info.signalSemaphoreInfoCount = 1;
+    submit_info.pSignalSemaphoreInfos = &signal_semaphore;
+
+    if ((vr = VK_CALL(vkQueueSubmit2(vk_queue, 1, &submit_info, VK_NULL_HANDLE))))
+        ERR("Failed to submit semaphore waits, vr %d.\n");
```
HansKristian-Work: Missing vr argument.

libs/vkd3d/vkd3d_private.h
doitsujin: derp

libs/vkd3d/command.c
doitsujin: Fixed.

libs/vkd3d/command.c
doitsujin: derp v2

libs/vkd3d/command.c
doitsujin: Think this is the same one as above?

HansKristian-Work (approved):
LGTM. Very nice rework overall.

PR #2246, opened 2024-12-06
Attempt to workaround Outlaws hang on NV

### LINKED ISSUE #2092: Star Wars outlaws Crashes
Please describe your issue as accurately as possible. Include screenshots or videos if relevant

### Software information
Name of the game, settings used etc.

### System information
- GPU: RTX 4070 / 5700XT
- Driver:560.35.03
- Wine version: proton GE 9-11
- VKD3D-Proton version: 2.13
[kernel_logs.txt](https://github.com/user-attachments/files/16769743/kernel_logs.txt)
[vkd3d log](https://github.com/user-attachments/files/16769744/vkd3d.log)

HansKristian-Work:
Fix #2092.

This doesn't make much sense, but it seems to fix the hang for me.
Breadcrumbs are pointing to "nothing", suggesting that the hang happens in internal driver code which does who knows what. Hopefully this is a viable workaround and we can report something actionable to NV.

### DIFF
--- a/include/vkd3d.h
+++ b/include/vkd3d.h
@@ -111,6 +111,7 @@ extern "C" {
 #define VKD3D_CONFIG_FLAG_INSTRUCTION_QA_CHECKS (1ull << 53)
 #define VKD3D_CONFIG_FLAG_TRANSFER_QUEUE (1ull << 54)
 #define VKD3D_CONFIG_FLAG_NO_GPU_UPLOAD_HEAP (1ull << 55)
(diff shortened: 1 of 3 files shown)


doitsujin:
Ah well, looks like this fixes the game for people.

LGTM if deemed ready, if we ever end up using this more we should probably add a `WARN` in case something is submitted twice though.

HansKristian-Work:
Ye, it's not hard to add validation on our end if we need this workaround more, but it really should be fixed in driver. I've sent a report to NV about this.

PR #2248, opened 2024-12-09
vkd3d: Don't signal fences after device lost or breadcrumb timeout.

HansKristian-Work:
We don't want to let application keep going, thinking it can start freeing memory. This is potentially dangerous, and also muddles up validation.


HansKristian-Work:
Derp, we already did this.

Issue #2261, opened 2024-12-27
Regression on drivers without sparse binding support (Turnip)


valentineburley:
Turnip doesn't currently support sparse binding. Since https://github.com/HansKristian-Work/vkd3d-proton/commit/40e4b22fdc6d34a8b05334835c5d77c12c7feaf3, vkd3d-proton tests are all crashing in [MesaCI](https://gitlab.freedesktop.org/mesa/mesa/-/jobs/68303028#L2780). This issue most likely occurs on all drivers without sparse.

### Software information
vkd3d-proton-tests

### System information
- GPU: Qualcomm Adreno (any)
- Driver: Turnip (main)
- Wine version:  any
- VKD3D-Proton version: 2.14

### Log files
```
Thread 1 "d3d12" received signal SIGSEGV, Segmentation fault.
0x0000007ff7a16d34 in ?? () from ~/vkd3d-proton-tests/x64/libvkd3d-proton-d3d12core.so
(gdb) bt
#0  0x0000007ff7a16d34 in ?? () from ~/vkd3d-proton-tests/x64/libvkd3d-proton-d3d12core.so
#1  0x0000007ff7996358 in ?? () from ~/vkd3d-proton-tests/x64/libvkd3d-proton-d3d12core.so
#2  0x0000007ff79968d0 in ?? () from ~/vkd3d-proton-tests/x64/libvkd3d-proton-d3d12core.so
#3  0x0000007ff7997194 in ?? () from ~/vkd3d-proton-tests/x64/libvkd3d-proton-d3d12core.so
#4  0x0000007ff79e27a8 in ?? () from ~/vkd3d-proton-tests/x64/libvkd3d-proton-d3d12core.so
#5  0x0000007ff797b3a8 in ?? () from ~/vkd3d-proton-tests/x64/libvkd3d-proton-d3d12core.so
#6  0x0000007ff7fd1344 in ?? () from ~/vkd3d-proton-tests/x64/libvkd3d-proton-d3d12.so
#7  0x0000007ff7fd1864 in D3D12CreateDevice () from ~/vkd3d-proton-tests/x64/libvkd3d-proton-d3d12.so
#8  0x0000003000152388 in create_device () at ../../vkd3d-proton-src/tests/d3d12_crosstest.h:587
#9  0x000000300015255c in init_adapter_info () at ../../vkd3d-proton-src/tests/d3d12_crosstest.h:668
[...]

PR #2262, opened 2024-12-29
vkd3d: Fix crash on hardware that doesn't support sparse.

Sonicadvance1:
Ever since commit 40e4b22fdc6d34a8b05334835c5d77c12c7feaf3, vkd3d would try and allocate a sparse queue unconditionally even if the hardware didn't support sparse. The cause of this comes from the logic inside of `d3d12_device_create_vkd3d_queues` which would skip any queues that were `VK_QUEUE_FAMILY_IGNORED` but was initializing `queue_family_indices` to zero. Zero is a valid queue index which would cause logic later to always try and allocate a queue unconditionally. The simple fix is to initialize these indices to be VK_QUEUE_FAMILY_IGNORED (which is ~0U) instead of zero.

Fixes a crash on Turnip (and any other driver that doesn't support sparse), when running any d3d12 game.

### DIFF
--- a/libs/vkd3d/device.c
+++ b/libs/vkd3d/device.c
@@ -2808,7 +2808,10 @@ static HRESULT d3d12_device_create_vkd3d_queues(struct d3d12_device *device,
     device->unique_queue_mask = 0;
     device->queue_family_count = 0;
     memset(device->queue_families, 0, sizeof(device->queue_families));
-    memset(device->queue_family_indices, 0, sizeof(device->queue_family_indices));
+    for (i = 0; i < VKD3D_QUEUE_FAMILY_COUNT; i++)
+    {
+        device->queue_family_indices[i] = VK_QUEUE_FAMILY_IGNORED;
+    }
 
     for (i = 0, k = 0; i < VKD3D_QUEUE_FAMILY_COUNT; i++)
     {


Sonicadvance1:
Just noticed there was a bug report open for this #2261

doitsujin:
I still don't really understand the issue, why exactly is there a crash in the first place?

Sonicadvance1:
> I still don't really understand the issue, why exactly is there a crash in the first place?
> 
> Unconditionally creating a sparse queue is obviously unintended, but I don't really see why it would be causing issues.

Previously, this chains in to `d3d12_device_reserve_internal_sparse_queue` which checks if the queue family is ignored or not. It doesn't get ignored so it calls `d3d12_device_allocate_vkd3d_queue` passing in `device->queue_families[VKD3D_QUEUE_FAMILY_SPARSE_BINDING]` which is a nullptr, The first loop inside that function dereferences the nullptr and crashes. 

If the list was properly ignored then `d3d12_device_allocate_vkd3d_queue` never gets called.

Alternatively this can be thought that because `d3d12_device_create_vkd3d_queues` previously skips ignored queue family indices, it always initialized all `device->queue_families` array members to nullptr with all `device->queue_family_indices` indices being set to valid index zero. So any time that a `device->queue_family_indices` member was checked for invalid or not, there was a logic desync between the valid index versus a queue family that wasn't ever initialized.

doitsujin:
Yeah it wasn't obvious where the crash location was, this explains it.

HansKristian-Work:
Actually, looking at the code some more, looking at queue_family_indices is just straight up broken.

`device->queue_family_indices[device->queue_family_count++] = info->vk_family_index;`

This is the compacted list of family indices that is used for purposes of CONCURRENT sharing, so the d3d12_device_reserve_internal_sparse_queue function is just checking for the wrong thing.

PR #2268, opened 2025-01-03
Queue profile improvements

HansKristian-Work:
Makes it easier to debug CPU submission overhead, etc.


HansKristian-Work:
The queue refactoring turned out to be a bad idea after all, so dropping for now.

PR #2269, opened 2025-01-04
Only allocate device memory if explicitly requested

### LINKED ISSUE #2258: HVV -> Sysmem fallback logic is flawed
Current NVK exposes the following memory types, in this order:
```
[0] DEVICE_LOCAL
[1] DEVICE_LOCAL | HOST_VISIBLE | HOST_COHERENT
[2] HOST_VISIBLE | HOST_COHERENT | HOST_CACHED
```

There is no uncached system memory type (yet; will probably get added).

When exceeding our fixed HVV budget with allocations from memory type 1, our fallback logic would only scan for `DEVICE_LOCAL | HOST_VISIBLE` and land on memory type 1 again instead of memory type 2, thus erroring out and causing games to crash unless `no_upload_hvv` is used, as seen in `https://gitlab.freedesktop.org/mesa/mesa/-/issues/11177`.

The proper fix here would be to have more sophisticated logic for allocating device memory, and feed back the memory type mask of failed allocations to the caller. There are more scenarios that are theoretically allowed by the spec where our current logic would brea, (e.g. two identical HVV types.

doitsujin:
Matches DXVK behaviour and should fix #2258, but needs testing.

The idea here is that if we want to allocate system memory (as a fallback or otherwise, e.g. `UPLOAD` heaps in an app that also uses GPU upload heaps), we should never try to allocate from device-local memory unless all supported memory types for a given allocation are device-local.

Also fixes the possible case where a fallback allocation with no property flags would attempt to allocate from `DEVICE_LOCAL` again. This only hasn't been an issue so far because most drivers allow oversubscribing VRAM, but this is not guarateed.

### DIFF
--- a/libs/vkd3d/memory.c
+++ b/libs/vkd3d/memory.c
@@ -998,7 +998,7 @@ HRESULT vkd3d_allocate_device_memory(struct d3d12_device *device,
              * E.g. there exists no memory type that is not DEVICE_LOCAL and covers both RT and DS.
              * For this case, we have no choice but to not allocate,
              * and defer actual memory allocation to CreatePlacedResource() time.
-			 * NVIDIA bug reference for fixing this case: 2175829. */
+             * NVIDIA bug reference for fixing this case: 2175829. */
             WARN("Memory allocation failed, but it is not possible to fallback to system memory here. Deferring allocation.\n");
             return hr;
         }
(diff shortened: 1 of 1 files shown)


Sid127:
Confirmed, games no longer run into "out of video memory" errors on NVK (current git as of writing this comment)
Tested with Amid Evil DX12 and Marvel Rivals

HansKristian-Work (approved):
LGTM.

PR #2287, opened 2025-01-15
Experimental workgraphs - Phase 3

HansKristian-Work:
More general refactoring.

### DIFF
--- a/libs/vkd3d/raytracing_pipeline.c
+++ b/libs/vkd3d/raytracing_pipeline.c
@@ -1507,24 +1507,6 @@ static bool vkd3d_stage_is_global_group(VkShaderStageFlagBits stage)
     }
 }
 
-static VkShaderModule create_shader_module(struct d3d12_device *device, const void *data, size_t size)
-{
-    const struct vkd3d_vk_device_procs *vk_procs = &device->vk_procs;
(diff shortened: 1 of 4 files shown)


libs/vkd3d/raytracing_pipeline.c
```diff
@@ -373,6 +373,16 @@ static void STDMETHODCALLTYPE d3d12_state_object_properties_SetPipelineStackSize
     object->pipeline_stack_size = stack_size_in_bytes;
 }
 
+static D3D12_PROGRAM_IDENTIFIER * STDMETHODCALLTYPE d3d12_state_object_properties_GetProgramIdentifier(
+        d3d12_state_object_properties_iface *iface,
+        D3D12_PROGRAM_IDENTIFIER *ret,
+        LPCWSTR pProgramName)
+{
+    TRACE("iface %p, ret %p, pProgramName %s!\n", iface, ret, debugstr_w(pProgramName));
```
doitsujin, line 381: Should be `FIXME` / add "stub" if this is an actual stub

doitsujin: in that case the ! at the end is a bit weird.

doitsujin (approved):

PR #2288, opened 2025-01-15
Experimental workgraph - Phase 4

HansKristian-Work:
This is the first one with some real meat to it.

- Adds a missing meta .md document I forgot to include in the first rebase.
- Adds meta operations for workgraphs. The high-level details are explained in the .md, but the TLDR is that we have four shaders:

### Setup GPU input

Only used when processing a GPU_INPUT entry point. Just generates indirect dispatches.

### distribute_workgroups

- Counts number of records produced by each node, prefix sums it and allocates space for the expanded offsets.
- Sets up indirect command for each node. TODO: Potentially allocate a EXT_dgc entry for non-empty nodes.
- Sets up one indirect command for payload expander.

### distribute_payload_offsets

Loops over { node index, payload offset, count } entries and expands this into a simple array of u32 payload offsets which will be consumed by the next stage's nodes. The space for these offsets were allocated by distribute_workgroups.

### complete_compaction

This was recently added. Adjusts the broadcast amplification rate based on the data we analyzed in payload_offsets, once per node.


docs/workgraphs.md
```diff
@@ -1269,3 +1269,75 @@ NVIDIA performance is also quite underwhelming here.
...
+is another spin on the classic tile-deferred renderer.
+It's a little more realistic since the binner and classifier is emitting to Wave32 broadcast nodes,
+where each node is specialized in working on a specific material type. It's not using MaxDispatchGrid either,
+which should make scheduling very simple to deal with.
+
+#### RTX 4070 (native)
+
+GPU performance is extremely similar between workgraphs and plain compute. However, this is slightly concerning,
+since the compute shading path is [quite inefficient](https://github.com/NVIDIAGameWorks/donut_examples/blob/main/examples/work_graphs/deferred_shading.hlsl).
+It evaluates the materials in a big switch block, where the workgraph path can execute straight forward code.
```
Josh015, line 1285:
```suggestion
It evaluates the materials in a big switch block, where the workgraph path can execute straightforward code.
```

HansKristian-Work: I'm not particularly interested in drive by grammar review here.

libs/vkd3d/shaders/cs_workgraph_distribute_workgroups.comp
```diff
@@ -0,0 +1,204 @@
...
+
+layout(buffer_reference, buffer_reference_align = 8, std430) buffer NodeAtomics
+{
+	uint payload_atomic;
+	uint fused_atomic;
+	uint node_counts[];
+};
+
+// Sharing across threads requires coherent, even within a subgroup.
+layout(buffer_reference, buffer_reference_align = 16, std430) coherent buffer IndirectCommandsBuffer
```
doitsujin, line 29: `subgroupcoherent` is a thing, although for that to do anything we'd also need to opt-into vulkan memory model for these shaders (`#pragma use_vulkan_memory_model`).

Might be worth looking into so drivers don't pessimize code gen too much.

libs/vkd3d/shaders/cs_workgraph_distribute_payload_offsets.comp
```diff
@@ -0,0 +1,265 @@
...
+		output_offset = subgroupBroadcastFirst(output_offset);
+	}
+
+	uvec4 ballot = subgroupBallot(count != 0u);
+
+	while (any(notEqual(ballot, uvec4(0))))
+	{
+		uint lane = subgroupBallotFindLSB(ballot);
+
+		// TODO: Is there a more elegant way that is just as fast and fully portable?
```
doitsujin, line 97: only thing that would come to mind is rewriting the loop as something like

```glsl
bool thread_mask = count != 0;

while (subgroupAny(thread_mask)) {
  uint lane = subgroupBallotFindLSB(subgroupBallot(thread_mask));
  thread_mask = thread_mask && gl_SubgroupInvocationID > lane;
  ...
}
```

HansKristian-Work: I think that comparison would require vector ops.

libs/vkd3d/meta.c
```diff
@@ -2044,6 +2044,92 @@ static HRESULT vkd3d_sampler_feedback_ops_init(struct vkd3d_sampler_feedback_res
...
+        map_entries[i].size = sizeof(uint32_t);
+        map_entries[i].constantID = i;
+    }
+
+    spec_info.mapEntryCount = ARRAY_SIZE(map_entries);
+    spec_info.pMapEntries = map_entries;
+    spec_info.pData = spec_data;
+    spec_info.dataSize = ARRAY_SIZE(map_entries) * sizeof(uint32_t);
+    spec_data[0] = device->device_info.vulkan_1_1_properties.subgroupSize;
+    spec_data[1] = device->device_info.vulkan_1_1_properties.subgroupSize;
```
doitsujin, line 2096: shouldn't we need to use `maxSubgroupSize` here for correctness, or alternatively set a required subgroup size? We use SPIR-V 1.6 for built-in shaders, so we'd implicitly get `VARYING_SUBGROUP_SIZE` as well.

HansKristian-Work: device->device_info.vulkan_1_1_properties.subgroupSize should be equal to maxSubgroupSize, but I guess it won't hurt to use size control to avoid the != 0 checks.

doitsujin (approved):
Should be good now.

PR #2296, opened 2025-01-20
Add workaround to skip NULL sparse tile updates.

HansKristian-Work:


libs/vkd3d/command.c
```diff
@@ -17422,8 +17422,20 @@ static void STDMETHODCALLTYPE d3d12_command_queue_UpdateTileMappings(ID3D12Comma
...
             if (range_flags)
+            {
                 range_flag = range_flags[range_idx];
 
+                /* AMDGPU seems to have a strange bug where remapping a page to NULL can cause an impossible
+                 * page table issue where it's now possible to fault on a PRT page. */
+                if (range_flag == D3D12_TILE_RANGE_FLAG_NULL &&
+                        (vkd3d_config_flags & VKD3D_CONFIG_FLAG_SKIP_NULL_SPARSE_TILES) &&
+                        command_queue->device->device_info.properties2.properties.vendorID == VKD3D_VENDOR_ID_AMD &&
+                        !(vkd3d_config_flags & VKD3D_CONFIG_FLAG_SKIP_DRIVER_WORKAROUNDS))
```
doitsujin, line 17433: Since this is another kernel bug, can we introduce a device bit similar to `amdgpu_broken_clearvram` for this? *Surely* we expect this to get fixed in some future kernel version, right?

HansKristian-Work: We certainly can. It's not really confirmed what is going on yet though, but it's the only logical explanation after a month of suffering.

doitsujin (approved):

PR #2299, opened 2025-01-21
Experimental workgraphs - Phase 5

HansKristian-Work:
Hooks up some required helpers.

- Node metadata parsing
- Helper for workgraph compatible pipeline layout. Root parameters are now placed in push UBO and workgraph specific metadata is placed in push constants. This is better than the alternative, since we need to update workgraph block every node execution, but the push UBO can be reused across all nodes. There are some TODOs, but that can be left for later.


libs/vkd3d/state.c
```diff
@@ -1560,6 +1560,69 @@ HRESULT d3d12_root_signature_create_local_static_samplers_layout(struct d3d12_ro
...
+    /* If we're already using push UBO block, we just need to modify the push range. */
+    /* TODO: Local sampler set. */
+    uses_push_ubo = !!(root_signature->compute.flags & VKD3D_ROOT_SIGNATURE_USE_PUSH_CONSTANT_UNIFORM_BLOCK);
+    range.stageFlags = VK_SHADER_STAGE_COMPUTE_BIT;
+    range.offset = 0;
+    range.size = sizeof(struct vkd3d_shader_node_input_push_signature);
+
+    if (root_signature->root_descriptor_push_mask)
+    {
+        FIXME("The root signature is already using push descriptors, cannot add another push descriptor set. Make sure to use VKD3D_CONFIG=force_raw_va_cbv on NVIDIA.\n");
```
doitsujin, line 1582: Do any of the existing demos hit this path, and is it *likely* that any real shipping apps will?

If so then this sounds like it's going to turn into a bit of a maintenance nightmare with app workarounds...

doitsujin (approved):
just one question

libs/vkd3d/state.c
HansKristian-Work: One of them does, indeed. I think the real solution will be to augment the push descriptor set to support both root CBVs + the root params CBV, but that's a significant refactor that is somewhat orthogonal.

PR #2303, opened 2025-01-22
Clock calibration fixes

doitsujin:
winevulkan somehow screws up the EXT, and there was a scary bug where we'd pass an uninitialized struct to the function on the fallback path.


zmike (approved):

libs/vkd3d/device.c
```diff
@@ -80,12 +80,12 @@ static const struct vkd3d_optional_extension_info optional_device_extensions[] =
     VK_EXTENSION(KHR_SHADER_MAXIMAL_RECONVERGENCE, KHR_shader_maximal_reconvergence),
     VK_EXTENSION(KHR_SHADER_QUAD_CONTROL, KHR_shader_quad_control),
     VK_EXTENSION(KHR_COMPUTE_SHADER_DERIVATIVES, KHR_compute_shader_derivatives),
+    VK_EXTENSION(KHR_CALIBRATED_TIMESTAMPS, KHR_calibrated_timestamps),
```
HansKristian-Work, line 83: Should we not check both? The KHR one is fairly recent.

HansKristian-Work (changes requested):

libs/vkd3d/device.c
HansKristian-Work: Seems like the cutoff is Mesa 24.0 and NV 535. Maybe that's okay?

HansKristian-Work: Can we confirm that stable Deck ships the KHR?

doitsujin: vulkaninfo says yes.

On the Nvidia side we kinda dropped <535 already anyway, so I don't think that's a concern here. The KHR is like one and a half years old already anyway.

HansKristian-Work (approved):

PR #2307, opened 2025-01-24
Batch sparse binds

HansKristian-Work:


doitsujin:
For hazard tracking we could also just write a bitmask of dirty tiles for each resource and do a look-up on that. The downside would be an extra malloc of `O(n_tiles_in_resource)`, but we have that exact level of runtime complexity in `UpdateTileMappings` already anyway.

Since we're passing around the tile indices already, this should be relatively easy to do and has the advantage that we get 100% accurate hazard detection in linear time.

HansKristian-Work:
Yes, a bitmask tracker would be simpler and way faster. I'll rewrite to that on Monday.

ryzendew:
i can Confirm FF16 Stutter issue is gone and the game is super smooth now

HansKristian-Work:
> i can Confirm FF16 Stutter issue is gone and the game is super smooth now

Which GPU and where in the game did you test? In Lostwing where I tested the game breaks batching due to back-to-back sparse -> signal -> sparse -> signal ad nauseam.

doitsujin (approved):

PR #2314, opened 2025-01-27
vkd3d: enable ds<->color copies using maintenance8

zmike:
this avoids renderpass copies for allowed formats, though only DIRECT/GRAPHICS queue is supported by maintenance8

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -7720,7 +7720,39 @@ struct d3d12_image_copy_barrier
     bool dst_is_depth_stencil;
 };
 
-static void d3d12_command_list_copy_image_barrier_flags(struct d3d12_image_copy_barrier *barrier,
+/* maintenance8 enables copies between:
+    • 32-bit depth (VK_FORMAT_D32_SFLOAT, VK_FORMAT_D32_SFLOAT_S8_UINT)
+        ◦ VK_FORMAT_R32_SFLOAT, VK_FORMAT_R32_SINT, VK_FORMAT_R32_UINT
+    • 24-bit depth (VK_FORMAT_X8_D24_UNORM_PACK32, VK_FORMAT_D24_UNORM_S8_UINT)
+        ◦ VK_FORMAT_R32_SFLOAT, VK_FORMAT_R32_SINT, VK_FORMAT_R32_UINT
+    • 16-bit depth (VK_FORMAT_D16_UNORM, VK_FORMAT_D16_UNORM_S8_UINT)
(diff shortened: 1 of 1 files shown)


libs/vkd3d/command.c
```diff
@@ -7720,10 +7720,90 @@ struct d3d12_image_copy_barrier
...
+    if ((ds_format->vk_aspect_mask & ds_bits) == 0 ||
+        (color_format->vk_aspect_mask & VK_IMAGE_ASPECT_COLOR_BIT) == 0)
+        return false;
+
+    /* try the stencil copy case first: always 8bit */
+    if (ds_format->vk_aspect_mask & VK_IMAGE_ASPECT_STENCIL_BIT &&
+        color_format->byte_count == 1)
+        return d3d12_command_list_check_ds_color_copy_compatibility_8bit(color_format);
+
+    switch (ds_format->vk_format)
```
HansKristian-Work, line 7788: Explicitly checking formats is overzealous. The list in maintenance8 is meant to be exhaustive to any possible format D3D12 could use. I think this can be just simplified to "one is color, another is depth-stencil".

libs/vkd3d/command.c
```diff
@@ -7720,10 +7720,90 @@ struct d3d12_image_copy_barrier
...
+        default:
+            return false;
+    }
+}
+
 static void d3d12_command_list_copy_image_barrier_flags(struct d3d12_image_copy_barrier *barrier,
         const struct d3d12_resource *dst_resource, const struct vkd3d_format *dst_format,
         const struct d3d12_resource *src_resource, const struct vkd3d_format *src_format,
-        const VkImageCopy2 *region, bool overlapping_subresource)
+        const VkImageCopy2 *region, bool overlapping_subresource, bool allow_ds_color_copies)
```
doitsujin, line 7752: a cleaner way would be to just pass the command list as the first argument, so we can check for queue and maintenance8 support through that. The function is called `d3d12_command_list_*` anyway.

doitsujin (commented):
Style nit.

As for what Hans-Kristian said, I'm not sure off the top of my head if something like D32->RGBA8 is legal or not. In D3D11 it definitely is, but at the same time I don't think our current fallback would work with that.

I also don't think we should ever encounter D24S8 here, D3D12 largely treats depth-stencil as multi-planar and CopyResource doesn't allow mismatched formats, so if there's an aspect mismatch here we should only ever have one single aspect set on each image.

libs/vkd3d/command.c
zmike: The z is for zealous.

HansKristian-Work (approved):
LGTM. Passes test on RADV and new VVL doesn't trip errors in the edge case copy test, so should be fine.

PR #2323, opened 2025-01-31
vkd3d: add missing memory flush on cpu clear operations

djdeath:
Non coherent host visible memory touched from the host needs to be flushed before it gets used on the device.

Fixes VKD3D_TEST_FILTER=test_execute_indirect ./build/tests/d3d12 on Anv/Meteorlake which appears to be the only platform to have a memory type that is HOST_VISIBLE | HOST_CACHED memory type (without HOST_COHERENT).

Found-by: José Roberto de Souza <jose.souza@intel.com>

### DIFF
--- a/libs/vkd3d/memory.c
+++ b/libs/vkd3d/memory.c
@@ -489,9 +489,21 @@ static void vkd3d_memory_transfer_queue_clear_allocation(struct vkd3d_memory_tra
 
     if (allocation->cpu_address)
     {
+        const struct d3d12_device *device = queue->device;
+        const struct vkd3d_vk_device_procs *vk_procs;
+        VkMappedMemoryRange mapped_range = { 0 };
+
+        vk_procs = &device->vk_procs;
+
+        mapped_range.memory = allocation->device_allocation.vk_memory;
+        mapped_range.offset = allocation->offset;
+        mapped_range.size = allocation->resource.size;
+
         /* Probably faster than doing this on the GPU
          * and having to worry about synchronization */
         memset(allocation->cpu_address, 0, allocation->resource.size);
+
+        VK_CALL(vkFlushMappedMemoryRanges(device->vk_device, 1, &mapped_range));
     }
     else if (allocation->resource.vk_buffer)
     {
--- a/libs/vkd3d/resource.c
+++ b/libs/vkd3d/resource.c
@@ -2187,6 +2187,8 @@ static HRESULT STDMETHODCALLTYPE d3d12_resource_WriteToSubresource(d3d12_resourc
     struct d3d12_resource *resource = impl_from_ID3D12Resource2(iface);
     struct vkd3d_subresource_layout *subresource_layout;
     struct d3d12_device *device = resource->device;
+    const struct vkd3d_vk_device_procs *vk_procs;
+    VkMappedMemoryRange mapped_range = { 0 };
     struct vkd3d_format_footprint footprint;
     const struct vkd3d_format *format;
     uint32_t plane_idx;
@@ -2194,6 +2196,8 @@ static HRESULT STDMETHODCALLTYPE d3d12_resource_WriteToSubresource(d3d12_resourc
     VkOffset3D offset;
     uint8_t *dst_data;
     D3D12_BOX box;
+
+    vk_procs = &device->vk_procs;
 
     TRACE("iface %p, src_data %p, src_row_pitch %u, src_slice_pitch %u, "
             "dst_sub_resource %u, dst_box %s.\n",
@@ -2258,6 +2262,12 @@ static HRESULT STDMETHODCALLTYPE d3d12_resource_WriteToSubresource(d3d12_resourc
     vkd3d_format_copy_data(format, src_data, src_row_pitch, src_slice_pitch, dst_data,
             subresource_layout->row_pitch, subresource_layout->depth_pitch, extent.width, extent.height, extent.depth);
 
+    mapped_range.memory = resource->mem.device_allocation.vk_memory;
+    mapped_range.offset = resource->mem.offset + subresource_layout->offset;
+    mapped_range.size = (extent.depth - 1) * subresource_layout->depth_pitch +
+      extent.height * subresource_layout->row_pitch;
+    VK_CALL(vkFlushMappedMemoryRanges(device->vk_device, 1, &mapped_range));
+
     return vkd3d_memory_transfer_queue_write_subresource(&device->memory_transfers,
             resource, dst_sub_resource, offset, extent);
 }
@@ -2267,12 +2277,17 @@ static HRESULT STDMETHODCALLTYPE d3d12_resource_ReadFromSubresource(d3d12_resour
         UINT src_sub_resource, const D3D12_BOX *src_box)
 {
     struct d3d12_resource *resource = impl_from_ID3D12Resource2(iface);
+    struct d3d12_device *device = resource->device;
     struct vkd3d_subresource_layout *subresource_layout;
+    const struct vkd3d_vk_device_procs *vk_procs;
+    VkMappedMemoryRange mapped_range = { 0 };
     struct vkd3d_format_footprint footprint;
     const struct vkd3d_format *format;
     uint32_t plane_idx;
     uint8_t *src_data;
     D3D12_BOX box;
+
+    vk_procs = &device->vk_procs;
 
     TRACE("iface %p, dst_data %p, dst_row_pitch %u, dst_slice_pitch %u, "
             "src_sub_resource %u, src_box %s.\n",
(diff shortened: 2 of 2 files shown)


libs/vkd3d/memory.c
```diff
@@ -1325,6 +1337,10 @@ static HRESULT vkd3d_memory_allocation_init(struct vkd3d_memory_allocation *allo
         return hr;
     }
 
+    if ((type_flags & VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT) &&
```
HansKristian-Work, line 1340: Shouldn't this be the type we actually end up allocating? This seems very wrong.

libs/vkd3d/memory.c
```diff
@@ -492,6 +492,18 @@ static void vkd3d_memory_transfer_queue_clear_allocation(struct vkd3d_memory_tra
...
+        if (allocation->flags & VKD3D_ALLOCATION_FLAG_HOST_NON_COHERENT)
+        {
+            struct d3d12_device *device = queue->device;
+            const struct vkd3d_vk_device_procs *vk_procs = &device->vk_procs;
+            VkMappedMemoryRange mapped_range = { 0 };
+            mapped_range.memory = allocation->device_allocation.vk_memory;
+            mapped_range.offset = allocation->offset;
+            mapped_range.size = allocation->resource.size;
+
+            VK_CALL(vkFlushMappedMemoryRanges(device->vk_device, 1, &mapped_range));
```
HansKristian-Work, line 505: Might make more sense to just unconditionally call FlushMappedMemoryRanges. It seems like we do that for ID3D12Resource::Map/Unmap.

However, I think we're likely missing flushes in way more places going by this fix ...

djdeath: Yeah, I'm afraid of that too.

I'll scout around `->cpu_address` accesses.

libs/vkd3d/memory.c
djdeath: Dropped, doing the flush unconditionally now.

libs/vkd3d/resource.c
```diff
@@ -2258,6 +2260,12 @@ static HRESULT STDMETHODCALLTYPE d3d12_resource_WriteToSubresource(d3d12_resourc
     vkd3d_format_copy_data(format, src_data, src_row_pitch, src_slice_pitch, dst_data,
             subresource_layout->row_pitch, subresource_layout->depth_pitch, extent.width, extent.height, extent.depth);
 
+    mapped_range.memory = resource->mem.device_allocation.vk_memory;
+    mapped_range.offset = resource->mem.offset + subresource_layout->offset;
+    mapped_range.size = (extent.depth - 1) * subresource_layout->depth_pitch +
+      extent.height * subresource_layout->row_pitch;
+    VK_CALL(vkFlushMappedMemoryRanges(device->vk_device, 1, &mapped_range));
```
zehortigoza, line 2267:
```
D3D12_RANGE range;

range.Begin = 0;
range.End = resource->res.size;
d3d12_resource_flush_range(resource, 0, &range);
```

libs/vkd3d/memory.c
djdeath: Added vkd3d_format_copy_data locations too.

libs/vkd3d/resource.c
```diff
@@ -2326,6 +2337,12 @@ static HRESULT STDMETHODCALLTYPE d3d12_resource_ReadFromSubresource(d3d12_resour
     src_data += subresource_layout->offset + vkd3d_format_get_data_offset(format,
             subresource_layout->row_pitch, subresource_layout->depth_pitch, src_box->left, src_box->top, src_box->front);
 
+    mapped_range.memory = resource->mem.device_allocation.vk_memory;
+    mapped_range.offset = resource->mem.offset + subresource_layout->offset;
+    mapped_range.size = (src_box->back - src_box->front - 1) * subresource_layout->depth_pitch +
+      (src_box->bottom - src_box->top) * subresource_layout->row_pitch;
+    VK_CALL(vkInvalidateMappedMemoryRanges(device->vk_device, 1, &mapped_range));
```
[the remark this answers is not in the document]
djdeath: I don't know that the resource was written only from the host.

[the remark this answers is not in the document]
zehortigoza: I think it should be callers responsibility to flush after any CPU write and invalidate or flush CPU caches before sending buffer to GPU.

HansKristian-Work:
One question I have though. HOST + COHERENT is guaranteed to exist according to spec, and in most of these cases, I expect that type is used. Is this particular Intel iGPU non-compliant or something?

HansKristian-Work:
What's the exact error line you see when running vkd3d-proton tests?

djdeath:
> What's the exact error line you see when running vkd3d-proton tests?

It's flaky, so it's not always the same, but usually the first : 
```
    check_sub_resource_uint(context.render_target, 0, queue, command_list, 0xff00ff00, 0);
```

https://github.com/HansKristian-Work/vkd3d-proton/blob/master/tests/d3d12_command.c#L3442

zehortigoza:
> One question I have though. HOST + COHERENT is guaranteed to exist according to spec, and in most of these cases, I expect that type is used. Is this particular Intel iGPU non-compliant or something?
> 
> Thinking about it, it's possible the only time non-COHERENT memory may be allocated is for CUSTOM or READBACK user heaps. Internally in vkd3d-proton for staging uploads I don't think we should ever allocate non-COHERENT memory.

This devices provides 2 memory types:
- host visible + coherent
- host visible + cached

In the proton tests that I ran usually only the render target buffer is allocated with the non-coherent memory type.

HansKristian-Work:
And what is the order that those two types are returned?

djdeath:
> And what is the order that those two types are returned?

https://gitlab.freedesktop.org/mesa/mesa/-/blob/main/src/intel/vulkan/i915/anv_device.c?ref_type=heads#L189

0 : DEVICE_LOCAL | HOST_VISIBLE | HOST_COHERENT
1 : DEVICE_LOCAL | HOST_VISIBLE | HOST_CACHED

HansKristian-Work:
It should be impossible for a render target to pick HOST_VISIBLE | CACHED in that scenario. I'm confused why this PR even helps then.

djdeath:
I think it's not the render target but rather the buffer in which the RT data is copied into.

That buffer gets cleared on the CPU internally before use, but the memset writes there are not getting flushed, until it's read back. So GPU copy & memset get garbbled.

HansKristian-Work:
> That buffer gets cleared on the CPU internally before use, but the memset writes there are not getting flushed, until it's read back. So GPU copy & memset get garbbled.

Ok, now it makes sense. So this should limit any workaround to:

- Resource clears
- ID3D12Resource::Map/Unmap
- WriteToSubresource / ReadFromSubresource

libs/vkd3d/memory.c
```diff
@@ -489,9 +489,18 @@ static void vkd3d_memory_transfer_queue_clear_allocation(struct vkd3d_memory_tra
 
     if (allocation->cpu_address)
     {
+        struct d3d12_device *device = queue->device;
+        const struct vkd3d_vk_device_procs *vk_procs = &device->vk_procs;
```
HansKristian-Work, line 493: Move this line above. Need inverse :christmas_tree: order of declarations.

libs/vkd3d/resource.c
```diff
@@ -2187,13 +2187,15 @@ static HRESULT STDMETHODCALLTYPE d3d12_resource_WriteToSubresource(d3d12_resourc
     struct d3d12_resource *resource = impl_from_ID3D12Resource2(iface);
     struct vkd3d_subresource_layout *subresource_layout;
     struct d3d12_device *device = resource->device;
+    const struct vkd3d_vk_device_procs *vk_procs = &device->vk_procs;
```
HansKristian-Work, line 2190: inverse :christmas_tree:

HansKristian-Work: Seems like you need to split up the lines in two here.

djdeath:
> > That buffer gets cleared on the CPU internally before use, but the memset writes there are not getting flushed, until it's read back. So GPU copy & memset get garbbled.
> 
> Ok, now it makes sense. So this should limit any workaround to:
> 
> * Resource clears
> * ID3D12Resource::Map/Unmap
> * WriteToSubresource / ReadFromSubresource

Map/Unmap already have their invalidate/flush it seems.

libs/vkd3d/memory.c
djdeath: hopefully I've done that right

HansKristian-Work (approved):
LGTM

PR #2333, opened 2025-02-07
Fence worker callback + work around MHW sparse bug

doitsujin:
The callback refactor moves all the hardcoded stuff that we do in the fence worker to callbacks, and stores callback data within the `vkd3d_fence_wait_info` structure. 

For fence waits it doesn't look like there was any reason at all to keep the fence objects alive, so I removed that part.

Lastly, this adds some tracking to keep sparse resources alive through page table updates; otherwise, Monster Hunter Wilds screws up synchronization somewhere and will either crash or hang on Nvidia when starting the benchmark with DLSS FG enabled.


HansKristian-Work (approved):
LGTM. Nice cleanup.

PR #2344, opened 2025-02-14
vkd3d: Remove shader quirk for Rebirth.

HansKristian-Work:
The game updated today.

The shader hash in question changed to f6ce99fa20cc4da2 for reference, and it seems like the game now adds a proper UAV barrier.


doitsujin (approved):
yay

PR #2346, opened 2025-02-17
Enable the SSBO fast path for NVK

gfxstrand:
NVK should basically follow the same rules as the proprietary driver here.  We advertise 16B alignments so that we can freely vectorize inside the compiler.  However, as long as you are loading scalars, we shouldn't up-convert those to vectors behind your back.


HansKristian-Work:
Commit should be prepended with `vkd3d: ` in this case.

gfxstrand:
Having given this some more thought, I think I'd rather smash `minStorageBufferOffsetAlignment = 4` in NVK.  That way we can make sure NIR isn't going to do anything that would break things.

https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/33613

PR #2348, opened 2025-02-18
Implement ID3D12DeviceExt2 for cubin 64bit support

Saancreed:
I don't quite see the advantage of `pNext` just yet over simply reusing [NVAPI's own mechanism](https://github.com/NVIDIA/nvapi/blob/ce6d2a183f9559f717e82b80333966d19edb9c8c/nvapi.h#L18272-L18273) but maybe I'm too used to working with NVIDIA's APIs to know any better.

Temporary marked as draft until I update dxvk-nvapi's side to match.


Saancreed:
Corresponding dxvk-nvapi PR is here: https://github.com/jp7677/dxvk-nvapi/pull/276 (still quite WIP though)

But I suppose this is now ready for review… cc: @liam-middlebrook

libs/vkd3d/device_vkd3d_ext.c
```diff
@@ -258,7 +283,123 @@ static HRESULT STDMETHODCALLTYPE d3d12_device_vkd3d_ext_GetVulkanQueueInfoEx(d3d
...
+    {
+        case D3D12_GET_CUDA_INDEPENDENT_DESCRIPTOR_OBJECT_SURFACE:
+            imageViewHandleInfo.descriptorType = VK_DESCRIPTOR_TYPE_STORAGE_IMAGE;
+            break;
+        case D3D12_GET_CUDA_INDEPENDENT_DESCRIPTOR_OBJECT_TEXTURE:
+            FIXME("TEXTURE object type not supported.\n");
+            return E_FAIL;
+        case D3D12_GET_CUDA_INDEPENDENT_DESCRIPTOR_OBJECT_SAMPLER:
+            imageViewHandleInfo.descriptorType = VK_DESCRIPTOR_TYPE_SAMPLED_IMAGE;
+            break;
```
Saancreed, line 392: @liam-middlebrook I'd appreciate a closer look at this, maybe you'll have an idea how to handle the `TEXTURE` case.

Zebra2711: base on [#descriptorsets-combinedimagesampler](https://registry.khronos.org/vulkan/specs/latest/html/vkspec.html#descriptorsets-combinedimagesampler) and [nvapi.h#L18388](https://github.com/NVIDIA/nvapi/blob/ce6d2a183f9559f717e82b80333966d19edb9c8c/nvapi.h#L18388) i guess it could be `VK_DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER`?

liam-middlebrook: I believe that `D3D12_GET_CUDA_INDEPENDENT_DESCRIPTOR_OBJECT_TEXTURE` corresponds to `VK_DESCRIPTOR_TYPE_SAMPLED_IMAGE`.

It is D3D12_GET_CUDA_INDEPENDENT_DESCRIPTOR_`OBJECT_SAMPLER` that you cannot support grabbing the handle of independently, as `vkGetImageViewHandle64NVX()` does not allow you to query a sampler. This support would be useful for kernels which have `.target texmode_independent` set in their PTX, which as I recall is used by OpenCL predominantly.

From a DLSS/NGX perspective this is fine to not support. We only ever independently query surfaces, as `texmode_independent` does not work.

Saancreed: Thanks, I've adjusted the code for this, should be correct now :crossed_fingers:

PR #2351, opened 2025-02-19
Add test proving D3D12 AMD driver doing questionable things with bindless CBV

HansKristian-Work:


doitsujin:
Does the test pass on nv/warp (/intel?) native?

HansKristian-Work:
> Does the test pass on other d3d12 drivers too or do we need to restrict it to amd?

I'd imagine it passes on NV too, but not going to swap GPUs just to test it on Intel as well. I restricted the test to only run on AMD, since that's the focus.

PR #2352, opened 2025-02-20
Starfield workaround for flickering volumetrics

HansKristian-Work:


doitsujin (approved):
i do enjoy that spaghetti cleanup.

PR #2358, opened 2025-02-26
script: Update installscript

SveSop:
This will make it work with wine >=10.2 and older

Some "major" changes to wine-10.2 dropped the usage of `wine64` binary, so a lot of scripts broke as they depended on checking system32/syswow64 usage when using wine64/wine respectively.
This should do some checks to verify folder structures so it is useful again.

### DIFF
--- a/setup_vkd3d_proton.sh
+++ b/setup_vkd3d_proton.sh
@@ -43,42 +43,65 @@ if [ -n "$WINEPREFIX" ] && ! [ -f "$WINEPREFIX/system.reg" ]; then
 fi
 
 # find wine executable
-export WINEDEBUG=-all
+export WINEDEBUG="${WINEDEBUG:--all}"
 # disable mscoree and mshtml to avoid downloading
 # wine gecko and mono
-export WINEDLLOVERRIDES="mscoree,mshtml="
+export WINEDLLOVERRIDES="${WINEDLLOVERRIDES:-}${WINEDLLOVERRIDES:+,}mscoree,mshtml="
 
-wine="wine"
-wine64="wine64"
-wineboot="wineboot"
+wine_path="$(which wine 2>/dev/null)"
+wine64_path="$(which wine64 2>/dev/null)"
 
-# $PATH is the way for user to control where wine is located (including custom Wine versions).
-# Pure 64-bit Wine (non Wow64) requries skipping 32-bit steps.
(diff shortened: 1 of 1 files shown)


HansKristian-Work:
Seems reasonable to me, but I'd like someone that's more familiar with this script to give a nod that it works.

SveSop:
https://aur.archlinux.org/cgit/aur.git/tree/setup_dxvk.sh?h=dxvk-msvc-git#n55

Could also be useful, using a slightly different approach that is probably going to be a bit more more tested than what i did? 😄
I see recent updates, so might not be fully 100% (last couple of hours ago).

SveSop:
Hmpf.. Depending on how deep we wanna go, it does indeed have a couple of snafu's (same with dxvk script).

Case: System installed wine is wine-10.2, custom wine version is < 10.2 and 64-bit only. This will choose `/usr/bin/wine` (assuming it is in the path ofc) and run wineboot -u on the prefix = borked.

This turned out to be more complicated than at first glance tbh, and needs more work :cry:

SveSop:
This could do it, in the case that the system installed wine (in path) is different than the custom wine, as the user would be given a choice and hopefully choose the correct one.

```
wine_path="$(which wine 2>/dev/null)"
wine64_path="$(which wine64 2>/dev/null)"

if [ -n "$wine_path" ] && [ -n "$wine64_path" ] && [ "$(dirname "$wine_path")" != "$(dirname "$wine64_path")" ]; then
    echo "Multiple Wine installations detected:"
    echo "1) $wine_path"
    echo "2) $wine64_path"

    while true; do
        read -p "Select the Wine binary to use (1 or 2): " choice
        case "$choice" in
          1) wine="wine"; break ;;
          2) wine="wine64"; break ;;
          *) echo "Invalid choice. Please enter 1 or 2." ;;
        esac
    done
else
    if [ -n "$wine_path" ]; then
      wine="wine"
    fi
    if [ -n "$wine64_path" ]; then
      wine="wine64"
    fi
fi

winever=$($wine --version | grep wine)
if [ -z "$winever" ]; then
    echo "$wine:"' Not a wine executable. Check your $wine.' >&2
    exit 1
fi
echo "Using: $winever"
```

IroAlexis:
Can't we use `readelf -h /path/to/bin/wine` to determine the architecture of the binary header ?

HansKristian-Work (approved):

PR #2361, opened 2025-02-27
Validate GS input topology

doitsujin:
Clanfolk uses a GS that consumes points in a pipeline that inputs triangles. This (sort of) works in D3D11 on some drivers, but fails runtime validation in D3D12 and is illegal in Vulkan anyway - which causes the game to not render half its geometry when enabling one of the graphics options, but that's *actually the correct behaviour*.

Also fixes an oversight where a GS that consumes triangles but outputs something else would be tagged as outputting triangles, because `SpvExecutionModeTriangles` can mean different things. That said, we probably never hit this bug in the real world anyway and if we did, it would only affect line rasterization behaviour in some very specific edge case.


libs/vkd3d/state.c
```diff
@@ -4364,7 +4364,37 @@ static HRESULT d3d12_pipeline_state_graphics_create_shader_stages(
...
 
-static void d3d12_pipeline_state_graphics_handle_meta(struct d3d12_pipeline_state *state,
+static bool d3d12_pipeline_state_validate_gs_input_toplogy(struct d3d12_pipeline_state *state,
+        const struct vkd3d_shader_meta *gs_meta, uint32_t geometry_meta)
+{
+    struct d3d12_graphics_pipeline_state *graphics = &state->graphics;
+
+    switch (gs_meta->gs_input_topology)
+    {
+        case VK_PRIMITIVE_TOPOLOGY_POINT_LIST:
```
HansKristian-Work, line 4374: Is this correct for tessellation? I think geometry_meta needs to reflect the meta of the TES here, not PSO input.

HansKristian-Work: Ah. No, geometry_meta is updated on line 4453.

doitsujin: we're checking against TES in topologies that can actually be generated by TES (`geometry_meta` will implicitly reflect that). There shouldn't be any way for TES to emit points.

HansKristian-Work (approved):
LGTM

PR #2370, opened 2025-03-03
Fix xfb validation errors

doitsujin:
There's some jank with how xfb is supposed to work, the TL;DR here is that every xfb buffer referenced by `BeginTransformFeedbackEXT` needs to have a valid binding, and we can't bind `NULL` buffers (not allowed per spec, even with null descriptors enabled).

Additionally, the spec explicitly bans changing pipelines while xfb is active, so just end the render pass when that happens. The most common usage pattern by far is to do a single draw per xfb pass anyway, so this won't suddenly become a source of perf issues in the few games that use xfb.

Fixes ##2368.


HansKristian-Work (approved):
LGTM

PR #2391, opened 2025-03-10
vkd3d-shader: Fix cull/clip distance used as shader input.

### LINKED ISSUE #2379: VVL error in test_clip_distance_dxbc
> 202f6:err:vkd3d_debug_messenger_callback: vkCreateGraphicsPipelines(): pCreateInfos[0] (SPIR-V Interface) VK_SHADER_STAGE_GEOMETRY_BIT declared input at Location 2 Component 0 but it is not an Output declared in VK_SHADER_STAGE_VERTEX_BIT.
The Vulkan spec states: Any user-defined variables shared between the OpEntryPoint of two shader stages, and declared with Input as its Storage Class for the subsequent shader stage, must have all Location slots and Component words declared in the preceding shader stage's OpEntryPoint with Output as the Storage Class (https://docs.vulkan.org/spec/latest/appendices/spirvenv.html#VUID-RuntimeSpirv-OpEntryPoint-08743)

doitsujin:
Fixes #2379.

Highly cursed in that we simply didn't support clip/cull as shader input at all, and multi-dimensional arrays were also not really supported - except in PS because PS inputs use different instructions compared to everything else, but we weren't declaring it correctly there either.

Can't really reuse any of the existing code for clip/cull either because outputs are handled very differently from inputs.

Domain shaders are entirely excluded here because we simply pass vec4 arrays around without checking for semantics. The only thing that's semi-interesting there is tess factors.

Should probably be tested with real games as well; test suite passes though and I'm not seeing any spirv-val errors.

### DIFF
--- a/libs/vkd3d-shader/spirv.c
+++ b/libs/vkd3d-shader/spirv.c
@@ -2364,6 +2364,8 @@ struct vkd3d_dxbc_compiler
     uint32_t private_output_variable_array_idx[MAX_REG_OUTPUT + 1]; /* 1 entry for oDepth */
     uint32_t private_output_variable_write_mask[MAX_REG_OUTPUT + 1]; /* 1 entry for oDepth */
     uint32_t epilogue_function_id;
+    uint32_t input_clip_distance_id;
+    uint32_t input_cull_distance_id;
 
     uint32_t binding_idx;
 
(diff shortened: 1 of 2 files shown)


libs/vkd3d-shader/spirv.c
```diff
@@ -4776,10 +4829,13 @@ static uint32_t vkd3d_dxbc_compiler_emit_input(struct vkd3d_dxbc_compiler *compi
     const struct vkd3d_shader_signature *shader_signature;
     const struct vkd3d_shader_register *reg = &dst->reg;
     uint32_t type_id, ptr_type_id, float_type_id;
+    unsigned int clip_cull_index, clip_cull_size;
```
HansKristian-Work, line 4832: Warnings about potential uninitialized use.

HansKristian-Work (commented):
Did this PR fix any VVL errors triggering? Should probably remove some vkd3d_mute-s in test suite if so.

doitsujin:
Right, will add that later.

HansKristian-Work (approved):

PR #2397, opened 2025-03-13
vkd3d: Create local copy of muted VUID strings.

doitsujin:
Rockstar's launcher unloads d3d12.dll before initializing it a second time, but d3d12core remains loaded. That causes us to retain stale pointers to VUID strings and explanations, which in turn leads to a segfault when setting up VVL exceptions again.

The easy way out is to simply retain a local copy of the strings involved and also allow duplicates.


HansKristian-Work (approved):

HansKristian-Work:
Jeez ...

PR #2400, opened 2025-03-14
Fix GS topology validation with point tessellation

doitsujin:
Forgot that this is a thing...

Still need to verify on Windows that this does the right thing, but Pioneers of Pagonia relies on it because why *wouldn't* that game use tess + gs all at once.


HansKristian-Work (approved):
LGTM when the test is confirmed to pass on Windows.

doitsujin:
Confirmed it passes.

HansKristian-Work (approved):

PR #2407, opened 2025-03-19
vkd3d: Expose 16-bit min precision support.

doitsujin:
War Thunder doesn't render anything otherwise.

Should be safe to do this since min16 stuff doesn't *really* have any guarantees, and we implement this properly in DXIL if the device supports 16-bit things anyway.


libs/vkd3d/device.c
```diff
@@ -8140,7 +8140,7 @@ static void d3d12_device_caps_init_feature_options(struct d3d12_device *device)
 
     options->OutputMergerLogicOp = features->logicOp;
     /* Currently not supported */
-    options->MinPrecisionSupport = D3D12_SHADER_MIN_PRECISION_SUPPORT_NONE;
+    options->MinPrecisionSupport = D3D12_SHADER_MIN_PRECISION_SUPPORT_16_BIT;
```
HansKristian-Work, line 8143: Probably remove the comment above?

HansKristian-Work (approved):

PR #2408, opened 2025-03-20
Implement rudimentary buffer sync validation

HansKristian-Work:
Deep dark magic full of crimes against GPUs.

Exposed a lot of UB in Marvel Rivals, but none of them seemed to be root cause of hang, so :crying_cat_face:


doitsujin (approved):
Fairly self-contained, should be good.

PR #2410, opened 2025-03-21
vkd3d: Always use smallest supported alignment for placed SRV / UAV.

doitsujin:
Works around an issue with Wreckfest on RADV where the game aliases an RG32 image with BC1 when uploading its virtual terrain textures.

Draft because I should write a test for this, and another question is whether we just want to do this by default or hide behind an app profile.


libs/vkd3d/resource.c
```diff
@@ -951,9 +951,19 @@ static HRESULT vkd3d_get_image_create_info(struct d3d12_device *device,
         uint32_t candidate_alignment = desc->Alignment ?
                 desc->Alignment : D3D12_DEFAULT_RESOURCE_PLACEMENT_ALIGNMENT;
 
-        /* Only consider alignments that are <= to the requested alignment. */
-        while (candidate_alignment && !(candidate_alignment & supported_alignment))
-            candidate_alignment >>= 1;
+        if (desc->Flags & (D3D12_RESOURCE_FLAG_ALLOW_RENDER_TARGET | D3D12_RESOURCE_FLAG_ALLOW_DEPTH_STENCIL))
```
HansKristian-Work, line 954: I'm a little wary about making this the default when we don't fully understand the performance implication. Would it be possible to make it a config flag first? If we run into any such cases in the wild we can re-evaluate.

libs/vkd3d/resource.c
```diff
@@ -951,9 +951,19 @@ static HRESULT vkd3d_get_image_create_info(struct d3d12_device *device,
...
+            /* Only consider alignments that are <= to the requested alignment. */
+            while (candidate_alignment && !(candidate_alignment & supported_alignment))
+                candidate_alignment >>= 1;
+        }
+        else
+        {
+            /* For non-render target resources, always use the smallest supported alignment in order
+             * to support certain aliasing edge cases that are technically undefined behaviour in
+             * D3D12, but happen to work on native drivers anyway. */
+            candidate_alignment = supported_alignment & -supported_alignment;
```
HansKristian-Work, line 965: I think we should try to avoid the 256B alignment which some GPUs expose here, and stick to the lowest value that D3D12 can require, i.e. 4 KiB.

HansKristian-Work: Should probably also only do this for either BC or 8/16 byte pixel formats which could in theory alias like this.

libs/vkd3d/device.c
```diff
@@ -698,7 +701,7 @@ static const struct vkd3d_shader_quirk_hash re_hashes[] = {
 
 static const struct vkd3d_shader_quirk_info re_quirks = {
     re_hashes, ARRAY_SIZE(re_hashes), 0,
-};
+7};
```
HansKristian-Work, line 704: ?

HansKristian-Work (approved):

PR #2425, opened 2025-03-31
tests: fix small-texture allocation size tests for Adreno

zdobersek:
This fix tries to address failures on Adreno (via Turnip) in the `test_suballocate_small_textures_size` test case. Failures were initially observed on very narrow textures, but there's enough other conditions at play (15% of tests were failing) that it might not make sense to provide specific exceptions for expected failures.

Adreno-specific computation of the expected size is provided. It follows Freedreno's layout computation, but doesn't necessarily address different cases we don't hit in this test. The computation is used for both Adreno blob and Turnip, though it was only tested on the latter (vkd3d doesn't run cleanly on blob last I heard).

### DIFF
--- a/tests/d3d12_resource.c
+++ b/tests/d3d12_resource.c
@@ -2130,14 +2130,52 @@ void test_suballocate_small_textures_size(void)
                      * It is unknown what native does here. */
                     is_bugged_thin = is_radv && is_thin && levels >= 4;
 
-                    /* This assumes tight packing without compression metadata.
-                     * Generally compression is not allowed for placed non-RTV/DSV in D3D12. */
-                    expected_size = (config->width * config->height * config->bpp) / 8;
+                    if (is_adreno_device(device))
+                    {
+                        /* Adreno has layout rules because of which the required memory size
+                         * can't necessarily be captured in some conservative estimate. The
+                         * calculation below should provide the exact required size, but doesn't
+                         * address specific things we don't hit here (e.g. 3D textures and UBWC). */
+                        unsigned int block_size, pitchalign, layer_size, l;
 
-                    /* Be a bit conservative and allow 2x overflow for mipmaps. Tuned so native implementations pass. */
-                    if (levels > 1)
-                        expected_size = 2 * expected_size;
-                    expected_size *= layers;
+                        block_size = format_size(tests[i].format);
+                        pitchalign = block_size * 64;
+                        if (block_size == 2)
+                            pitchalign *= 2;
+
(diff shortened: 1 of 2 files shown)


tests/d3d12_resource.c
```diff
@@ -2130,14 +2130,47 @@ void test_suballocate_small_textures_size(void)
...
 
-                    /* This assumes tight packing without compression metadata.
-                     * Generally compression is not allowed for placed non-RTV/DSV in D3D12. */
-                    expected_size = (config->width * config->height * config->bpp) / 8;
-
-                    /* Be a bit conservative and allow 2x overflow for mipmaps. Tuned so native implementations pass. */
-                    if (levels > 1)
-                        expected_size = 2 * expected_size;
-                    expected_size *= layers;
+                    if (is_adreno_device(device)) {
```
HansKristian-Work, line 2133: Braces are on next line.

tests/d3d12_resource.c
```diff
@@ -2130,14 +2130,47 @@ void test_suballocate_small_textures_size(void)
...
+                        /* Adreno has layout rules because of which the required memory size
+                         * can't necessarily be captured in some conservative estimate. The
+                         * calculation below should provide the exact required size, but doesn't
+                         * address specific things we don't hit here (e.g. 3D textures and UBWC). */
+                        unsigned layer_size = 0;
```
HansKristian-Work, line 2138: unsigned int

HansKristian-Work (changes requested):
Just style nits, otherwise ok to have in.

tests/d3d12_resource.c
```diff
@@ -2130,14 +2130,50 @@ void test_suballocate_small_textures_size(void)
...
+                                nblocksy = align(nblocksy, 16);
+                            if ((l + 1) == levels)
+                                nblocksy = align(nblocksy, 4);
+
+                            layer_size += align(nblocksx * block_size, pitchalign) * nblocksy;
+                        }
+
+                        /* Specific alignment is required for layer size of any non-3D texture. */
+                        expected_size = align(layer_size, 4096) * layers;
+                    } else {
```
HansKristian-Work, line 2167: Nit

HansKristian-Work (changes requested):

tests/d3d12_resource.c
zdobersek: Right, sorry. Fixed.

HansKristian-Work (approved):

PR #2427, opened 2025-04-02
Implement AGS WMMA

HansKristian-Work:


libs/vkd3d/device.c
```diff
@@ -9517,6 +9526,66 @@ bool d3d12_device_validate_shader_meta(struct d3d12_device *device, const struct
...
+        bool supports_f32_16x16x16_f16 = false;
+        bool supports_f16_16x16x16_f16 = false;
+        bool supports_u8_a = false;
+        bool supports_u8_b = false;
+        uint32_t i, count;
+
+        /* There are no sub-capabilities (yet at least).
+         * Validate that we support everything that dxil-spirv can emit. */
+        if (VK_CALL(vkGetPhysicalDeviceCooperativeMatrixPropertiesKHR(
+                device->vk_physical_device, &count, NULL)) != VK_SUCCESS)
```
doitsujin: does it make sense to cache this at a device level or do we expect queries to change depending on the shader in the future?

libs/vkd3d/device.c
```diff
@@ -9528,61 +9528,68 @@ bool d3d12_device_validate_shader_meta(struct d3d12_device *device, const struct
...
             }
         }
-
-        vkd3d_free(props);
-        if (!supports_f16_16x16x16_f16 || !supports_f32_16x16x16_f16 || !supports_u8_a || !supports_u8_b)
+        else
         {
-            ERR("Missing sufficient features to expose WMMA.\n");
-            return false;
+            INFO("Allowing cooperative matrices to go through despite not being supported. RDoc workaround for now.\n");
```
doitsujin, line 9599: Seems a bit scary given that our own tests currently have no way to query for support and rely on pipeline compiles to fail; this is probably going to cause crashes on e.g. rdna2?

We have a renderdoc check somewhere to enable `HOST_CACHED` and friends already, could perhaps reuse that for now for the time being.

HansKristian-Work: Given how I'm done debugging, I can probably just get rid of that commit.

libs/vkd3d/device.c
HansKristian-Work: It could make sense to cache it.

PR #2432, opened 2025-04-07
tests: expect bugged sRGB multisample resolves on Adreno

zdobersek:
Adjust the test_multisample_resolve_strongly_typed unit test to address Adreno drivers ignoring sRGB during multisample attachment resolves, which is allowed per spec.

### DIFF
--- a/tests/d3d12_copy.c
+++ b/tests/d3d12_copy.c
@@ -2303,6 +2303,11 @@ void test_multisample_resolve_strongly_typed(void)
 
         transition_resource_state(context.list, resolve_dst, D3D12_RESOURCE_STATE_RESOLVE_DEST, D3D12_RESOURCE_STATE_COPY_SOURCE);
 
+        /* Under certain conditions resolves will happen through renderpass attachments.
+         * In that case Adreno drivers will ignore sRGB as the resolve format, which is
+         * allowed but should be worked around. */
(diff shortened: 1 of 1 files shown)


tests/d3d12_copy.c
```diff
@@ -2303,6 +2303,10 @@ void test_multisample_resolve_strongly_typed(void)
 
         transition_resource_state(context.list, resolve_dst, D3D12_RESOURCE_STATE_RESOLVE_DEST, D3D12_RESOURCE_STATE_COPY_SOURCE);
 
+        /* Under certain conditions resolves will happen through renderpass attachments.
+         * In that case Adreno drivers will ignore sRGB as the resolve format, which is allowed. */
+        bug_if(is_adreno_device(context.device) && tests[i].resolve_format == DXGI_FORMAT_R8G8B8A8_UNORM_SRGB &&
```
HansKristian-Work, line 2308: This should be todo_if, since it's not a driver bug, but a vkd3d-proton issue we have to workaround.

zdobersek:
> > which is allowed per spec
> 
> It is "should:" in spec though, so drivers should strive to do that. The "should:" is there for historical reasons.

I changed the commit message to say it's tolerated by the spec.

HansKristian-Work (approved):

PR #2455, opened 2025-04-30
Test and fix some large heap behaviour

doitsujin:
Not exhaustive. TLDR this tests a 6GB heap with:
- UAV + copies involving small placed buffers at large offsets
- UAV + copies involving one large placed buffer covering the entire allocation
- binding tiles to a small sparse buffer

This currently doesn't test other buffer uses, or creating placed images.

Heap creation is allowed to fail (and does so on WARP), however Assetto Corsa EVO crashes if we do that, so not supporting large heaps isn't exactly an option.

Creating a placed buffer to cover the entire heap seems to be required in D3D12 and does work, however AMD native crashes when trying to create view descriptors at ≥4GB *buffer* offsets. Smaller placed buffers on a large heap work as expected.

On the vkd3d-proton side, we exceed `maxBufferSize` on mesa drivers as well as amdvlk since the limit is <4GB, which is technically out of spec, but with the exception of one trivial RADV issue where buffer view offsets are treated as 32-bits, this happens to work anyway, so I'm not sure if it's really worth reworking the entire global buffer thing, especially since there's nothing stopping an app from using huge buffers anyway.

### DIFF
--- a/libs/vkd3d/memory.c
+++ b/libs/vkd3d/memory.c
@@ -343,10 +343,13 @@ static HRESULT vkd3d_memory_transfer_queue_flush_locked(struct vkd3d_memory_tran
         switch (transfer->op)
         {
             case VKD3D_MEMORY_TRANSFER_OP_CLEAR_ALLOCATION:
-                VK_CALL(vkCmdFillBuffer(vk_cmd_buffer,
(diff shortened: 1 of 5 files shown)


HansKristian-Work (approved):
LGTM

PR #2463, opened 2025-05-08
Implement NULL RTAS heap descriptors correctly.

HansKristian-Work:
Should resolve GPU hang in SottR.


HansKristian-Work:
https://gitlab.freedesktop.org/mesa/mesa/-/issues/12143#note_2902200 suggests it fixes the issue.

PR #2464, opened 2025-05-08
Add decompression shader for DirectStorage metacommand implementation

doitsujin:
Now that the upstream implementation is dual-licensed as Apache/MIT, we can finally do this.

This allows us to always expose the metacommand rather than relying on the fallback shader in DStorage itself, which is a) slow and b) broken in version 1.2.3, which has famously been causing us issues in Monster Hunter Wilds.

Draft for now because this could do with some testing on more hardware; RDNA2 works in real games and Lavapipe passes our test so it *probably* should work on anything Mesa at least.

This is quite heavily modified in that I put everything into one file and rewrote all the wave op fallbacks to be subgroup size-agnostic, so this *heavily* relies on driver constant-folding at least `gl_SubgroupSize` for good code gen (i think on Intel we even get `gl_NumSubgroups` as a compile-time constant), but that should also really just work on any hardware we care about. On AMD we manually force wave32.

Edit: @Blisto91 has been doing some successful testing on Intel BMG. Also tested on Renoir (Vega) and Nvidia.

### DIFF
--- a/libs/vkd3d/meta_commands.c
+++ b/libs/vkd3d/meta_commands.c
@@ -157,9 +157,7 @@ static bool vkd3d_check_meta_command_support(struct d3d12_device *device, REFGUI
 {
     if (!memcmp(command_id, &IID_META_COMMAND_DSTORAGE, sizeof(*command_id)))
     {
-        return device->device_info.memory_decompression_features.memoryDecompression &&
-                (device->device_info.memory_decompression_properties.decompressionMethods & VK_MEMORY_DECOMPRESSION_METHOD_GDEFLATE_1_0_BIT_NV) &&
-                (device->meta_ops.dstorage.vk_emit_nv_memory_decompression_regions_pipeline);
+        return device->meta_ops.dstorage.vk_emit_nv_memory_decompression_regions_pipeline != VK_NULL_HANDLE;
     }
 
     return false;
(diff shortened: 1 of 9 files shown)


libs/vkd3d/meta_commands.c
```diff
@@ -400,6 +405,8 @@ static void d3d12_meta_command_exec_dstorage(struct d3d12_meta_command *meta_com
             parameters->status_buffer_va, parameters->status_buffer_size,
             parameters->stream_count);
 
+    d3d12_command_list_debug_mark_begin_region(list, "DStorage");
```
HansKristian-Work, line 407: I think we're missing a root parameter invalidate here?

libs/vkd3d/shaders/cs_gdeflate.comp
```diff
@@ -0,0 +1,1053 @@
...
+        {
+            mask <<= gl_SubgroupSize * gl_SubgroupID;
+
+            if (subgroupElect())
+                g_tmp[gl_SubgroupID] = mask;
+
+            barrier();
+
+            for (uint32_t i = 1u; i < gl_NumSubgroups; i++)
+                mask |= g_tmp[i ^ gl_SubgroupID];
```
HansKristian-Work, line 198: Nice trick.

libs/vkd3d/meta_commands.c
doitsujin: `ExecuteMetaCommands` calls `d3d12_command_list_invalidate_all_state`, that should be enough?

HansKristian-Work: Ah, right ...

HansKristian-Work (approved):

Issue #2471, opened 2025-05-14
Helldivers 2 missing texture


EnigmaRaptor:
The Artillery shells in the secondary objective SEAF Artillery do not have the color coded tip textures they are supposed to have.  instead they are a shiny silver color.  This impacts gameplay due to the lack of visual indicator for players to know which type of artillery round they are seeing or picking up.

### Software information
Helldivers 2 (persistent issue across month of updates.  Only present in Wine/Proton, not a bug present in Windows)

### System information
- GPU: 2080 Ti
- Driver: 570.144
- Wine version: ?  Using Steam on Nobara 42
- VKD3D-Proton version: Proton Hotfix

EnigmaRaptor:
Sure.  Here's a sample screenshot and a reference image of the artillery shells in the game.  You can see that a couple of the shells were splattered with bug blood in my screenshot and the blood applied to the shells does have the proper coloring, but the shells themselves are completely grey metallic.  Each type of shell should have a different color tip, and other texturing on it as shown in the reference image.

My screenshot:
![Image](https://github.com/user-attachments/assets/b79d4d88-011e-472b-93ed-3b3500516a6a)


Reference Image: (There's also a static charge shell not shown here)
![Image](https://github.com/user-attachments/assets/26880276-b0ce-4597-9fea-69330d561c34)

PR #2480, opened 2025-05-21
Deal with edge cases in TEX2D/TEX2DARRAY aliasing on NVIDIA

HansKristian-Work:
Closes https://github.com/HansKristian-Work/vkd3d-proton/issues/2471.

This isn't well defined on Vulkan, but it is on D3D sadly. Fortunately, we can make it work fairly easily.


libs/vkd3d/resource.c
```diff
@@ -5838,6 +5838,22 @@ static void vkd3d_create_buffer_srv(vkd3d_cpu_descriptor_va_t desc_va,
...
 
+static void vkd3d_texture_view_desc_fixup(struct d3d12_device *device, struct vkd3d_texture_view_desc *desc)
+{
+    if (device->device_info.properties2.properties.vendorID == VKD3D_VENDOR_ID_NVIDIA)
+    {
+        /* D3D allows some reinterpretation between Texture2D and Texture2DArray.
+         * Texture2D in shader can read a resource with 1 array layer,
+         * and Texture2DArray can read a Texture2D descriptor.
+         * NVIDIA does not correctly deal with Texture2DArray unless we always emit 2D_ARRAY views.
+         * Other implementations don't seem to care, so just emit the natural 2D view. */
```
doitsujin, line 5849: I'd feel safer with a `WARN` or something here reminding us that this workaround exists in case things break in the future, other than that, looks... good I guess. Not a fan of hacks like these but it's not like we can do anything here without potentially sacrificing perf.

HansKristian-Work: Added.

doitsujin (approved):

PR #2495, opened 2025-06-05
Force zero clears

HansKristian-Work:
Resolves some fireflies as seen in SF6 in some cases.


doitsujin (approved):
ah well, guess we're doing this now.

I suppose the inteniton is to keep the remaining `CREATE_NOT_ZEROED` code around for internal use?

HansKristian-Work:
> I suppose the inteniton is to keep the remaining CREATE_NOT_ZEROED code around for internal use?

Yes, and it's also used for suballocator, since we clear the suballocations, there's no need to zero clear the initial backing. Also, internal allocations, and this way it's easier to add potential fine grained behavior down the line should that be needed.

PR #2499, opened 2025-06-10
Add support for EXT_shader_float8 / NV_cooperative_matrix2

HansKristian-Work:
Completes FSR4.


Lawstorant:
Maybe with some spoofing because for now, the blackbox dll is checking for RDNA4 card.

libs/vkd3d/device.c
```diff
@@ -2628,34 +2647,59 @@ static bool vkd3d_supports_minimum_coopmat_caps(struct d3d12_device *device)
...
-        if (fmt->BType == VK_COMPONENT_TYPE_UINT8_KHR)
-            supports_u8_b = true;
-        if (fmt->CType == VK_COMPONENT_TYPE_UINT8_KHR)
-            supports_u8_c = true;
+        if (fp8)
+        {
+            if (fmt->AType == VK_COMPONENT_TYPE_FLOAT8_E4M3_EXT)
+                supports_8bit_a = true;
+            if (fmt->BType == VK_COMPONENT_TYPE_FLOAT8_E4M3_EXT)
+                supports_8bit_b = true;
```
doitsujin, line 2655: Considering how the mode enums work, wouldn't it make more sense to check for the specific combinations that we need to support (i.e. something along the lines of `à == FP8 && b == FP8 && c == FP8 / FP32`)?

HansKristian-Work: The mode bits are a bit awkward in that if at least one matmul supports the type, it's valid, so I think it makes sense to do it like this.

PR #2507, opened 2025-06-16
Add support for Opacity Micromaps / DXR 1.2, part 1

Saancreed:
Changes from initial PR:

* Extension is not enabled unless `VKD3D_CONFIG=dxr12` is passed.
* vkd3d-shader's OMM extension is not yet forwarded to dxil-spirv.
* EXT_opacity_micromap features are not logged at TRACE level.
* Some minor reorders were made so micromap stuff is at the end of extensions, features and procs lists.


libs/vkd3d/va_map.c
```diff
@@ -300,6 +300,55 @@ VkAccelerationStructureKHR vkd3d_va_map_place_acceleration_structure(struct vkd3
...
+    struct vkd3d_view_map *old_view_map;
+    struct vkd3d_view_map *view_map;
+    const struct vkd3d_view *view;
+    struct vkd3d_view_key key;
+
+    resource = vkd3d_va_map_deref_mutable(va_map, va);
+    if (!resource || !resource->va)
+        return VK_NULL_HANDLE;
+
+    view_map = vkd3d_atomic_ptr_load_explicit(&resource->view_map, vkd3d_memory_order_acquire);
```
HansKristian-Work: Can this code to get the view map be pulled out to a helper? Seems like code duplication.

Saancreed: Sure. I could also offer reworking this into something that better supports D3D12's having us guess if it's RTAS or OMM at given VA and

* Remove `VKD3D_VIEW_TYPE_OPACITY_MICROMAP`, so RTAS and OMMs share the same keys in the map, and perhaps even rename `VKD3D_VIEW_TYPE_ACCELERATION_STRUCTURE` to `VKD3D_VIEW_TYPE_ACCELERATION_STRUCTURE_OR_OPACITY_MICROMAP`.
* Store an enum/flag somewhere, perhaps in `vkd3d_view.buffer`, that indicates that given view of `VKD3D_VIEW_TYPE_ACCELERATION_STRUCTURE` type is actually a micromap.
* In the future, add a function like

```c
enum vkd3d_va_map_read_result
{
    VKD3D_VA_MAP_READ_RESULT_FAILED,
    VKD3D_VA_MAP_READ_RESULT_ACCELERATION_STRUCTURE,
    VKD3D_VA_MAP_READ_RESULT_OPACITY_MICROMAP,
};

vkd3d_va_map_read_result vkd3d_va_map_try_read_rtas(struct vkd3d_va_map *va_map,
        struct d3d12_device *device,
        VkDeviceAddress va,
        VkAccelerationStructureKHR *acceleration_structure,
        VkMicromapEXT *micromap);
```

… that can retrieve either RTAS or OMM with half as many map lookups.

HansKristian-Work: That would make sense. No particular need to have a result here. Just write VK_NULL_HANDLE or a valid handle to each output and caller can decide how to make sense of it.

libs/vkd3d/va_map.c
```diff
@@ -288,16 +292,44 @@ VkAccelerationStructureKHR vkd3d_va_map_place_acceleration_structure(struct vkd3
...
     key.u.buffer.format = NULL;
 
-    view = vkd3d_view_map_create_view(view_map, device, &key);
+    view = vkd3d_view_map_create_view2(view_map, device, &key, rtas_is_omm);
     if (!view)
-        return VK_NULL_HANDLE;
-    return view->vk_acceleration_structure;
+        return;
+    
+    if (view->info.buffer.rtas_is_micromap)
```
HansKristian-Work, line 305: Is the implication here that is a VA has been used for RTAS, it can never be used for OMM and vice versa?

libs/vkd3d/vkd3d_private.h
```diff
@@ -1212,6 +1216,7 @@ struct vkd3d_view
         {
             VkDeviceSize offset;
             VkDeviceSize size;
+            bool rtas_is_micromap;
```
HansKristian-Work: This is not hashed, right? Should probably be commented.

libs/vkd3d/va_map.c
Saancreed: Oh hmm, that's tricky. I don't think D3D12 forbids it, so I guess we'll need some mechanism that destroys existing AS/OMM and creates the other one if we receive a call in a context where we are sure that VA is supposed to be AS or OMM, like `BuildRaytracingAccelerationStructure`.

I'll think about how to address this… meanwhile it would be nice to have tests that verify expected behavior :slightly_smiling_face:

libs/vkd3d/vkd3d_private.h
Saancreed: It's by design not a part of the key. Sure, I'll leave a comment about it.

libs/vkd3d/va_map.c
HansKristian-Work: I don't think this is really solveable with existing EXT_omm. I think it's fine to leave it as is as long as we mark conflicts with a FIXME or similar. It's an experimentally enabled feature after all, so we can allow it.

HansKristian-Work: It's fine to defer this to later. No need to solve every possible edge case here.

Saancreed: Added some FIXMEs for now.

PR #2510, opened 2025-06-17
Test and handle RGB9E5 write mask behaviour

doitsujin:
TLDR radv is borked out of the box when rendering RGB9E5 with `RGB_` write mask, so we need to explicitly enable alpha. Spec doesn't say anything about alpha, so I'm assuming it's *supposed* to be ignored, but might as well work around this for now.

See mesa issue [13371](https://gitlab.freedesktop.org/mesa/mesa/-/issues/13371).


doitsujin:
Draft because apparently I might be reading the spec wrong and setting alpha is banned, which means there is no spec-compliant way to render to RGB9E5 at all.

tests/d3d12_render_target.c
```diff
@@ -2139,3 +2139,226 @@ void test_mismatching_rtv_dsv_size(void)
 
     destroy_test_context(&context);
 }
+
+
+void decode_rgb9e5(uint32_t packed, float color[3]) {
```
HansKristian-Work: Lots of brace placement nits here.

doitsujin: derp, i was half asleep when writing this. will fix later.

PR #2512, opened 2025-06-17
Add support for Opacity Micromaps / DXR 1.2, part 2

Saancreed:
Micromap construction and immediate postbuild info, plus some preparations for emit postbuild info.


libs/vkd3d/opacity_micromap.c
```diff
@@ -0,0 +1,286 @@
...
+
+    switch (desc->InfoType)
+    {
+        case D3D12_RAYTRACING_ACCELERATION_STRUCTURE_POSTBUILD_INFO_COMPACTED_SIZE:
+            vk_query_type = VK_QUERY_TYPE_MICROMAP_COMPACTED_SIZE_EXT;
+            type_index = VKD3D_QUERY_TYPE_INDEX_OMM_COMPACTED_SIZE;
+            stride = sizeof(uint64_t);
+            break;
+        case D3D12_RAYTRACING_ACCELERATION_STRUCTURE_POSTBUILD_INFO_SERIALIZATION:
+            vk_query_type = VK_QUERY_TYPE_MICROMAP_SERIALIZATION_SIZE_EXT;
```
HansKristian-Work, line 174: Just flag this FIXME and write 0 too. We know it's broken on NV atm, and it's not a very exciting query anyway.

libs/vkd3d/command.c
```diff
@@ -16180,6 +16180,106 @@ static void d3d12_command_list_flush_rtas_batch(struct d3d12_command_list *list)
     d3d12_command_list_clear_rtas_batch(list);
 }
 
+static void d3d12_command_list_build_raytracing_opacity_micromap_array(struct d3d12_command_list *list,
+        const D3D12_BUILD_RAYTRACING_ACCELERATION_STRUCTURE_DESC *desc, UINT num_postbuild_info_descs,
+        const D3D12_RAYTRACING_ACCELERATION_STRUCTURE_POSTBUILD_INFO_DESC *postbuild_info_descs)
+{
+    const struct vkd3d_vk_device_procs *vk_procs = &list->device->vk_procs;
```
HansKristian-Work: Unused variable warning.

HansKristian-Work (changes requested):
Some nits, but otherwise seems good.

libs/vkd3d/command.c
Saancreed: Right, it's used only with `VKD3D_ENABLE_BREADCRUMBS`, I'll just move it to a smaller scope inside that ifdef.

libs/vkd3d/opacity_micromap.c
Saancreed: What about compacted size though? It might be more exciting but it's just as broken on NV.

Plus, I wouldn't want to rely on one vendor's implementation being currently broken; at worst we'll see some errors from query pool creation failures (which should hopefully just start working in the future). I'd test this with RADV's MR but the only relevant hardware I have for that is a Steam Deck.

HansKristian-Work: It is? :| yikes. Well, ok.

HansKristian-Work (approved):

PR #2515, opened 2025-06-18
Add support for Opacity Micromaps / DXR 1.2, part 3

Saancreed:
Some helpers to retrieve placed RTAS or OMM from va map, emit postbuild info and copying. The last one shuffles a few things around to avoid multiple lookups.


libs/vkd3d/acceleration_structure.c
```diff
@@ -408,12 +409,14 @@ void vkd3d_acceleration_structure_emit_postbuild_info(
 
     for (i = 0; i < count; i++)
     {
-        vk_acceleration_structure = vkd3d_va_map_place_acceleration_structure(
-                &list->device->memory_allocator.va_map, list->device, addresses[i]);
+        vkd3d_va_map_try_read_rtas(&list->device->memory_allocator.va_map,
```
HansKristian-Work: This is broken. If application records an RTAS copy command before the build (it is valid if submitted out of order), things will break. If no object is found, you'll need to explicitly place an RTAS (as it's the common case) and use that. Flag a FIXME in this case, since this is one of the reasons EXT_omm is kinda broken as-is.

libs/vkd3d/command.c
```diff
@@ -16533,12 +16535,28 @@ static void STDMETHODCALLTYPE d3d12_command_list_CopyRaytracingAccelerationStruc
...
+    else if (src_omm)
+    {
+        vkd3d_opacity_micromap_copy(list, dst_data, src_omm, mode);
+        VKD3D_BREADCRUMB_AUX64(dst_data);
+        VKD3D_BREADCRUMB_AUX64(src_data);
+        VKD3D_BREADCRUMB_AUX32(mode);
+        VKD3D_BREADCRUMB_COMMAND(COPY_OMM);
+    }
+    else
+        ERR("Invalid src address #%"PRIx64" for RTAS / OMM copy.\n", src_data);
```
HansKristian-Work: Same concern. If you cannot find anything resort to RTAS placement and flag a FIXME.

libs/vkd3d/acceleration_structure.c
Saancreed: Another day I become more convinced that NVAPI's micromap API was much more reasonable…

HansKristian-Work (approved):
LGTM

PR #2519, opened 2025-06-19
Add support for Opacity Micromaps / DXR 1.2, part 4

Saancreed:
Final part, deals with attaching micromaps to triangles in RTAS and mapping pipeline creation flag. Current test results from Nvidia:

```
d3d12: ======== test_raytracing_opacity_micro_map end ==========
d3d12: 45783 tests executed (9 failures, 12 successful todo, 0 skipped, 9 todo, 0 bugs).
```

As before, failures are all query-related.


libs/vkd3d/acceleration_structure.c
```diff
@@ -227,6 +244,69 @@ bool vkd3d_acceleration_structure_convert_inputs(const struct d3d12_device *devi
...
+                            break;
+                        case DXGI_FORMAT_R32_UINT:
+                            omm->indexType = VK_INDEX_TYPE_UINT32;
+                            break;
+                        case DXGI_FORMAT_R16_UINT:
+                            omm->indexType = VK_INDEX_TYPE_UINT16;
+                            break;
+                        case DXGI_FORMAT_R8_UINT:
+                            FIXME_ONCE("Using UINT8 as OMM index format is technically out of spec.\n");
+                            omm->indexType = VK_INDEX_TYPE_UINT8;
```
Saancreed, line 286: Fwiw this triggers a validation error:

```
Validation Error: [ VUID-VkAccelerationStructureTrianglesOpacityMicromapEXT-indexType-parameter ] | MessageID = 0xd160e0a1
vkGetAccelerationStructureBuildSizesKHR(): pBuildInfo[0].pGeometries[0].geometry.triangles.pNext<VkAccelerationStructureTrianglesOpacityMicromapEXT>.indexType (VK_INDEX_TYPE_UINT8) requires the extensions VK_KHR_index_type_uint8 or VK_EXT_index_type_uint8.
The Vulkan spec states: indexType must be a valid VkIndexType value (https://docs.vulkan.org/spec/latest/chapters/accelstructures.html#VUID-VkAccelerationStructureTrianglesOpacityMicromapEXT-indexType-parameter)
```

Should I enable that extension to silence it?

HansKristian-Work: That's somewhat surprising. 8-bit indices should just be banned wholesale by EXT_omm.

HansKristian-Work: I wouldn't bother right now since this is only optionally enabled.

libs/vkd3d/acceleration_structure.c
```diff
@@ -36,6 +36,10 @@ static VkBuildAccelerationStructureFlagsKHR d3d12_build_flags_to_vk(
         vk_flags |= VK_BUILD_ACCELERATION_STRUCTURE_PREFER_FAST_BUILD_BIT_KHR;
     if (flags & D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_PREFER_FAST_TRACE)
         vk_flags |= VK_BUILD_ACCELERATION_STRUCTURE_PREFER_FAST_TRACE_BIT_KHR;
+    if (flags & D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_ALLOW_OMM_LINKAGE_UPDATE)
+        vk_flags |= VK_BUILD_ACCELERATION_STRUCTURE_ALLOW_OPACITY_MICROMAP_DATA_UPDATE_EXT;
```
HansKristian-Work: These should not use the deprecated alias, missing _BIT_EXT.

Also, I think the use of DATA_UPDATE here is wrong.

- VK_BUILD_ACCELERATION_STRUCTURE_ALLOW_OPACITY_MICROMAP_UPDATE_BIT_EXT specifies that the opacity micromaps associated with the specified acceleration structure may change with an acceleration structure update.

This maps to OMM_LINKAGE_UPDATE I think, not DATA_UPDATE.

DATA_UPDATE should always be set for OMMs I think, since D3D12 allows that always.

Saancreed: > These should not use the deprecated alias, missing _BIT_EXT.

I'll fix this once vulkan-headers are updated. The revision referenced by the current master doesn't have `_BIT_EXT` flags yet. Should I wait for shader_float8 branch to land or should I update headers on my own?

> Also, I think the use of DATA_UPDATE here is wrong.

DXR spec is… somewhat contradicting itself on that:

> It is also possible to assign a different OMM Array and OMM indices when updating a BLAS. The new AS build flag [D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_ALLOW_OMM_LINKAGE_UPDATE](https://microsoft.github.io/DirectX-Specs/d3d/Raytracing.html#d3d12_raytracing_acceleration_structure_build_flags) lets the driver know that OMM contents may change with BLAS updates.

> `D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_ALLOW_OMM_LINKAGE_UPDATE`
> The AS supports updating OMM contents (fields of [D3D12_RAYTRACING_GEOMETRY_OMM_LINKAGE_DESC](https://microsoft.github.io/DirectX-Specs/d3d/Raytracing.html#d3d12_raytracing_geometry_omm_linkage_desc)).
> While any part of the linkage description can be changed, switching between having an OMM linkage at all and not having one isn’t permitted.
> If an acceleration structure with OMM linkage is having just it’s positions updated (not the OMM linkage), so `ALLOW_OMM_LINKAGE_UPDATE` isn’t specified, the update description still has to provide the original OMM linkage description.

And from `D3D12_RAYTRACING_GEOMETRY_OMM_LINKAGE_DESC`:

> `D3D12_GPU_VIRTUAL_ADDRESS_AND_STRIDE OpacityMicromapIndexBuffer`
> Buffer specifying which OMM index to use for each triangle. Some values have special meaning and are used to encode OMMs with uniform state (see [D3D12_RAYTRACING_OPACITY_MICROMAP_SPECIAL_INDEX](https://microsoft.github.io/DirectX-Specs/d3d/Raytracing.html#d3d12_raytracing_opacity_micromap_special_index)). This input is optional. If NULL, there is a 1:1 mapping between input triangles and OMM Array entries, as long as the OMM Array has at least as many entries as the number of triangles + OpacityMicromapBaseLocation. For BLAS updates, this input buffer must match that of the original build if the [D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_ALLOW_OMM_LINKAGE_UPDATE](https://microsoft.github.io/DirectX-Specs/d3d/Raytracing.html#d3d12_raytracing_acceleration_structure_build_flags) build flag is cleared.

Notably, none of the other members have that "this must match that of the original build if the `ALLOW_OMM_LINKAGE_UPDATE` build flag is cleared.

So we have one place that claims that updating _anything_ from OMM linkage requires the flag, and another that specifies only one member that requires the flag to update. I guess I'll look how Nvidia maps the flags between RHIs.

libs/vkd3d/acceleration_structure.c
```diff
@@ -36,6 +36,12 @@ static VkBuildAccelerationStructureFlagsKHR d3d12_build_flags_to_vk(
         vk_flags |= VK_BUILD_ACCELERATION_STRUCTURE_PREFER_FAST_BUILD_BIT_KHR;
     if (flags & D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_PREFER_FAST_TRACE)
         vk_flags |= VK_BUILD_ACCELERATION_STRUCTURE_PREFER_FAST_TRACE_BIT_KHR;
+    if (flags & D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_ALLOW_OMM_LINKAGE_UPDATE)
+        /* D3D12 spec isn't clear on what is allowed to be updated and when */
+        vk_flags |= VK_BUILD_ACCELERATION_STRUCTURE_ALLOW_OPACITY_MICROMAP_UPDATE_BIT_EXT |
```
HansKristian-Work, line 42: Wrap this in braces, this looks goofy.

HansKristian-Work (approved):

PR #2526, opened 2025-06-25
Implement AMD anti-lag interface

HansKristian-Work:


libs/vkd3d/device_vkd3d_ext.c
```diff
@@ -763,3 +763,128 @@ CONST_VTBL struct ID3DLowLatencyDeviceVtbl d3d_low_latency_device_vtbl =
...
+    {
+        /* Mode setting for v2. */
+        if (v2->uiSize != sizeof(*v2))
+        {
+            ERR("Invalid size for API structure.\n");
+            return E_INVALIDARG;
+        }
+
+        /* This structure only seems to flag certain things, and does not modify the mode?
+         * There isn't much we can do with this struct right now I think other than logging ... */
```
doitsujin, line 866: Have we seen an app try to use this with FSR frame gen? Wonder what it's doing in that case.

HansKristian-Work: GoT can trigger it, yes and it flags stuff in a way that looks sensible enough. The end result is that we don't sleep at all due to low latency submits and staggered, so it's effectively a noop overall. Not ideal, but at least no regression.

PR #2531, opened 2025-06-28
vkd3d-shader: Make range inclusive for preferred wave size

lax1dude:
The problems don't seem to fully go away for me unless I change the code from #2497 to this, maybe the min/max subgroup size range was supposed to be inclusive here?

### DIFF
--- a/libs/vkd3d-shader/dxil.c
+++ b/libs/vkd3d-shader/dxil.c
@@ -1308,14 +1308,14 @@ int vkd3d_shader_compile_dxil(const struct vkd3d_shader_code *dxbc,
         {
             /* We don't expect min and max heuristics to have conflicting ideas. */
             if (heuristic_max_wave_size &&
-                    compiler_args->max_subgroup_size > heuristic_max_wave_size &&
+                    compiler_args->max_subgroup_size >= heuristic_max_wave_size &&
                     compiler_args->min_subgroup_size <= heuristic_max_wave_size)
             {
                 wave_size_preferred = heuristic_max_wave_size;
             }
             else if (heuristic_min_wave_size &&
                     compiler_args->max_subgroup_size >= heuristic_min_wave_size &&
-                    compiler_args->min_subgroup_size < heuristic_min_wave_size)
+                    compiler_args->min_subgroup_size <= heuristic_min_wave_size)
             {
                 wave_size_preferred = heuristic_min_wave_size;
             }


lax1dude:
Related Mesa issue: [https://gitlab.freedesktop.org/mesa/mesa/-/issues/13356](https://gitlab.freedesktop.org/mesa/mesa/-/issues/13356)

libs/vkd3d-shader/dxil.c
```diff
@@ -1308,14 +1308,14 @@ int vkd3d_shader_compile_dxil(const struct vkd3d_shader_code *dxbc,
         {
             /* We don't expect min and max heuristics to have conflicting ideas. */
             if (heuristic_max_wave_size &&
-                    compiler_args->max_subgroup_size > heuristic_max_wave_size &&
+                    compiler_args->max_subgroup_size >= heuristic_max_wave_size &&
```
HansKristian-Work, line 1311: Which of these two lines actually fix the issue? It's problematic to add this as-is since it means that a lot of shaders on ANV will now be promoted to Wave32 (slow).

tpalli: Looks like with Talos and Oblivion both lines need the change, I tried applying only one or another but neither fixed artifacts on their own. I guess this change could be limited to only UE5 engine (?) .. but of course we might be hitting this with something else too.

HansKristian-Work: Adding this as-is is not acceptable, but if this commit fixes things it definitely points to wave16 vs 32 shenanigans. A targeted fix must be found here. It's likely another heuristic that has to be found.

lax1dude: It looks like these two compiler_args values are just the subgroup min/max limits from the vulkan device properties, do you think you could explain why it validates the max/min heuristic wave sizes against a range that excludes the upper and lower limit reported by the device? That would help me understand this problem a lot better, I can't find any clues in the commits or PRs why it doesn't validate them against the full range.

lax1dude: I kind of had the wrong idea of what was being expressed by these heuristics last week, but I think I understand it now. The min wave size heuristic is the value intended to deal with this bug, yet my changes could only have affected when the max wave size heuristic is selected here, since both values can only be 0 or 32 and the hardware defines 16 and 32 as the min and max subgroup sizes. If I want to fix it for real I need to express it through some additional heuristic on the min wave size.

lax1dude: @HansKristian-Work The MicropolyRasterize compute shaders in UE5 contain intrinsics that would only work properly in Wave32, but have an undefined min/max wave size in the DXIL, and are also not triggering the heuristic to force it to use Wave32. The reason it fails to trigger the heuristic is because another separate part of the shader issues a barrier. Ignoring the barriers when checking for the min wave size heuristic makes the bug go away for me.

Do you have any thoughts for me, I haven't been able to determine the reason for why it needs to avoid the heuristic on shaders that also use barriers for workgroup shared memory.

**Edit:** Alternatively the ANV devs might just force Wave32 when vkd3d sets the pipeline's varying subgroup size bit on these UE5 games, since that also has the effect of promoting this shader and other problematic ones to Wave32: [https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/35852](https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/35852)

```patch
diff --git a/opcodes/dxil/dxil_waveops.cpp b/opcodes/dxil/dxil_waveops.cpp
index f28880d..bcf58ab 100644
--- a/opcodes/dxil/dxil_waveops.cpp
+++ b/opcodes/dxil/dxil_waveops.cpp
@@ -540,8 +540,7 @@ bool emit_wave_read_lane_at_instruction(Converter::Impl &impl, const llvm::CallI
                            impl.execution_model == spv::ExecutionModelMeshEXT ||
                            impl.execution_model == spv::ExecutionModelTaskEXT)
                        {
-                               if (!impl.shader_analysis.has_group_shared_barrier &&
-                                   impl.execution_mode_meta.workgroup_threads[0] == 32 &&
+                               if (impl.execution_mode_meta.workgroup_threads[0] == 32 &&
                                    impl.execution_mode_meta.workgroup_threads[1] == 1 &&
                                    impl.execution_mode_meta.workgroup_threads[2] == 1 &&
                                    lane_is_wave32_masked(lane))

```

phonkloverrr:
yo ik this has nothing to do with whatever is above me but I have a request for eaglercraft helpers, hh_the_og was abusing  powers and banning ppl in unfair ways and grieding bases and shi

phonkloverrr:
hah

PR #2563, opened 2025-08-11
vkd3d: Guard against invalid VA lookups in more cases.

HansKristian-Work:


libs/vkd3d/command.c
```diff
@@ -11267,9 +11267,19 @@ static void d3d12_command_list_set_push_descriptor_info(struct d3d12_command_lis
...
+
+            if (resource)
+            {
+                descriptor->info.buffer.buffer = resource->vk_buffer;
+                descriptor->info.buffer.offset = gpu_address - resource->va;
+                descriptor->info.buffer.range = min(resource->size - descriptor->info.buffer.offset, max_range);
+            }
+            else
+            {
+                descriptor->info.buffer.buffer = VK_NULL_HANDLE;
```
doitsujin, line 11279: is there a reason why everything else has an `ERR` but this doesn't?

doitsujin (commented):
question / nit

libs/vkd3d/command.c
HansKristian-Work: Since this is called in response to hoisting and not direct application call, it felt more appropriate.

libs/vkd3d/command.c
```diff
@@ -11441,9 +11451,17 @@ static void STDMETHODCALLTYPE d3d12_command_list_IASetIndexBuffer(d3d12_command_
...
-        list->index_buffer.size = view->SizeInBytes;
+        if (resource)
+        {
+            list->index_buffer.buffer = resource->vk_buffer;
+            list->index_buffer.offset = view->BufferLocation - resource->va;
+            list->index_buffer.size = view->SizeInBytes;
+        }
+        else
+        {
+            ERR("Invalid VA lookup.\n");
```
HansKristian-Work: Should probably demote this to WARN, not ERR.

doitsujin (approved):

PR #2589, opened 2025-08-24
Make sure idl generated headers get installed

has207:
(it's not super pretty but seems to work)

### DIFF
--- a/include/meson.build
+++ b/include/meson.build
@@ -19,3 +19,14 @@ vkd3d_idl = [
 ]
 
 vkd3d_header_files = idl_generator.process(vkd3d_idl)
+
+# Install generated IDL headers
+foreach idl_file : vkd3d_idl
+  header_name = idl_file.split('.')[0] + '.h'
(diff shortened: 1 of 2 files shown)


include/meson.build
```diff
@@ -18,4 +18,20 @@ vkd3d_idl = [
...
+# even when called with single item so we process
+# each idl file in a loop generating and installing
+# them to avoid calling process() twice
+vkd3d_header_files = []
+foreach idl_file : vkd3d_idl
+  header_file = idl_generator.process(idl_file)
+  vkd3d_header_files += header_file
+
+  header_name = idl_file.split('.')[0] + '.h'
+  custom_target('install_' + header_name,
```
HansKristian-Work: You'll want to gate this behind Linux build. It's confusing to install these headers on a Windows build since you should just use normal D3D headers there. See build.meson:186. Otherwise, it seems okay to install the IDL headers.

has207: Done. PTAL

PR #2591, opened 2025-08-25
Disable forced anti-lag in CP77

HansKristian-Work:
For whatever reason, the game always attempts to use anti-lag when
supported. It's not possible to disable it, and it's causing issues
for some users in the wild, so the only reasonable thing is to disable
it.

https://gitlab.freedesktop.org/mesa/mesa/-/issues/13724


shelterx:
Works fine, thank you!

PR #2597, opened 2025-08-28
vkd3d: Enable low SSBO alignment path on NVK too.

HansKristian-Work:
Enables descriptor buffers too.


HansKristian-Work:
This breaks NVK hard since it realigns descriptors, so need a different approach.

PR #2598, opened 2025-08-28
vkd3d: Add mechanism to detect and retain references to certain GPU resources while in-flight

HansKristian-Work:


libs/vkd3d/resource.c
```diff
@@ -2113,6 +2113,23 @@ static HRESULT STDMETHODCALLTYPE d3d12_resource_SetPrivateDataInterface(d3d12_re
...
+    struct d3d12_resource *resource = impl_from_ID3D12Resource2(iface);
+
+    /* Disgusting workaround, but we've seen many titles screwing up their FSR implementation
+     * with use-after-free. This is set right after resource creation. */
+    const WCHAR fsr_prefix[] = u"FSR3UPSCALER";
+
+    if (vkd3d_wcslen(str) >= 12 && memcmp(fsr_prefix, str, 12 * sizeof(WCHAR)) == 0)
+    {
+        WARN("FSR resource detected. Forcing retained GPU reference to work around broken integration code in either game or UE5.\n");
+        resource->flags |= VKD3D_RESOURCE_RETAINED_GPU_REFERENCE;
```
K0bin, line 2137: Potential thread safety issue. Should be fine in practice because it only gets assigned at the creation of the resource and here.

libs/vkd3d/resource.c
```diff
@@ -3516,6 +3533,64 @@ static void d3d12_resource_wait_for_sparse_init(struct d3d12_resource *resource)
...
+void d3d12_resource_decref_weak(struct d3d12_resource *resource)
+{
+    /* To be able to detect a destroyed resource, we need to hold on to the d3d12_resource memory a bit longer.
+     * Effectively, we have a weak_ptr system in place. Only bother going through this if
+     * we enable the weak_ptr retain path. This should only be enabled in debug builds and/or special workaround
+     * cases. ID3D12GraphicsCommandList can retain a weak reference until it is Reset. */
+#ifdef VKD3D_ENABLE_BREADCRUMBS
+    const bool can_have_weak_references = true;
+#else
+    bool can_have_weak_references = !!(resource->flags & VKD3D_RESOURCE_RETAINED_GPU_REFERENCE);
```
K0bin: Make this one const too :P

libs/vkd3d/resource.c
HansKristian-Work: Not really a concern since this is a targeted workaround.

libs/vkd3d/resource.c
HansKristian-Work: ... sure.

libs/vkd3d/command.c
```diff
@@ -18551,17 +18529,42 @@ static void STDMETHODCALLTYPE d3d12_command_queue_ExecuteCommandLists(ID3D12Comm
     sub.execute.num_command_allocators = command_list_count;
     sub.execute.retained_resources = retained_resources;
     sub.execute.num_retained_resources = num_retained_resources;
+    for (i = 0; i < num_retained_resources; i++)
+    {
+        if (!vkd3d_atomic_uint32_load_explicit(&retained_resources[i]->internal_refcount, vkd3d_memory_order_relaxed) ||
+                d3d12_resource_incref(retained_resources[i]) == 1)
+        {
+            ERR("Trying to execute resource which has already been destroyed. This will likely hang GPU at some point ... Skipping ECL.\n");
+            /* TODO: We could mark device as lost, or just ignore this submit. Hard to tell
```
doitsujin: Does Mafia hit this path? Sounds like dropping the device would break the game?

HansKristian-Work: It does not. I added this mostly for completeness. Maybe it's sufficient to just detect if the load is 0 and error out, and not try to do any weird shenanigans with rewriting the refcount?

doitsujin (approved):

PR #2606, opened 2025-09-04
descriptor-qa: Add support for timestamp validation to catch use-before-alloc

HansKristian-Work:


HansKristian-Work:
CI caught some printf warnings.

PR #2607, opened 2025-09-04
Fix some edge cases for embedded root signatures

### LINKED ISSUE #2605: Investigate behavior of CreatePipeline from implicit root signature
We only consider vertex shader in d3d12_pipeline_create_private_root_signature, but apparently it should be possible to place it in any stage. Needs test coverage and trivial fix if so.

HansKristian-Work:
Fix #2605.


libs/vkd3d/state.c
```diff
@@ -5522,29 +5522,72 @@ static HRESULT d3d12_pipeline_create_private_root_signature(struct d3d12_device
...
+        ID3D12RootSignature *object = NULL;
 
-    *root_signature = impl_from_ID3D12RootSignature(object);
-    d3d12_root_signature_inc_ref(*root_signature);
-    ID3D12RootSignature_Release(object);
-    return hr;
+        if (!bytecode->BytecodeLength)
+            continue;
+
+        if (SUCCEEDED(ID3D12Device12_CreateRootSignature(&device->ID3D12Device_iface, 0,
```
doitsujin: does an error return here imply "shader doesn't define root sig"?

HansKristian-Work: ye, it fails if there is no root signature. I could guard it with a check for "is there even any RS in here" to make it clearer, but it's guts deep in vkd3d-shader ...

PR #2618, opened 2025-09-15
Remove legacy DXBC compiler

doitsujin:
*Should* be everything hopefully.

The one function in _main.c was copy-pasted over from spirv.c so we can nuke that file. dxbc.c still cointains all the signature / rootsig parsing.


HansKristian-Work (approved):
:partying_face:  I guess VKD3D_CONFIG_DISABLE_DXBC_SPIRV can be removed now as well.

PR #2623, opened 2025-09-18
Extended mismatch robustness tests

HansKristian-Work:
A very particular set of tests which aims to further expose how native drivers deal with aliasing of typed buffers and raw.

The rules for AMD are quite simple: Structured == typed, since structured buffers pull stride from descriptor. Raw buffers use a different mechanism for robustness. The most surprising effects is when aliasing byte address and typed/structured. Reading typed as raw means robustness is completely disabled (!).
The other way around is stride = 0, and writes all conflict and reads all read the same data. Completely useless behavior, so applications cannot rely on this particular UB. This opens up possibilities to treat BAB and Structured quite differently in vkd3d-proton.

NV is highly unusual, and there is (weak) evidence to suggest that D3D12 driver uses some form of texel buffers internally since:

Structured and raw aliasing "works", except that robustness range does not work as you would expect. The encoded size seems to be in terms of elements for structured, and dwords for byte address buffers. For strides larger than 16b, the element size seems actually be split into "texel elements" which can neatly subdivide the full stride. For non-aligned strides, the "element" count changes dramatically, e.g.:

- Stride 4: Behaves as R32_UINT
- 8: RG32_UINT
- 12: RGB32_UINT (prime dwords, but RGB32)
- 16: RGBA32_UINT
- 20: R32_UINT (prime dwords)
- 24: RGB32_UINT (RG32 would have worked too, but driver seems to pick largest format that can subdivide)
- 28: R32_UINT (prime dwords)
- 32: RGBA32_UINT

This behavior only seems to apply to SRVs, for UAVs, it behaves in a more scalar way, likely to have the option to write scalars if need be.

vkd3d-proton behavior is adjusted to better match NV behavior, since for SRVs at least, it seems fairly reliable to read texel buffers <-> structured with proper type size. Right now we only get usable behavior for 4 byte aliasing.

Also fixed a long standing bug with 16-bit structured. No content I know of has ever hit that particular case, so it has been broken until now.


tests/d3d12_robustness.c
```diff
@@ -906,7 +906,170 @@ void test_undefined_structured_raw_read_typed_dxil(void)
...
+                D3D12_RESOURCE_STATE_UNORDERED_ACCESS);
+    }
+
+    uav_desc.Buffer.StructureByteStride = 16;
+    uav_desc.Buffer.FirstElement = 0;
+    uav_desc.Buffer.NumElements = 64;
+
+    for (i = 0; i < 16; i++)
+    {
+        /* Commit every possible GPU crime. */
```
doitsujin, line 1248: :frog:

PR #2624, opened 2025-09-19
Shared fence initial value

Vingian:
Some shared fences expect that an initial value has been set.


HansKristian-Work (approved):

HansKristian-Work:
Good to know, thanks.

PR #2633, opened 2025-09-24
Add support for unified image layouts and improve copy batching situation

HansKristian-Work:
Main idea is that we can avoid some needless transitions.

- For DEPTH, we did vendor ID hackery to fake this already. This will extend the hack to be more general (pun intended).
- For copies where we had to consider state decay before, we can massively simplify copy batching. Should translate to real-world performance improvements.

Main idea for copies is that we can unify the WAW tracker that we use for buffers. The uglier wart that remains is for placed resources where we may have to transition from UNDEFINED, but we only need UNDEFINED -> GENERAL, no secondary barrier after the copies. When placed resources are used, we can avoid the transition.

For committed resources we avoid UNDEFINED -> something as much as possible to avoid extra barriers.

The two exploratory tests exercise the new code somewhat exhaustively and I've confirmed the behavior looks as good as it can get in RenderDoc.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -8833,12 +8833,40 @@ static bool vk_image_copy_subresource_overlaps(const ID3D12Resource *res, const 
             subres->mipLevel == other_subres->mipLevel;
 }
 
+static bool vk_image_copy_subresource_fuse_aspects(
+        ID3D12Resource *dst, ID3D12Resource *src, const VkImageCopy2 *copy,
+        ID3D12Resource *orig_dst, ID3D12Resource *orig_src, VkImageCopy2 *orig_copy)
+{
+    /* If we can find an identical copy, just fuse the aspects. */
+    if (dst == orig_dst && src == orig_src &&
+            /* regions have to match */
+            memcmp(&copy->extent, &orig_copy->extent, sizeof(VkExtent3D)) == 0 &&
+            memcmp(&copy->dstOffset, &orig_copy->dstOffset, sizeof(VkExtent3D)) == 0 &&
+            memcmp(&copy->srcOffset, &orig_copy->srcOffset, sizeof(VkExtent3D)) == 0 &&
+            /* subresource has to match */
+            copy->dstSubresource.baseArrayLayer == orig_copy->dstSubresource.baseArrayLayer &&
+            copy->dstSubresource.mipLevel == orig_copy->dstSubresource.mipLevel &&
+            copy->srcSubresource.baseArrayLayer == orig_copy->srcSubresource.baseArrayLayer &&
+            copy->srcSubresource.mipLevel == orig_copy->srcSubresource.mipLevel &&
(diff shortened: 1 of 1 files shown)


libs/vkd3d/command.c
```diff
@@ -8194,13 +8216,21 @@ static void d3d12_command_list_copy_image_transition_images(struct d3d12_command
...
+                &region->dstSubresource,
+                VK_PIPELINE_STAGE_2_COPY_BIT, VK_ACCESS_2_NONE,
+                writes_full_subresource ? VK_IMAGE_LAYOUT_UNDEFINED : dst_resource->common_layout,
+                barrier.dst.stages, dst_copy_pass_access, barrier.dst.layout);
+    }
 
-    if (!overlapping_subresource)
+    /* Source textures never have to be transitioned with unified layouts,
+     * since we know they're in GENERAL, and we already did TRANSFER_READ_BIT.
+     * If we have to add non-transfer reads due to fallback copies, that warrants a barrier as well. */
```
doitsujin, line 8237: I suppose the latter is due to fallback copies potentially doing their own barriers *with* layout transitions regardless of unified layouts?

doitsujin (commented):
I wonder if there's a universe where we could simplify things w.r.t. copy batching if we just slapped `GENERAL` on more things by default.

Currently, `vk_common_image_layout_from_d3d12_desc` will default to `SHADER_READ_ONLY_OPTIMAL` for *most* images, but even without unified layouts I don't think there's an expectation that `GENERAL` is necessarily worse for images that can't also be used as a render target, which leaves the question whether use cases involving render targets are common enough that all this insane complexity is actually worth it.

libs/vkd3d/command.c
HansKristian-Work: Main thing is that image barriers can now do access = TRANSFER_READ early, but in the off-case where they need to be read as SHADER_READ_BIT or attachments, we have to make those visible too late. COPY_DEST might also need TRANSFER_READ for the overlapping resolve case which is extremely bizarre and rare too.

In the old path, we never did visibility when entering COPY states, since we had to transition away from the layout anyway to do the copy.

doitsujin:
> Main problem here is RDNA2 MSAA, which would force a decompress.

Yes, but I don't think we expect non-render target use cases to benefit from `OPTIMAL` layouts in the first place, and I'd also expect the majority of use cases for back-to-back copies to be buffer->texture copies or render target -> some cube map or whatever. Are we still expecting a lot of layout transition spam involving RTs?

HansKristian-Work:
Right, you mean that for non-RTV/DSV enabled resources we could just use common_layout = GENERAL anyway? The main trouble I suppose is that the non-unified layouts path will have to consider both cases which could be mix and match in worst case.

doitsujin:
Probably useful to split this up a bit, I had looked at everything up to the madness that is `vkd3d: Optimize resource copies for unified layout. 79f8034` before starting the layout discussion.

HansKristian-Work:
Superseded by split-up PRs.

PR #2637, opened 2025-09-26
Implement tight resource alignment

doitsujin:
Should probably be sanity-checked in a bunch of games to avoid regressions, especially those that are known to be weird with resource allcation (Starfield?).


libs/vkd3d/device.c
```diff
@@ -5508,6 +5508,24 @@ static HRESULT STDMETHODCALLTYPE d3d12_device_CheckFeatureSupport(d3d12_device_i
...
+            D3D12_FEATURE_DATA_TIGHT_ALIGNMENT *data = feature_data;
+
+            if (feature_data_size != sizeof(*data))
+            {
+                WARN("Invalid size %u.\n", feature_data_size);
+                return E_INVALIDARG;
+            }
+
+            /* Trivial vulkan catch-up, always supported */
+            data->SupportTier = D3D12_TIGHT_ALIGNMENT_TIER_1;
```
HansKristian-Work: I don't think we can expose this on devices with bufferImageGranularity that is larger than 4 KiB (i.e. older NV).

HansKristian-Work: That's only a problem if the device also exposes TIER_2 heaps though I guess.

HansKristian-Work (commented):
Small nit, otherwise LGTM.

libs/vkd3d/device.c
```diff
@@ -7824,7 +7824,11 @@ static D3D12_RESOURCE_ALLOCATION_INFO* STDMETHODCALLTYPE d3d12_device_GetResourc
         if (desc->Dimension == D3D12_RESOURCE_DIMENSION_BUFFER)
         {
             resource_info.SizeInBytes = desc->Width;
-            resource_info.Alignment = D3D12_DEFAULT_RESOURCE_PLACEMENT_ALIGNMENT;
+
+            if (desc->Flags & D3D12_RESOURCE_FLAG_USE_TIGHT_ALIGNMENT)
```
HansKristian-Work, line 7862: How does this interact with RTAS? I think for VKD3D_RESOURCE_ACCELERATION_STRUCTURE we cannot return less than 256b.

libs/vkd3d/resource.c
```diff
@@ -1122,6 +1130,12 @@ HRESULT vkd3d_get_image_allocation_info(struct d3d12_device *device,
      * align the image ourselves. */
     target_alignment = desc->Alignment ? desc->Alignment : D3D12_DEFAULT_RESOURCE_PLACEMENT_ALIGNMENT;
 
+    /* Tight alignment enforces small alignment for eligible resources */
+    if ((desc->Flags & D3D12_RESOURCE_FLAG_USE_TIGHT_ALIGNMENT) &&
+            d3d12_resource_supports_small_resource_alignment(desc, vkd3d_get_format(device, desc->Format,
+                    !!(desc->Flags & D3D12_RESOURCE_FLAG_ALLOW_DEPTH_STENCIL))))
+        target_alignment = D3D12_SMALL_RESOURCE_PLACEMENT_ALIGNMENT;
```
HansKristian-Work, line 1145: Looking over this, I can't immediately see a code path for being able to expose the minimum image alignment to application. E.g. if driver only needs 256 byte alignment or something for an image, it seems like we at minimum return 4 KiB to application here?

libs/vkd3d/device.c
doitsujin: Need to check, wasn't aware that there's a dedicated flag in D3D12.

libs/vkd3d/resource.c
doitsujin: `target_alignment` is used to pad allocations if the driver-required alignment is *larger*. If the driver gives us 256b alignment, this does nothing.

The intention behind this is to force 4k alignment for small resources so that we don't return 64k.

libs/vkd3d/device.c
HansKristian-Work: I think it depends on either the resource state in legacy model or resource flag for RTAS in enhanced barriers.

libs/vkd3d/device.c
HansKristian-Work: If we expose TIER_2, we have to ensure that any returned image alignment is >= bufferImageGranularity I think.

doitsujin: Think we can't even rely on tier 1 since we might still create linear images for weird formats.

libs/vkd3d/device.c
doitsujin: Native doesn't do anything here, but I went with the 256 byte route now anyway.

libs/vkd3d/device.c
```diff
@@ -7855,6 +7855,7 @@ static D3D12_RESOURCE_ALLOCATION_INFO* STDMETHODCALLTYPE d3d12_device_GetResourc
         info->Alignment = max(info->Alignment, resource_info.Alignment);
     }
 
+    info->SizeInBytes = align(info->SizeInBytes, info->Alignment);
```
HansKristian-Work, line 7899: align64?

HansKristian-Work: nvm, it's size_t already and allocating >4 GB resources on 32-bit is just nope.

PR #2640, opened 2025-10-01
tests: Add test for dynamically indexed root constants.

doitsujin:
FF Tactics relies on this to sort-of work. Currently we fail pipeline compilation.

Not sure how to fix this properly since the relevant part of dxil-spirv doesn't know if the input is dxbc or not, and I also don't know if this is a difference between legacy vs non-legacy cbuffer load or not since there doesn't seem to be a way to emit the former, or if this is *actually* just a dxbc vs dxil check. I don't want to break dxil validation either since games may be relying on this.


tests/d3d12_shaders.c
```diff
@@ -15005,3 +15005,113 @@ void test_derivative_hoisting_dxil(void)
...
+    resource_desc.DepthOrArraySize = 1u;
+    resource_desc.MipLevels = 1u;
+    resource_desc.SampleDesc.Count = 1u;
+    resource_desc.Layout = D3D12_TEXTURE_LAYOUT_ROW_MAJOR;
+    resource_desc.Flags = D3D12_RESOURCE_FLAG_ALLOW_UNORDERED_ACCESS;
+
+    hr = ID3D12Device_CreateCommittedResource(context.device, &heap_properties,
+        D3D12_HEAP_FLAG_NONE, &resource_desc, D3D12_RESOURCE_STATE_UNORDERED_ACCESS,
+        NULL, &IID_ID3D12Resource, (void**)&uav_resource);
+    ok(hr == S_OK, "Failed to create UAV resource, hr %#x.\n");
```
HansKristian-Work: missing hr in printf.

tests/d3d12_shaders.c
```diff
@@ -15005,3 +15005,113 @@ void test_derivative_hoisting_dxil(void)
...
+    memset(&pso_desc, 0, sizeof(pso_desc));
+    pso_desc.pRootSignature = context.root_signature;
+    pso_desc.CS = use_dxil ? cs_root_constant_indexing_dxil : cs_root_constant_indexing_dxbc;
+
+    /* For DXIL, there seems to be some runtime validation going on */
+    hr = ID3D12Device_CreateComputePipelineState(context.device,
+            &pso_desc, &IID_ID3D12PipelineState, (void**)&context.pipeline_state);
+
+    if (use_dxil)
+        ok(hr == E_INVALIDARG, "Got hr %#x, expected E_INVALIDARG.\n", hr);
```
HansKristian-Work: dxil-spirv will just allow it, so hr == S_OK if running against vkd3d-proton should be fine.

tests/shaders/shaders/cs_root_constant_indexing.cs_5_0.cs_6_0.hlsl
```diff
@@ -0,0 +1,16 @@
+cbuffer data_t : register(b0)
+{
+    uint4 in_data[12];
+}
+
+RWStructuredBuffer<uint2> out_data : register(u0);
+
+[numthreads(12,1,1)]
+void main(in uint3 tid : SV_DISPATCHTHREADID)
```
HansKristian-Work: I suppose test should use GROUPID instead since it more closely matches FFT case?

doitsujin: For the initial test I deliberately used this to check whether the non-uniform case works, will need to partially rewrite the test anyway.

tests/d3d12_shaders.c
doitsujin: simply made this todo for now, it *does* fail on native after all.

HansKristian-Work (approved):

PR #2641, opened 2025-10-02
dxil-spirv: Update submodule.

HansKristian-Work:
Fixes FF Tactics issues.


doitsujin (approved):
Can confirm it runs with this.

PR #2649, opened 2025-10-08
Implement shared resources using the D3DKMT API.

rbernon:
The D3D12 side of https://github.com/doitsujin/dxvk/pull/5257, same description.

### DIFF
--- a/include/private/vkd3d_d3dkmt.h
+++ b/include/private/vkd3d_d3dkmt.h
@@ -167,6 +167,25 @@ typedef struct _D3DDDI_OPENALLOCATIONINFO2
     ULONG_PTR Reserved[6];
 } D3DDDI_OPENALLOCATIONINFO2;
 
+typedef struct _D3DKMT_OPENRESOURCE
+{
+    D3DKMT_HANDLE hDevice;
+    D3DKMT_HANDLE hGlobalShare;
+    UINT NumAllocations;
+    union
+    {
+        D3DDDI_OPENALLOCATIONINFO *pOpenAllocationInfo;
+        D3DDDI_OPENALLOCATIONINFO2 *pOpenAllocationInfo2;
+    };
+    void *pPrivateRuntimeData;
+    UINT PrivateRuntimeDataSize;
+    void *pResourcePrivateDriverData;
+    UINT ResourcePrivateDriverDataSize;
+    void *pTotalPrivateDriverDataBuffer;
+    UINT TotalPrivateDriverDataBufferSize;
+    D3DKMT_HANDLE hResource;
+} D3DKMT_OPENRESOURCE;
+
 typedef struct _D3DKMT_OPENRESOURCEFROMNTHANDLE
 {
     D3DKMT_HANDLE hDevice;
@@ -191,6 +210,28 @@ typedef struct _D3DKMT_OPENSYNCOBJECTFROMNTHANDLE
     HANDLE hNtHandle;
     D3DKMT_HANDLE hSyncObject;
 } D3DKMT_OPENSYNCOBJECTFROMNTHANDLE;
+
+typedef struct _D3DKMT_QUERYRESOURCEINFO
+{
+    D3DKMT_HANDLE hDevice;
+    D3DKMT_HANDLE hGlobalShare;
+    void *pPrivateRuntimeData;
+    UINT PrivateRuntimeDataSize;
+    UINT TotalPrivateDriverDataSize;
+    UINT ResourcePrivateDriverDataSize;
+    UINT NumAllocations;
+} D3DKMT_QUERYRESOURCEINFO;
+
+typedef struct _D3DKMT_QUERYRESOURCEINFOFROMNTHANDLE
+{
+    D3DKMT_HANDLE hDevice;
+    HANDLE hNtHandle;
+    void *pPrivateRuntimeData;
+    UINT PrivateRuntimeDataSize;
+    UINT TotalPrivateDriverDataSize;
+    UINT ResourcePrivateDriverDataSize;
+    UINT NumAllocations;
+} D3DKMT_QUERYRESOURCEINFOFROMNTHANDLE;
 
 typedef struct _UNICODE_STRING {
     USHORT Length;        /* bytes */
@@ -316,9 +357,14 @@ EXTERN_C WINBASEAPI NTSTATUS WINAPI D3DKMTDestroyKeyedMutex(const D3DKMT_DESTROY
 EXTERN_C WINBASEAPI NTSTATUS WINAPI D3DKMTDestroySynchronizationObject(const D3DKMT_DESTROYSYNCHRONIZATIONOBJECT *desc);
 EXTERN_C WINBASEAPI NTSTATUS WINAPI D3DKMTEscape(const D3DKMT_ESCAPE *desc);
 EXTERN_C WINBASEAPI NTSTATUS WINAPI D3DKMTOpenAdapterFromLuid(D3DKMT_OPENADAPTERFROMLUID *desc);
+EXTERN_C WINBASEAPI NTSTATUS WINAPI D3DKMTOpenResource2(D3DKMT_OPENRESOURCE *desc);
 EXTERN_C WINBASEAPI NTSTATUS WINAPI D3DKMTOpenResourceFromNtHandle(D3DKMT_OPENRESOURCEFROMNTHANDLE *desc);
 EXTERN_C WINBASEAPI NTSTATUS WINAPI D3DKMTOpenSyncObjectFromNtHandle(D3DKMT_OPENSYNCOBJECTFROMNTHANDLE *desc);
+EXTERN_C WINBASEAPI NTSTATUS WINAPI D3DKMTQueryResourceInfo(D3DKMT_QUERYRESOURCEINFO *desc);
+EXTERN_C WINBASEAPI NTSTATUS WINAPI D3DKMTQueryResourceInfoFromNtHandle(D3DKMT_QUERYRESOURCEINFOFROMNTHANDLE *desc);
 EXTERN_C WINBASEAPI NTSTATUS WINAPI D3DKMTShareObjects(UINT count, const D3DKMT_HANDLE *handles, OBJECT_ATTRIBUTES *attr, UINT access, HANDLE *handle);
+
+extern HRESULT d3d12_device_open_resource_descriptor(struct d3d12_device *device, HANDLE handle, D3D12_RESOURCE_DESC1 *desc);
 
 #endif  /* _WIN32 */
 
--- a/libs/vkd3d/d3dkmt.c
+++ b/libs/vkd3d/d3dkmt.c
@@ -193,6 +193,209 @@ void d3d12_resource_close_export_kmt(struct d3d12_resource *resource, struct d3d
     D3DKMTDestroyAllocation(&destroy);
 }
 
+HRESULT d3d12_device_open_resource_descriptor(struct d3d12_device *device, HANDLE handle, D3D12_RESOURCE_DESC1 *desc)
+{
+    D3DKMT_DESTROYALLOCATION destroy = {0};
+    union d3dkmt_desc d3dkmt = {0};
+    UINT size;
+
+    if ((UINT_PTR)handle & 0xc0000000)
+    {
+        D3DDDI_OPENALLOCATIONINFO2 alloc = {0};
+        D3DKMT_QUERYRESOURCEINFO query = {0};
+        D3DKMT_OPENRESOURCE open = {0};
+
+        query.hDevice = device->kmt_local;
+        query.hGlobalShare = (UINT_PTR)handle;
+        query.pPrivateRuntimeData = &d3dkmt;
+        query.PrivateRuntimeDataSize = sizeof(d3dkmt);
+
+        if (D3DKMTQueryResourceInfo(&query) != STATUS_SUCCESS)
+            return E_INVALIDARG;
+        if (query.PrivateRuntimeDataSize < sizeof(d3dkmt.dxgi) || query.PrivateRuntimeDataSize > sizeof(d3dkmt))
+        {
+            WARN("Unsupported shared resource runtime data size %#x\n", query.PrivateRuntimeDataSize);
+            return E_NOTIMPL;
+        }
+
+        open.hDevice = device->kmt_local;
+        open.hGlobalShare = (UINT_PTR)handle;
+        open.NumAllocations = 1;
+        open.pOpenAllocationInfo2 = &alloc;
+        open.pPrivateRuntimeData = &d3dkmt;
+        open.PrivateRuntimeDataSize = query.PrivateRuntimeDataSize;
+
+        if (D3DKMTOpenResource2(&open) != STATUS_SUCCESS)
+        {
+            WARN("Failed to open shared resource handle %p\n", handle);
+            return E_INVALIDARG;
+        }
+        size = open.PrivateRuntimeDataSize;
+        destroy.hResource = open.hResource;
+    }
+    else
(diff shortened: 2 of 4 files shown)


libs/vkd3d/device.c
```diff
@@ -6563,13 +6563,33 @@ static HRESULT STDMETHODCALLTYPE d3d12_device_CreateSharedHandle(d3d12_device_if
...
             iface, object, attributes, access, debugstr_w(name), handle);
 
+    attr.Length = sizeof(attr);
+    attr.SecurityDescriptor = (void *)attributes;
+    if (name)
+    {
+        DWORD session, len, name_len = wcslen(name);
+
+        ProcessIdToSessionId(GetCurrentProcessId(), &session);
+        len = swprintf(buffer, ARRAY_SIZE(buffer), L"\\Sessions\\%u\\BaseNamedObjects\\", session);
```
HansKristian-Work, line 6566: WCHAR is uint16_t on Linux build, but L"" is 32-bit on native. I assume all of this Win32 stuff can be ifdeffed out on Linux anyway?

include/private/vkd3d_d3dkmt.h
```diff
@@ -22,6 +22,8 @@
 #ifdef _WIN32
 #include "vkd3d_windows.h"
 
+#include <d3d9.h>
```
HansKristian-Work, line 25: Why d3d9.h?

libs/vkd3d/device.c
```diff
@@ -6582,6 +6602,12 @@ static HRESULT STDMETHODCALLTYPE d3d12_device_CreateSharedHandle(d3d12_device_if
             return DXGI_ERROR_INVALID_CALL;
         }
 
+        if (!D3DKMTShareObjects(1, &resource->kmt_local, &attr, access, handle))
```
HansKristian-Work: Any particular reason why we're also setting the legacy shared metadata if this succeeds?

libs/vkd3d/resource.c
```diff
@@ -3677,6 +3678,72 @@ void d3d12_resource_decref_weak(struct d3d12_resource *resource)
...
 }
 
+static void d3d12_resource_open_kmt(struct d3d12_resource *resource, struct d3d12_device *device,
+        struct vkd3d_memory_allocation *allocation)
+{
+#ifdef _WIN32
+    const struct vkd3d_vk_device_procs *vk_procs = &device->vk_procs;
+    VkMemoryGetWin32HandleInfoKHR win32_handle_info;
+    D3DDDI_OPENALLOCATIONINFO2 alloc = {0};
+    D3DKMT_OPENRESOURCEFROMNTHANDLE open = {0};
```
HansKristian-Work: inverse :christmas_tree:

libs/vkd3d/resource.c
```diff
@@ -4146,6 +4215,9 @@ HRESULT d3d12_resource_create_committed(struct d3d12_device *device, const D3D12
         if (FAILED(hr = vkd3d_allocate_memory(device, &device->memory_allocator, &allocate_info, allocation)))
             goto fail;
 
+        if (heap_flags & D3D12_HEAP_FLAG_SHARED && export_info.handleTypes == VK_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_WIN32_BIT)
```
HansKristian-Work: We use parens around bitwise & in this case.

libs/vkd3d/resource.c
```diff
@@ -3677,6 +3678,72 @@ void d3d12_resource_decref_weak(struct d3d12_resource *resource)
...
+    open.PrivateRuntimeDataSize = 0;
+    open.pTotalPrivateDriverDataBuffer = &dummy;
+    open.TotalPrivateDriverDataBufferSize = 0;
+
+    win32_handle_info.sType = VK_STRUCTURE_TYPE_MEMORY_GET_WIN32_HANDLE_INFO_KHR;
+    win32_handle_info.pNext = NULL;
+    win32_handle_info.memory = allocation->device_allocation.vk_memory;
+    win32_handle_info.handleType = VK_EXTERNAL_MEMORY_HANDLE_TYPE_OPAQUE_WIN32_BIT;
+
+    if ((vr = VK_CALL(vkGetMemoryWin32HandleKHR(device->vk_device, &win32_handle_info, &open.hNtHandle))))
```
HansKristian-Work: Is this for importing? "open_kmt" would imply importing to me at least. I'm very confused why we're querying a Win32 HANDLE here. That implies we've already allocated memory.

libs/vkd3d/vkd3d_private.h
```diff
@@ -671,6 +671,7 @@ struct d3d12_shared_fence
     D3D12_FENCE_FLAGS d3d12_flags;
 
     VkSemaphore timeline_semaphore;
+    D3DKMT_HANDLE kmt_local;
```
HansKristian-Work, line 674: Are "0" kmt handles invalid objects?

libs/vkd3d/device.c
```diff
@@ -4218,6 +4219,36 @@ static void vkd3d_null_rtas_allocation_cleanup(
     }
 }
 
+static void d3d12_device_open_kmt(struct d3d12_device *device)
+{
+#ifdef _WIN32
+    D3DKMT_OPENADAPTERFROMLUID open_adapter = {0};
+    open_adapter.AdapterLuid = device->adapter_luid;
+
+    if (!D3DKMTOpenAdapterFromLuid(&open_adapter))
```
HansKristian-Work: Is this actually checking for success? We tend to use SUCCEEDED() macros in these 0-is-success cases.

libs/vkd3d/command.c
```diff
@@ -1508,6 +1509,43 @@ static ULONG STDMETHODCALLTYPE d3d12_shared_fence_AddRef(d3d12_fence_iface *ifac
...
+    win32_handle_info.sType = VK_STRUCTURE_TYPE_SEMAPHORE_GET_WIN32_HANDLE_INFO_KHR;
+    win32_handle_info.pNext = NULL;
+    win32_handle_info.semaphore = fence->timeline_semaphore;
+    win32_handle_info.handleType = VK_EXTERNAL_SEMAPHORE_HANDLE_TYPE_D3D12_FENCE_BIT;
+
+    if ((vr = VK_CALL(vkGetSemaphoreWin32HandleKHR(device->vk_device, &win32_handle_info, &open.hNtHandle))))
+        ERR("Failed to get exported semaphore handle, vr %d.\n", vr);
+    else
+    {
+        if ((status = D3DKMTOpenSyncObjectFromNtHandle(&open)))
```
HansKristian-Work: If we're holding onto a KMT object here, what does that accomplish for a sync object which has no external metadata? I thought the HANDLE itself would be enough.

libs/vkd3d/device.c
HansKristian-Work: If the NT API calls are different from normal Win32 api calls, explicit KMTCall() == NTSUCCESS or whatever it is would help.

libs/vkd3d/device.c
rbernon: Right, then this is already in an ifdef _WIN32 afaics?

include/private/vkd3d_d3dkmt.h
rbernon: This is for the D3D9 resource descriptors, which have needs a couple of D3D9 types (format, resource types). The import descriptor conversion at https://github.com/HansKristian-Work/vkd3d-proton/pull/2649/commits/dbc9f23499f5cd6887aa1228b18c38d68ff4faf6#diff-b0ee8131c26d1b6d9bc7a22a04d7827985d12ed2197fbe8f6d47a130f11ea2a8R6863 will need it.

libs/vkd3d/device.c
rbernon: This is the other way around, STATUS_SUCCESS == 0, see below.

libs/vkd3d/vkd3d_private.h
rbernon: Yes

libs/vkd3d/device.c
rbernon: SUCCEEDED is for HRESULT, this is returning a NTSTATUS, which also have success == 0 but otherwise are a different type and set of values.

libs/vkd3d/command.c
```diff
@@ -1508,6 +1509,43 @@ static ULONG STDMETHODCALLTYPE d3d12_shared_fence_AddRef(d3d12_fence_iface *ifac
...
+    D3DKMT_OPENSYNCOBJECTFROMNTHANDLE open = {0};
+    NTSTATUS status;
+    VkResult vr;
+
+    win32_handle_info.sType = VK_STRUCTURE_TYPE_SEMAPHORE_GET_WIN32_HANDLE_INFO_KHR;
+    win32_handle_info.pNext = NULL;
+    win32_handle_info.semaphore = fence->timeline_semaphore;
+    win32_handle_info.handleType = VK_EXTERNAL_SEMAPHORE_HANDLE_TYPE_D3D12_FENCE_BIT;
+
+    if ((vr = VK_CALL(vkGetSemaphoreWin32HandleKHR(device->vk_device, &win32_handle_info, &open.hNtHandle))))
```
HansKristian-Work: I have similar concerns as for memory objects. It looks as-if we're importing, but this is actually export? Maybe shared_fence_open_kmt name should be clarified so it's obvious it is exporting.

libs/vkd3d/resource.c
rbernon: No, this is called after exporting, and it "opens" the local KMT handle to the allocated resource, similar to other "open_kmt" functions also open the corresponding local KMT handle for the objects. I can change the name if it's confusing.

libs/vkd3d/device.c
```diff
@@ -6646,6 +6710,216 @@ static inline bool handle_is_kmt_style(HANDLE handle)
 {
     return ((ULONG_PTR)handle & 0x40000000) && ((ULONG_PTR)handle - 2) % 4 == 0;
 }
+
+static HRESULT d3d12_device_open_resource_descriptor(struct d3d12_device *device, HANDLE handle, D3D12_RESOURCE_DESC1 *desc)
```
HansKristian-Work: Could this live in the separate shared_metadata.c file?

libs/vkd3d/command.c
rbernon: Same thing as above, this is to open a local KMT handle, which are necessary to create the exported handles in CreateSharedHandle with D3DKMTShareObjects.

libs/vkd3d/resource.c
```diff
@@ -3677,6 +3678,129 @@ void d3d12_resource_decref_weak(struct d3d12_resource *resource)
...
+        }
+
+        CloseHandle(open.hNtHandle);
+    }
+
+    if (resource->kmt_local)
+    {
+        struct d3dkmt_d3d12_desc desc =
+        {
+            .d3d11.dxgi =
```
HansKristian-Work: We don't use C99 designated initializers since the rest of the code base we inherited never used it.

libs/vkd3d/command.c
rbernon: See elsewhere, local handles are required for the D3DKMTShareObjects calls, to create the actual Win32 shared handles (for the object name, rights, attributes etc...).

libs/vkd3d/resource.c
HansKristian-Work: "open_export_kmt" would clarify it a lot.

libs/vkd3d/device.c
HansKristian-Work: Right ... Please make the code explicitly check for success, since this pattern reminds me of the BOOL WINAPI stuff instead of 0-based success.

libs/vkd3d/device.c
rbernon: Sure, though IMO this is quite specific to the implementation of `d3d12_device_OpenSharedHandle` so I'm not sure it deserves a separate file? IMO that file is very much the old API and it is meant to be dropped eventually.

Still I can move it there if you think it's better.

libs/vkd3d/device.c
HansKristian-Work: We override WCHAR to uint16_t on Linux and use u"" to get forced 16-bit wchar since we have to consume that.

HansKristian-Work: Right, derp. This is already in ifdef, it was outside the diff window.

libs/vkd3d/device.c
HansKristian-Work: A new file dedicated to helpers for the new KMT API is fine too. Can rename the old one to shared_metadata_legacy.c or something and make a new shared_metadata_kmt.c if that helps. The shared resource stuff should be as out of sight as possible and the more code can be moved to separate files, the better.

libs/vkd3d/command.c
HansKristian-Work: Ok, so in theory they could be created on-demand every time there's a CreateSharedHandle() call, and destroyed after exporting? It's fine to keep the object around, but probably good to comment why the object lingers for non-obvious reasons.

rbernon: The vulkan specification states that when NT handle are used, vkGetMemoryWin32HandleKHR must not be called more than once, so I don't think we can call it lazily. https://registry.khronos.org/vulkan/specs/latest/man/html/VkMemoryGetWin32HandleInfoKHR.html#VUID-VkMemoryGetWin32HandleInfoKHR-handleType-00663

libs/vkd3d/d3dkmt.c
```diff
@@ -48,6 +48,35 @@ void d3d12_device_close_kmt(struct d3d12_device *device)
...
+    VkSemaphoreGetWin32HandleInfoKHR win32_handle_info;
+    D3DKMT_OPENSYNCOBJECTFROMNTHANDLE open = {0};
+    VkResult vr;
+
+    win32_handle_info.sType = VK_STRUCTURE_TYPE_SEMAPHORE_GET_WIN32_HANDLE_INFO_KHR;
+    win32_handle_info.pNext = NULL;
+    win32_handle_info.semaphore = fence->timeline_semaphore;
+    win32_handle_info.handleType = VK_EXTERNAL_SEMAPHORE_HANDLE_TYPE_D3D12_FENCE_BIT;
+
+    if ((vr = VK_CALL(vkGetSemaphoreWin32HandleKHR(device->vk_device, &win32_handle_info, &open.hNtHandle))))
```
HansKristian-Work, line 65: I think there is a problem here if the KMT API is not supported. If KMT is supported, we query this once, and do KMT -> HANDLE exports every time CreateSharedHandle is used, but in legacy path we end up calling this twice, violating

> VUID-VkSemaphoreGetWin32HandleInfoKHR-handleType-01127
If handleType is defined as an NT handle, [vkGetSemaphoreWin32HandleKHR](https://registry.khronos.org/vulkan/specs/latest/html/vkspec.html#vkGetSemaphoreWin32HandleKHR) must be called no more than once for each valid unique combination of semaphore and handleType

I suppose we need to hold on to open.hNtHandle if KMT API doesn't work, and use DuplicateHandle() in CreateSharedHandle instead?

libs/vkd3d/command.c
HansKristian-Work: Found a bug in this area for legacy path.

libs/vkd3d/d3dkmt.c
HansKristian-Work: Is it possible to use device->kmt_handle as a sentinel here for new vs old Wine behavior?

rbernon: Yeah, did that for both resource and semaphore. The old shared resource metadata code arguably breaks the contract, but I don't think it's worth fixing.

PR #2663, opened 2025-10-20
tests: Add test for ClearUAV without heap bound.

HansKristian-Work:
Verifies that native drivers never look at the GPU heap for ClearUAV, so we can never make use of it since there is very high chance of game breakage. D3D12 validation complains violently, but runtime doesn't trip device lost.


HansKristian-Work:
I haven't verified that, but I guess I can try Intel iGPU on my laptop ...

PR #2678, opened 2025-10-27
vkd3d: Avoid some redundant KMT destroy calls.

HansKristian-Work:
If device doesn't support it, we shouldn't spam it.

### DIFF
--- a/libs/vkd3d/d3dkmt.c
+++ b/libs/vkd3d/d3dkmt.c
@@ -55,7 +55,11 @@ void d3d12_shared_fence_open_export_kmt(struct d3d12_shared_fence *fence, struct
     D3DKMT_OPENSYNCOBJECTFROMNTHANDLE open = {0};
     VkResult vr;
 
-    if (!device->kmt_local) return; /* D3DKMT API isn't supported */
+    if (!device->kmt_local)
+    {
+        /* D3DKMT API isn't supported */
+        return;
+    }
 
     win32_handle_info.sType = VK_STRUCTURE_TYPE_SEMAPHORE_GET_WIN32_HANDLE_INFO_KHR;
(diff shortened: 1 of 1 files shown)


libs/vkd3d/d3dkmt.c
```diff
@@ -187,10 +190,13 @@ void d3d12_resource_open_export_kmt(struct d3d12_resource *resource, struct d3d1
 
 void d3d12_resource_close_export_kmt(struct d3d12_resource *resource, struct d3d12_device *device)
 {
-    D3DKMT_DESTROYALLOCATION destroy = {0};
-    destroy.hDevice = device->kmt_local;
-    destroy.hResource = resource->kmt_local;
-    D3DKMTDestroyAllocation(&destroy);
+    if (device->kmt_local && resource->kmt_local)
```
rbernon: Checking the resource handle should be enough.

HansKristian-Work: Fixed.

libs/vkd3d/d3dkmt.c
```diff
@@ -199,6 +205,8 @@ HRESULT d3d12_device_open_resource_descriptor(struct d3d12_device *device, HANDL
     union d3dkmt_desc d3dkmt = {0};
     UINT size;
 
+    if (!device->kmt_local) return E_NOTIMPL; /* D3DKMT API isn't supported */
```
doitsujin: weird formating?

HansKristian-Work: Inherited from previous PR. Can fix I guess.

HansKristian-Work: Done.

PR #2679, opened 2025-10-28
vkd3d: Add -Denable_extended_emulation

HansKristian-Work:
Only enables WMMA if FP8 is natively supported. FP16 emulation paths can be re-enabled with build option. (Official Proton builds will not enable it).

This decision is above my paygrade and it's done to work toward being able to support FSR4 more officially in Proton.


doitsujin (approved):

shmerl:
AMD supposedly will allow using FSR 4.1 on RDNA 2 and RDNA 3 GPUs. Will that unblock FP16 emulation?

[Source](https://x.com/jackhuynh/status/2054904153013387273).

mbriar:
The "official" FSR4.1 on RDNA3 will apparently use int8, just like the version that was leaked months ago. So it's unrelated to the fp16 emulation path in vkd3d-proton.

PR #2681, opened 2025-10-29
Attempt to resolve pulsing GPU signals and rewinds.

HansKristian-Work:
Resolve some rare deadlocks in Ninja Gaiden 4.


libs/vkd3d/command.c
```diff
@@ -952,16 +953,71 @@ static void d3d12_fence_signal_external_events_locked(struct d3d12_fence *fence,
...
+                }
+            }
+
+            need_ticket_rescan = false;
+        }
+
+        if (!has_unsatisfied_wait)
+            return false;
+
+        if (pending_value > fence->max_pending_virtual_timeline_value)
```
doitsujin, line 994: I'm not entirely sure which purpose the whole `max_pending_virtual_timeline_value` serves now, given that it is basically the source of all our problems. Since we remove any queued wait from the ticket list once we queue up a signal or signal on CPU, isn't it enough to just wait until that happens?

HansKristian-Work: The ticket list is only removed when the wait is actually complete. What we want to wait for is for a wait to either:

- Materialize into a concrete physical timeline. GPU might not be done. That's what max_pending does.
- Get actually signalled. That's what ticket being removed signals.

The main thing is that we cannot use the max_pending to detect this, because with rewinds the wait might be satisfied then pulled back to 0 again, so when this thread wakes up, it doesn't know there has been a proper signal, and deadlocks.

doitsujin: Ok, I understand why that distinction exists now, we can't check the fence value after the fact because it may have been rewound so removing the ticket on queueing signals won't work.

That being said, since we check tickets first, `max_pending` is only a function of `pending_updates` when we check it, which in turn means that this function could also just call `d3d12_fence_get_physical_wait_value_locked` in order to simplify this quite a bit, so `max_pending` can be nuked after all.

There's a `FIXME` in that function that would need to be removed, and similarly, `d3d12_fence_can_elide_wait_semaphore_locked` still checks the current fence value, which should also be removed or turned into an assert since the condition should never be true now.

HansKristian-Work: Right, I can probably fuse together the three separate functions we have now ... There's no reason to check for "can elide" and then query the physical timeline separately. I'll make a follow-up commit in this PR that cleans that up.

HansKristian-Work: Done.

doitsujin (approved):
Much clearer with the cleanup.

PR #2688, opened 2025-11-04
Unified image layouts - Phase 4/5

HansKristian-Work:
Since there is no longer a need to transition from common layout into transfer optimal layouts, we can lower copies immediately instead of batching the copies with extra barriers.

The idea behind this PR is to make image copies look more like buffer copies where we copy images right away without batching, and rely on tracking to figure out when we have to emit WAW memory barriers.

There are some exceptions to this, which makes it somewhat more complicated:

- Images which are placed may require us to handle aliasing, where UNDEFINED -> GENERAL has to happen. In this case, we have to enter the batching path since we really want to batch those barriers.
- Rather than unconditionally do layout transitions, the idea now is to only do so when non-committed resources are used.
- To support both styles seamlessly, after we add a copy to a batch, we can check if that copy can trivially be flushed out right away. This is the case when there is no potential image layout transition in play, and no extra memory barriers are required to resolve the copy. Once we've entered the batching path, we keep going with that until the batch is flushed for simplicity.
- Full image copies can now participate in the batching, which was not the case before. We simply model it as touching UINT32_MAX subresource in the WAW tracker.

In the batching tests, we avoid a lot of barriers now where before there would be bad pessimization. Also, since we're lowering the copies right away, it's possible to mix and match copy types without breaking batching and tripping barriers.

It's a quite large commit, but it's not easy to see how to split it up meaningfully since it's effectively a rewrite of how the batching works.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -81,6 +81,7 @@ static void d3d12_command_list_decay_optimal_dsv_resource(struct d3d12_command_l
         const struct d3d12_resource *resource, uint32_t plane_optimal_mask,
         struct d3d12_command_list_barrier_batch *batch);
 static void d3d12_command_list_end_transfer_batch(struct d3d12_command_list *list);
+static void d3d12_command_list_end_transfer_batch_if_trivial(struct d3d12_command_list *list);
 static void d3d12_command_list_end_wbi_batch(struct d3d12_command_list *list);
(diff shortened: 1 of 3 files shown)


doitsujin (approved):
Guess the tests are thorough enough to ensure correctness, but the complexity at this point is nasty.

PR #2689, opened 2025-11-05
Unified image layouts - Phase 5/5

HansKristian-Work:
This optimizes depth-stencil copies when they are copied as separate subresources.

Both NV and RADV (recently) optimize for this case.

This completes unified image layout bringup.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -8938,12 +8938,34 @@ static bool vk_image_copy_subresource_overlaps(const ID3D12Resource *res, const 
             subres->mipLevel == other_subres->mipLevel;
 }
 
+static bool vk_image_copy_subresource_can_fuse_aspects(
+        ID3D12Resource *dst, ID3D12Resource *src, const VkImageCopy2 *copy,
+        ID3D12Resource *orig_dst, ID3D12Resource *orig_src, const VkImageCopy2 *orig_copy)
+{
+    /* If we can find an identical copy, just fuse the aspects. */
+    return dst == orig_dst && src == orig_src &&
+            /* regions have to match */
+            memcmp(&copy->extent, &orig_copy->extent, sizeof(VkExtent3D)) == 0 &&
+            memcmp(&copy->dstOffset, &orig_copy->dstOffset, sizeof(VkExtent3D)) == 0 &&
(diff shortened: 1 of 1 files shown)


libs/vkd3d/command.c
```diff
@@ -8983,12 +9012,20 @@ static void STDMETHODCALLTYPE d3d12_command_list_CopyTextureRegion(d3d12_command
                     subres = &copy_info.copy.buffer_image.imageSubresource;
                     other_subres = &other_info->copy.buffer_image.imageSubresource;
                     assert(subres->layerCount == 1 && other_subres->layerCount == 1);
-                    alias = vk_image_copy_subresource_equals(copy_info.dst.pResource, subres,
+                    alias = vk_image_copy_subresource_overlaps(copy_info.dst.pResource, subres,
                             other_info->dst.pResource, other_subres);
                     break;
                 case VKD3D_BATCH_TYPE_COPY_IMAGE:
+                    if (vk_image_copy_subresource_fuse_aspects(
```
HansKristian-Work: Staring a bit at this, I think there's a potential issue if we fuse, then later find there was an alias. Need a small rethink.

doitsujin: Probably need to decouple the fuse check from actually modifying the aspect mask?

HansKristian-Work: Yes.

libs/vkd3d/command.c
```diff
@@ -9170,6 +9213,16 @@ static void d3d12_command_list_end_transfer_batch_if_trivial(struct d3d12_comman
 
     dst_resource = impl_from_ID3D12Resource(list->transfer_batch.batch[0].dst.pResource);
 
+    copy = &list->transfer_batch.batch[0].copy.image;
+    if (list->transfer_batch.batch_type == VKD3D_BATCH_TYPE_COPY_IMAGE &&
+        copy->dstSubresource.aspectMask == copy->srcSubresource.aspectMask &&
+        copy->dstSubresource.aspectMask != dst_resource->format->vk_aspect_mask)
```
doitsujin, line 9220: Should probably explicitly check for `aspectMask & (DEPTH | STENCIL)` as well since we might encounter planar images where we will most likely not be able to do any batching?

HansKristian-Work: Good point ...

HansKristian-Work: done

PR #2695, opened 2025-11-10
tests: also use todo_if in test_custom_border_color_limits for Turnip

zdobersek:
In test_custom_border_color_limits Turnip also hits the custom border color sampler limit of 4K samplers, so let's make use of the existing conditional todo.

### DIFF
--- a/tests/d3d12_descriptors.c
+++ b/tests/d3d12_descriptors.c
@@ -6165,7 +6165,7 @@ void test_custom_border_color_limits(void)
         /* NV will fail around 4k unique samplers. */
         if (is_nvidia_device(context.device))
             is_todo = flat_index >= 4000;
-        else if (is_amd_vulkan_device(context.device))
+        else if (is_amd_vulkan_device(context.device) || is_adreno_device(context.device))
             is_todo = flat_index >= 4096;
         else
             is_todo = false;


tests/d3d12_descriptors.c
```diff
@@ -6165,7 +6165,7 @@ void test_custom_border_color_limits(void)
         /* NV will fail around 4k unique samplers. */
         if (is_nvidia_device(context.device))
             is_todo = flat_index >= 4000;
-        else if (is_amd_vulkan_device(context.device))
+        else if (is_amd_vulkan_device(context.device) || is_adreno_device(context.device))
```
zdobersek, line 6168: `is_adreno_device()` covers both Turnip and the prop driver. I don't test the prop driver, and I'm not aware of anyone else doing it. The Vulkan HW database lists 4000 for `maxCustomBorderColorSamplers` on one version of the prop driver, so this condition isn't entirely correct for that case.

It might make sense to limit Adreno-specific test tweaks to only Turnip. Let me know, there's only a handful of them currently in the code.

doitsujin: I don't particularly care about proprietary qcom given that even DXVK is more or less non-functional on it, I'd be very surprised if vkd3d-proton worked on it at all in any capacity. Don't think we need to complicate things for the time being.

KostasGangstar-AI: > I don't particularly care about proprietary qcom given that even DXVK is more or less non-functional on it, I'd be very surprised if vkd3d-proton worked on it at all in any capacity. Don't think we need to complicate things for the time being.

I don't have a big sample from DX12 but I tried a few games in DXVK on my Snapdragon Elite and various DXVK versions worked nearly with zero graphical glitches in different prop and DXVK version combos because some combos have bugs, Hades and STALKER Enhanced Edition are the only DX12 games I tried and they both worked. Maybe Adreno 8xx prop drivers are good for PC emulation.

DXVK:

Screenshot_2025-11-08-19-03-37-787_com ludashi benchmark
Screenshot_2025-11-10-14-19-53-484_com ludashi benchmark
Screenshot_2025-11-07-09-16-08-565_com ludashi benchmark
Screenshot_2025-11-03-11-05-05-206_com ludashi benchmark

vkd3d proton:

Screenshot_2025-11-02-15-41-08-564_com ludashi benchmark

Edit: I wrote nearly because I had wrong colors in Fallout New Vegas and Hades 1 but I didn't mess with drivers and DXVK versions for those games so maybe I would fix them with other combos

PR #2703, opened 2025-11-13
Misc test suite fixes

### LINKED ISSUE #2699: test_line_rasterization failing on drivers without smooth line support
This is observed on Turnip, which can't support smooth lines on Adreno hardware.

Two test combinations in `test_line_rasterization` end up using `D3D12_LINE_RASTERIZATION_MODE_ALPHA_ANTIALIASED` mode, which is only translatable to Vulkan if the smooth line feature is present -- if not, it will use the default VK line rasterization mode, making these tests fail.

I don't think there's an options struct member that would cover this.

HansKristian-Work:
Fix #2699.

### DIFF
--- a/tests/d3d12_descriptors.c
+++ b/tests/d3d12_descriptors.c
@@ -6454,8 +6454,8 @@ void test_large_buffer_descriptors(void)
 
     static const uint32_t offsets[4] = {
         3ull * 1024 * 1024 * 1024 / 4 - 1,
-        3ull * 1024 * 1024 * 1024 / 2 - 1,
-        1ull * 1024 * 1024 * 1024 / 2 - 1 - 32 * 1024,
+        3ull * 1024 * 1024 * 1024 / 2 - 2,
+        1ull * 1024 * 1024 * 1024 / 2 - 2 - 32 * 1024,
         3ull * 1024 * 1024 * 1024,
     };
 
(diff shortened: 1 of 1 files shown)


tests/d3d12_crosstest.h
```diff
@@ -599,6 +604,35 @@ static ID3D12Device *create_device(void)
...
+    ID3D12DeviceExt_GetVulkanHandles(ext, &vk_instance, &vk_physical_device, &vk_device);
+    ID3D12DeviceExt_Release(ext);
+
+    pfn_vkGetPhysicalDeviceFeatures2
+            = (void *)pfn_vkGetInstanceProcAddr(vk_instance, "vkGetPhysicalDeviceFeatures2");
+    ok(pfn_vkGetPhysicalDeviceFeatures2, "vkGetPhysicalDeviceFeatures2 is NULL.\n");
+
+    memset(&device_features2, 0, sizeof(device_features2));
+    device_features2.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_FEATURES_2;
+    device_features2.pNext = pnext;
```
doitsujin, line 631: do we ever expect to have a use case where we need to query `DEVICE_FEATURES_2` itself?

Can probably leave this as-is for now and just special-case `pNext->sType == VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_FEATURES_2` if we need it though.

HansKristian-Work: If we do, we can certainly just change the function.

PR #2725, opened 2025-11-27
Implement deferred clears

doitsujin:
Draft because the test needs more, well, test cases, and some more real-world testing.

Merges render target clears into render passes, and uses `LOAD_OP_DONT_CARE` for images that were just discarded. The general idea is that we're super conservative around placed/reserved resources due to potential aliasing memes, but for committed render targets we should be able to delay everything until the respective image is either used or transitioned out of render target state.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -4140,7 +4140,7 @@ static VkPipelineStageFlags2 vk_queue_shader_stages(struct d3d12_device *device,
     return queue_shader_stages;
 }
 
-static void d3d12_command_list_flush_clears(struct d3d12_command_list *list,
+static void d3d12_command_list_discard_attachment(struct d3d12_command_list *list,
         struct d3d12_resource *resource, const VkImageSubresourceRange *subresources);
 
 static void d3d12_command_list_discard_attachment_barrier(struct d3d12_command_list *list,
@@ -4157,7 +4157,7 @@ static void d3d12_command_list_discard_attachment_barrier(struct d3d12_command_l
     if ((list->type == D3D12_COMMAND_LIST_TYPE_DIRECT) &&
             (resource->desc.Flags & D3D12_RESOURCE_FLAG_ALLOW_RENDER_TARGET))
     {
-        d3d12_command_list_flush_clears(list, resource, subresources);
+        d3d12_command_list_discard_attachment(list, resource, subresources);
 
         stages = VK_PIPELINE_STAGE_2_COLOR_ATTACHMENT_OUTPUT_BIT;
         access = VK_ACCESS_2_COLOR_ATTACHMENT_WRITE_BIT | VK_ACCESS_2_COLOR_ATTACHMENT_READ_BIT;
@@ -4166,7 +4166,7 @@ static void d3d12_command_list_discard_attachment_barrier(struct d3d12_command_l
     else if ((list->type == D3D12_COMMAND_LIST_TYPE_DIRECT) &&
             (resource->desc.Flags & D3D12_RESOURCE_FLAG_ALLOW_DEPTH_STENCIL))
     {
-        d3d12_command_list_flush_clears(list, resource, subresources);
+        d3d12_command_list_discard_attachment(list, resource, subresources);
 
         stages = VK_PIPELINE_STAGE_2_EARLY_FRAGMENT_TESTS_BIT | VK_PIPELINE_STAGE_2_LATE_FRAGMENT_TESTS_BIT;
         access = VK_ACCESS_2_DEPTH_STENCIL_ATTACHMENT_WRITE_BIT | VK_ACCESS_2_DEPTH_STENCIL_ATTACHMENT_READ_BIT;
@@ -5023,9 +5023,35 @@ static void d3d12_command_list_fuse_attachment_clear(struct d3d12_command_list *
         VkRenderingAttachmentInfo *attachment, struct d3d12_resource *resource,
         struct vkd3d_view *view, VkImageAspectFlagBits aspect)
 {
+    struct vkd3d_deferred_discard *discard;
+    VkImageSubresourceRange subresources;
     struct vkd3d_deferred_clear *clear;
     VkExtent3D extent;
     unsigned int i;
+
+    subresources = vk_subresource_range_from_view(view);
+
+    if (aspect & (VK_IMAGE_ASPECT_DEPTH_BIT | VK_IMAGE_ASPECT_STENCIL_BIT))
+        subresources.aspectMask &= aspect;
+
+    for (i = 0u; i < list->deferred_discard_count; i++)
+    {
+        discard = &list->deferred_discards[i];
+
+        /* We can only discard if the discard covers *all* view subresources */
+        if (discard->resource == resource && (discard->subresources.aspectMask & aspect) &&
+                discard->subresources.baseMipLevel <= subresources.baseMipLevel &&
+                discard->subresources.baseMipLevel + discard->subresources.levelCount >= subresources.baseMipLevel + subresources.levelCount &&
+                discard->subresources.baseArrayLayer <= subresources.baseArrayLayer &&
+                discard->subresources.baseArrayLayer + discard->subresources.layerCount >= subresources.baseArrayLayer + subresources.layerCount)
+        {
+
+            attachment->loadOp = VK_ATTACHMENT_LOAD_OP_DONT_CARE;
+
+            *discard = list->deferred_discards[--list->deferred_discard_count];
+            break;
+        }
+    }
 
     for (i = 0u; i < list->deferred_clear_count; i++)
     {
@@ -5093,6 +5119,27 @@ static bool d3d12_command_list_resource_overlaps_attachment(struct d3d12_command
     return false;
 }
 
+static void d3d12_command_list_flush_discards(struct d3d12_command_list *list,
+        struct d3d12_resource *resource, const VkImageSubresourceRange *subresources)
+{
+    unsigned int src, dst;
+
+    for (src = 0, dst = 0; src < list->deferred_discard_count; src++)
+    {
+        const struct vkd3d_deferred_discard *discard = &list->deferred_discards[src];
+
+        if (!vkd3d_check_subresource_overlap(resource, subresources, discard->resource, &discard->subresources))
+        {
+            if (dst < src)
+                list->deferred_discards[dst] = list->deferred_discards[src];
+
+            dst++;
+        }
+    }
+
+    list->deferred_discard_count = dst;
+}
+
 static void d3d12_command_list_flush_clears(struct d3d12_command_list *list,
         struct d3d12_resource *resource, const VkImageSubresourceRange *subresources)
 {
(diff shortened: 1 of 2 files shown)


libs/vkd3d/command.c
```diff
@@ -4976,6 +4983,158 @@ static bool d3d12_command_list_gather_pending_queries(struct d3d12_command_list
...
+static bool vkd3d_check_subresource_overlap(struct d3d12_resource *a, const VkImageSubresourceRange *a_subresources,
+        struct d3d12_resource *b, const VkImageSubresourceRange *b_subresources)
+{
+    /* NULL resource means all resources */
+    if (!a || !b)
+        return true;
+
+    if (a != b)
+    {
+        return d3d12_resource_may_alias_other_resources(a) &&
```
HansKristian-Work: I think we can detect if two placed resources can possibly alias each other by checking the bound heaps, right? The throw-hands-in-the-air case is if any one of the resources are sparse and the other is placed (since aliasing is a thing).

libs/vkd3d/command.c
```diff
@@ -12507,6 +12711,65 @@ static bool d3d12_barrier_subresource_range_covers_resource(const struct d3d12_r
...
+
+            return;
+        }
+
+        /* Submit clears if any clear overlaps with the given view */
+        clear_subresources = vk_subresource_range_from_view(clear->view);
+
+        if (vkd3d_check_subresource_overlap(resource, &subresources, clear->resource, &clear_subresources))
+        {
+
```
HansKristian-Work: weird newline

libs/vkd3d/command.c
```diff
@@ -4976,6 +4983,158 @@ static bool d3d12_command_list_gather_pending_queries(struct d3d12_command_list
...
+static bool d3d12_command_list_resource_overlaps_attachment(struct d3d12_command_list *list,
+        struct d3d12_resource *resource, const VkImageSubresourceRange *subresources)
+{
+    unsigned int i;
+
+    if (vkd3d_check_rtv_overlap(&list->dsv, resource, subresources))
+        return true;
+
+    for (i = 0u; i < ARRAY_SIZE(list->rtvs); i++)
```
HansKristian-Work: We should loop over colorAttachmentCount now I think.

libs/vkd3d/command.c
```diff
@@ -4976,6 +4983,158 @@ static bool d3d12_command_list_gather_pending_queries(struct d3d12_command_list
...
+        const struct vkd3d_deferred_clear *clear = &list->deferred_clears[src];
+        clear_subresources = vk_subresource_range_from_view(clear->view);
+
+        if (d3d12_command_list_resource_overlaps_attachment(list, clear->resource, &clear_subresources))
+        {
+            d3d12_command_list_execute_deferred_clear(list, clear);
+        }
+        else
+        {
+            if (dst < src)
```
HansKristian-Work, line 5192: I guess you do it like this to preserve the ordering of clears?

libs/vkd3d/command.c
```diff
@@ -12507,6 +12711,65 @@ static bool d3d12_barrier_subresource_range_covers_resource(const struct d3d12_r
...
+    {
+        clear = &list->deferred_clears[i];
+
+        if (clear->resource != resource)
+            continue;
+
+        if (clear->view == view)
+        {
+            /* Deduplicate clears or add a clear aspect */
+            clear->clear_aspects |= clear_aspects;
```
HansKristian-Work, line 12847: I think this may fall foul of the same case as copy batching had with aspect batching.

Think a situation where

- Clear depth layer { 0, 1 } - view A
- Clear stencil layer { 1, 2 } - view B
- Clear depth-stencil layer { 0, 1 } - view A - hazards with view B.

libs/vkd3d/command.c
```diff
@@ -4976,6 +4983,158 @@ static bool d3d12_command_list_gather_pending_queries(struct d3d12_command_list
...
+
+            if (clear->clear_aspects & aspect)
+            {
+                attachment->loadOp = VK_ATTACHMENT_LOAD_OP_CLEAR;
+                attachment->clearValue = clear->clear_value;
+
+                clear->clear_aspects &= ~aspect;
+            }
+
+            if (!(clear->clear_aspects))
```
HansKristian-Work: the parens here are a bit goofy.

libs/vkd3d/command.c
```diff
@@ -4976,6 +4983,158 @@ static bool d3d12_command_list_gather_pending_queries(struct d3d12_command_list
...
+                *clear = list->deferred_clears[--list->deferred_clear_count];
```
HansKristian-Work, line 5077: There are parts of the code which preserve order, while this does not. Is that intentional?

libs/vkd3d/command.c
HansKristian-Work: Technically we would need to see an aliasing barrier to make the two placed resources case work, unless they have the exact same offset with same resource desc, but that's probably being a bit overzealous.

libs/vkd3d/command.c
```diff
@@ -12507,6 +12711,65 @@ static bool d3d12_barrier_subresource_range_covers_resource(const struct d3d12_r
...
+        d3d12_command_list_flush_clears(list, NULL, NULL);
+    }
+
+    clear = &list->deferred_clears[list->deferred_clear_count++];
+    clear->resource = resource;
+    clear->view = view;
+    clear->clear_aspects = clear_aspects;
+    clear->clear_value = *clear_value;
+
+    if (d3d12_command_list_resource_overlaps_attachment(list, resource, &subresources))
```
HansKristian-Work: When does this happen? I would expect us to hit inline clears in that case.

libs/vkd3d/command.c
```diff
@@ -4976,6 +4983,158 @@ static bool d3d12_command_list_gather_pending_queries(struct d3d12_command_list
...
+    return vkd3d_check_subresource_overlap(resource, subresources,
+            rtv->resource, &rtv_subresources);
+}
+
+static bool d3d12_command_list_resource_overlaps_attachment(struct d3d12_command_list *list,
+        struct d3d12_resource *resource, const VkImageSubresourceRange *subresources)
+{
+    unsigned int i;
+
+    if (vkd3d_check_rtv_overlap(&list->dsv, resource, subresources))
```
HansKristian-Work, line 5110: Should this check be run if it's not in an active render pass?

libs/vkd3d/command.c
```diff
@@ -5081,6 +5107,25 @@ static bool d3d12_command_list_resource_overlaps_attachment(struct d3d12_command
...
+        const struct vkd3d_deferred_discard *discard = &list->deferred_discards[src];
+
+        if (!vkd3d_check_subresource_overlap(resource, subresources, discard->resource, &discard->subresources))
+        {
+            if (dst < src)
+                list->deferred_discards[dst] = list->deferred_discards[src];
+
+            dst++;
+        }
+    }
```
HansKristian-Work, line 5138: Missing update to deferred_discard_count?

libs/vkd3d/command.c
```diff
@@ -5011,10 +5011,36 @@ static void d3d12_command_list_fuse_attachment_clear(struct d3d12_command_list *
...
+
+    if (aspect & (VK_IMAGE_ASPECT_DEPTH_BIT | VK_IMAGE_ASPECT_STENCIL_BIT))
+        subresources.aspectMask &= aspect;
+
+    for (i = 0u; i < list->deferred_discard_count; i++)
+    {
+        discard = &list->deferred_discards[i];
+
+        /* We can only discard if the discard covers *all* view subresources */
+        if (discard->resource == resource && (discard->subresources.aspectMask & aspect) &&
```
HansKristian-Work, line 5042: aspect check is incomplete?

libs/vkd3d/command.c
doitsujin: This happens when a bound RTV covers e.g. only layer 0 but the app clears a view that covers layers 0-5 of the same resource, or a view that covers layer 0 as well but uses a different format.

Aliasing memes too.

libs/vkd3d/command.c
doitsujin: Right, need to check for overlaps first.

libs/vkd3d/command.c
HansKristian-Work: Right, it's the "over-clear" scenario ... Makes sense I guess.

HansKristian-Work: In the outer scope we hit `attachment_idx < 0 || !writable` and end the render pass anyway (but with false argument), and I guess we hit that case because the clear isn't fully contained in any active render pass.

libs/vkd3d/command.c
doitsujin: The ordering doesn't *really* matter, can always change the other paths to do it like we do here.

libs/vkd3d/command.c
doitsujin: `aspect` is a FlagBits type here, i.e. only ever one aspect.

libs/vkd3d/command.c
doitsujin: Yeah the original intent was to not do that and just defer even if we're inside a render pass. I need to write more tests to make sure that those cases actually work.

libs/vkd3d/command.c
doitsujin: derp

libs/vkd3d/command.c
doitsujin: This can be called from `d3d12_command_list_begin_render_pass` *before* we mark the render pass as active.

If anything we should check some of the callers, but at worst those will suspend an already ended render pass i.e. do nothing, so it shouldn't affect correctness at least.

libs/vkd3d/command.c
```diff
@@ -12507,6 +12711,71 @@ static bool d3d12_barrier_subresource_range_covers_resource(const struct d3d12_r
...
+    clear = &list->deferred_clears[list->deferred_clear_count++];
+    clear->resource = resource;
+    clear->view = view;
+    clear->clear_aspects = clear_aspects;
+    clear->clear_value = *clear_value;
+
+    /* May need to transition active render targets here if there is overlap */
+    if ((list->rendering_info.state_flags & VKD3D_RENDERING_ACTIVE) &&
+            d3d12_command_list_resource_overlaps_attachment(list, resource, &subresources))
+        d3d12_command_list_end_current_render_pass(list, false);
```
HansKristian-Work, line 12876: Deferring clears will not work if there are subsequent draws. In that case, where we clear late, we clear actual draws we shouldn't have, so I think this placement is correct.

libs/vkd3d/command.c
HansKristian-Work: I see, we're actually checking depth and stencil separate in different calls.

libs/vkd3d/command.c
doitsujin: Not sure I understand the comment, but TLDR is that we end the pass here so that the `start_render_pass` call for the next draw would execute the problematic clear first.

doitsujin: updated comment for clarity.

PR #2731, opened 2025-12-01
vkd3d: Use render_target_count directly instead of deducing it.

HansKristian-Work:
Cleans up a lot of jank where we iterate over all 8 RTs.

Pull this commit early from the suspend-resume work since we use this new pattern for clear fuse.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -3285,7 +3285,7 @@ static int d3d12_command_list_find_attachment_view(struct d3d12_command_list *li
     }
     else
     {
-        for (i = 0; i < D3D12_SIMULTANEOUS_RENDER_TARGET_COUNT; i++)
+        for (i = 0; i < list->rendering_info.info.colorAttachmentCount; i++)
         {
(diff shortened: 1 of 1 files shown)


libs/vkd3d/command.c
```diff
@@ -17159,6 +17155,9 @@ static void STDMETHODCALLTYPE d3d12_command_list_BeginRenderPass(d3d12_command_l
         }
     }
 
+    /* TODO: Does BeginRenderPass clobber OMSetRenderTargets? */
```
doitsujin: Spec doesn't say anything but iirc at least one implementation behaves as-if all render targets are unbound after a render pass.

HansKristian-Work: k, so it's effectively UB then. We already clobber the render targets anyway, so can just keep it as-is.

doitsujin: actually, yeah, `d3d12_command_list_EndRenderPass` resets all RTV state so i think we need to set the rtv count to 0 as well?

HansKristian-Work: Oh, it does, right ...

PR #2732, opened 2025-12-01
Implement experimental view instancing over multiview - Phase 1

HansKristian-Work:
Initial refactorings.

### DIFF
--- a/libs/vkd3d/state.c
+++ b/libs/vkd3d/state.c
@@ -4780,6 +4780,69 @@ static bool d3d12_graphics_pipeline_state_needs_noop_fs(
             device->device_info.vulkan_1_2_properties.driverID == VK_DRIVER_ID_MESA_RADV;
 }
 
+static bool d3d12_pipeline_state_validate_view_instancing(struct d3d12_device *device,
+        struct d3d12_graphics_pipeline_state *graphics, const struct d3d12_pipeline_state_desc *desc)
+{
+    /* Need to map ViewIndex (32-bit) back to 2-bit ViewID (max view instancing in D3D12 is 4). */
+    uint32_t limit = min(device->device_info.vulkan_1_1_properties.maxMultiviewViewCount, 32 / 2);
+    unsigned int i;
+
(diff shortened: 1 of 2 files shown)


libs/vkd3d/state.c
```diff
@@ -4780,6 +4780,69 @@ static bool d3d12_graphics_pipeline_state_needs_noop_fs(
             device->device_info.vulkan_1_2_properties.driverID == VK_DRIVER_ID_MESA_RADV;
 }
 
+static bool d3d12_pipeline_state_validate_view_instancing(struct d3d12_device *device,
+        struct d3d12_graphics_pipeline_state *graphics, const struct d3d12_pipeline_state_desc *desc)
+{
+    /* Need to map ViewIndex back to 2-bit ViewID (max view instancing in D3D12 is 4). */
+    uint32_t limit = min(device->device_info.vulkan_1_1_properties.maxMultiviewViewCount, 16);
```
doitsujin: where is the 16 coming from?

libs/vkd3d/state.c
```diff
@@ -4780,6 +4780,69 @@ static bool d3d12_graphics_pipeline_state_needs_noop_fs(
...
+    unsigned int i;
+
+    if (desc->view_instancing_desc.Flags & D3D12_VIEW_INSTANCING_FLAG_ENABLE_VIEW_INSTANCE_MASKING)
+    {
+        /* There are three potential strategies for this approach.
+         * - Push down a u32 which nops out the draw call based on ViewID.
+         *   This is somewhat questionable w.r.t. queries, UAV and potentially Xfb.
```
doitsujin, line 4794: XFB is banned with multiview in Vulkan, and is at the very least UB in d3d12 due to interactioins with ExecuteIndirect, so I don't think we have to worry about that.

Tier 2 also doesn't make any hard guarantees on how geometry stages are executed, so I think masking in the shader that does View ID stuff is a valid implementation if we ever need to support this.

libs/vkd3d/state.c
HansKristian-Work: 32 bits viewMask / 2-bit ViewID mask.

HansKristian-Work: Clarified.

doitsujin (approved):

PR #2737, opened 2025-12-02
Suspend resume - Phase 2

HansKristian-Work:
Adds the test case as well as the simple OMSetRenderTarget filtering. Next phase will have more of the real meat.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -12764,6 +12764,69 @@ static void d3d12_command_list_invalidate_ds_state(struct d3d12_command_list *li
     }
 }
 
+static bool d3d12_command_list_filter_set_render_targets(struct d3d12_command_list *list,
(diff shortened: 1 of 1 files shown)


libs/vkd3d/command.c
```diff
@@ -12764,6 +12764,69 @@ static void d3d12_command_list_invalidate_ds_state(struct d3d12_command_list *li
...
+    }
+    else
+    {
+        VkFormat prev_dsv_format;
+
+        if (rtv_desc->view != list->dsv.view)
+            return false;
+
+        /* We may keep the same DSV, but plane write masks may have changed.
+         * We don't have to split render passes just for this however. */
```
doitsujin, line 12821: i suppose the `pso_invalidates_rp` thing handles all the cases where we do need to split?

HansKristian-Work: ye, if the depth layout itself changes, the render pass is invalidated. This invalidation is only to make sure dynamic state is updated.

PR #2739, opened 2025-12-03
Suspend resume - Phase 3

HansKristian-Work:
Implements the initial scaffolding for suspend-resume.

Basic idea is that if the vendor ID is considered "tiler-adjacent", we speculate that:

- Every renderpass that is begun is SUSPENDING. It's possible the command list will Close() before we end the render pass. In this case, we may or may not be able to link with a subsequent RESUME, so record a "fixup" command buffer that does RESUME -> END. On ExecuteCommandLists, we may or may not insert this command list.

- For resume, we need to record if there has been incompatible action commands. If nothing has happened, we can attempt a RESUME with similar fixup command lists if it doesn't work out.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -21186,6 +21186,10 @@ static bool d3d12_command_queue_needs_staggered_submissions_locked(struct d3d12_
     if (vkd3d_config_flags & VKD3D_CONFIG_FLAG_NO_STAGGERED_SUBMIT)
         return false;
 
+    /* Cannot meaningfully stagger submits if we're doing suspend resume style render passes. */
+    if (command_queue->device->workarounds.tiler_suspend_resume)
+        return false;
+
     current_time_ns = vkd3d_get_current_time_ns();
 
     /* Make sure current command queue is recognized as busy */


libs/vkd3d/vkd3d_private.h
```diff
@@ -3032,6 +3033,34 @@ struct d3d12_command_list_iteration
...
+struct d3d12_command_list_render_pass_suspend_resume
+{
+    struct d3d12_command_list_render_pass_suspend_resume_compat resume, suspend;
+
+    /* If true, we have performed an action command, so it's not possible to attempt a resume. */
+    bool block_resume;
+
+    /* If true, we have complex fixup work happening during resume,
+     * which blocks any hope of fusing. */
+    bool complex_resume;
```
HansKristian-Work, line 3057: In future PRs, complex_resume and complex_suspend will go away, since there's lots of work to hoist and sink work outside a render pass chain.

libs/vkd3d/command.c
```diff
@@ -5315,6 +5359,113 @@ static void d3d12_command_list_discard_attachment(struct d3d12_command_list *lis
     discard->subresources = *subresources;
 }
 
+static void d3d12_command_list_copy_render_pass_suspend_resume_compat(
+        struct d3d12_command_list *list, struct d3d12_command_list_render_pass_suspend_resume_compat *compat)
+{
+    unsigned int i;
```
HansKristian-Work, line 5368: As a follow-up, I'll need to check what EndRenderPass() does here since it will clear out the rendering info.

HansKristian-Work: Fixed.

libs/vkd3d/command.c
```diff
@@ -17317,15 +17318,22 @@ static void STDMETHODCALLTYPE d3d12_command_list_EndRenderPass(d3d12_command_lis
...
-    memset(&list->dsv, 0, sizeof(list->dsv));
-    list->rendering_info.info.colorAttachmentCount = 0;
+
+    if (!suspending)
+    {
+        d3d12_command_list_resolve_render_pass_attachments(list);
+
+        /* Bound render targets are implicitly unbound after the render pass */
+
+        /* Do not clear this when suspending, since we need the compatibility information.
```
HansKristian-Work, line 17597: We call end_current_render_pass here, which would break suspend-resume, so I'll need more tests to cover that scenario. I've been focusing mostly on legacy rendering for now.

libs/vkd3d/command.c
```diff
@@ -4461,6 +4495,9 @@ static void d3d12_command_list_flush_subresource_updates(struct d3d12_command_li
     /* Do not emit another barrier here, let the caller handle that as
      * necessary. If this is used at the end of a command buffer, no
      * barrier is necessary as submissions will emit a host barrier. */
+
+    if (list->rendering_info.state_flags & VKD3D_RENDERING_END_OF_COMMAND_LIST)
+        list->cmd.suspend_resume.complex_suspend = true;
```
doitsujin, line 4500: do we need to set complex_suspend when flushing clears as well?

HansKristian-Work: Right ... I missed that since I wrote this before clear flushing.

HansKristian-Work: Fixed.

PR #2743, opened 2025-12-04
Suspend resume - Phase 5

HansKristian-Work:
Implements query reset hoisting to start of ECL. Also misc fixes thrown in.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -6127,6 +6127,8 @@ static HRESULT d3d12_command_list_build_init_commands(struct d3d12_command_list 
             dep_info.pMemoryBarriers = &barrier;
 
             VK_CALL(vkCmdPipelineBarrier2(iteration->vk_post_indirect_barrier_commands, &dep_info));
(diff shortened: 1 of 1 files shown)


libs/vkd3d/command.c
```diff
@@ -6243,17 +6243,18 @@ static HRESULT STDMETHODCALLTYPE d3d12_command_list_Close(d3d12_command_list_ifa
     return S_OK;
 }
 
-static bool d3d12_command_list_find_query(struct d3d12_command_list *list,
+static bool vkd3d_find_query(
+        const struct vkd3d_query_range *query_ranges, size_t query_ranges_count,
```
doitsujin, line 6266: maybe a bit of a nit, but I'd much prefer to have a dedicated struct the for query range array + capacity + size here since we're using it in quite a few different places, to keep things more readable.

HansKristian-Work: ack

doitsujin (approved):

PR #2744, opened 2025-12-05
Suspend resume - Phase 6/6

HansKristian-Work:
Implements command buffer reordering for cleanup and post-indirect commands.

With this, we have implemented best-effort support for OMSetRenderTarget style rendering on tilers I think. Only potential improvement I can think of is to let Discard() turn into STORE_OP_DONT_CARE somehow, but that will be rather difficult to deal with. It should be much easier to deal with in the actual render pass API.


doitsujin (approved):
The complexity with all the different command buffers is certainly, uh, *something*, at this point.

PR #2746, opened 2025-12-08
Fix deferred clear synchronization and some VVL errors

doitsujin:
Some fallout from the King of Bones issue.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -5149,35 +5149,39 @@ static bool vkd3d_check_subresource_overlap(struct d3d12_resource *a, const VkIm
 }
 
 static void d3d12_command_list_fuse_attachment_clear(struct d3d12_command_list *list,
-        VkRenderingAttachmentInfo *attachment, struct d3d12_resource *resource,
-        struct vkd3d_view *view, VkImageAspectFlagBits aspect,
+        VkRenderingAttachmentInfo *attachment, VkRenderingAttachmentInfo *stencil_attachment,
+        struct d3d12_resource *resource, struct vkd3d_view *view, VkImageAspectFlags aspect_mask,
         struct d3d12_command_list_barrier_batch *batch)
 {
-    struct vkd3d_deferred_discard *discard;
     VkImageSubresourceRange subresources;
-    struct vkd3d_deferred_clear *clear;
     bool requires_barrier;
     VkExtent3D extent;
     unsigned int i;
 
     subresources = vk_subresource_range_from_view(view);
 
-    if (aspect & (VK_IMAGE_ASPECT_DEPTH_BIT | VK_IMAGE_ASPECT_STENCIL_BIT))
-        subresources.aspectMask &= aspect;
+    if (aspect_mask & (VK_IMAGE_ASPECT_DEPTH_BIT | VK_IMAGE_ASPECT_STENCIL_BIT))
+        subresources.aspectMask &= aspect_mask;
 
     for (i = 0u; i < list->deferred_discard_count; i++)
     {
-        discard = &list->deferred_discards[i];
+        struct vkd3d_deferred_discard *discard = &list->deferred_discards[i];
 
-        /* We can only discard if the discard covers *all* view subresources */
-        if (discard->resource == resource && (discard->subresources.aspectMask & aspect) &&
+        /* Need to look at the view aspect mask here rather than the aspect_mask parameter since the
+         * latter would be COLOR_BIT for planar images, whereas discards refer to resource aspects. */
+        VkImageAspectFlags discard_aspects = discard->subresources.aspectMask & subresources.aspectMask;
+
+        if (discard->resource == resource && discard_aspects &&
                 discard->subresources.baseMipLevel <= subresources.baseMipLevel &&
                 discard->subresources.baseMipLevel + discard->subresources.levelCount >= subresources.baseMipLevel + subresources.levelCount &&
                 discard->subresources.baseArrayLayer <= subresources.baseArrayLayer &&
                 discard->subresources.baseArrayLayer + discard->subresources.layerCount >= subresources.baseArrayLayer + subresources.layerCount)
         {
+            if (discard_aspects != VK_IMAGE_ASPECT_STENCIL_BIT)
+                attachment->loadOp = VK_ATTACHMENT_LOAD_OP_DONT_CARE;
 
-            attachment->loadOp = VK_ATTACHMENT_LOAD_OP_DONT_CARE;
+            if (discard_aspects & VK_IMAGE_ASPECT_STENCIL_BIT)
+                stencil_attachment->loadOp = VK_ATTACHMENT_LOAD_OP_DONT_CARE;
 
             *discard = list->deferred_discards[--list->deferred_discard_count];
             break;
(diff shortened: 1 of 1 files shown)


libs/vkd3d/command.c
```diff
@@ -5210,53 +5212,73 @@ static void d3d12_command_list_fuse_attachment_clear(struct d3d12_command_list *
...
-                    if (aspect & (VK_IMAGE_ASPECT_DEPTH_BIT | VK_IMAGE_ASPECT_STENCIL_BIT))
+                memset(&vk_image_barrier, 0, sizeof(vk_image_barrier));
+                vk_image_barrier.sType = VK_STRUCTURE_TYPE_IMAGE_MEMORY_BARRIER_2;
+                vk_image_barrier.image = clear->resource->res.vk_image;
+                vk_image_barrier.oldLayout = VK_IMAGE_LAYOUT_UNDEFINED;
+                vk_image_barrier.subresourceRange = vk_subresource_range_from_view(clear->view);
+
+                if (aspect & (VK_IMAGE_ASPECT_DEPTH_BIT | VK_IMAGE_ASPECT_STENCIL_BIT))
+                {
+                    if (!emitted_dsv_barrier)
```
HansKristian-Work: If we fuse ClearDSV(depth) and ClearDSV(stencil) into one clear entry, is it even possible to hit the path where emitted_dsv_clear is true?

HansKristian-Work: Could it happen if ClearDSV(depth, layers = 2), ClearDSV(stencil, layers = 3) perhaps? I assume those would not be possible to fuse into one entry.

doitsujin: I confused myself here, this is nonsense indeed.

The intention here was to only emit any sort of d/s transition once and handle both aspects in one go, but since we call `fuse_attachment_clear` twice, this check doesn't actually do that. Need to go back to transitioning each aspect individually.

libs/vkd3d/command.c
```diff
@@ -5210,53 +5212,73 @@ static void d3d12_command_list_fuse_attachment_clear(struct d3d12_command_list *
...
+                                    : vk_separate_stencil_layout(vk_image_barrier.newLayout);
+                        }
+
                         vk_image_barrier.srcStageMask = VK_PIPELINE_STAGE_2_LATE_FRAGMENT_TESTS_BIT;
                         vk_image_barrier.dstStageMask = VK_PIPELINE_STAGE_2_EARLY_FRAGMENT_TESTS_BIT |
                                 VK_PIPELINE_STAGE_2_LATE_FRAGMENT_TESTS_BIT;
                         vk_image_barrier.srcAccessMask = VK_ACCESS_2_DEPTH_STENCIL_ATTACHMENT_WRITE_BIT;
                         vk_image_barrier.dstAccessMask = VK_ACCESS_2_DEPTH_STENCIL_ATTACHMENT_WRITE_BIT;
+
+                        vk_image_barrier.subresourceRange.aspectMask = clear->clear_aspects;
```
HansKristian-Work: For a placed DSV, this seems a bit iffy.

When this is called with DEPTH, we'll do a full resource barrier UNDEFINED -> DS. Then, it's called with STENCIL, and we'll do a separate stencil UNDEFINED -> DS. I think it makes more sense to handle depth and stencil attachment info together perhaps?

doitsujin: see above.

libs/vkd3d/command.c
HansKristian-Work: I think it makes more sense to fuse since metadata clear on AMD at least likes to do those together.

libs/vkd3d/command.c
doitsujin: Should be fixed now by transitioning aspects one by one, that just happens to be broken on proprietary AMD.

libs/vkd3d/command.c
```diff
@@ -5376,9 +5376,35 @@ static void d3d12_command_list_flush_clears(struct d3d12_command_list *list,
...
 
+    memset(&barrier, 0, sizeof(barrier));
+    barrier.sType = VK_STRUCTURE_TYPE_MEMORY_BARRIER_2;
+
+    memset(&dep_info, 0, sizeof(dep_info));
+    dep_info.sType = VK_STRUCTURE_TYPE_DEPENDENCY_INFO;
+    dep_info.memoryBarrierCount = 1u;
+    dep_info.pMemoryBarriers = &barrier;
+
+    if (list->rendering_info.info.colorAttachmentCount)
```
HansKristian-Work: Could probably move this down to the barrier.srcStageMask path? Seems redundant to set all this up if it's not going to be used.

doitsujin: done

libs/vkd3d/command.c
doitsujin: done

libs/vkd3d/command.c
```diff
@@ -6764,24 +6846,24 @@ static bool d3d12_command_list_update_rendering_info(struct d3d12_command_list *
...
-            {
-                d3d12_command_list_fuse_attachment_clear(list, &rendering_info->stencil,
-                    list->dsv.resource, list->dsv.view, VK_IMAGE_ASPECT_STENCIL_BIT, &barrier);
-            }
+                dsv_optimal_aspects |= VK_IMAGE_ASPECT_STENCIL_BIT;
 
             rendering_info->info.pStencilAttachment = &rendering_info->stencil;
         }
+
+        if (dsv_optimal_aspects)
```
HansKristian-Work, line 6854: Does this mean we cannot fuse clears for read-only DSVs? Is that a necessary restriction? iirc the only thing we use read-only DSV for is dynamic depth/stencil test.

doitsujin: Some sort of restriction is necessary in the sense that we cannot legally clear `DEPTH` if `DEPTH` is going to be in a read-only layout during the render pass. If we use `GENERAL` for everything anyway then we don't hit the restriction.

I'm not sure what to really do here since we (have to) determine rendering dsv layout based on PSO state etc before all of this happens.

libs/vkd3d/command.c
```diff
@@ -5204,59 +5211,84 @@ static void d3d12_command_list_fuse_attachment_clear(struct d3d12_command_list *
...
+                    if (clear_aspects & VK_IMAGE_ASPECT_STENCIL_BIT)
+                        plane_write_mask |= VKD3D_STENCIL_PLANE_OPTIMAL;
 
-                        vk_image_barrier.newLayout = d3d12_resource_pick_layout(clear->resource,
-                                VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL);
+                    vk_image_barrier.newLayout = dsv_plane_optimal_mask_to_layout(
+                            d3d12_command_list_notify_dsv_writes(list, clear->resource, clear->view, plane_write_mask),
+                            resource->format->vk_aspect_mask);
+
+                    if (vk_image_barrier.newLayout != list->dsv_layout)
```
HansKristian-Work: This feels fishy. Given this is called in update_rendering_info, we have already called promote_dsv_layout(), and there's code earlier that blocks clear fuse if we don't have list->dsv_plane_optimal_mask.

I can't think of a scenario where this barrier is actually needed.

libs/vkd3d/command.c
HansKristian-Work: ... reading the code some more, it's probably required since we've already locked in the dsv_layout when update_graphics_pipeline was called.

libs/vkd3d/command.c
HansKristian-Work: One thing that can clean this up as well is that in defer_attachment_clear() we don't register the DSV writes, which means when we lock in the layout, we will not have the knowledge that there have been writes to those aspects.

One potential scenario that could trigger this:

- ClearDSV(D), gets deferred. This *should* probably register that depth was written.
- OMSetRenderTargets(D)
- SetPipeline(read-only depth)
- Draw()

I think in this scenario, we will not get the clear fuse we want.

doitsujin: well this code was already there, I guess if the layout can't change if we don't add any optimal planes then we can just assume `list->dsv_layout`.

HansKristian-Work: Hm ... Just using list->dsv_layout should work I think?

HansKristian-Work: When I wrote the UNDEFINED on clear on path I think I missed the fact that we don't clear non-optimal planes.

PR #2751, opened 2025-12-11
Add present timing support

HansKristian-Work:


mbriar:
Will this eventually allow to implement SyncInterval > 1 with proper frame pacing? I believe right now it would use a framelimiter that won't vsync window reliably? It looks like it would require presentAtAbsoluteTimeSupported or presentAtRelativeTimeSupported, so from the current state of https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/38770 that means x11/xwayland or a wayland compositor with commit_timing support, is that correct?

libs/vkd3d/device.c
```diff
@@ -2482,6 +2487,24 @@ static void vkd3d_physical_device_info_init(struct vkd3d_physical_device_info *i
         vk_prepend_struct(&info->features2, &info->shader_mixed_float_dot_product_features);
     }
 
+    if (vulkan_info->KHR_present_id2)
```
doitsujin, line 2508: as mentioned on Discord, these are broken on Nvidia 590 so should probably gate it behind present_timing support or 595 drivers.

libs/vkd3d/swapchain.c
```diff
@@ -2063,6 +2300,31 @@ static void dxgi_vk_swap_chain_recreate_swapchain_in_present_task(struct dxgi_vk
...
+        dxgi_vk_swap_chain_poll_time_domains(chain);
+        dxgi_vk_swap_chain_poll_calibration(chain);
+
+        /* For the first timing requests, ensure we're requesting a sensible time domain. */
+        if (chain->timing.time_domains_count)
+            chain->timing.feedback.present_time_domain_id = chain->timing.time_domain_ids[0];
+
+        /* This could lead to problems if we have IMMEDIATE, but we ignore present timing for IMMEDIATE or MAILBOX,
+         * so there is no real risk of overflow. */
+        if (VK_CALL(vkSetSwapchainPresentTimingQueueSizeEXT(vk_device, chain->present.vk_swapchain,
```
doitsujin, line 2322: do we need to do anything special if this fails or does this just mean that we can't query timing info?

Just wondering if we need to opt-out of timing for the swapchain in that case.

libs/vkd3d/swapchain.c
```diff
@@ -2606,12 +2867,155 @@ static VkResult dxgi_vk_swap_chain_try_acquire_next_image(struct dxgi_vk_swap_ch
...
+            chain->timing.feedback.present_count &&
+            chain->timing.refresh_duration >= DELTA_NS &&
+            (chain->present.timing_absolute || chain->present.timing_relative);
+
+    if (use_present_timing_target && frame_limiter_ns > chain->timing.refresh_duration)
+    {
+        uint64_t effective_interval = chain->timing.refresh_duration;
+        uint64_t align;
+
+        if (chain->timing.refresh_interval != 0 && chain->timing.refresh_interval != UINT64_MAX)
```
doitsujin: `UINT64_MAX` was VRR?

libs/vkd3d/swapchain.c
```diff
@@ -2771,11 +3204,38 @@ static void dxgi_vk_swap_chain_signal_waitable_handle(struct dxgi_vk_swap_chain
...
+        pthread_mutex_lock(&chain->timing.lock);
+        dxgi_vk_swap_chain_poll_time_domains(chain);
+        dxgi_vk_swap_chain_poll_calibration(chain);
+        pthread_mutex_unlock(&chain->timing.lock);
+    }
+    else if (chain->present.present_count % 64 == 0)
+    {
+        /* Regularly recalibrate. */
+        pthread_mutex_lock(&chain->timing.lock);
+        dxgi_vk_swap_chain_poll_calibration(chain);
```
doitsujin, line 3236: Do we really need to do this that often?

Maybe I'm scared for no reason but I was under the impression that calibration was supposed to be slow and more of a one-time thing (bar clocks wrapping around, and in a present_timing context obviously whenever the timing domains update), but if this works fine / is necessary to avoid deviation then fair enough.

doitsujin (commented):
pain, couple questions but this looks sensible.

libs/vkd3d/swapchain.c
HansKristian-Work: Clock drift is a thing. RDNA3 APUs are kinda extreme right now w.r.t. that. Polling once a second is reasonable I think. Polling every frame would be more on the questionable side though.

libs/vkd3d/swapchain.c
HansKristian-Work: interval of UINT64_MAX is defined to be VRR yes.

libs/vkd3d/device.c
HansKristian-Work: Will add a commit on top to filter it.

libs/vkd3d/swapchain.c
HansKristian-Work: It would mean we cannot use present timing indeed. It really shouldn't happen though unless driver is OOM or something. We'll end up seeing PRESENT_QUEUE_FULL errors in QueuePresent instead, so it should be robust, and that might be better than "silently" failing to use present timing.

libs/vkd3d/swapchain.c
doitsujin: urgh, okay.

doitsujin (approved):

PR #2758, opened 2025-12-16
DSV layout fixes

### LINKED ISSUE #2749: Revisit DSV layout tracking shenanigans
The clear fuse work exposed a lot of weakness in the tracking, need to revisit that whole mess.

- Clear fuse does not properly promote optimal layouts.
- DSV images with layers > 1 or mips > 1 seem to have some weakness where we don't get the correct optimal layout in all cases. Needs more test cases ...

Ideally we should be able to get rid of all this mess.

HansKristian-Work:
Surely we'll get it right this time :frog:

Fix #2749.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -4007,7 +4007,7 @@ static void d3d12_command_list_load_attachment(struct d3d12_command_list *list, 
 {
     const struct vkd3d_vk_device_procs *vk_procs = &list->device->vk_procs;
     VkRenderingAttachmentInfo attachment_info, stencil_attachment_info;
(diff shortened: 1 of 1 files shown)


libs/vkd3d/command.c
```diff
@@ -6832,7 +6850,7 @@ static bool d3d12_command_list_update_rendering_info(struct d3d12_command_list *
     {
         assert(list->dsv.view);
 
-        dsv_optimal_aspects = 0u;
+        dsv_aspects = 0u;
```
doitsujin: this is equivalent to `list->dsv.view->format->vk_aspect_mask` now and can probably be removed.

doitsujin (approved):

PR #2761, opened 2025-12-18
Improve RenderPass API handling w.r.t. suspend-resume

HansKristian-Work:


libs/vkd3d/command.c
```diff
@@ -3436,9 +3436,16 @@ static void d3d12_command_list_sync_tiler_renderpass_writes(struct d3d12_command
...
+                rendering_info->pDepthAttachment &&
+                rendering_info->pDepthAttachment->resolveMode != VK_RESOLVE_MODE_NONE;
+        bool stencil_resolve =
+                rendering_info->pStencilAttachment &&
+                rendering_info->pStencilAttachment->resolveMode != VK_RESOLVE_MODE_NONE;
+
         memset(&vk_barrier, 0, sizeof(vk_barrier));
         vk_barrier.sType = VK_STRUCTURE_TYPE_MEMORY_BARRIER_2;
-        if (rendering_info->colorAttachmentCount)
+        if (rendering_info->colorAttachmentCount || depth_resolve || stencil_resolve)
```
doitsujin, line 3448: huh, does depth-stencil resolve actually imply `COLOR_ATTACHMENT` stage/access?

HansKristian-Work: Indeed. It's a known weird corner of the spec.

HansKristian-Work: 8.7 in spec:

> Access to the single sample attachment can be synchronized with [VK_ACCESS_COLOR_ATTACHMENT_READ_BIT](https://registry.khronos.org/vulkan/specs/latest/html/vkspec.html#VkAccessFlagBits) and [VK_ACCESS_COLOR_ATTACHMENT_WRITE_BIT](https://registry.khronos.org/vulkan/specs/latest/html/vkspec.html#VkAccessFlagBits). These pipeline stage and access types are used whether the attachments are color or depth/stencil attachments.

libs/vkd3d/command.c
```diff
@@ -17734,16 +17777,172 @@ static void d3d12_command_list_resolve_render_pass_attachments(struct d3d12_comm
...
+        return false;
+    }
+
+    if (aspect & (VK_IMAGE_ASPECT_DEPTH_BIT | VK_IMAGE_ASPECT_STENCIL_BIT))
+    {
+        const VkPhysicalDeviceVulkan12Properties *vk12 = &list->device->device_info.vulkan_1_2_properties;
+        VkResolveModeFlagBits vk_mode = vk_resolve_mode_from_d3d12(args->ResolveMode);
+
+        /* We could do a bit better here if we correlate depth and stencil attachments,
+         * but I think all relevant implementations support this? */
```
doitsujin: Turnip doesn't seem to:
```c
   p->independentResolveNone  = false;
   p->independentResolve      = false;
```

Think we need to be at least semi-smart here and make sure that stencil either doesn't exist or uses the exact same mode as depth, I don't think we're ever going to reasonably see anything other than `SAMPLE_ZERO` resolves here anyway.

HansKristian-Work: Fun ... I'll need to think about it some more.

HansKristian-Work:
Added a comment at the end which deals with independent resolves in a smarter way.

libs/vkd3d/command.c
```diff
@@ -5587,25 +5611,30 @@ static void d3d12_command_list_copy_render_pass_suspend_resume_compat(
...
         compat->layouts[D3D12_SIMULTANEOUS_RENDER_TARGET_COUNT + 1] = list->rendering_info.info.pStencilAttachment->imageLayout;
         compat->load_ops[D3D12_SIMULTANEOUS_RENDER_TARGET_COUNT + 1] = list->rendering_info.info.pStencilAttachment->loadOp;
+        compat->resolve_views[D3D12_SIMULTANEOUS_RENDER_TARGET_COUNT + 1] =
+                list->rendering_info.info.pStencilAttachment->resolveImageView;
     }
 
     compat->views[D3D12_SIMULTANEOUS_RENDER_TARGET_COUNT + 2] = list->rendering_info.vrs.imageView;
     compat->layouts[D3D12_SIMULTANEOUS_RENDER_TARGET_COUNT + 2] = list->rendering_info.vrs.imageLayout;
-    compat->load_ops[D3D12_SIMULTANEOUS_RENDER_TARGET_COUNT + 2] = VK_ATTACHMENT_LOAD_OP_LOAD;
+    /* LOAD and STORE is fixed for VRS. */
```
doitsujin, line 5640: slightly confusing comment since VRS doesn't really have load/store ops (and if anything it would be LOAD+NONE semantics since it's read-only), but yeah, tracking isn't really useful.

PR #2773, opened 2025-12-31
vkd3d: Update shader quirk hash for Wuthering Waves 3.0

guilherme-chaves:
Once again, the shader hash changed after the latest game update.

Renderdoc capture: https://drive.google.com/file/d/1MjpUgXZ7Fxqfj9U7lMAg3cDJXq9zIuma/view

Tested with Wine 10.20 on Lutris with SteamOS=1 environment variable set.

### DIFF
--- a/libs/vkd3d/device.c
+++ b/libs/vkd3d/device.c
@@ -857,7 +857,7 @@ static const struct vkd3d_shader_quirk_info death_stranding_quirks = {
 
 static const struct vkd3d_shader_quirk_hash wuthering_waves_hashes[] = {
     /* LightGridInjectionCS. Forgets to UAV barrier after ClearCS. */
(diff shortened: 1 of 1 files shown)


guilherme-chaves:
From the Renderdoc capture, it looks like they added an UAV barrier between the ClearCS and LightGridInjectionCS, but it wasn't enough to stop the flickering.

Renderdoc Capture

https://github.com/user-attachments/assets/ba73a46c-7009-4939-99b4-1cfdc8ac979e

xqlenty1:
@guilherme-chaves Thank you! I've managed to clone and compile your fork of vkd3d, then copied cachy proton in user directory and swaped the dlls! Now its fixed, I'll try to figure out how to change the string myself before the next version so I can fix it right away, thanks one more time!

PR #2782, opened 2026-01-07
tests: add an Adreno-specific bug_if in test_deferred_clears

zdobersek:
The subtest of two aliased render targets, where one is cleared and the other is tested for the correct content, fails on Adreno because the hardware uses incompatible image compression encodings for the underlyng Vulkan image formats. The test otherwise passes if image compression is force-disabled in Turnip through debug options.


zdobersek:
Right. The existing comment already hints at this not being entirely valid, and the D3D placed resource documentation also describes this kind of usage as something unsupported.

PR #2783, opened 2026-01-08
tests: adjust timeout in test_fence_pending_signal_cpu_rewind on slow…

zdobersek:
…er GPUs

Adreno GPUs will take a bit more time on the heavy-load dispatch used in the test_fence_pending_signal_cpu_rewind test case, so an increased event timeout value should be used there.

### DIFF
--- a/tests/d3d12_sync.c
+++ b/tests/d3d12_sync.c
@@ -1819,7 +1819,7 @@ static void test_fence_pending_signal_cpu_rewind(bool use_shared)
 
             /* The heavy-GPU-load dispatch will take more time on weaker GPUs */
             if (is_adreno_device(context.device))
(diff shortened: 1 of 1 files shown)


tests/d3d12_sync.c
```diff
@@ -1815,6 +1815,11 @@ static void test_fence_pending_signal_cpu_rewind(bool use_shared)
         else
         {
             unsigned int wait_result;
+            unsigned int wait_timeout = 2000;
+
+            /* The heavy-GPU-load dispatch will take more time on weaker GPUs */
+            if (is_adreno_device(context.device))
+               wait_timeout = 8000;
```
HansKristian-Work:
```suggestion
                wait_timeout = 8000;
```

PR #2792, opened 2026-01-12
Add dynamic rendering local read extensions to allow tilers to go brrrrr

HansKristian-Work:


libs/vkd3d/resource.c
```diff
@@ -984,6 +984,12 @@ static HRESULT vkd3d_get_image_create_info(struct d3d12_device *device,
         image_info->extent = d3d12_resource_desc_get_padded_feedback_extent(desc);
     }
 
+    if (device->tiler_optimizations.enable && image_info->usage & (
+        VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT | VK_IMAGE_USAGE_DEPTH_STENCIL_ATTACHMENT_BIT))
```
doitsujin: DSV can opt-out of `SAMPLED_BIT`, do we want to respect that here?

docs/tiler_optimization_extensions.md
```diff
@@ -0,0 +1,257 @@
...
+{
+    void InputAttachmentPixelBarrier();
+    void SetRootSignatureInputAttachments(D3D12_GPU_DESCRIPTOR_HANDLE handle);
+    void SetInputAttachmentFeedback(UINT render_target_concurrent_mask, BOOL depth_concurrent, BOOL stencil_concurrent);
+}
+```
+
+### `void InputAttachmentPixelBarrier()`
+
+While an image as in `RENDER_TARGET` or `DEPTH_WRITE` resource states (or equivalent in enhanced barriers),
```
K0bin: This is supposed to say 'an image **is** in`RENDER_TARGET`', right?

Also: maybe also mention D3D12 enhanced barrier texture layouts.

libs/vkd3d/resource.c
HansKristian-Work: Yes.

docs/tiler_optimization_extensions.md
HansKristian-Work: Right.

K0bin: You forgot to fix the typo.

HansKristian-Work: Yes, because I haven't pushed anything yet, just acking the issue.

PR #2805, opened 2026-01-26
Prototype implementation of VK_EXT_descriptor_heap

HansKristian-Work:
Bringup implementation during spec development. This will have to be rewritten to work alongside at least descriptor buffer paths which will be a glorious mess ...

This branch can be used to compare existing DB with new heap paths for performance, correctness, etc. Once sufficient gains are realized, work to productize this stuff could happen.


Bossiskeyboard:
here is the proton log of black myth wukong
[steam-3132990.log](https://github.com/user-attachments/files/24908019/steam-3132990.log)


[bmw_shader_trace.log](https://github.com/user-attachments/files/24908020/bmw_shader_trace.log)

doitsujin:
Well yeah you're using a Proton build without the necessary winevulkan patches (there currently is none, needs upstream wine 11.1).

Like I said, setup issue on your end, please keep those out of PR discussions.

morgwai:
@HansKristian-Work thanks for the clarification! In such case, would it be possible to temporarily limit commenting ability only to project members and collaborators?

@HansKristian-Work, @doitsujin I would also like to say BIG thank you for your amazing work, while I'm here BTW :)

HansKristian-Work:
This is superseded by https://github.com/HansKristian-Work/vkd3d-proton/pull/2943. Branch will remain valid for comparison until it's merged properly.

PR #2817, opened 2026-02-11
Return `S_FALSE` on NULL output COMPtr in `D3D12GetInterface()`

MarijnS95:
The output pointer in `D3D12GetInterface()` is allowed to be `NULL` according to `_COM_Outptr_opt_`, which is useful in cases where the caller wants to check if a certain CLSID/IID is supported by the implementation (which returns a positive success code `S_FALSE` or `E_NOINTERFACE` when it is not available), without actually returning an owned interface object.

https://learn.microsoft.com/en-us/windows/win32/api/d3d12/nf-d3d12-d3d12getinterface

### DIFF
--- a/libs/d3d12core/main.c
+++ b/libs/d3d12core/main.c
@@ -1001,7 +1001,8 @@ static HRESULT d3d12core_D3D12GetInterface(REFCLSID rcslid, REFIID iid, void **d
     {
         if (IsEqualGUID(iid, &IID_IVKD3DCoreInterface))
         {
-            if (debug) {
+            if (debug)
+            {
                 *debug = (void*)&d3d12core_interface_instance;
                 return S_OK;
             }
@@ -1013,7 +1014,8 @@ static HRESULT d3d12core_D3D12GetInterface(REFCLSID rcslid, REFIID iid, void **d
     {
         if (IsEqualGUID(iid, &IID_IVKD3DDebugControlInterface))
         {
-            if (debug) {
+            if (debug)
+            {
                 *debug = (void*)&vkd3d_debug_control_instance;
                 return S_OK;
             }


libs/d3d12core/main.c
```diff
@@ -1001,7 +1001,7 @@ static HRESULT d3d12core_D3D12GetInterface(REFCLSID rcslid, REFIID iid, void **d
     {
         if (IsEqualGUID(iid, &IID_IVKD3DCoreInterface))
         {
-            *debug = (void*)&d3d12core_interface_instance;
+            if (debug) *debug = (void*)&d3d12core_interface_instance;
```
HansKristian-Work: Don't think this is actually correct. Normally when there's NULL output, you get S_FALSE return. Is D3D12GetInterface special?

build-win32.txt
```diff
@@ -5,6 +5,12 @@ ar = 'i686-w64-mingw32-ar'
 strip = 'i686-w64-mingw32-strip'
 widl-mingw-tools-fallback = 'i686-w64-mingw32-widl'
 
+[built-in options]
```
HansKristian-Work: Where is this coming from and why is it part of this PR?

MarijnS95: I'm sorry, this is automatically done by the arch package and I didn't notice it in the commit; will remove asap.

libs/d3d12core/main.c
MarijnS95: You may be right, I was just investigating possible ways to detect vkd3d-proton by checking the custom CLSID and didn't yet test this on Windows either, will check it soon.

build-win32.txt
HansKristian-Work: If this is solving a real problem, please make it at least a separate commit or separate PR as we follow "one logical change per commit".

MarijnS95: I think it's just expanding the custom `C(XX)FLAGS` (supposedly taken from Proton) into the Meson project (because it doesn't work via env vars?): https://aur.archlinux.org/cgit/aur.git/tree/vkd3d-proton-extraopts.patch?h=vkd3d-proton-mingw-git

Not something that should be in this PR, not even a separate commit.

libs/d3d12core/main.c
MarijnS95: You are entirely right. I only thought about the nullability portion of this function (where it would currently segfault) but not that the error code would be different (though both `S_OK` and `S_FALSE` are `SUCCEEDED()` because of `>= 0`).

Fixed that now.

libs/d3d12core/main.c
```diff
@@ -1001,17 +1001,23 @@ static HRESULT d3d12core_D3D12GetInterface(REFCLSID rcslid, REFIID iid, void **d
     {
         if (IsEqualGUID(iid, &IID_IVKD3DCoreInterface))
         {
-            *debug = (void*)&d3d12core_interface_instance;
-            return S_OK;
+            if (debug) {
```
HansKristian-Work:
```suggestion
            if (debug)
            {
```

libs/d3d12core/main.c
```diff
@@ -1001,17 +1001,23 @@ static HRESULT d3d12core_D3D12GetInterface(REFCLSID rcslid, REFIID iid, void **d
...
         }
     }
 
     if (IsEqualGUID(rcslid, &CLSID_VKD3DDebugControl))
     {
         if (IsEqualGUID(iid, &IID_IVKD3DDebugControlInterface))
         {
-            *debug = (void*)&vkd3d_debug_control_instance;
-            return S_OK;
+            if (debug) {
```
HansKristian-Work:
```suggestion
            if (debug)
            {
```

libs/d3d12core/main.c
```diff
@@ -1001,17 +1001,23 @@ static HRESULT d3d12core_D3D12GetInterface(REFCLSID rcslid, REFIID iid, void **d
...
         if (IsEqualGUID(iid, &IID_IVKD3DCoreInterface))
         {
-            *debug = (void*)&d3d12core_interface_instance;
-            return S_OK;
+            if (debug) {
+                *debug = (void*)&d3d12core_interface_instance;
+                return S_OK;
+            }
+            return S_FALSE;
         }
```
MarijnS95, line 1009: There appears to be a helper named `return_interface()` already that takes care of this, but it also calls `IUnknown_Release()` which should be a no-op here on these static objects.

However that has the caveat that it checks for `NULL` and returns `S_FALSE` _before_ checking the IID; should that behaviour apply here too? (Didn't test on Windows yet). But it'll also call `QueryInterface` if the IID didn't match, which again should be harmless but might be undesired?

```suggestion
        return return_interface((void*)&d3d12core_interface_instance, &IID_IVKD3DCoreInterface, iid, debug);
```

---

Separately (while searching for existing `S_FALSE` patterns) I found a few cases that use `return_interface()` but also manually call `_Release()`:

https://github.com/HansKristian-Work/vkd3d-proton/blob/56c3c98b4f0eacda5e7c18685521ac26d3d2cde5/libs/vkd3d/device.c#L7292-L7301

That seems to be introduced in #843 _before_ `return_interface()` handled `S_FALSE`; can I clean those up?

HansKristian-Work: > should that behaviour apply here too?

uh ... I don't really care, since these are referring to private APIs.

> can I clean those up?

Sure.

MarijnS95: Also seems `return_interface()` is defined in `d3d12`, is `d3d12core` supposed to have access to that?

HansKristian-Work: Huh? it's in libs/vkd3d/utils.c, which is part of the internal helper library.

MarijnS95: (This was all in reply to "should that behaviour apply here too")

Yeah, and the `D3D12GetInterface()` function from this PR is in `d3d12core`, while `return_interface()` is defined in `d3d12` (declared in `vkd3d_private.h`) so the util would have to move or be duplicated. Seems like a bit too much hassle.

PR #2832, opened 2026-02-18
Implement independent devices

### LINKED ISSUE #2827: Implement independent device features
Some ISV requested this feature a while back and I forgot about it ... until now that is.

Should not be too hard since we have to go out of our way to make singleton device work and implementing this should mean just ... not doing that.

HansKristian-Work:
Fix #2827.

https://microsoft.github.io/DirectX-Specs/d3d/IndependentDevices.html

### DIFF
--- a/libs/d3d12core/main.c
+++ b/libs/d3d12core/main.c
@@ -1065,6 +1065,7 @@ static HRESULT STDMETHODCALLTYPE d3d12_device_factory_QueryInterface(ID3D12Devic
     if (IsEqualGUID(iid, &IID_ID3D12DeviceConfiguration) ||
         IsEqualGUID(iid, &IID_ID3D12DeviceConfiguration1))
     {
+        /* DeviceConfiguration shares the refcount with the base interface. */
         struct d3d12_device_factory *factory = impl_from_ID3D12DeviceFactory(iface);
         d3d12_device_factory_AddRef(iface);
         *object = &factory->ID3D12DeviceConfiguration1_iface;
--- a/libs/vkd3d/device.c
+++ b/libs/vkd3d/device.c
@@ -10494,6 +10494,8 @@ HRESULT d3d12_device_create(struct vkd3d_instance *instance,
         forced_singletons = vkd3d_get_env_var("ENABLE_VULKAN_RENDERDOC_CAPTURE", env, sizeof(env)) &&
                 strcmp(env, "1") == 0;
 
+        INFO("Forcing singleton device due to RenderDoc being enabled.\n");
+
         if (forced_singletons &&
             (create_info->device_factory_flags &
                 (D3D12_DEVICE_FACTORY_FLAG_ALLOW_RETURNING_EXISTING_DEVICE |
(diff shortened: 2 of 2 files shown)


libs/d3d12/main.c
```diff
@@ -212,10 +212,127 @@ HRESULT WINAPI DLLEXPORT D3D12EnableExperimentalFeatures(UINT feature_count,
...
+    FIXME("iface %p, SDKVersion %u, SDKPath %s stub!\n", iface, SDKVersion, SDKPath);
+    /* Native seems to just return S_OK here even for garbage inputs, *shrug*. */
+    return S_OK;
+}
+
+static HRESULT STDMETHODCALLTYPE d3d12_sdk_configuration_CreateDeviceFactory(ID3D12SDKConfiguration1 *iface,
+    UINT SDKVersion, LPCSTR SDKPath, REFIID iid, void **ppvFactory)
+{
+    TRACE("iface %p, SDKVersion %u, SDKPath %s, iid %s, ppvFactory %p!\n", iface, SDKVersion, SDKPath,
+            debugstr_guid(iid), ppvFactory);
```
MarijnS95, line 277: Should `SDKVersion` maybe be checked against `<= D3D12_SDK_VERSION` so that we can't just query anything, or is that too fragile?

HansKristian-Work: Since we're not actually loading d3d12core.dll runtime from agilitysdk, it's too brittle to fail this call due to that.

MarijnS95: Was there any significance then in having and exporting `D3D12SDKVersion`, some games check it after all?

HansKristian-Work: Yes, some games are known to fail if we report unexpectedly low SDK version (Windows can promote to its own d3d12core.dll if newer, but should never degrade it), so we always expose the latest stable version number to mitigate.

libs/vkd3d/device.c
```diff
@@ -10374,17 +10385,53 @@ bool d3d12_device_validate_shader_meta(struct d3d12_device *device, const struct
...
+    if (!create_info->independent || forced_singletons)
     {
-        TRACE("Returned existing singleton device %p.\n", object);
+        if ((object = d3d12_find_device_singleton(create_info->adapter_luid)))
+        {
+            TRACE("Returned existing singleton device %p.\n", object);
+
+            if (reject_existing_device)
+            {
+                pthread_mutex_unlock(&d3d12_device_map_mutex);
```
doitsujin, line 10518: could probably do with an `INFO` here (and the other error-out path below) to make life easier when we do inevitably hit this path for whatever reason.

libs/d3d12core/main.c
```diff
@@ -993,10 +1006,274 @@ static const d3d12core_interface d3d12core_interface_instance =
...
+        d3d12_device_factory_AddRef(iface);
+        *object = iface;
+        return S_OK;
+    }
+
+    if (IsEqualGUID(iid, &IID_ID3D12DeviceConfiguration) ||
+        IsEqualGUID(iid, &IID_ID3D12DeviceConfiguration1))
+    {
+        struct d3d12_device_factory *factory = impl_from_ID3D12DeviceFactory(iface);
+        d3d12_device_factory_AddRef(iface);
```
doitsujin, line 1070: should this AddRef the configuration iface instead?
edit: nvm they seem to share a ref count, but it's not immediately obvious.

HansKristian-Work: Fixed.

libs/vkd3d/device.c
HansKristian-Work: Fixed.

PR #2846, opened 2026-02-25
Insert proper clamping for out of range ClearRTV/UAV values

HansKristian-Work:


libs/vkd3d/command.c
```diff
@@ -13800,12 +13800,93 @@ static void STDMETHODCALLTYPE d3d12_command_list_ClearDepthStencilView(d3d12_com
...
+        case VK_FORMAT_B10G11R11_UFLOAT_PACK32:
+            for (i = 0; i < 2; i++)
+                color[i] = clamp_float_to_narrow_float(in_color[i], 65024.0f, true, false);
+            color[2] = clamp_float_to_narrow_float(in_color[2], 64512.0f, true, false);
+            color[3] = 0.0f;
+            break;
+
+        case VK_FORMAT_E5B9G9R9_UFLOAT_PACK32:
+            for (i = 0; i < 3; i++)
+                color[i] = clamp_float_to_narrow_float(in_color[i], 511.0f * 64.0f * 1024.0f, true, true);
```
doitsujin: where's the 1024 multiplier coming from?

HansKristian-Work: I wrote out 2^16 as 2^6 * 2^10 since I can never remember the full constant :v

doitsujin: max float value for this should be `511*2^(16-9)` = `511*128` though, hence the question.

HansKristian-Work: ... So much for trusting the internet :frog:

PR #2857, opened 2026-03-03
vkd3d: Workaround missing compute barrier in HFW.

HansKristian-Work:


llandwerlin-intel:
2 different problems here, the atomics on R11G11B10 surfaces is not this.

As far as I can tell there is only a single shader with an untyped clear/initialization not synchronized properly in this title.

PR #2870, opened 2026-03-11
Misc debug improvements for enhanced barriers

HansKristian-Work:


tests/d3d12_enhanced_barriers.c
```diff
@@ -1594,7 +1594,8 @@ void test_enhanced_barrier_subresource(void)
 
     /* Try different edge cases of the subresources struct. */
 
-    /* NumPlanes = 0 -> nothing happens */
+    /* NumPlanes = 0 -> nothing happens. However, there is content in the wild that uses it for some bizarre reason,
```
doitsujin, line 1597: What's the observed usage exactly? I wonder if we can test native behaviour with a transition from `UNDEFINED` + `D3D12_TEXTURE_BARRIER_FLAG_DISCARD` in some way, i.e. whether it actually affects any subresources or if it just behaves like a global barrier or something.

HansKristian-Work: The observed usage is that everything else looks as expected in the barrier, but NumPlanes = 0, so it's skipped. This check was originally added because it looked like d3d12 validation layers skipped the call if NumPlanes was 0, but that's somewhat weak proof, since it doesn't skip NumLayers = 0.

PR #2873, opened 2026-03-12
Transfer queue - phase 1

HansKristian-Work:
Adds tests for all edge cases we want to ponder. Also adds COMPUTE queue tests for completeness' sake.


tests/d3d12_copy.c
```diff
@@ -3816,53 +3842,83 @@ void test_copy_queue_render_target(void)
...
+        src.PlacedFootprint.Footprint.Height = 64;
+        src.PlacedFootprint.Footprint.Depth = 1;
+        src.PlacedFootprint.Footprint.RowPitch = 256;
+        src.PlacedFootprint.Footprint.Format = DXGI_FORMAT_R8G8B8A8_UNORM;
+    }
 
     ID3D12GraphicsCommandList_Reset(context.copy_list, context.copy_allocator, NULL);
-    get_texture_readback_with_command_list(tex, 0, &rb, context.copy_queue, context.copy_list);
+    ID3D12GraphicsCommandList_CopyResource(context.copy_list, copy_tex, tex);
+    /* COPY queue does not seem to like COPY_SOURCE for some reason. */
```
doitsujin, line 3863: Does this imply issues on native?

If the [doc](https://learn.microsoft.com/en-us/windows/win32/api/d3d12/ne-d3d12-d3d12_resource_states) for `COMMON` is anything to go by, copy queue seems to mandate `COMMON` for *all* access types somehow.

tests/d3d12_test_utils.h
```diff
@@ -666,8 +684,19 @@ static inline void get_texture_readback_with_command_list(ID3D12Resource *textur
                 &IID_ID3D12Resource, (void **)&src_resource);
         assert_that(hr == S_OK, "Failed to create texture, hr %#x.\n", hr);
 
-        ID3D12GraphicsCommandList_ResolveSubresource(command_list,
-                src_resource, 0, texture, sub_resource, resource_desc.Format);
+        if (format_is_depth_stencil(resource_desc.Format) &&
+            SUCCEEDED(ID3D12GraphicsCommandList_QueryInterface(command_list, &IID_ID3D12GraphicsCommandList1, (void **)&list1)))
+        {
+            ID3D12GraphicsCommandList1_ResolveSubresourceRegion(list1,
+                    src_resource, 0, 0, 0, texture, sub_resource, NULL, resource_desc.Format, D3D12_RESOLVE_MODE_MIN);
```
doitsujin, line 691: does this even work properly on native AMD? IIRC resolve modes other than `SAMPLE_ZERO` were kind of broken there.

tests/d3d12_copy.c
HansKristian-Work: It seems like native expects COMMON state, yes.

tests/d3d12_test_utils.h
HansKristian-Work: That could explain why depth msaa tests are failing, but I don't think that invalidates the test.

PR #2874, opened 2026-03-12
vkd3d: Support VK_PRESENT_MODE_FIFO_LATEST_READY_KHR

silvanshade:
This PR introduces two changes:

The first change allows selecting `VK_PRESENT_MODE_FIFO_LATEST_READY_KHR` and prefers it before `VK_PRESENT_MODE_MAILBOX_KHR` when finding the unlocked present mode.

The second change adds a new option `VKD3D_CONFIG=tear_free`. If specified, this option changes the behavior when finding the unlocked mode to prefer tear-free modes, so FIFO Latest Ready or Mailbox will be chosen before Immediate.

The new option is intended to be similar in behavior to [`dxvk.tearFree`](https://github.com/doitsujin/dxvk/blob/3730b707223a273fb1774f2f2ac3572d68c4dbe9/src/dxvk/dxvk_options.h#L44-L46). This also makes it possible to override an application to use an unlocked tear-free mode:
```
DXVK_CONFIG="dxgi.syncInterval = 0" VKD3D_CONFIG=tear_free
```

Tested working on an NVIDIA GeForce RTX 5090 with driver 595.45.04.

Related: https://github.com/doitsujin/dxvk/pull/5550

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,6 +156,7 @@ commas or semicolons.
       so it should not be a real issue even on lower VRAM cards.
     - `force_host_cached` - Forces all host visible allocations to be CACHED, which greatly accelerates captures.
     - `no_invariant_position` - Avoids workarounds for invariant position. The workaround is enabled by default.
+    - `tear_free` - Prefer tear-free present modes when Vsync is disabled. Selects FIFO Latest Ready or Mailbox before Immediate.
  - `VKD3D_DEBUG` - controls the debug level for log messages produced by
    vkd3d-proton. Accepts the following values: none, err, info, fixme, warn, trace.
  - `VKD3D_SHADER_DEBUG` - controls the debug level for log messages produced by
--- a/include/vkd3d.h
+++ b/include/vkd3d.h
@@ -120,6 +120,7 @@ extern "C" {
 #define VKD3D_CONFIG_FLAG_DEFER_RESOURCE_DESTRUCTION (1ull << 60)
 #define VKD3D_CONFIG_FLAG_PREFER_THIN_UAV_TILING (1ull << 61)
 #define VKD3D_CONFIG_FLAG_EXTENDED_DEBUG_UTILS (1ull << 62)
+#define VKD3D_CONFIG_FLAG_TEAR_FREE (1ull << 63)
 
 struct vkd3d_instance;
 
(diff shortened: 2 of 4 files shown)


HansKristian-Work:
What is the motivation for this? FIFO latest ready is meaningless without present timing, and even then I don't understand what problem it's solving.

silvanshade:
> > The motivation for adding VKD3D_CONFIG=tear_free is because there currently is no way (that I am aware of) to override the application to use a tear-free mode.
> 
> That is more reasonable. If you can isolate that to a separate PR, and make it more like VKD3D_SWAPCHAIN_PRESENT_MODE=mailbox, or VKD3D_SWAPCHAIN_CONFIG or something, I think that'd be better than adding another vkd3d_config flag which is very crowded.

I've created a new PR which implements this.

> > In the absence of present timing, FIFO latest ready behaves similarly to Mailbox. Preferring it over Mailbox where supported seems like a reasonable strategy.
> 
> D3D12 applications have no way to express any of this, so I don't really see the point right now.

I don't think the motivation for forward looking compatibility with `present_timing` needs to be limited to whether that would function at the D3D12 application layer.

From what I understand of the existing code base, there are already areas where `present_timing` integration could be beneficial.

For example, take [`dxgi_vk_swap_chain_delay_next_frame`](https://github.com/HansKristian-Work/vkd3d-proton/blob/a0cc484f561950cd2bdb5ea7979f953843b02f71/libs/vkd3d/swapchain.c#L2723). This appears to be maintaining a heuristic frame window to estimate display cadence, compute deadlines based on estimated intervals, and busy-waiting with a sleep fallback to meet deadlines.

Other parts of the code involve latency management and use of `present_wait`.

Support for `present_timing` seems like it could simplify some of that and also improve accuracy.

> From what I can tell latest_ready is not quite equal to MAILBOX when present timing is not used, since based on my reading of the spec it's not clear that presentation requests can retire before vblank when FPS is very high, so using latest_ready might cap FPS at some multiple of swapchain size, which is not what we want.
> 
> Maybe there is some use for this later if we want to add higher-than-refresh-rate frame limiter on VRR displays or something outside the scope of D3D12, but I'm wary about adding dead code that may or may not be useful later.

The spec does state this, which would seem to support your concern:

> This extension adds a new present mode, `VK_PRESENT_MODE_FIFO_LATEST_READY_EXT`, which allows the presentation engine to dequeue multiple present requests during vblank, in FIFO order, until encountering a request that is not ready. The image of the last present request to be dequeued is presented to the display. This can increase an application’s frame rate up to `(swapchainImageCount - 1) * refreshRate` compared to `VK_PRESENT_MODE_FIFO_KHR`.

However, in my testing of the actual implementation in the NVIDIA driver, it does not seem to show such a limitation. In all cases I have tested, FPS at very high rates is essentially identical to Mailbox.

For an extreme example, I've included screenshots of `vkcube`: Mailbox reports `13897` FPS and FIFO Latest Ready reports `14237` FPS, both exceeding any plausible `swapchain_size * refresh_rate` threshold.

Perhaps it is more realistic to consider the `(swapchainImageCount - 1) * refreshRate` figure as a guarantee for how much better FIFO Latest Ready should be in the worst case rather than an upper bound on what a conformant implementation can achieve.

It is a valid concern and maybe a good reason not to prefer the mode by default, though I still think supporting it is well motivated given the potential for `present_timing` integration at the vkd3d layer.

mailbox
fifo-latest-ready

PR #2878, opened 2026-03-14
vkd3d: Implement VKD3D_SWAPCHAIN_PRESENT_MODE env setting

silvanshade:
This PR is a follow-up to https://github.com/HansKristian-Work/vkd3d-proton/pull/2874

A new environment setting `VKD3D_SWAPCHAIN_PRESENT_MODE` is introduced, which accepts a (case insensitive) shortened version of a Vulkan present mode name.

The name can optionally also contain the prefix `VK_PRESENT_MODE_` and/or the suffix `_KHR`, so it is possible to specify either `VKD3D_SWAPCHAIN_PRESENT_MODE=mailbox` or VKD3D_SWAPCHAIN_PRESENT_MODE=vk_present_mode_mailbox_khr` for example.

I've also enabled FIFO Latest Ready as a possible present mode. Using it requires explicitly specifying it through the new env var setting so now it should have minimal impact on the rest of the code base.

### DIFF
--- a/libs/vkd3d/swapchain.c
+++ b/libs/vkd3d/swapchain.c
@@ -224,6 +224,7 @@ struct dxgi_vk_swap_chain
         VkPresentModeKHR unlocked_present_mode;
         VkPresentModeKHR selected_present_mode;
         bool compatible_unlocked_present_mode;
+        bool override_present_mode;
 
         /* Info about the current low latency state of the swapchain */
         uint32_t low_latency_present_mode_count;
@@ -1847,7 +1848,6 @@ static void dxgi_vk_swap_chain_recreate_swapchain_in_present_task(struct dxgi_vk
     VkDevice vk_device = chain->queue->device->vk_device;
     VkCommandPoolCreateInfo command_pool_create_info;
     VkSwapchainCreateInfoKHR swapchain_create_info;
-    bool has_present_mode_override = false;
     VkPresentModeKHR present_mode_group[2];
     char present_mode_env[VKD3D_PATH_MAX];
     VkSurfaceCapabilitiesKHR surface_caps;
@@ -1903,6 +1903,8 @@ static void dxgi_vk_swap_chain_recreate_swapchain_in_present_task(struct dxgi_vk
     memset(&swapchain_create_info, 0, sizeof(swapchain_create_info));
     swapchain_create_info.sType = VK_STRUCTURE_TYPE_SWAPCHAIN_CREATE_INFO_KHR;
 
+    chain->present.override_present_mode = false;
+
     if (vkd3d_get_env_var("VKD3D_SWAPCHAIN_PRESENT_MODE", present_mode_env, sizeof(present_mode_env)))
     {
         VkPresentModeKHR candidate_present_mode;
@@ -1912,7 +1914,7 @@ static void dxgi_vk_swap_chain_recreate_swapchain_in_present_task(struct dxgi_vk
             {
                 chain->present.selected_present_mode = candidate_present_mode;
                 chain->present.compatible_unlocked_present_mode = false;
-                has_present_mode_override = true;
+                chain->present.override_present_mode = true;
                 INFO("Overriding swapchain present mode to %s.\n", present_mode_env);
             }
             else
@@ -1922,7 +1924,7 @@ static void dxgi_vk_swap_chain_recreate_swapchain_in_present_task(struct dxgi_vk
             WARN("Ignoring unrecognized value for VKD3D_SWAPCHAIN_PRESENT_MODE=%s.\n", present_mode_env);
     }
 
-    if (!has_present_mode_override)
+    if (!chain->present.override_present_mode)
     {
         chain->present.compatible_unlocked_present_mode =
                 dxgi_vk_swap_chain_find_compatible_unlocked_present_mode(chain,
@@ -2080,7 +2082,8 @@ static bool dxgi_vk_swap_chain_request_needs_swapchain_recreation(
             request->dxgi_format != last_request->dxgi_format ||
             request->target_min_image_count != last_request->target_min_image_count ||
             ((!!request->swap_interval) != (!!last_request->swap_interval) &&
-                    !chain->present.compatible_unlocked_present_mode);
+                    !chain->present.compatible_unlocked_present_mode &&
+                    !chain->present.override_present_mode);
 }
 
 static void dxgi_vk_swap_chain_present_signal_blit_semaphore(struct dxgi_vk_swap_chain *chain, uint64_t present_count)


libs/vkd3d/vkd3d_private.h
```diff
@@ -6696,4 +6698,52 @@ static inline void vkd3d_mapped_memory_range_align(const struct d3d12_device *de
     range->size = min(range->size, size - range->offset);
 }
 
+static inline bool vkd3d_parse_swapchain_present_mode(const char *string, VkPresentModeKHR *present_mode) {
```
HansKristian-Work: This really does not belong in vkd3d_private.h and it's very overengineered for what it needs to do. It really does not need to do anything more than simple strcmp. There is also no need to support any random combination of "vk_khr" blah strings. The code here is surprisingly hard to follow as well for something that should be very simple.

libs/vkd3d/vkd3d_private.h
```diff
@@ -6696,4 +6698,52 @@ static inline void vkd3d_mapped_memory_range_align(const struct d3d12_device *de
...
+        const char *name;
+        VkPresentModeKHR value;
+    };
+
+    static const struct present_mode_entry present_mode_table[] = {
+        { "IMMEDIATE", VK_PRESENT_MODE_IMMEDIATE_KHR },
+        { "MAILBOX", VK_PRESENT_MODE_MAILBOX_KHR },
+        { "FIFO", VK_PRESENT_MODE_FIFO_KHR },
+        { "FIFO_RELAXED", VK_PRESENT_MODE_FIFO_RELAXED_KHR },
+        { "SHARED_DEMAND_REFRESH", VK_PRESENT_MODE_SHARED_DEMAND_REFRESH_KHR },
```
HansKristian-Work: SHARED modes are separate extensions and are completely irrelevant to D3D use cases.

libs/vkd3d/swapchain.c
```diff
@@ -1812,6 +1812,8 @@ static void dxgi_vk_swap_chain_recreate_swapchain_in_present_task(struct dxgi_vk
     VkSurfaceFormatKHR surface_format;
     VkImageViewCreateInfo view_info;
     VkPresentModeKHR present_mode;
+    bool has_present_mode_override = false;
```
HansKristian-Work: Variable declaration is in inverse christmas-tree order w.r.t. length.

libs/vkd3d/swapchain.c
```diff
@@ -1859,43 +1861,64 @@ static void dxgi_vk_swap_chain_recreate_swapchain_in_present_task(struct dxgi_vk
...
+            if (present_mode == VK_PRESENT_MODE_IMMEDIATE_KHR &&
+                    !dxgi_vk_swap_chain_check_present_mode_support(chain, present_mode))
+            {
+                if (dxgi_vk_swap_chain_check_present_mode_support(chain, VK_PRESENT_MODE_MAILBOX_KHR))
+                    present_mode = VK_PRESENT_MODE_MAILBOX_KHR;
+                else
+                    present_mode = VK_PRESENT_MODE_FIFO_KHR;
+            }
+
+            chain->present.present_mode_forces_fifo = present_mode == VK_PRESENT_MODE_FIFO_KHR;
```
HansKristian-Work: When overriding present modes, the present-wait path selection is dropped. E.g. if IMMEDIATE or MAILBOX is forced, I think vkd3d-proton will keep using present wait to unblock progress, which is very much not what is supposed to happen.

include/private/vkd3d_common.h
```diff
@@ -250,6 +250,17 @@ static inline int ascii_strcasecmp(const char *a, const char *b)
     return c_a - c_b;
 }
 
+static inline ssize_t ascii_hasprefix_strcasecmp(const char *string, const char *prefix)
```
HansKristian-Work: unnecessary

libs/vkd3d/swapchain.c
HansKristian-Work: You kinda need the opposite of this case. If SwapInterval >= 1 translates to unlocked mode, present wait must be disengaged as-if SwapInterval was 0.

silvanshade: Thanks for clarifying.

I think I've addressed this now by introducing the `present_mode_pacing_should_wait` predicate and storing the actual present mode in `selected_present_mode` (which is updated accordingly during the swapchain maintenance section).

This should cause the present-wait path to be taken only for `FIFO` and `FIFO_RELAXED` when SwapInterval >= 1 (or if `low_latency_state.mode` is set).

libs/vkd3d/swapchain.c
```diff
@@ -24,6 +24,43 @@
 #include "vkd3d_private.h"
 #include "vkd3d_timestamp_profiler.h"
 
+static inline bool vkd3d_swapchain_present_mode_parse(const char *string, VkPresentModeKHR *present_mode)
+{
+    struct present_mode_entry {
```
HansKristian-Work: Braces go on next line.

libs/vkd3d/swapchain.c
```diff
@@ -2483,16 +2539,14 @@ static void dxgi_vk_swap_chain_present_iteration(struct dxgi_vk_swap_chain *chai
     present_info.pWaitSemaphores = &chain->present.vk_release_semaphores[swapchain_index];
     present_info.pResults = &vk_result;
 
-    /* Even if application requests IMMEDIATE mode, the WSI implementation may not support it.
-     * In this case, we should still opt for using frame latency object to avoid catastrophic latency. */
-    swapchain_is_fifo = chain->request.swap_interval > 0 || chain->present.present_mode_forces_fifo;
+    pacing_should_wait = chain->request.swap_interval > 0 && present_mode_pacing_should_wait(chain->present.selected_present_mode);
```
HansKristian-Work: This is wrong. It should only depend on selected present mode. FIFO might be forced on SwapInterval == 0 which is still a concern.

Also, the selected present mode is not updated according to swap_interval properly. It gets updated below it seems, but that's too late.

HansKristian-Work (changes requested):
Bugs

libs/vkd3d/swapchain.c
```diff
@@ -24,6 +24,44 @@
...
 
+static inline bool vkd3d_swapchain_present_mode_parse(const char *string, VkPresentModeKHR *present_mode)
+{
+    struct present_mode_entry
+    {
+        const char *name;
+        VkPresentModeKHR value;
+    };
+
+    static const struct present_mode_entry present_mode_table[] = {
```
HansKristian-Work: I didn't just mean that line when I pointed out braces.

libs/vkd3d/swapchain.c
silvanshade: > This is wrong. It should only depend on selected present mode. FIFO might be forced on SwapInterval == 0 which is still a concern.

I did consider this case but I had been thinking about it from the perspective of someone explicitly setting both FIFO on and setting `swap_interval = 0` (through DXGI override). Since that would seem unusual, maybe they wanted to disable the usual pacing path for some sort of debugging purpose.

I have changed it to depend only on `selected_present_mode`.

> Also, the selected present mode is not updated according to swap_interval properly. It gets updated below it seems, but that's too late.

I did notice this also. But the original code seems to have this sort of delayed behavior too:

```C
/* Even if application requests IMMEDIATE mode, the WSI implementation may not support it.
 * In this case, we should still opt for using frame latency object to avoid catastrophic latency. */
swapchain_is_fifo = chain->request.swap_interval > 0 || chain->present.present_mode_forces_fifo;
```

If a compatible unlocked mode has been found (which would cause `chain->present.present_mode_forces_fifo = false`), doesn't this make the choice to ignore the present-wait path (modulo the low-latency stuff) even though the present mode (which starts in FIFO) has not yet been changed to the unlocked one?

I wasn't sure if that was intentional or a bug so I opted to not rearrange the assignments.

I will changed it to update the `selected_present_mode` before the present-wait conditional.

libs/vkd3d/swapchain.c
```diff
@@ -24,6 +24,44 @@
...
+        {"IMMEDIATE", VK_PRESENT_MODE_IMMEDIATE_KHR},
+        {"MAILBOX", VK_PRESENT_MODE_MAILBOX_KHR},
+        {"FIFO", VK_PRESENT_MODE_FIFO_KHR},
+        {"FIFO_RELAXED", VK_PRESENT_MODE_FIFO_RELAXED_KHR},
+        {"FIFO_LATEST_READY", VK_PRESENT_MODE_FIFO_LATEST_READY_KHR},
+    };
+    static const size_t present_mode_table_len = ARRAY_SIZE(present_mode_table);
+
+    for (size_t i = 0; i < present_mode_table_len; i++) {
```
HansKristian-Work: Here

libs/vkd3d/swapchain.c
```diff
@@ -24,6 +24,44 @@
...
+        if (strcmp(string, present_mode_table[i].name) == 0) {
```
HansKristian-Work: And here.

libs/vkd3d/swapchain.c
silvanshade: I've changed it to how you suggested.

There are several places in the code base that do not use this styling:

https://github.com/search?q=repo%3AHansKristian-Work%2Fvkd3d-proton+%2F%5C%5B%5C%5D%5B+%5Ct%5D%2B%3D%5B+%5Ct%5D%2B%5C%7B%5Cn%2F&type=code

The others were an oversight that I think I'd already fixed by the time you responded.

silvanshade:
> With style nits fixed it seems good to go. I'll need to do some local testing to be 100% sure it doesn't break anything first though.

Please also see https://github.com/HansKristian-Work/vkd3d-proton/pull/2894.

PR #2904, opened 2026-03-23
Misc fixes to make test suite pass on Turnip

HansKristian-Work:
Fixes some bad assumptions by test and a few real vkd3d-proton bugs were uncovered.


doitsujin (approved):
The render pass stuff is a headache and a half but it *looks* plausible

PR #2915, opened 2026-03-31
vkd3d: Disable load_store_op compatibility for Turnip.

HansKristian-Work:
It doesn't work like we want right now and we need a vendor extension to sort this out to do it properly.
Keep the hackery for IMR for now to keep things testable until we have a more robust solution.


ghost:
@HansKristian-Work Which vendor extension are you waiting for? Asking as a Turnip/Adreno user trying to run DX12 games on Android via Winlator.

PR #2930, opened 2026-04-09
vkd3d: Fix DLSS Frame Generation causing Reflex desync and game freezes

rackaracka123:
When presents stall (e.g. Wayland compositor workspace switch) or when DLSS Frame Generation presents interpolated frames without new latency markers, the monotonicity guard in the present path bumps the present ID to a non-stride-aligned value (e.g. N*10000+1 instead of N*10000).

Previously, dxgi_vk_swap_chain_get_latency_info() would zero ALL latency reports if ANY report had a non-aligned presentID. This caused Streamline's DLSS-G module to conclude that Reflex was not active (eDLSSGStatusFailReflexNotDetectedAtRuntime), disabling Frame Generation and freezing the game.

Fix by skipping non-aligned entries instead of discarding all reports. Valid application-submitted frames are still reported correctly, keeping Streamline's Reflex detection working alongside DLSS Frame Generation.

Fixes freezes in Crimson Desert (and likely other Streamline/DLSS-G titles) when DLSS Frame Generation is enabled under Proton, particularly when switching workspaces on Wayland compositors like Hyprland.

### DIFF
--- a/libs/vkd3d/swapchain.c
+++ b/libs/vkd3d/swapchain.c
@@ -3299,14 +3299,16 @@ void dxgi_vk_swap_chain_get_latency_info(struct dxgi_vk_swap_chain *chain, D3D12
                 D3D12_FRAME_REPORT *report;
 
                 /* If the frame ID isn't a natural aligned value,
-                 * we assume it's a fake frame that the application never submitted a marker for.
-                 * Ignore it. */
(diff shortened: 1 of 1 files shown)


HansKristian-Work (approved):
LGTM

PR #2938, opened 2026-04-15
Adjust VK_NV_low_latency2 usage so downstream components can map vkQueuePresent calls to the corresponding low latency frame id

wcampbell-nv:
**Motivation:** In order to measure latency statistics in applications integrating NVIDIA Reflex we need to map vkQueuePresent calls to the corresponding low latency frame id. We can then measure latency from the SIMULATION_START marker to the time at which the vkQueuePresent lands (queried from VK_EXT_present_timing). This can be done in native applications by looking for PRESENT_START and _END markers surrounding the present call, but vkd3d-proton interferes with this by deferring present calls to a background thread.

This change has vkd3d-proton submit OUT_OF_BAND_PRESENT markers on the background thread surrounding the vkQueuePresent call with the appropriate low latency frame id for the frame. We get the frame id by looking at the latency markers on the app thread during the DXGI present call, and pass it down to the background presentation callback.

This change also simplifies the VkPresentIdKHR selection logic, no longer attempting to align the present id to the low latency frame id. The VkPresentIdKHR frame attribution scheme does not work with DLSS Frame Generation, which submits multiple OUT_OF_BAND presents with the same low latency frame id. This contradicts the VK_NV_low_latency2 spec but works fine in practice as the NVIDIA driver has always treated these as separate IDs. We plan to make this explicit in a future revision of the VK_NV_low_latency2 spec.

This change should have no functional impact with today's drivers.

**Testing:** Tested with a Vulkan layer that prints latency statistics by attributing vkQueuePresent calls to low latency frame IDs as follows:

- When vkQueuePresent is called in between OUT_OF_BAND_PRESENT_START and _END markers, use their frame id. (vkd3d-proton will always take this path)
- Otherwise take the frame id of the most recent PRESENT_START marker that has not already been mapped to a vkQueuePresent call. (vkd3d-proton mirrors this logic during the DXGI present call to determine the frame id)
- If no such marker exists then Reflex is off or not implemented correctly, this latency measurement scheme is not possible.

Saw reasonable results in several titles with various combinations of Reflex on and off, DLSS-FG and NVIDIA Smooth Motion.

 - Cyberpunk 2077, Arc Raiders and Dying Light 2

Also tested in some Vulkan native titles to confirm this vkQueuePresent to latency marker mapping scheme works there as well.

### DIFF
--- a/libs/vkd3d/swapchain.c
+++ b/libs/vkd3d/swapchain.c
@@ -1098,16 +1098,44 @@ static HRESULT STDMETHODCALLTYPE dxgi_vk_swap_chain_Present(IDXGIVkSwapChain2 *i
     request->dxgi_hdr_metadata = chain->user.dxgi_hdr_metadata;
     request->modifies_hdr_metadata = chain->user.modifies_hdr_metadata;
     request->begin_frame_time_ns = chain->user.begin_frame_time_ns;
-    request->low_latency_frame_id = vkd3d_atomic_uint64_load_explicit(
-            &chain->queue->device->frame_markers.present, vkd3d_memory_order_acquire);
-
-    if (chain->debug_latency && request->low_latency_frame_id)
-        INFO("Presenting with low latency frame ID: %"PRIu64".\n", request->low_latency_frame_id);
+    request->low_latency_frame_id = 0;
 
     chain->user.modifies_hdr_metadata = false;
 
     if (chain->queue->device->vk_info.NV_low_latency2)
     {
+        struct vkd3d_device_frame_markers *markers = &chain->queue->device->frame_markers;
+
+        /* When DLSS Frame Generation is enabled each DXGI Present will be
+         * called between OUT_OF_BAND_PRESENT START and END markers. We'll
+         * deliver these markers on our background present thread surrounding the
+         * corresponding vkQueuePresentKHR call, so that the presentation can
+         * be mapped to the correct low_latency_frame_id. We can have multiple
+         * presents mapping to the same id. */
+        request->low_latency_frame_id = vkd3d_atomic_uint64_load_explicit(
+                &markers->out_of_band_present, vkd3d_memory_order_acquire);
+        if (!request->low_latency_frame_id)
+        {
+            /* If NVIDIA Reflex is active with frame gen off, the DXGI Present should
+             * be surrounded by PRESENT markers. We'll deliver OUT_OF_BAND_PRESENT
+             * markers with a matching frame id on our background present thread,
+             * again so that the vkQueuePresentKHR call can be mapped to
+             * the correct low_latency_frame_id.
+             *
+             * We also sanitize the input to ensure we don't send multiple presents
+             * mapping to the same id in this frame gen off case. */
+            uint64_t current_present_marker_id = vkd3d_atomic_uint64_load_explicit(
+                    &markers->present, vkd3d_memory_order_acquire);
+
+            spinlock_acquire(&chain->queue->device->low_latency_swapchain_spinlock);
+            if (current_present_marker_id > markers->consumed_present_id)
+            {
+                request->low_latency_frame_id = current_present_marker_id;
+                markers->consumed_present_id = request->low_latency_frame_id;
+            }
+            spinlock_release(&chain->queue->device->low_latency_swapchain_spinlock);
+        }
+
         pthread_mutex_lock(&chain->present.low_latency_state_update_lock);
         request->requested_low_latency_state = chain->requested_low_latency_state;
         request->low_latency_update_requested = chain->low_latency_update_requested;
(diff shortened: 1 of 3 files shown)


libs/vkd3d/swapchain.c
```diff
@@ -3380,11 +3405,23 @@ void dxgi_vk_swap_chain_set_latency_sleep_mode(struct dxgi_vk_swap_chain *chain,
...
+        uint64_t frameID, VkLatencyMarkerNV marker, bool from_app)
 {
     const struct vkd3d_vk_device_procs *vk_procs = &chain->queue->device->vk_procs;
     VkSetLatencyMarkerInfoNV latency_marker_info;
 
+    if (from_app && (marker == VK_LATENCY_MARKER_OUT_OF_BAND_PRESENT_START_NV
+                     || marker == VK_LATENCY_MARKER_OUT_OF_BAND_PRESENT_END_NV))
+    {
+        /* Record the frame ID for OUT_OF_BAND_PRESENT markers. We'll send
+         * these markers later on our background present thread. */
```
HansKristian-Work, line 3418: Who is "our" referring to here? NV reflex or vkd3d-proton?

libs/vkd3d/swapchain.c
```diff
@@ -2694,11 +2704,26 @@ static void dxgi_vk_swap_chain_present_iteration(struct dxgi_vk_swap_chain *chai
...
+                VK_LATENCY_MARKER_OUT_OF_BAND_PRESENT_START_NV, false);
+    }
+
     vk_queue = vkd3d_queue_acquire(chain->queue->vkd3d_queue);
     VKD3D_REGION_BEGIN(queue_present);
     vr = VK_CALL(vkQueuePresentKHR(vk_queue, &present_info));
     VKD3D_REGION_END(queue_present);
     vkd3d_queue_release(chain->queue->vkd3d_queue);
+
+    if (chain->queue->device->vk_info.NV_low_latency2
```
HansKristian-Work, line 2720: If should be on one line.

libs/vkd3d/swapchain.c
```diff
@@ -2694,11 +2704,26 @@ static void dxgi_vk_swap_chain_present_iteration(struct dxgi_vk_swap_chain *chai
     else
         use_present_id = false;
 
+    if (chain->queue->device->vk_info.NV_low_latency2
```
HansKristian-Work, line 2707: If should be on one line.

libs/vkd3d/swapchain.c
```diff
@@ -2650,39 +2683,16 @@ static void dxgi_vk_swap_chain_present_iteration(struct dxgi_vk_swap_chain *chai
...
-            spinlock_acquire(&chain->queue->device->low_latency_swapchain_spinlock);
-            chain->present.present_id = max(chain->present.present_id, markers->consumed_present_id + 1);
-            markers->consumed_present_id = chain->present.present_id;
-            spinlock_release(&chain->queue->device->low_latency_swapchain_spinlock);
-        }
-
-        if (chain->debug_latency)
-            INFO("Presenting with frame ID: %"PRIu64".\n", chain->present.present_id);
+        /* Ensure present ID is increasing monotonically. */
+        chain->present.present_id = max(chain->present.present_id+1, present_count);
```
HansKristian-Work, line 2695:
```suggestion
        chain->present.present_id = max(chain->present.present_id + 1, present_count);
```

libs/vkd3d/swapchain.c
```diff
@@ -2694,11 +2704,26 @@ static void dxgi_vk_swap_chain_present_iteration(struct dxgi_vk_swap_chain *chai
     else
         use_present_id = false;
 
+    if (chain->queue->device->vk_info.NV_low_latency2
+        && chain->request.low_latency_frame_id)
+    {
+        dxgi_vk_swap_chain_set_latency_marker(chain, chain->request.low_latency_frame_id,
```
HansKristian-Work, line 2710: We still have the LOW_LATENCY_FRAME_ID_STRIDE in low_latency_device_SetLatencyMarker. Should this not be removed now? It seems like LL2 frame id is not actually coupled with presentID anymore in this setup which makes all that mess go away.

HansKristian-Work: If I understand it correctly, is OUT_OF_BAND_PRESENT_START_NV/END_NV a way to decouple present ID from LL2 frame ID and that's why all this stuff works fine now?

HansKristian-Work (changes requested):
Nits and comments, but overall seems like a good cleanup if it works. I'll assume it does since it's NV proposing this change.

Locking LL2 ID to present ID has been rather miserable and fixing that would remove most of the problems we have with LL2 I think.

libs/vkd3d/swapchain.c
HansKristian-Work: Does OUT_OF_BAND_PRESENT_START_NV work only for out of band queues, or is it designed to work for threaded presents in general?

wcampbell-nv: I've posted a new revision removing LOW_LATENCY_FRAME_ID_STRIDE.

> If I understand it correctly, is OUT_OF_BAND_PRESENT_START_NV/END_NV a way to decouple present ID from LL2 frame ID and that's why all this stuff works fine now?

Yes, the new convention is that VkPresentIdKHR is totally separate from the LL2 frame IDs. The PRESENT and OUT_OF_BAND_PRESENT markers are used to determine the LL2 frame ID of a vkQueuePresent call, following the scheme I describe in the 'Testing' section of the PR.

> Does OUT_OF_BAND_PRESENT_START_NV work only for out of band queues, or is it designed to work for threaded presents in general?

It is designed to work for threaded presents in general.

libs/vkd3d/swapchain.c
wcampbell-nv: "our" as in "vkd3d-proton's"

wcampbell-nv:
> Locking LL2 ID to present ID has been rather miserable and fixing that would remove most of the problems we have with LL2 I think.

There are some issues with vkQueueSubmit attribution remaining, which can cause stutter/subpar latency in games with frame gen enabled. Fixing this will require changes to both the NVIDIA driver and vkd3d-proton. We're still working out the details on this.

HansKristian-Work (approved):

PR #2940, opened 2026-04-16
vkd3d: Implicitly bind RTPSO root signature for DispatchRays when none is set

danielbanariba:
## Summary

Some DXR applications call `DispatchRays` without explicitly calling `SetComputeRootSignature`, relying on the RTPSO's embedded global root signature. Native D3D12 drivers (tested: NVIDIA) handle this implicitly.

**Without this fix**, vkd3d-proton silently drops every `DispatchRays` call with:
```
warn: d3d12_command_list_DispatchRays: Failed to update raygen state, ignoring dispatch.
```

**Root cause** (two issues in `d3d12_command_list_update_raygen_pipeline`):

1. The degenerate variant (`pipelines[0]`, NULL root signature fallback) matches the NULL bound root signature but has `pipeline == VK_NULL_HANDLE`. The loop `break`s after this match, never reaching the real variant.

2. Even when the correct variant is found via fallback, `compute_bindings.root_signature` remains NULL, so `d3d12_command_list_update_descriptors` cannot bind any descriptors — resulting in black frames.

**This fix:**
- Skips degenerate variants with NULL pipeline handles instead of breaking the loop
- When bound RS is NULL/empty, falls back to the first variant with a valid pipeline and implicitly binds its global root signature for correct descriptor layout

## Test case

**sm64rt** (Super Mario 64 with RT64 path tracer) — legacy DXR 1.0 application that calls `DispatchRays` without `SetComputeRootSignature`.

| | Before | After |
|---|---|---|
| DispatchRays | Every call dropped silently | All calls succeed |
| Rendering | Black screen (only 2D HUD visible) | Full path-traced scene |
| Stability | N/A | Stable, no crashes, no validation errors |

**Hardware**: NVIDIA GeForce RTX 3090 Ti, driver 595.58.03, Vulkan 1.4.329

**Environment**: Wine 11.6, DXVK 2.7.1 (dxgi), Arch Linux (CachyOS kernel 6.19.12)

## Debug data that confirmed the root cause

```
bound_rs = 0x0000000000000000 (hash = 0x0000000000000000)  ← app never calls SetComputeRootSignature

variant[0]: rs hash = 0x0000000000000000, pipeline = NULL   ← degenerate, matches but unusable
variant[1]: rs hash = 0xd94d12186c0f2fb7, pipeline = VALID  ← real variant, doesn't match NULL
```

After the fallback binds variant[1]'s root signature, subsequent frames match variant[1] directly through the normal path (self-correcting).

## Risk assessment

- **Low risk**: The fallback only triggers when `compute_bindings.root_signature` is NULL/empty — games that correctly call `SetComputeRootSignature` are unaffected
- **Matches native D3D12 behavior**: NVIDIA's driver implicitly uses the RTPSO's root signature when none is explicitly bound
- The degenerate variant skip (`break` moved inside `if (pipeline)`) is a strict improvement — breaking on a NULL pipeline was always a bug

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -7139,16 +7139,49 @@ static bool d3d12_command_list_update_raygen_pipeline(struct d3d12_command_list 
                 list->rt_state->pipelines[i].global_root_signature,
                 list->compute_bindings.root_signature))
         {
-            /* We might have degenerate variants. */
+            /* We might have degenerate variants with no compiled pipeline.
+             * Skip them and continue searching for a valid one. */
             if (list->rt_state->pipelines[i].pipeline)
+            {
                 list->rt_state_variant = &list->rt_state->pipelines[i];
-            break;
+                break;
+            }
         }
     }
 
     if (!list->rt_state_variant)
     {
-        WARN("Global root signature compatible with the RTPSO.\n");
+        /* Some applications (e.g. RT64) call DispatchRays without explicitly calling
+         * SetComputeRootSignature, relying on the RTPSO's embedded global root signature.
+         * Native D3D12 drivers handle this implicitly. When the bound root signature is
+         * NULL or empty, fall back to the first variant with a valid pipeline and
+         * implicitly bind its global root signature for correct descriptor layout. */
+        if (!list->compute_bindings.root_signature ||
+                list->compute_bindings.root_signature->layout_compatibility_hash == 0)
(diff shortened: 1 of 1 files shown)


HansKristian-Work:
Broken application, but guess we need to workaround then <_< sigh.

HansKristian-Work (approved):
sigh

PR #2945, opened 2026-04-17
vkd3d: Fix shader interface key for quirks above 32-bit.

HansKristian-Work:
Ridiculous ...


doitsujin (commented):
Ah well, I guess we should make sure to use pointer-based APIs going forward to at least get a compiler warning.

doitsujin (approved):

PR #2947, opened 2026-04-18
vkd3d: Recognize DX core adapter in d3d12_get_adapter().

gofman:


gofman:
That fixes Age of Wonders 4 (Steam ID 1669000) which regressed in Proton 11 after some minor advancements in dxcore support. The game now enumerates DX Core adapters, tries to QueryInterface(IID_IDXGIAdapter) which fails (I tested that also fails on Windows), and then passes IDXCoreAdapter for d3d12 device creation. I tested on Windows that D3D12CreateDevice() succeeds with DX core adapter.

Ultimately dxcore's place is somewhat close to dxgi. Wine's implementation uses wined3d directly (that works now with vkd3d-proton too as soon as the only implemented dxcore functionality is enumerating adapters and those adapters don't have anything but luid). Going forward with a more complete implementation, dxcore should probably exist as a part of dxvk (similar to dxgi.dll), and then vkd3d-proton would probably grow a more direct interfacing with it.

include/private/dxcore_interface.h
```diff
@@ -0,0 +1,319 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
+ */
+
+#ifndef __DXCORE_INTERFACE__
+#define __DXCORE_INTERFACE__
```
HansKristian-Work, line 20: Why is this not the idl file?

libs/d3d12core/main.c
```diff
@@ -388,6 +390,37 @@ static HRESULT d3d12_get_adapter(IDXGIAdapter **dxgi_adapter, IUnknown *adapter)
...
+        {
+            WARN("Failed to get InstanceLuid property from dxcore adapter, hr %#x.\n", hr);
+            goto done;
+        }
+        if (FAILED(hr = CreateDXGIFactory1(&IID_IDXGIFactory4, (void **)&factory)))
+        {
+            WARN("Failed to create DXGI factory, hr %#x.\n", hr);
+            goto done;
+        }
+        for (i = 0; !IDXGIFactory4_EnumAdapters(factory, i, dxgi_adapter); ++i)
```
HansKristian-Work: SUCCEEDED() instead of false check I think?

libs/d3d12core/main.c
```diff
@@ -388,6 +390,37 @@ static HRESULT d3d12_get_adapter(IDXGIAdapter **dxgi_adapter, IUnknown *adapter)
...
+        {
+            if (FAILED(hr = IDXGIAdapter_GetDesc(*dxgi_adapter, &desc)))
+            {
+                IDXGIAdapter_Release(*dxgi_adapter);
```
HansKristian-Work, line 413: Looks like a crash if it also releases on line 436?

libs/d3d12core/main.c
```diff
@@ -388,6 +390,37 @@ static HRESULT d3d12_get_adapter(IDXGIAdapter **dxgi_adapter, IUnknown *adapter)
...
+                goto done;
+            }
+            if (!memcmp(&luid, &desc.AdapterLuid, sizeof(luid)))
+                goto done;
+            IDXGIAdapter_Release(*dxgi_adapter);
```
HansKristian-Work, line 419: Should probably be nulled out once it's release. If this is the last iteration of the loop, I'm not sure EnumAdapters will null the output.

HansKristian-Work (changes requested):
Comments.

include/private/dxcore_interface.h
gofman: There is no IDL file (only header) in Windows SDK and thus in Wine. It is of course possible to craft an idl but IMO easier to keep in sync this way.

libs/d3d12core/main.c
gofman: Done.

libs/d3d12core/main.c
gofman: That is dxcore_adapter freed at 'done', not this dxgi_adapter.

libs/d3d12core/main.c
gofman: The value is not used on error return, I wasn't sure if nullifying output interfaces on errors is assumed for internal local functions. Updated.

HansKristian-Work (approved):

PR #2951, opened 2026-04-20
Update command.c. Deduplicate query pool resets on NVIDIA

ghost:
For VK_DRIVER_ID_NVIDIA_PROPRIETARY (unless VKD3D_CONFIG_FLAG_SKIP_DRIVER_WORKAROUNDS is set), track query resets per (VkQueryPool, index) in command lists and avoid emitting redundant vkCmdResetQueryPool() calls.
Limit this optimization to NVIDIA proprietary drivers for now, since behavior and performance impact on other GPUs is currently unknown.

In local testing this reduces command stream overhead and improves performance in games on GTX 1660 (driver 595.58.03, CachyOS, kernel 6.19.11-arch1-1 stable).

Tested in games:

Frostpunk 2 (Unreal Engine 5), low settings, FSR set to Ultra Performance:
before: up to 142 FPS
after: up to 145 FPS
Civilization VI (DX12), high settings:
average frame time (before): 12.7 ms
average frame time (after): 12.7 ms
99th percentile frame time (before): 19.7 ms
99th percentile frame time (after): 18.7 ms

It was discovered after three weeks of researching the repository code with the help of Codex, GPT-5.4 High. I’m stating honestly that this code was written by that model and verified by me for correctness, meaning full system stability and the absence of any visual artifacts.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -6968,16 +6968,16 @@ static inline bool d3d12_device_inline_query_resets(const struct d3d12_device *d
     return device->device_info.vulkan_1_2_properties.driverID == VK_DRIVER_ID_NVIDIA_PROPRIETARY;
 }
 
-static bool d3d12_command_list_track_inline_query_reset(struct d3d12_command_list *list,
-                                                        VkQueryPool vk_pool, uint32_t index)
+static void d3d12_command_list_track_inline_query(struct d3d12_command_list *list,
+        VkQueryPool vk_pool, uint32_t index)
 {
     size_t pos;
 
+    /* Keep query bookkeeping intact even when resets are emitted inline. */
     if (d3d12_command_list_find_query(list, vk_pool, index, &pos))
-        return false;
+        return;
 
     d3d12_command_list_insert_query_range(list, &pos, vk_pool, index, 1, 0);
-    return true;
 }
 static void d3d12_command_list_reset_rtv_resolves(struct d3d12_command_list *list)
 {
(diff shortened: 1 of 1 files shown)


libs/vkd3d/command.c
```diff
@@ -16064,7 +16083,12 @@ static void STDMETHODCALLTYPE d3d12_command_list_BeginQuery(d3d12_command_list_i
 
         /* We're using the query heap directly, so we must reset it on the GPU.
          * At most we can hoist the query reset to init_command_buffer. */
-        if (!d3d12_command_list_reset_query(list, query_heap->vk_query_pool, index))
+        if (d3d12_device_inline_query_resets(list->device))
+        {
+            if (d3d12_command_list_track_inline_query_reset(list, query_heap->vk_query_pool, index))
+                VK_CALL(vkCmdResetQueryPool(list->cmd.vk_command_buffer, query_heap->vk_query_pool, index, 1));
```
HansKristian-Work: This is completely broken. Why are you attempting to not reset the query here? If the query has already been used, it must be reset if it is to be reused.

HansKristian-Work (changes requested):
I really have no idea how this is supposed to fix anything. In fact, this is just breaking stuff for no reason.

ghost:
I figured out where the error was. After fixing it, I don’t really see much difference. We’ll probably close the PR.

Tests after the fix:
Frostpunk FPS is the same as in the original code.
Civilization VI(DX12) FPS hasn’t changed. However, the 99th percentile is still better than in the repository baseline. Repeated runs show an improvement of about 1 ms — from 19.3 ms to 18.3 ms.

I suggest shelving the code for now, although it’s still a valid approach.

HansKristian-Work:
The idea that you can avoid "redundant" QueryPoolReset calls makes no sense to me (it would violate VUs) and the code is still just confusing. If you cannot fully understand and explain yourself why this PR works, then it's not a good PR and I won't spend time on it. I'll close this PR for now. I think you're placebo-ing yourself into thinking there are gains here.

ghost:
Closed.

Just a quick clarification, as the fix actually changed the whole point of this PR. It's not about deduplication anymore. Leaving this here in case anyone needs it.
On NVIDIA proprietary driver, emitting vkCmdResetQueryPool inline
immediately before each query reduces 99th percentile frame time
in Civilization VI (DX12) by ~1ms (19.3ms → 18.3ms, 10 runs).
Average frame time and throughput are unchanged.

The mechanism is not fully understood. 

Tested on GTX 1660, driver 595.58.03.

PR #2952, opened 2026-04-20
Use VkFence for ID3D12Fence signaling on NV prop driver

### LINKED ISSUE #2950: Horizon Zero Dawn Remastered performance regression on descriptor-heap-test branch
### Software information
Horizon Zero Dawn Remastered

### System information
- GPU: RTX 3080
- Driver: Nvida 595.58.03
- VKD3D-Proton version: proton-cachyos-10.0-20260320-slr-x86_64_v3 and proton-cachyos-10.0-20260321-slr-x86_64_v3

The performance on the [descriptor-heap-test](https://github.com/HansKristian-Work/vkd3d-proton/tree/refs/heads/descriptor-heap-test) was very good on my system, but after commit https://github.com/HansKristian-Work/vkd3d-proton/commit/97a0c5d0bfb47321f3a1fd52c1709e58feba7290 where code is merged from `master` the performance regressed pretty badly for some reason. I've swapped different versions of `d3d12.dll` and `d3d12core.dll` to find the commit, unfortunately it's a merge commit so it's hard to pinpoint the culprit.

I'm aware of the fact that `descriptor-heap-rebase` is the main branch now for descriptor-heap and that descriptor-heap-test is mainly a prototype. My main reason for reporting this is to bring attention to possibility that the code from the `master` branch may cause slow downs:

Image
On commit 3864e83d7935a6016d92c025a30362a0992237c0 i get 79 fps

Image
On commit 97a0c5d0bfb47321f3a1fd52c1709e58feba7290 only 54 fps

HansKristian-Work:
NVIDIA proprietary apparently cannot deal well with timeline semaphores despite it being a 8 year old API. Adding the transfer queue probably reintroduced CPU overhead problems that we had before we moved to single timeline per vkd3d_queue :(

Reintroduce terrible old school virtualization where we assign VkFence per timeline value like it's 2017 again.

Fix #2950.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -22404,6 +22404,11 @@ static bool d3d12_command_queue_needs_cpu_waits_locked(struct d3d12_command_queu
     if (vkd3d_config_flags & VKD3D_CONFIG_FLAG_NO_STAGGERED_SUBMIT)
         return false;
 
+    /* If we risk massive stalls due to broken timeline semaphores,
+     * it's better to just go ahead. */
+    if (d3d12_device_has_slow_cpu_timeline_semaphores(command_queue->device))
+        return false;
+
     current_time_ns = vkd3d_get_current_time_ns();
 
     for (i = 0; i < command_queue->vkd3d_queue->command_queue_count; i++)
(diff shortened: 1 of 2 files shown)


doitsujin (approved):
pain, i suppose the plan is to nuke this path again once we can be reasonably sure that a fixed driver is rolled out to most users?

HansKristian-Work:
Right, I think we'll carry the workaround until a fix hits stable drivers.

PR #2966, opened 2026-04-27
vkd3d: Fix stale OMM pNext after RTAS batch realloc.

chrisd-work:
This addresses a use-after-free bug that occurs in the following situation:

1. `geometry.triangles.pNext` is set to `rtas_batch->omm_infos[i]` when converting `OMM_TRIANGLES`.
2. Enough triangles get converted afterwards that `rtas_batch->omm_infos` is realloced and its base pointer changes.

In this case `vkCmdBuildAccelerationStructuresKHR` will dereference a stale pointer.

This bug is limited to when DXR 1.2 is enabled, and OMM is used by an application.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -19713,6 +19713,7 @@ static void d3d12_command_list_flush_rtas_batch(struct d3d12_command_list *list)
 {
     const struct vkd3d_vk_device_procs *vk_procs = &list->device->vk_procs;
     struct d3d12_rtas_batch_state *rtas_batch = &list->rtas_batch;
+    VkAccelerationStructureGeometryKHR *geometry;
     unsigned int i, geometry_index, usage_index;
 
     if (!rtas_batch->build_info_count && !rtas_batch->omm_build_info_count)
@@ -19756,6 +19757,20 @@ static void d3d12_command_list_flush_rtas_batch(struct d3d12_command_list *list)
         rtas_batch->omm_build_infos[i].pUsageCounts = &rtas_batch->omm_usage_infos[usage_index];
 
         usage_index += usage_count;
+    }
+
+    /* pNext pointers on Geometry AS. Currently limited to Opacity micromaps */
+    for (i = 0; i < rtas_batch->geometry_info_count; i++)
+    {
+        geometry = &rtas_batch->geometry_infos[i];
+
+        if (geometry->geometry.triangles.pNext)
+        {
+            assert(geometry->geometryType == VK_GEOMETRY_TYPE_TRIANGLES_KHR);
+            assert(((const VkBaseInStructure *)&rtas_batch->omm_infos[i])->sType ==
+                   VK_STRUCTURE_TYPE_ACCELERATION_STRUCTURE_TRIANGLES_OPACITY_MICROMAP_EXT);
+            geometry->geometry.triangles.pNext = &rtas_batch->omm_infos[i];
+        }
     }
 
     d3d12_command_list_end_current_render_pass(list, true);


libs/vkd3d/command.c
```diff
@@ -19626,6 +19627,20 @@ static void d3d12_command_list_flush_rtas_batch(struct d3d12_command_list *list)
...
 
+    /* pNext pointers on Geometry AS. Currently limited to Opacity micromaps */
+    for (i = 0; i < rtas_batch->geometry_info_count; i++)
+    {
+        geometry = &rtas_batch->geometry_infos[i];
+
+        if (geometry->geometry.triangles.pNext)
+        {
+            assert(geometry->geometryType == VK_GEOMETRY_TYPE_TRIANGLES_KHR);
+            assert(((const VkBaseInStructure *)geometry->geometry.triangles.pNext)->sType ==
```
HansKristian-Work: This seems broken. Aren't you potentially accessing freed memory in the assert here?

HansKristian-Work: I think it makes more sense to not setup pNext in line acceleration_structure.c:270 and then setup the pNext here if the respective omm struct is setup with a valid sType. It should default to something that is memset.

Also, no point in looping over all geometry infos if OOM is not enabled.

chrisd-work: 1. Sigh..... I was too eager to get this to a PR.

2. I will make this change. I have had some thoughts about tracking the reallocs, and if the base pointer changes, then and only the do the fixups. However, I am not sure if the fixups have been a problem on profiles or not.

chrisd-work: In this case actually, we have to set the pNext in the acceleration_structure code. The code assumes that if the pNext is set, it will be set with `VK_STRUCTURE_TYPE_ACCELERATION_STRUCTURE_TRIANGLES_OPACITY_MICROMAP_EXT`. Since the object that is pointing to the micromap is just a `VK_GEOMETRY_TYPE_TRIANGLES_KHR`, we have no way of knowing.

The asserts guard us against this changing later. In fact, I do worry that at some point using arrays everywhere here will not work. It just happens to work today.

HansKristian-Work: Where is this assumption made?

chrisd-work: This is guarded by two asserts I added to the code:

```
   assert(geometry->geometryType == VK_GEOMETRY_TYPE_TRIANGLES_KHR);
   assert(((const VkBaseInStructure *)&rtas_batch->omm_infos[i])->sType ==
                VK_STRUCTURE_TYPE_ACCELERATION_STRUCTURE_TRIANGLES_OPACITY_MICROMAP_EXT);
```

If this changes in the future, it will require much more sophisticated tracking. Notably, we could imagine a future where a different `pNext` struct exists that serves a similar purpose to `VK_STRUCTURE_TYPE_ACCELERATION_STRUCTURE_TRIANGLES_OPACITY_MICROMAP_EXT` but is of a different size. If that occurs, the current code will likely break, since it treats omm_infos as a fixed-stride array. I consider this a problem that isn't worth solving today, but the asserts give us at least some feedback if it arises in the future.

HansKristian-Work:
Is this PR still active? Considering I saw this commit being part of the big OMM PR.

chrisd-work:
Yes, I consider it active, responded to your last comment. I am re working the latter part of the OMM PR based on @Saancreed's feedback. Some of it is hitting earlier commits so I want to make sure it robust before I post the next pass. The good thing about @Saancreed's feedback is it has reduced the complexity of the PR quite a bit.

The biggest impact on the earlier commits is the definition of what does it mean for the OMM feature to be on. DXR 1.2 is a very wide net, SM 6.9, SER, and OMM at a minimum. Whereas a game only needs VK_KHR_opacity_micromap to support the NVAPI OMM extensions.

HansKristian-Work:
The assert is still completely broken I think. I'll just fix this in another way ...

HansKristian-Work:
Merged a different approach to solving this.

PR #2967, opened 2026-04-27
Add initial support for NV shader intrinsics and NV flavor of SER

Saancreed:
Since we're now doing AMD intrinsics too… The dxvk-nvapi side isn't wired up yet, but I'll deal with that soon.

I intend to keep this hidden behind an environment variable on nvapi side, with explicit opt-in for well-behaved titles, otherwise games like Cyberpunk 2077 will just break (it uses some atomic operations without checking if they're supported first :weary:), at least for as long as dxil-spirv has a problem with HitObjects used around Phi instructions.

### DIFF
--- a/libs/vkd3d-shader/dxil.c
+++ b/libs/vkd3d-shader/dxil.c
@@ -1147,6 +1147,19 @@ static int vkd3d_dxil_converter_set_options(dxil_spv_converter converter,
                 return VKD3D_ERROR_NOT_IMPLEMENTED;
             }
         }
+
+        if (compiler_args->nv_shader_extn_uav_slot != UINT32_MAX)
+        {
+            const dxil_spv_option_nvapi helper =
+                    { { DXIL_SPV_OPTION_NVAPI },
+                    DXIL_SPV_TRUE, compiler_args->nv_shader_extn_uav_slot, compiler_args->nv_shader_extn_uav_space };
+            if (dxil_spv_converter_add_option(converter, &helper.base) != DXIL_SPV_SUCCESS)
+            {
+                ERR("dxil-spirv does not support NVAPI.\n");
+                return VKD3D_ERROR_NOT_IMPLEMENTED;
+            }
+        }
+
         if (compiler_args->dual_source_blending)
         {
             static const dxil_spv_option_dual_source_blending helper =
--- a/libs/vkd3d/device.c
+++ b/libs/vkd3d/device.c
@@ -4722,6 +4722,8 @@ static void d3d12_device_destroy(struct d3d12_device *device)
 
     vkd3d_free((void *)device->vk_info.extension_names);
     VK_CALL(vkDestroyDevice(device->vk_device, NULL));
+    hash_map_free(&device->nv_shader_extns);
+    rwlock_destroy(&device->nv_shader_lock);
     rwlock_destroy(&device->fragment_output_lock);
     rwlock_destroy(&device->vertex_input_lock);
     pthread_mutex_destroy(&device->mutex);
@@ -10521,6 +10523,7 @@ static HRESULT d3d12_device_init(struct d3d12_device *device,
 
     device->adapter_luid = create_info->adapter_luid;
     device->removed_reason = S_OK;
+    device->has_nv_shader_extns = false;
 
     device->vk_device = VK_NULL_HANDLE;
 
@@ -10551,8 +10554,14 @@ static HRESULT d3d12_device_init(struct d3d12_device *device,
         goto out_free_vertex_input_lock;
     }
 
-    if (FAILED(hr = vkd3d_create_vk_device(device, create_info)))
+    if ((rc = rwlock_init(&device->nv_shader_lock)))
+    {
+        hr = hresult_from_errno(rc);
         goto out_free_fragment_output_lock;
+    }
+
+    if (FAILED(hr = vkd3d_create_vk_device(device, create_info)))
+        goto out_free_nv_shader_lock;
 
     if (FAILED(hr = vkd3d_private_store_init(&device->private_store)))
         goto out_free_vk_resources;
@@ -10631,6 +10640,11 @@ static HRESULT d3d12_device_init(struct d3d12_device *device,
             vkd3d_fragment_output_pipeline_desc_compare,
             sizeof(struct vkd3d_fragment_output_pipeline));
 
+    hash_map_init(&device->nv_shader_extns,
+            vkd3d_nv_shader_extn_entry_hash,
+            vkd3d_nv_shader_extn_entry_compare,
+            sizeof(struct vkd3d_nv_shader_extn_entry));
+
     if ((device->parent = create_info->parent))
         IUnknown_AddRef(device->parent);
 
@@ -10697,6 +10711,8 @@ out_free_vk_resources:
     VK_CALL(vkDestroyDevice(device->vk_device, NULL));
 out_free_instance:
     vkd3d_instance_decref(device->vkd3d_instance);
+out_free_nv_shader_lock:
+    rwlock_destroy(&device->nv_shader_lock);
 out_free_fragment_output_lock:
     rwlock_destroy(&device->fragment_output_lock);
 out_free_vertex_input_lock:
--- a/libs/vkd3d/device.c
+++ b/libs/vkd3d/device.c
@@ -4506,7 +4506,8 @@ HRESULT STDMETHODCALLTYPE d3d12_device_QueryInterface(d3d12_device_iface *iface,
     if (IsEqualGUID(riid, &IID_ID3D12DeviceExt)
             || IsEqualGUID(riid, &IID_ID3D12DeviceExt1)
             || IsEqualGUID(riid, &IID_ID3D12DeviceExt2)
-            || IsEqualGUID(riid, &IID_ID3D12DeviceExt3))
+            || IsEqualGUID(riid, &IID_ID3D12DeviceExt3)
+            || IsEqualGUID(riid, &IID_ID3D12DeviceExt4))
     {
         d3d12_device_vkd3d_ext_AddRef(&device->ID3D12DeviceExt_iface);
         *object = &device->ID3D12DeviceExt_iface;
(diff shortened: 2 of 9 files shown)


libs/vkd3d-shader/dxil.c
```diff
@@ -1147,6 +1147,19 @@ static int vkd3d_dxil_converter_set_options(dxil_spv_converter converter,
                 return VKD3D_ERROR_NOT_IMPLEMENTED;
             }
         }
+
+        if (~compiler_args->nv_shader_extn_uav_slot)
```
HansKristian-Work: ... what? Is this supposed to read slot != UINT32_MAX?

libs/vkd3d/device.c
```diff
@@ -2523,6 +2524,12 @@ static void vkd3d_physical_device_info_init(struct vkd3d_physical_device_info *i
         vk_prepend_struct(&info->features2, &info->present_timing_features);
     }
 
+    if (vulkan_info->NV_ray_tracing_invocation_reorder)
```
HansKristian-Work, line 2527: Why NV when we have EXT or is it a case that NVAPI one maps more to NV than EXT?

libs/vkd3d/device_vkd3d_ext.c
```diff
@@ -20,6 +20,53 @@
...
+bool vkd3d_nv_shader_extn_entry_compare(const void *key, const struct hash_map_entry *entry)
+{
+    const struct vkd3d_nv_shader_extn_entry *extn = (const void*)entry;
+    const unsigned int *thread_id = key;
+    return *thread_id == extn->thread_id;
+}
+
+struct vkd3d_nv_shader_extn d3d12_device_get_nv_shader_extn(struct d3d12_device *device)
+{
+    struct vkd3d_nv_shader_extn result = { ~(uint32_t)0, 0 };
```
HansKristian-Work: Just write UINT32_MAX

libs/vkd3d/device_vkd3d_ext.c
```diff
@@ -20,6 +20,53 @@
...
+
+    if ((rc = rwlock_lock_read(&device->nv_shader_lock)))
+    {
+        ERR("Failed to lock mutex, rc %d.\n", rc);
+        return result;
+    }
+
+    entry = (void*)hash_map_find(&device->nv_shader_extns, &key);
+
+    if (!entry || !~entry->extn.uav_slot)
```
HansKristian-Work: `!~` is weird. Use normal literals.

libs/vkd3d/device_vkd3d_ext.c
```diff
@@ -486,10 +487,32 @@ static HRESULT STDMETHODCALLTYPE d3d12_device_vkd3d_ext_SetAGSUAVSlot(d3d12_devi
...
+        case NV_EXTN_OP_HIT_OBJECT_GET_GEOMETRY_INDEX:
+        case NV_EXTN_OP_HIT_OBJECT_GET_HIT_KIND:
+        case NV_EXTN_OP_HIT_OBJECT_GET_RAY_DESC:
+        case NV_EXTN_OP_HIT_OBJECT_GET_ATTRIBUTES:
+        case NV_EXTN_OP_HIT_OBJECT_GET_SHADER_TABLE_INDEX:
+        case NV_EXTN_OP_HIT_OBJECT_LOAD_LOCAL_ROOT_TABLE_CONSTANT:
+        case NV_EXTN_OP_HIT_OBJECT_IS_HIT:
+        case NV_EXTN_OP_HIT_OBJECT_IS_NOP:
+        case NV_EXTN_OP_HIT_OBJECT_MAKE_NOP:
+            return device->device_info.ray_tracing_invocation_reorder_features_nv.rayTracingInvocationReorder
```
HansKristian-Work: Just put it on one line.

libs/vkd3d-shader/dxil.c
Saancreed: Yes, setting it to `0xFFFFFFFF` is supposed to be interpreted as disabling them, see https://github.com/NVIDIA/nvapi/blob/9b181ea572f680327fe01a14a0f1f41c78034104/nvapi.h#L11525. Which is why I reused that UINT32_MAX as a default sentinel/unset value.

HansKristian-Work: That is fine, but the issue I have is that using bitwise negate here is weird compared to != UINT32_MAX.

libs/vkd3d/device.c
Saancreed: It's mostly the latter. dxil-spirv currently maps NVAPI opcodes to NV extension, and the EXT one lacks a direct equivalent for `NV_EXTN_OP_HIT_OBJECT_MAKE_HIT` / `OpHitObjectRecordHitNV`.

libs/vkd3d/device.c
```diff
@@ -10551,8 +10554,14 @@ static HRESULT d3d12_device_init(struct d3d12_device *device,
         goto out_free_vertex_input_lock;
     }
 
-    if (FAILED(hr = vkd3d_create_vk_device(device, create_info)))
+    if ((rc = rwlock_init(&device->nv_shader_lock)))
```
HansKristian-Work: I would like these init/deinit things of hashmaps to be moved to device_ext.c and a struct that holds this information. It's a bit too ugly to have in the main code path here I think.

d3d12_device_init_vendor_hacks() is a suitable place to call that helper I think.

libs/vkd3d-shader/dxil.c
Saancreed: Right, I'll switch to just UINT32_MAX then.

libs/vkd3d/device.c
Saancreed: Fair enough, I'll see what I can do.

Saancreed: Hm, this becomes a bit tricky if we consider that initialization of rwlock can fail (even though I have no idea how would it be possible with how `pthread_rwlock_init` is implemented in e.g. glibc). Handling this "correctly" would require either propagating error from that helper, to `d3d12_device_init_vendor_hacks`, to `vkd3d_create_vk_device`, and add a new late-error code path there… or not bother at all, and add some kind of `bool initialized;` to that struct and check it every time before allowing any DeviceExt function to succeed.

Which one do you prefer? I'm leaning towards `bool initialized;` but it will probably be a little ugly either way.

libs/vkd3d/device_vkd3d_ext.c
```diff
@@ -436,7 +514,106 @@ static HRESULT STDMETHODCALLTYPE d3d12_device_vkd3d_ext_SetAGSUAVSlot(d3d12_devi
...
+    {
+        ERR("Failed to lock mutex, rc %d.\n", rc);
+        return E_FAIL;
+    }
+
+    entry = (void*)hash_map_find(&device->vendor_hacks.nv_shader.map, &key);
+
+    if (entry)
+    {
+        entry->extn.uav_slot = uav_slot;
```
chrisd-work: This is concerning behavior in general. The hash map appears to be under a read lock, and yet it is being modified. If another thread is reading the same key at the same time it may see a torn state for uav_slot / uav_space.

Saancreed: Hmm, you're right, although because the map is keyed on current thread ID (so entries for keys other than `0` are touched only by threads owning them), this is only a problem when modifying the entry under `0`, meaning one thread is currently executing the process-wide version of `SetNvShaderExtnSlotSpace`, and another is compiling shaders, which sounds inherently racey anyway.

But if `SetNvShaderExtnSlotSpace` has some internal synchronization guarantees, I suppose the process-wide configuration can be moved outside the map and writes to it can be switched to be done while holding exclusive lock. I'll address this later today.

libs/vkd3d/device.c
HansKristian-Work: I wouldn't bother with failure here. Just log an error. If it somehow is able to fail that means the process is completely hosed anyway and not even sure SRWLock can fail init on Windows. InitializeSRWLock() returns void on Win32.

libs/vkd3d/device_vkd3d_ext.c
HansKristian-Work: Don't bother with read-write locks here. This is only used when compiling PSOs and that's hardly a hot path.

Saancreed:
Corresponding dxvk-nvapi PR is here: https://github.com/jp7677/dxvk-nvapi/pull/356, hence undrafting…

libs/vkd3d/device_vkd3d_ext.c
```diff
@@ -436,7 +514,93 @@ static HRESULT STDMETHODCALLTYPE d3d12_device_vkd3d_ext_SetAGSUAVSlot(d3d12_devi
...
+{
+    struct d3d12_device *device = d3d12_device_from_ID3D12DeviceExt(iface);
+    TRACE("iface %p, op_code %"PRIu32".\n", iface, op_code);
+
+    if (!device->vendor_hacks.nv_shader.initialized)
+        return FALSE;
+
+    switch (op_code)
+    {
+        case NV_EXTN_OP_HIT_OBJECT_TRACE_RAY:
```
HansKristian-Work, line 528: I don't think dxil-spirv currently supports these though? Best to add these when they're actually supported.

HansKristian-Work (changes requested):
Only concern now is the exposure of lots of opcodes that we actually don't support yet, but otherwise this is good.

libs/vkd3d/device_vkd3d_ext.c
Saancreed: It does: https://github.com/HansKristian-Work/dxil-spirv/blob/65ca7bdf60394edb8777900b166b4d1d07a00235/opcodes/dxil/dxil_nvapi.cpp#L1175

Saancreed:
Together with linked dxvk-nvapi PR and `DXVK_NVAPI_D3D12_NV_SHADER_EXTN=1` in the environment (currently needed for safety) it already brings some improvement, for example in Black Myth Wukong (checked using the benchmark tool).

Before:

Screenshot_20260429_231735_noSER

After:

Screenshot_20260429_231026_SER

libs/vkd3d/device_vkd3d_ext.c
HansKristian-Work: Oh does it now ... ok I guess.

HansKristian-Work (approved):

PR #2971, opened 2026-04-28
vkd3d: Depth attachment resolves happen in COLOR stages and ACCESS.

HansKristian-Work:
Weird, but spec says what it says. Fixes sync-val error.


doitsujin (approved):
right, sigh.

PR #2989, opened 2026-05-06
Descriptor heap - Phase 1 / ????

HansKristian-Work:
It begins. Includes:

- Some test updates to test more paths in legacy vs proper heap. We usually didn't bind descriptor heaps in meta-shader related tests, and this adds test coverage.
- Implements the initial bindless state setup which covers the intended binding model.

There is a dummy config for enabling heap which is purely for stepping through the code with a debugger. It will break in half as advertised if attempted.


HansKristian-Work:
Preemptively locking due to spam on all other heap PRs.

HansKristian-Work:
Apparently Github broke their lock function, yay.

doitsujin:
Review comments also broke for some reason, so....

https://github.com/HansKristian-Work/vkd3d-proton/blob/696109a827d35486594bc2462ae217fa71f8da16/libs/vkd3d/state.c#L7485-L7491

What exactly is the reasoning behind the 2*bufferSize logic here?

Asking because this kind of doesn't really match what we want for texel buffer + ssbo packing (TBO size can be greater than bufferDescriptorSize) or UAV counter packing with all the different special paths involved.

HansKristian-Work:
> What exactly is the reasoning behind the 2*bufferSize logic here?

I noticed the 2 * buffer size being used while rebasing so fixed up that part, but missed this case. Will fix.

doitsujin (approved):

PR #2995, opened 2026-05-07
Descriptor heap - Phase 2

HansKristian-Work:
Focuses on allocating the descriptor heap buffers.
Also adds the interface for a command allocator to temporarily borrow a resource index from the heap for use in e.g. meta shaders, since swapping the heap is a bad time for real this time.

### DIFF
--- a/libs/vkd3d/resource.c
+++ b/libs/vkd3d/resource.c
@@ -215,6 +215,8 @@ HRESULT vkd3d_create_buffer(struct d3d12_device *device,
             | VK_BUFFER_USAGE_TRANSFER_DST_BIT
             | VK_BUFFER_USAGE_UNIFORM_BUFFER_BIT
             | VK_BUFFER_USAGE_STORAGE_BUFFER_BIT
+            | VK_BUFFER_USAGE_UNIFORM_TEXEL_BUFFER_BIT
+            | VK_BUFFER_USAGE_STORAGE_TEXEL_BUFFER_BIT
             | VK_BUFFER_USAGE_INDEX_BUFFER_BIT
             | VK_BUFFER_USAGE_VERTEX_BUFFER_BIT
             | VK_BUFFER_USAGE_INDIRECT_BUFFER_BIT;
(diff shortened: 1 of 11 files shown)


libs/vkd3d/resource.c
```diff
@@ -8201,6 +8201,144 @@ CONST_VTBL struct ID3D12DescriptorHeapVtbl d3d12_descriptor_heap_vtbl =
...
+    else
+    {
+        alloc_size = device->bindless_state.sampler_size;
+        alloc_size *= descriptor_count;
+
+        if (descriptor_heap->desc.Flags & D3D12_DESCRIPTOR_HEAP_FLAG_SHADER_VISIBLE)
+        {
+            alloc_size = align64(alloc_size, device->device_info.descriptor_heap_properties.samplerHeapAlignment);
+            descriptor_heap->descriptor_buffer.reserved_offset = alloc_size;
+            alloc_size += device->device_info.descriptor_heap_properties.minSamplerHeapReservedRangeWithEmbedded;
```
doitsujin, line 8338: Should probably consider large sampler heaps (w/o embedded) here as well since that's a thing now.

Might be worth setting up the descriptor heap limits in `D3D12_OPTIONS19` first and then just validate against those? If we want to report proper limits there we'd have to use more or less the same logic anyway.

libs/vkd3d/resource.c
```diff
@@ -215,6 +215,8 @@ HRESULT vkd3d_create_buffer(struct d3d12_device *device,
             | VK_BUFFER_USAGE_TRANSFER_DST_BIT
             | VK_BUFFER_USAGE_UNIFORM_BUFFER_BIT
             | VK_BUFFER_USAGE_STORAGE_BUFFER_BIT
+            | VK_BUFFER_USAGE_UNIFORM_TEXEL_BUFFER_BIT
+            | VK_BUFFER_USAGE_STORAGE_TEXEL_BUFFER_BIT
```
doitsujin, line 219: Do we still care about legacy at all? If so then we'll have to use `VkBufferUsageFlags2CreateInfoKHR` on buffer view srv/uav creation, otherwise VVL will be unhappy when the app uses a texel buffer format that's only supported for `UNIFORM` but not `STORAGE` (think something like RGB32 on Nvidia), ran into that with DXVK.

libs/vkd3d/resource.c
HansKristian-Work: I don't think we can usefully expose the larger sampler heap because meta shaders may use static samplers. I guess it's technically not impossible to add a reservation interface to sampler descriptor heap too, but it'll massively increase complexity for no good reason.

doitsujin: From what I see it's only the swapchain blit fallback that uses a sampler at all, which uses legacy push descriptors anyway.

Maybe I'm missing something, but I don't really see any internal use case for any kind of samplers besides that.

libs/vkd3d/resource.c
HansKristian-Work: Addressed as a follow-up commit.

libs/vkd3d/resource.c
HansKristian-Work: Added expanded limits when heap and GPU upload heap is supported.

HansKristian-Work: It's used by sampler feedback at least. It'll probably make more sense later as the reviews go through. If we really find a way to drop those, we can maybe relax things.

doitsujin (approved):

PR #2997, opened 2026-05-08
Descriptor heap - Phase 3

HansKristian-Work:
Refactors out image view creation to setup of info structs and the creation. For heap, we will need to only extract the creation info when creating descriptors.

The last commit reworks the meta struct to encode all relevant view info in 8 bytes. This allows us to keep the existing view for legacy paths, but also allow heap path to work well. Technically, the legacy paths could have gone through the re-materialization path, but that's a bit silly and would be a minor perf regression in theory, so don't. There's no reason to squeeze the info struct below 16 bytes since it unions with the buffer one, which is also 16 byte.

### DIFF
--- a/libs/vkd3d/resource.c
+++ b/libs/vkd3d/resource.c
@@ -6452,18 +6452,16 @@ static void vkd3d_texture_view_desc_fixup(struct d3d12_device *device, struct vk
     }
 }
 
-static struct vkd3d_view *vkd3d_create_texture_uav_view(struct d3d12_device *device,
-        struct d3d12_resource *resource, const D3D12_UNORDERED_ACCESS_VIEW_DESC *desc)
+static bool vkd3d_setup_texture_uav_view(struct d3d12_device *device,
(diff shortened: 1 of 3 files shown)


libs/vkd3d/resource.c
```diff
@@ -7257,11 +7257,54 @@ static void vkd3d_create_buffer_uav(vkd3d_cpu_descriptor_va_t desc_va, struct d3
...
+    }
+    else
+    {
+        meta->first_array_slice = vk_subresource_range->baseArrayLayer;
+        meta->array_size = vk_subresource_range->layerCount;
+    }
+
+    meta->vk_view_type = vk_view_type;
+    if (desc && desc->ViewDimension == D3D12_UAV_DIMENSION_TEXTURE2D)
+        meta->plane_slice = desc->Texture2D.PlaneSlice;
```
doitsujin, line 7297: 2D_ARRAY has PlaneSlice too.

HansKristian-Work: Derp. Fixed.

doitsujin (approved):

PR #2998, opened 2026-05-08
libs/vkd3d: optimize memory transfer queue with inline buffers

dt10812:
This pr optimizes the memory transfer queue by adding a small inline buffer (32 entries) to the transfer queue structure, which reduces heap allocation overhead and improves CPU cache locality. 

I added : 

- transfers_inline to struct vkd3d_memory_transfer_queue. 

- implemented vkd3d_transfer_queue_reserve_transfers to handle the transition from inline to heap memory.

- fixed compiler warnings regarding uninitialized variables in command.c and ensure c90 compliance in memory.c

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -6098,7 +6098,8 @@ static void d3d12_command_list_end_rendering(struct d3d12_command_list *list)
     else if (suspend_resume)
     {
         VkResolveModeFlagBits rt_resolves[D3D12_SIMULTANEOUS_RENDER_TARGET_COUNT];
-        VkResolveModeFlagBits depth_resolve, stencil_resolve;
+        VkResolveModeFlagBits depth_resolve = VK_RESOLVE_MODE_NONE;
+        VkResolveModeFlagBits stencil_resolve = VK_RESOLVE_MODE_NONE;
         uint32_t i;
 
         /* We attempted to suspend, but we need to emit some other code, so that's not going to work.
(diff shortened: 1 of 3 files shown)


HansKristian-Work:
What is this even solving? Is there real world data that this is improving anything? Afaik this code is colder than Antarctica in practice.

> ensure c90 compliance in memory.c

Which issue? If it's a real bug it should be a separate PR that only solves that.

mati865:
Yeah... https://github.com/doitsujin/dxvk/pull/5619#issuecomment-4395380568

ITC-linux:
AI slop...

dt10812:
@HansKristian-Work fairpoint on the performance impact. My initial thought was to reduce heap churn during high frequency asset streaming in games, but I understand if the overhead is considered negligible in practice.

For the C90 compliance: I faced mixed declarations and code warnings during MinGW compilation when edited the code so I use the term 'ensure' to have proof that my code is not broken.

Also I swear this is not AI slop, you can see why I'm saying this in [doitsujin/dxvk#5619] (https://github.com/doitsujin/dxvk/pull/5619#issuecomment-4395380568). It's very rude to say that a person, who have a coding style that looks like AI an AI slop that does not understand anything in the code. I coded using this style before I even use AI but everyone keep accusing me and the DXVK situation was not the first time, I faced these before when contributing to Apple's Webkit but they do not care at all but the community keeps accusing me, the Linux kernel is one of the only place where they don't accuse me because they considered my code clean and also, how can AI read multiple files and write correct and precise code, I edited command.c which is a 1MB file and I'm sure AI will forget 50% what that file does. Because Linux kernel community respect me so I keep moving on to other Linux related projects and I faced those accusations again. @HansKristian-Work  I hope you understand me because I faced too much false critics.

llyyr:
I'm fully aware that my response is off topic, and probably a waste of time for anyone to read.

> Linux kernel is one of the only place where they don't accuse me

https://lore.kernel.org/all/?q=ducthinh100812%40gmail.com

The only reason nobody is accusing you of using LLMs in the lkml is because the patches you submit have syntax issues that make it not even compile. I'm impressed you've managed to waste people's time enough to get reviews on your changes that don't even compile while patches that are actually worthwhile continue to rot.

If you're doing this because you want to be helpful, consider triaging bug reports in your favorite project and maybe try to fix a specific bug in a project instead of asking the LLM to "optimize x". I'm sure there's plenty of issue in vkd3d/dxvk's issue trackers that can be reproduced on your hardware. Get the game, reproduce the bug, and try to fix it, if you want to do something helpful. People will be more open to helping you if your PRs are targeted fixes like that.

If you're doing this because you want to pad your resume, consider a new career.

HansKristian-Work:
This is just a complete shitshow so I'll close it.

dt10812:
I've put significant work into optimizing the transfer path and ensuring the c90 warnings. If there areissues with the implementation, i'm happy to address them, but I'd appreciate professional feedback.

PR #3003, opened 2026-05-10
vkd3d: Fix subresource layer range merging

ITC-linux:
d3d12_command_list_update_subresource_data() tracks which image subresource layers have been modified so they can later be copied back to the linear staging buffer.

When a new update overlaps an existing tracked entry, the two layer ranges are merged. The old merge calculation updated baseArrayLayer before computing the end of the merged range, and used subtraction instead of addition when deriving that end.

This could produce an incorrect merged layerCount: either too small, missing layers from the update, or very large due to unsigned underflow. In both cases the tracking list would be wrong, and modified layers could be left out of the copy back to the staging buffer.

The fix computes the merged end layer before overwriting baseArrayLayer, then derives the new layerCount from the merged start.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -4865,6 +4865,7 @@ static void d3d12_command_list_update_subresource_data(struct d3d12_command_list
 {
     struct vkd3d_subresource_tracking *entry;
     VkImageAspectFlags aspect, aspect_mask;
+    uint32_t merged_end;
     uint32_t i;
     bool skip;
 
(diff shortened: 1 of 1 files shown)


doitsujin:
does anything run into this in the real world?

ITC-linux:
I don't have a concrete reproducer. The calculation is provably wrong — it overwrites baseArrayLayer before using it in the same expression, and subtracts layerCount where it should add. The merged range ends up too small or wraps around.
Even if overlapping layer updates are uncommon, vkd3d-proton runs against a large number of D3D12 titles and it's hard to rule out. The fix is trivial and can't regress anything — it only affects the merge path, which already produces the wrong result.

doitsujin:
I mean, yes, but it's always nice to know if there's any specific movitation for any given fix other than "someone / some LLM looked at the code and found a bug" so we know if there's a need to backport it etc. If it's just the latter then fair enough.

ITC-linux:
No specific title — just me staring at the function until the subtraction stopped making sense. Consider it a free QA pass.

HansKristian-Work (approved):
LGTM. I cannot make sense of the previous logic either and the new code makes sense. Also ye, seconding that it's useful to know if a bug was found out of thin air or if it caught in the wild in general.

PR #3005, opened 2026-05-11
Support for VK_KHR_opacity_micromap

chrisd-work:
I have done my best to split the PR up into smaller commits. Each commit, has been verified to be buildable on a Linux box.

This lands many things related to Opacity Micromaps.

- Support for VK_KHR_opacity_micromap
- DXR 1.2 support
- NVAPI support for OMM (adapted from work by Krzysztof Bogacki krzysztof.bogacki@leancode.pl; Co-authored-by trailers on the 16 NVAPI commits reflect this)
* Cleanup

To test this you will need the following:

1. Testing this requires a driver that supports VK_KHR_opacity_micromap. NVIDIA provides a public beta driver https://developer.nvidia.com/vulkan-driver that supports this feature.
2. https://github.com/chrisd-work/dxil-spirv -- not ready for a PR, as the required submodules have not been merged to the main line.

Games tested:

- Black Myth: Wukong
- Cyberpunk 2077

A note on validation: the supporting VVL changes are still pending, and I have filed several issues against VVL upstream. Not only will VVL not function fully against this PR in its current state -- it is also unstable. Separately, VVL has become much more aggressive about overlapping-address checks. I am prototyping a feature that issues a second prebuild-size query so the reported ranges line up and those messages stop firing. That feature needs to be opt-in rather than always on, since the extra prebuild costs performance. The opt-in should also be independent of whether VVL itself is enabled, so all four combinations (feature on/off x VVL on/off) can be exercised.

In the interest of getting the PR in for review I have submitted it as the review will likely take some time due to the complexity. I am willing to split it into smaller PRs as well -- the fact that the PR is already split into 56 commits makes that possible, though reordering can be difficult.


HansKristian-Work:
This is a lot of stuff to digest. I would suggest reviewing and merging this in multiple as we do with massive PRs like this. Please take the first 4 commits and make a Phase 1 PR out of it and we can take it from there.

chrisd-work:
> This is a lot of stuff to digest. I would suggest reviewing and merging this in multiple as we do with massive PRs like this. Please take the first 4 commits and make a Phase 1 PR out of it and we can take it from there.

Yeah, I think we can do that at the start. It may get difficult in some parts to keep it down to 4, but since they are all buildable and I did my best to ensure that at any given point they shouldn't regress anything (bisectable), it should be possible.

I will keep this PR updated as we go through. It will become a way for us to track progress and see how much is left to merge before the feature is fully in.

include/vkd3d_device_vkd3d_ext.idl
```diff
@@ -163,6 +163,21 @@ interface ID3D12DeviceExt4 : ID3D12DeviceExt3
...
+    object,
+    local,
+    pointer_default(unique)
+]
+interface ID3D12DeviceExt5 : ID3D12DeviceExt4
+{
+    HRESULT SetCreatePipelineStateOptions(const void *params);
+    HRESULT CheckDriverMatchingIdentifierEx(void *params);
+    HRESULT GetRaytracingAccelerationStructurePrebuildInfoEx(void *params);
+    HRESULT GetRaytracingOpacityMicromapArrayPrebuildInfo(void *params);
```
Saancreed: While it was very convenient for prototyping to avoid using proper types, I have my doubts if `HRESULT Function(void *params);` is good enough for upstream, when it's really something closer to `NvAPI_Status Function(NVAPI_FUNCTION_PARAMS *params);`.

Would it be possible to maybe forward-declare necessary types?

Also, perhaps now some of these could be removed from public API surface and dxvk-nvapi would call core D3D12 methods instead of NVAPI ones? At least `SetCreatePipelineStateOptions` would have to stay, but `CheckDriverMatchingIdentifierEx` looks like something that can be trivially implemented on top of `ID3D12Device5::CheckDriverMatchingIdentifier` by mapping both `NVAPI_D3D12_SERIALIZED_DATA_RAYTRACING_ACCELERATION_STRUCTURE_EX` and `NVAPI_D3D12_SERIALIZED_DATA_RAYTRACING_OPACITY_MICROMAP_ARRAY_EX` to `D3D12_SERIALIZED_DATA_RAYTRACING_ACCELERATION_STRUCTURE`. Maybe the same applies to other methods?

libs/vkd3d/acceleration_structure.c
```diff
@@ -47,6 +47,34 @@ static VkBuildAccelerationStructureFlagsKHR d3d12_build_flags_to_vk(
...
+        vk_flags |= VK_BUILD_ACCELERATION_STRUCTURE_PREFER_FAST_BUILD_BIT_KHR;
+    if (flags & NVAPI_D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_PREFER_FAST_TRACE_EX)
+        vk_flags |= VK_BUILD_ACCELERATION_STRUCTURE_PREFER_FAST_TRACE_BIT_KHR;
+    if (flags & NVAPI_D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_ALLOW_OMM_UPDATE_EX)
+        vk_flags |= VK_BUILD_ACCELERATION_STRUCTURE_ALLOW_OPACITY_MICROMAP_UPDATE_BIT_KHR;
+    if (flags & NVAPI_D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAG_ALLOW_OMM_OPACITY_STATES_UPDATE_EX)
+    {
+        /* KHR collapsed the data-only-update path into the general OMM update bit; widen rather than drop the flag. */
+        vk_flags |= VK_BUILD_ACCELERATION_STRUCTURE_ALLOW_OPACITY_MICROMAP_UPDATE_BIT_KHR;
+    }
```
Saancreed: The commit message says

```
NVAPI's flag values overlap with D3D12's at most bit positions but
diverge at bit 6: NVAPI splits OMM update into linkage
(ALLOW_OMM_UPDATE_EX) and data (ALLOW_OMM_OPACITY_STATES_UPDATE_EX),
while D3D12 collapsed them into one flag (ALLOW_OMM_LINKAGE_UPDATE)
that maps to both Vulkan UPDATE bits. nvapi_build_flags_to_vk maps
each NVAPI flag explicitly rather than delegating to
d3d12_build_flags_to_vk to avoid the D3D12 helper's conservative
bit-6 expansion overriding the NVAPI app's narrower intent.
```

but in KHR there's only one bit, so we wouldn't actually lose anything by reusing `d3d12_build_flags_to_vk`, right?

include/vkd3d_command_list_vkd3d_ext.idl
```diff
@@ -40,3 +40,18 @@ interface ID3D12GraphicsCommandListExt1 : ID3D12GraphicsCommandListExt
 {
    HRESULT LaunchCubinShaderEx(D3D12_CUBIN_DATA_HANDLE *handle, UINT32 block_x, UINT32 block_y, UINT32 block_z, UINT32 smem_size, const void *params, UINT32 param_size, const void *raw_params, UINT32 raw_params_count);
 }
+
+[
+    uuid(d53b0028-afb4-4b65-a4f1-7b0daaa65b50),
```
Saancreed: This UUID should be regenerated to be another random UUIDv4; current value that I picked is just `uuidof(ID3D12GraphicsCommandListExt1)` incremented by one.

include/vkd3d_command_list_vkd3d_ext.idl
```diff
@@ -40,3 +40,18 @@ interface ID3D12GraphicsCommandListExt1 : ID3D12GraphicsCommandListExt
...
+    object,
+    local,
+    pointer_default(unique)
+]
+interface ID3D12GraphicsCommandListExt2 : ID3D12GraphicsCommandListExt1
+{
+   HRESULT BuildRaytracingAccelerationStructureEx(const void *params);
+   HRESULT BuildRaytracingOpacityMicromapArray(void *params);
+   HRESULT RelocateRaytracingOpacityMicromapArray(const void *params);
+   HRESULT EmitRaytracingOpacityMicromapArrayPostbuildInfo(const void *params);
```
Saancreed: Same as in device interface, maybe we don't actually need all of these…

include/vkd3d_device_vkd3d_ext.idl
chrisd-work: Agreed on the signatures -- I'll try to rework these to use `NvAPI_Status` and typed `NVAPI_*_PARAMS` via forward-declared types. The signatures should match NVAPI's published ABI exactly to avoid confusion.

On routing through core D3D12 methods:

- `CheckDriverMatchingIdentifierEx`: agreed, this one is genuinely a thin wrapper -- both EX serialized-data values can map to `D3D12_SERIALIZED_DATA_RAYTRACING_ACCELERATION_STRUCTURE` since the Vulkan-side query handles both AS kinds. Will land this.

- `EmitRaytracingOpacityMicromapArrayPostbuildInfo`: worth a look. The driver dispatches OMM-Array postbuild via the BLAS enum 1:1, so routing through `EmitRaytracingAccelerationStructurePostbuildInfo` is plausible without conversion overhead. Will audit.

- `BuildRaytracingAccelerationStructureEx` and the prebuild-info getters: this is where the duplication-vs-conversion trade-off bites. The NVAPI inputs carry fields D3D12's plain methods don't accept (OMM usage counts in geometry descs, OMM linkage on triangle inputs), so the conversion is non-trivial -- you'd either duplicate the builder (current shape) or have a NVAPI->D3D12 inputs translator plus the D3D12 builder, which needs temporary geometry-desc arrays in the worst case. I shared as much code between the two AS builders as I could without conflating their input shapes. Open to a deeper look if you see a cleaner factoring.

- `SetCreatePipelineStateOptions`: stays public per your earlier note.

include/vkd3d_command_list_vkd3d_ext.idl
chrisd-work: Agreed, I will do this.

libs/vkd3d/acceleration_structure.c
chrisd-work: You're right. This is a legacy of the EXT->KHR mapping. I will clean it up.

include/vkd3d_device_vkd3d_ext.idl
chrisd-work: I messed with `BuildRaytracingAccelerationStructureEx` and kin this evening, I think it might be easier push this into `dxvk-nvapi`. In addition, it may prevent having to push `nvapi.h` into vkd3d-proton. I will keep hacking at it.

HansKristian-Work:
I guess next steps is to rebase this and throw 4-5 commits or so into phase 2. The exact number of commits isn't fixed, just stop when some minor milestone has been reached in the series, like making the EXT/KHR unions. Although, adding tons of scaffolding in EXT just to remove it again right after will make this much harder to review than it needs to be. If it makes sense to squash some things to avoid that, that would help too I guess.

include/vkd3d_device_vkd3d_ext.idl
```diff
@@ -163,6 +163,21 @@ interface ID3D12DeviceExt4 : ID3D12DeviceExt3
...
 
+[
+    uuid(e51323f6-3541-4ddf-894d-8a8c70e3e427),
+    object,
+    local,
+    pointer_default(unique)
+]
+interface ID3D12DeviceExt5 : ID3D12DeviceExt4
+{
+    BOOL SetOpacityMicromapOverrideNVAPI(BOOL enable_opacity_micromap);
```
Saancreed: I think something halfway between the old version that accepted the whole versioned struct, and a new one with just a bool, like the following, would be best:

```cpp
typedef enum D3D12_NV_PIPELINE_CREATION_STATE_FLAGS
{
    D3D12_NV_PIPELINE_CREATION_STATE_FLAGS_NONE               = 0x0,
    D3D12_NV_PIPELINE_CREATION_STATE_FLAGS_ENABLE_OMM_SUPPORT = 0x1,
} D3D12_NV_PIPELINE_CREATION_STATE_FLAGS;

HRESULT SetCreatePipelineStateFlags(D3D12_NV_PIPELINE_CREATION_STATE_FLAGS flags);
```

This way we can reuse this entrypoint if we ever decide to extend this for Cluster AS or LSS support…

include/vkd3d_device_vkd3d_ext.idl
```diff
@@ -163,6 +163,21 @@ interface ID3D12DeviceExt4 : ID3D12DeviceExt3
...
+    void GetRaytracingAccelerationStructurePrebuildInfoNVAPI(
+        const D3D12_BUILD_RAYTRACING_ACCELERATION_STRUCTURE_INPUTS *desc,
+        D3D12_RAYTRACING_ACCELERATION_STRUCTURE_PREBUILD_INFO *info);
```
Saancreed: If the parameters are now exactly the same as in core D3D12 API now, do we even need this to exist? dxvk-nvapi could just call the method from `ID3D12Device5`.

The only thing that's different between the APIs is the source of `bool supports_opacity_micromap`, which in `ID3D12Device5` checks for more strict DXR 1.2 instead of just OMM extension, but perhaps we could just use only OMM extension check in both paths? This doesn't work as expected only with applications that attempt to use D3D12 OMM without checking for DXR 1.2 first, but that's an application bug and I'm not sure if we have to add guards against this…

Same applies to `BuildRaytracingAccelerationStructureNVAPI` and `EmitRaytracingAccelerationStructurePostbuildInfoNVAPI`.

include/vkd3d_command_list_vkd3d_ext.idl
```diff
@@ -40,3 +40,25 @@ interface ID3D12GraphicsCommandListExt1 : ID3D12GraphicsCommandListExt
...
+        const D3D12_BUILD_RAYTRACING_ACCELERATION_STRUCTURE_DESC *pDesc,
+        UINT NumPostbuildInfoDescs,
+        const D3D12_RAYTRACING_ACCELERATION_STRUCTURE_POSTBUILD_INFO_DESC *pPostbuildInfoDescs);
+
+    void EmitRaytracingAccelerationStructurePostbuildInfoNVAPI(
+        const D3D12_RAYTRACING_ACCELERATION_STRUCTURE_POSTBUILD_INFO_DESC* pDesc,
+        UINT NumSourceAccelerationStructures,
+        const D3D12_GPU_VIRTUAL_ADDRESS* pSourceAccelerationStructureData);
+ 
+    BOOL VerifyOpacityMicromapArrayNVAPI(D3D12_GPU_VIRTUAL_ADDRESS opacity_micromap_array);
```
Saancreed, line 52: It's unfortunate, but the API this attempts to cover (`NvAPI_D3D12_RelocateRaytracingOpacityMicromapArray`) is likely not something we'll be ever able to support in Vulkan without some kind of cursed `vkCmdRelocateCopiedAccelerationStructureNV` command. But it's good that nothing appears to rely on it in practice.

But I also don't have any idea about how to communicate this any better, so I guess it should stay as is. Maybe I'd change the return type to `HRESULT` in case we'd want to explicitly handle "micromaps not supported" case or something else, but no strong feelings about this.

That said, if this is unlikely to ever record any work on the command buffer and only ever does synchronous work, then maybe it should be converted to a method on `ID3D12DeviceExt5` for simplicity?

include/vkd3d_vk_includes.h
```diff
@@ -41,6 +41,7 @@ typedef enum VkImageLayout VkImageLayout;
 
 typedef enum D3D12_VK_EXTENSION
 {
+    D3D12_VK_EXT_OPACITY_MICROMAP   = 0x0,
```
Saancreed, line 44: Nitpicking, but with EXT_omm backend removed, this can be renamed to `D3D12_VK_KHR_OPACITY_MICROMAP` for accuracy.

chrisd-work:
Closing as other commits cover this. I am putting all commits here include cross repo commits in case anyone wants to look in the future:

VKD3D-proton:
Phase 1 - https://github.com/HansKristian-Work/vkd3d-proton/pull/3010
Phase 2 - https://github.com/HansKristian-Work/vkd3d-proton/pull/3038
Phase 3 - https://github.com/HansKristian-Work/vkd3d-proton/pull/3042
Phase 4 - https://github.com/HansKristian-Work/vkd3d-proton/pull/3052
Phase 5 - https://github.com/HansKristian-Work/vkd3d-proton/pull/3069
Phase 6 - https://github.com/HansKristian-Work/vkd3d-proton/pull/3071
Phase 7 - https://github.com/HansKristian-Work/vkd3d-proton/pull/3075
Phase 8 - https://github.com/HansKristian-Work/vkd3d-proton/pull/3084
Phase 9 - https://github.com/HansKristian-Work/vkd3d-proton/pull/3118
Phase 10 - https://github.com/HansKristian-Work/vkd3d-proton/pull/3119

dxil-spriv 
https://github.com/HansKristian-Work/dxil-spirv/pull/283

dxvk-nvapi
https://github.com/jp7677/dxvk-nvapi/pull/371

PR #3007, opened 2026-05-11
Rewrite how VKD3D_CONFIG flags are maintained

HansKristian-Work:
We exhausted 64-bits, so move to an expandable bitfield structure.


doitsujin (approved):
pain

HansKristian-Work:
Saw some warnings in CI about trying to print the u64 flags, so fixed that trivially.

PR #3008, opened 2026-05-11
Descriptor heap - Phase 4

HansKristian-Work:
This series focuses on creating descriptors on the heap. It's very similar to descriptor buffer embedded path.

Major difference is some details around how texel buffer <-> SSBO aliasing is done. We can support SSBO alignment > 4 here by dynamically falling back to texel buffer based on the stride. The shader mapping deals with this difference later. StructuredBuffer and BAB cannot reliably alias each other on native drivers, so it's probably fine. It might help enable heap on more drivers down the line.

Another difference is that we finally emit "proper" RTAS descriptors now, so no weird jank with embedded raw VAs, etc.

Samplers hit different paths for custom border colors as with heap we need explicit reservation.

Otherwise it should be fairly straight forward.

### DIFF
--- a/libs/vkd3d/resource.c
+++ b/libs/vkd3d/resource.c
@@ -8304,53 +8304,55 @@ HRESULT d3d12_create_static_sampler(struct d3d12_device *device,
     return hresult_from_vk_result(vr);
 }
 
-static HRESULT d3d12_create_sampler(struct d3d12_device *device,
-        const D3D12_SAMPLER_DESC2 *desc, VkSampler *vk_sampler)
+struct vkd3d_sampler_view_create_info
 {
-    const struct vkd3d_vk_device_procs *vk_procs = &device->vk_procs;
     VkSamplerCustomBorderColorCreateInfoEXT border_color_info;
+    VkSamplerCustomBorderColorIndexCreateInfoEXT index_create_info;
     VkSamplerReductionModeCreateInfoEXT reduction_desc;
     VkSamplerCreateInfo sampler_desc;
+};
+
+static void d3d12_setup_sampler_info(struct d3d12_device *device,
+        const D3D12_SAMPLER_DESC2 *desc, struct vkd3d_sampler_view_create_info *info)
+{
     uint32_t num_live_objects;
-    VkResult vr;
 
-    border_color_info.sType = VK_STRUCTURE_TYPE_SAMPLER_CUSTOM_BORDER_COLOR_CREATE_INFO_EXT;
-    border_color_info.pNext = NULL;
-    memcpy(border_color_info.customBorderColor.uint32, desc->UintBorderColor,
-            sizeof(border_color_info.customBorderColor.uint32));
-    border_color_info.format = VK_FORMAT_UNDEFINED;
+    memset(info, 0, sizeof(*info));
 
-    reduction_desc.sType = VK_STRUCTURE_TYPE_SAMPLER_REDUCTION_MODE_CREATE_INFO_EXT;
-    reduction_desc.pNext = NULL;
-    reduction_desc.reductionMode = vk_reduction_mode_from_d3d12(D3D12_DECODE_FILTER_REDUCTION(desc->Filter));
+    info->border_color_info.sType = VK_STRUCTURE_TYPE_SAMPLER_CUSTOM_BORDER_COLOR_CREATE_INFO_EXT;
+    memcpy(info->border_color_info.customBorderColor.uint32, desc->UintBorderColor,
+            sizeof(info->border_color_info.customBorderColor.uint32));
+    info->border_color_info.format = VK_FORMAT_UNDEFINED;
 
-    sampler_desc.sType = VK_STRUCTURE_TYPE_SAMPLER_CREATE_INFO;
-    sampler_desc.pNext = NULL;
-    sampler_desc.flags = 0;
-    sampler_desc.magFilter = vk_filter_from_d3d12(D3D12_DECODE_MAG_FILTER(desc->Filter));
-    sampler_desc.minFilter = vk_filter_from_d3d12(D3D12_DECODE_MIN_FILTER(desc->Filter));
-    sampler_desc.mipmapMode = vk_mipmap_mode_from_d3d12(D3D12_DECODE_MIP_FILTER(desc->Filter));
-    sampler_desc.addressModeU = vk_address_mode_from_d3d12(desc->AddressU);
-    sampler_desc.addressModeV = vk_address_mode_from_d3d12(desc->AddressV);
-    sampler_desc.addressModeW = vk_address_mode_from_d3d12(desc->AddressW);
-    sampler_desc.mipLodBias = desc->MipLODBias;
-    sampler_desc.anisotropyEnable = D3D12_DECODE_IS_ANISOTROPIC_FILTER(desc->Filter);
-    sampler_desc.maxAnisotropy = desc->MaxAnisotropy;
-    sampler_desc.compareEnable = D3D12_DECODE_IS_COMPARISON_FILTER(desc->Filter);
-    sampler_desc.compareOp = sampler_desc.compareEnable ? vk_compare_op_from_d3d12(desc->ComparisonFunc) : 0;
-    sampler_desc.minLod = desc->MinLOD;
-    sampler_desc.maxLod = desc->MaxLOD;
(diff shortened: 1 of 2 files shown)


libs/vkd3d/resource.c
```diff
@@ -1719,9 +1726,67 @@ void vkd3d_sampler_state_cleanup(struct vkd3d_sampler_state *state,
...
+    }
+
+    if (VK_CALL(vkRegisterCustomBorderColorEXT(device->vk_device, info, VK_FALSE, &i)) != VK_SUCCESS)
+    {
+        ERR("Failed to allocate custom border color index.\n");
+        i = UINT32_MAX;
+    }
+
+    /* Some drivers simply do not care about custom border colors and will just return the same value indefinitely.
+     * If we detect that drivers don't care, just skip the registration in the future. */
```
doitsujin: does this still happen on current drivers? I don't remember the discussion exactly but something about this screams "out of spec".

libs/vkd3d/resource.c
```diff
@@ -1719,9 +1726,67 @@ void vkd3d_sampler_state_cleanup(struct vkd3d_sampler_state *state,
...
+        {
+            i = state->border_colors[i].index;
+            goto unlock;
+        }
+    }
+
+    if (state->border_color_count == state->border_color_bank_size)
+    {
+        i = UINT32_MAX;
+        goto unlock;
```
doitsujin, line 1762: Think we should really have a `FIXME` here if we can run just out of available border colors.

Real apps probably won't run into this since this isn't d3d9 anymore, but at least in theory this is easy to break.

libs/vkd3d/resource.c
```diff
@@ -6226,6 +6226,168 @@ static bool vkd3d_buffer_view_get_aligned_view(
...
+        const D3D12_SHADER_RESOURCE_VIEW_DESC *desc)
+{
+    const struct vkd3d_vk_device_procs *vk_procs = &device->vk_procs;
+    struct vkd3d_descriptor_metadata_buffer_view view;
+    VkTexelBufferDescriptorInfoEXT texel_buffer_info;
+    VkResourceDescriptorInfoEXT desc_info;
+    struct d3d12_desc_split_embedded d;
+    VkDeviceAddressRangeEXT ssbo_range;
+    VkHostAddressRangeEXT desc_range;
+    uint8_t stack_payload[256];
```
doitsujin: this could probably do with a `VKD3D_MAX_DESCRIPTOR_SIZE` define or something

libs/vkd3d/resource.c
```diff
@@ -5854,6 +5854,57 @@ static void d3d12_descriptor_heap_write_null_descriptor_template(vkd3d_cpu_descr
...
+    desc_range.address = d.payload + device->bindless_state.packed_raw_buffer_offset;
+    desc_range.size = device->bindless_state.heap.ubo_size;
+    VK_CALL(vkWriteResourceDescriptorsEXT(device->vk_device, 1, &desc_info, &desc_range));
+
+    if (device->bindless_state.packed_raw_buffer_offset == 0 &&
+        device->bindless_state.heap.ubo_size < device->device_info.descriptor_heap_properties.bufferDescriptorSize)
+    {
+        uint8_t *padding = d.payload + device->bindless_state.packed_raw_buffer_offset;
+        padding += device->bindless_state.heap.ubo_size;
+        memset(padding, 0, device->device_info.descriptor_heap_properties.bufferDescriptorSize - device->bindless_state.heap.ubo_size);
```
doitsujin: Do we know how this performs vs native when writing descriptors directly into a GPU-visible heap?

Not sure if that's something games commonly do, but I wonder if there's a potential perf problem here with three different things writing write-combined memory in different ways, and whether we should be doing the stack payload thing from the srv buffer path everywhere instead.

libs/vkd3d/resource.c
```diff
@@ -8379,51 +8381,88 @@ static HRESULT d3d12_create_sampler(struct d3d12_device *device,
     if (num_live_objects > device->device_info.properties2.properties.limits.maxSamplerAllocationCount)
         FIXME_ONCE("Trying to create a sampler, but device limits are exhausted. Creation may fail.\n");
 
-    if (sampler_desc.borderColor == VK_BORDER_COLOR_FLOAT_CUSTOM_EXT ||
-            sampler_desc.borderColor == VK_BORDER_COLOR_INT_CUSTOM_EXT)
+    if (info->sampler_desc.borderColor == VK_BORDER_COLOR_FLOAT_CUSTOM_EXT ||
+            info->sampler_desc.borderColor == VK_BORDER_COLOR_INT_CUSTOM_EXT)
     {
-        uint32_t num_border_colors;
+        uint32_t num_legacy_border_colors = 0;
```
doitsujin, line 8375: I guess this indirectly addresses my `FIXME` complaint from above, but setting something called `legacy` on the heap path seems a bit hacky, could probably just be an out-of-bounds `bool` or something to indicate that we need to fall back?

libs/vkd3d/resource.c
```diff
@@ -7264,6 +7264,159 @@ VkDeviceAddress vkd3d_get_acceleration_structure_device_address(struct d3d12_dev
...
+    const struct vkd3d_bindless_state *bindless = &device->bindless_state;
+    const struct vkd3d_vk_device_procs *vk_procs = &device->vk_procs;
+    struct vkd3d_descriptor_metadata_buffer_view view;
+    VkTexelBufferDescriptorInfoEXT texel_buffer_info;
+    VkResourceDescriptorInfoEXT desc_info;
+    struct d3d12_desc_split_embedded d;
+    VkDeviceAddressRangeEXT ssbo_range;
+    VkHostAddressRangeEXT desc_range;
+    bool can_emit_sibling_typed;
+    uint8_t stack_payload[256];
```
doitsujin: dito wrt having a named define

libs/vkd3d/resource.c
```diff
@@ -7264,6 +7264,159 @@ VkDeviceAddress vkd3d_get_acceleration_structure_device_address(struct d3d12_dev
...
+        desc_range.size = device->device_info.descriptor_heap_properties.bufferDescriptorSize;
+
+        if (bindless->heap.uav_counter_embedded_offset == 0)
+        {
+            VK_CALL(vkWriteResourceDescriptorsEXT(device->vk_device, 1, &desc_info, &desc_range));
+        }
+        else
+        {
+            /* Ugly hackery. Place UAV counter BDA at end of payload. */
+            memcpy(stack_payload + bindless->heap.uav_counter_embedded_offset,
```
doitsujin, line 7400: more of a question for curiosity's sake right now, will this use `INDIRECT_ADDRESS` mappings or just load bda from heap memory in shader?

libs/vkd3d/resource.c
doitsujin: nvm, see below.

libs/vkd3d/resource.c
HansKristian-Work: Rodrigo suggested this, so I would think so.

HansKristian-Work: Tested, it does trip on latest NV beta. I'm not sure if this is out of spec or not ...

libs/vkd3d/resource.c
HansKristian-Work: The stack payload was suggested by Rodrigo. I guess I can move over to use that everywhere for completeness.

libs/vkd3d/resource.c
HansKristian-Work: True. Will fix.

libs/vkd3d/resource.c
HansKristian-Work: INDIRECT_ADDRESS mapping is not possible since it doesn't support arrays of descriptors. The plan is to use the redzone descriptor to read from the heap as an SSBO like we do already for DB path.

When root signatures have room, there is a path for INLINE redzone which lets us push heap size (for DESCRIPTOR_HEAP_ROBUSTNESS) + heap VA into push data and map the BDA to a UBO. dxil-spirv already has paths for this so it's just a matter of a little plumbing.

libs/vkd3d/resource.c
HansKristian-Work: Ye ...

libs/vkd3d/resource.c
HansKristian-Work: Added more comments.

doitsujin (approved):
should be good once CI passes.

PR #3010, opened 2026-05-12
Support VK_KHR_opacity_micromap - Phase 1

chrisd-work:
First of a series splitting the original KHR_OMM landing into smaller reviewable PRs. This one is the foundation: detect `VK_KHR_device_address_commands` and route acceleration structure view creation through `vkCreateAccelerationStructure2KHR` when the feature is enabled. v1 remains the fallback for older drivers.


libs/vkd3d/resource.c
```diff
@@ -5111,23 +5111,47 @@ bool vkd3d_create_acceleration_structure_view(struct d3d12_device *device, const
...
 
-    vr = VK_CALL(vkCreateAccelerationStructureKHR(device->vk_device, &create_info, NULL, &vk_acceleration_structure));
-    if (vr != VK_SUCCESS)
+    if (device->device_info.device_address_commands_features.deviceAddressCommands)
+    {
+        memset(&create_info_v2, 0, sizeof(create_info_v2));
+
+        create_info_v2.sType = VK_STRUCTURE_TYPE_ACCELERATION_STRUCTURE_CREATE_INFO_2_KHR;
+        create_info_v2.type = VK_ACCELERATION_STRUCTURE_TYPE_GENERIC_KHR;
+        create_info_v2.addressFlags = VK_ADDRESS_COMMAND_FULLY_BOUND_BIT_KHR
```
HansKristian-Work, line 5133: Is this necessarily true? What's stopping an application from using a sparse RTAS? It only says the memory is fully committed when accessed which I guess makes sense. PRT page for RTAS doesn't make too much sense.

libs/vkd3d/resource.c
```diff
@@ -5111,23 +5111,47 @@ bool vkd3d_create_acceleration_structure_view(struct d3d12_device *device, const
 {
     const struct vkd3d_vk_device_procs *vk_procs = &device->vk_procs;
     VkAccelerationStructureKHR vk_acceleration_structure;
-    VkAccelerationStructureCreateInfoKHR create_info;
-    VkDeviceAddress buffer_address;
-    VkDeviceAddress rtas_address;
+    VkAccelerationStructureCreateInfoKHR create_info_v1;
+    VkAccelerationStructureCreateInfo2KHR create_info_v2;
```
HansKristian-Work, line 5114: inverse🎄 -tree rule. Declarations are ordered in inverse length, so v2 line comes before v1, very minor nit but will likely come up later.

libs/vkd3d/resource.c
HansKristian-Work: Also, I think you're missing VK_ADDRESS_COMMAND_TRANSFORM_FEEDBACK_BUFFER_USAGE_BIT_KHR here. All VkBuffers should be created with that.

HansKristian-Work: Derp. I can't read.

HansKristian-Work:
This PR is triggering segfaults in RADV, that needs some digging.

HansKristian-Work:
Btw, it's probably unrelated to this PR, but the latest NV beta driver seems to have a bug where vkCreateQueryPool with type `VK_QUERY_TYPE_MICROMAP_COMPACTED_SIZE_EXT` fails with VK_ERROR_UNKNOWN. Happens when running test suite with VKD3D_CONFIG=dxr12.

Saancreed:
> Btw, it's probably unrelated to this PR, but the latest NV beta driver seems to have a bug where vkCreateQueryPool with type `VK_QUERY_TYPE_MICROMAP_COMPACTED_SIZE_EXT` fails with VK_ERROR_UNKNOWN. Happens when running test suite with VKD3D_CONFIG=dxr12.

Fwiw I reported this back in June last year (2025-06-14) and got a reply from NV engineer that he has a fix in flight that will take a while to get released… apparently it never did. But I suppose it's less of a problem if we aim to completely replace EXT with KHR ext anyway.

HansKristian-Work (approved):

chrisd-work:
@HansKristian-Work I will have to try the beta branch, since it wasn't released until last week I was using an internal version for testing. I will determine if it is a bug in the beta branch or TOT of our driver. If it is on TOT of our driver I will report this issue to the Vulkan team internally. Thanks for letting me know.

In terms of the inverse tree rule. I will make sure all future commits are adjusted to follow it.

PR #3013, opened 2026-05-12
Descriptor heap - Phase 5

HansKristian-Work:
First, sets up the optimized vtables for copying descriptors around. Similar idea to the things we have already, except expanded for more cases. (C++ templates sure would help here ...).

Afterwards, the process to implement ID3D12RootSignature is begun with the simpler parts to warm up for Phase 6.

The root signature implementation involves building up a DescriptorSetAndBinding mapping create info struct rather than creating a pipeline layout which is the biggest win of descriptor heap from a usability PoV.

The heap redzone layout is also introduced here.

### DIFF
--- a/libs/vkd3d/state.c
+++ b/libs/vkd3d/state.c
@@ -1261,31 +1261,66 @@ static HRESULT d3d12_root_signature_init_static_samplers(struct d3d12_root_signa
         const D3D12_ROOT_SIGNATURE_DESC2 *desc, struct vkd3d_descriptor_set_context *context,
         VkDescriptorSetLayout *vk_set_layout)
 {
-    VkDescriptorSetLayoutBinding *vk_binding_info, *vk_binding;
+    bool heap = d3d12_device_use_descriptor_heap(root_signature->device);
+    VkDescriptorSetLayoutBinding *vk_binding_info = NULL;
+    VkDescriptorSetAndBindingMappingEXT *vk_mapping;
     struct vkd3d_shader_resource_binding *binding;
+    VkDescriptorSetLayoutBinding *vk_binding;
     unsigned int i;
     HRESULT hr;
 
     if (!desc->NumStaticSamplers)
         return S_OK;
 
-    if (!(vk_binding_info = vkd3d_malloc(desc->NumStaticSamplers * sizeof(*vk_binding_info))))
+    if (!heap && !(vk_binding_info = vkd3d_malloc(desc->NumStaticSamplers * sizeof(*vk_binding_info))))
         return E_OUTOFMEMORY;
+
+    if (heap)
+    {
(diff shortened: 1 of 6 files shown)


libs/vkd3d/state.c
```diff
@@ -1399,6 +1401,141 @@ static void d3d12_root_signature_update_bind_point_layout(struct d3d12_bind_poin
...
+            mapping->bindingCount = 1;
+            mapping->sourceData.pushDataOffset = root_signature->heap.redzone_inline_heap_offset;
+
+            mapping = &root_signature->heap.mappings[root_signature->heap.mappings_count++];
+            memset(mapping, 0, sizeof(*mapping));
+            mapping->sType = VK_STRUCTURE_TYPE_DESCRIPTOR_SET_AND_BINDING_MAPPING_EXT;
+            mapping->resourceMask = VK_SPIRV_RESOURCE_TYPE_UNIFORM_BUFFER_BIT_EXT;
+            mapping->source = VK_DESCRIPTOR_MAPPING_SOURCE_PUSH_DATA_EXT;
+            mapping->descriptorSet = VKD3D_SHADER_GLOBAL_HEAP_VIRTUAL_DESCRIPTOR_SET;
+            mapping->firstBinding = VKD3D_SHADER_GLOBAL_HEAP_SIZE_BINDING;
```
doitsujin, line 1566: i suppose this is a single-dword pseudo-UBO?

libs/vkd3d/vkd3d_private.h
```diff
@@ -1999,6 +2006,24 @@ struct d3d12_root_signature
...
+        VkShaderDescriptorSetAndBindingMappingInfoEXT mapping_info;
+        VkDescriptorSetAndBindingMappingEXT *mappings;
+        size_t mappings_size;
+        size_t mappings_count;
+
+        struct
+        {
+            VkSamplerCreateInfo desc;
+            VkSamplerReductionModeCreateInfoEXT reduction;
+        } *vk_static_samplers_desc;
```
doitsujin: might as well make this a named struct and pass it to `setup_static_sampler_info` directly

doitsujin (commented):
cosmetic nit, unlike with EDB I actually understand what's going on here

libs/vkd3d/state.c
HansKristian-Work: Right, the idea is that push data is mapped to a very simple UBO that's basically:

```
layout(set = VIRTUAL_SET, binding = HEAP_SIZE_BINDING) uniform Size { uint count; };
```

and a plain BDA for the heap VA in GLOBAL_HEAP_BINDING.

libs/vkd3d/vkd3d_private.h
HansKristian-Work: I guess.

doitsujin (approved):

PR #3015, opened 2026-05-12
NULL out obviously bogus RTV/DSV formats.

### LINKED ISSUE #3014: VVL error with Hunt: Showdown 1896
The game is creating a graphics pipeline with VK_FORMAT_BC6H_UFLOAT_BLOCK as color attachment format which is obviously invalid.

See the related Mesa ticket here https://gitlab.freedesktop.org/mesa/mesa/-/work_items/15449

HansKristian-Work:
Fix #3014.

### DIFF
--- a/libs/vkd3d/state.c
+++ b/libs/vkd3d/state.c
@@ -5030,25 +5030,46 @@ static HRESULT d3d12_pipeline_state_init_graphics_create_info(struct d3d12_pipel
     for (i = 0; i < rt_count; ++i)
     {
         const D3D12_RENDER_TARGET_BLEND_DESC *rt_desc;
+        bool null_format = false;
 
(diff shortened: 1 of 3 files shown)


hakzsam:
Seems fixed with that.

doitsujin (approved):
Looks sensible.

Do we know what kind of color write mask the game uses here? Wondering if we should look at that as well going forward, but that would likely break all sorts of foz DBs.

HansKristian-Work:
Pretty sure it's non-zero, otherwise it would just have worked in RADV.

FWIW, we already clear out write masks to 0 if null RTV format is used, and this path will also force write mask to 0 for broken RTV formats, so we should be in the clear.

doitsujin:
I was wondering about the opposite direction, i.e. setting null RTV format if write mask is 0 to avoid some other theoretical edge cases, but yeah, can ignore until this becomes a problem.

PR #3019, opened 2026-05-13
Descriptor heap - Phase 7

HansKristian-Work:
Binds the descriptor heap. Unlike DB there is no "dummy" fallback heap, since we rely on drivers linking together the heaps properly on the drivers that care.

Also, sets up PSO properly.

Refactor some struct members to mark them as only being relevant for legacy model.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -2513,7 +2513,7 @@ static void d3d12_command_list_begin_new_sequence(struct d3d12_command_list *lis
 
     d3d12_command_list_invalidate_all_state(list);
     /* Extra special consideration since we're starting a fresh command buffer. */
-    list->descriptor_heap.buffers.heap_dirty = true;
+    list->descriptor_heap.buffers.global_heap_dirty = true;
     d3d12_command_list_debug_mark_label(list, "Split", 0.0f, 0.0f, 0.0f, 1.0f);
 }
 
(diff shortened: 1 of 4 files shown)


libs/vkd3d/command.c
```diff
@@ -13449,6 +13488,54 @@ static void vkd3d_pipeline_bindings_set_dirty_sets(struct vkd3d_pipeline_binding
...
+        unsigned int heap_count, ID3D12DescriptorHeap *const *heaps)
+{
+    VkDeviceAddress current_resource_va, current_sampler_va;
+    unsigned int i;
+
+    current_resource_va = list->descriptor_heap.buffers.resource.va;
+    current_sampler_va = list->descriptor_heap.buffers.sampler.va;
+
+    list->descriptor_heap.buffers.resource.heap = NULL;
+    list->descriptor_heap.buffers.sampler.heap = NULL;
```
doitsujin, line 13501: We'd reset these even if the call only sets one of the two heaps, is this intended?

Not sure what they are for anyway since they don't seem to be used in this series of commits.

HansKristian-Work: Pretty sure that SetDescriptorHeaps sets the currently active heaps.

> All previously set heaps are unset by the call. At most one heap of each shader-visible type can be set in the call.

from https://learn.microsoft.com/en-us/windows/win32/api/d3d12/nf-d3d12-id3d12graphicscommandlist-setdescriptorheaps.

HansKristian-Work: I guess there's a delta to the DB path here since it doesn't explicitly reset back to dummy buffer, but I think that's more of an artifact of there needing to be a descriptor buffer bound always for meta descriptor reasons.

doitsujin: Right, that's one part of the confusion, the other is that heap being null doesn't imply VA being null, but I suppose that is fine too as long as the latter is only used for actually binding the heap.

doitsujin (approved):

libs/vkd3d/command.c
HansKristian-Work: Ye, the heap pointer is only there to be able to refer back to the heap for meta reasons.

PR #3020, opened 2026-05-13
Workaround some ridiculous GPU hangs in Elden Ring

HansKristian-Work:
Elden Ring seems to hit some rather unfortunate edge cases where it
passes in UINT64_MAX VA for descriptor table in some rare cases which is obviously
completely broken, but by sheer luck it happens to be robust on native.
The subsequent commit demonstrates various aspects of this insanity.

There is no perfect solution here since we don't control page table
layouts and descriptor heap robustness does not apply to plain table
pointers unless we deliberately slow everything down by applying it to
*every* resource access, not just descriptor indexing.

The easiest fix is to just ignore these calls which fixes the game.
We can deliberate on more hardcore robustness solutions if required.

More diagnostics are added to be able to detect these cases going
forward.


tests/d3d12_descriptors.c
```diff
@@ -6736,3 +6736,131 @@ void test_static_sampler_dynamic_index(void)
     destroy_test_context(&context);
 }
 
+void test_descriptor_heap_broken_table_va(void)
+{
+    /* Somewhat of a sandbox to explore deep Eldritch horror UB with OOM table access.
```
LingMan: `/s/OOM/OOB/`?

doitsujin (approved):
pain

PR #3024, opened 2026-05-15
Descriptor heap - Phase 9

HansKristian-Work:
Now we're starting to modify the bindings reported to dxil-spirv so that it can use proper heap lowering done by driver instead of doing it ourselves as intended.

Also does some plumbing for SSBO alignment stuff. Idea that we did in CreateSRV/UAV is to emit a typed buffer only if alignment doesn't work to our advantage. For descriptor heap SSBO falls back to texel buffer path and root desc SSBO can fallback to BDA. This path isn't super well tested right now since the focus has been on AMD/NV prop which do not hit these paths. I did various spot checking during development to make sure the fallbacks can work.

A little goofy, but I realized that d3d12_root_signature_init_root_descriptors isn't actually used for local root sigs, but there's a separate entry point for that, so the handling for local root sigs that used to be there is just dead code.


libs/vkd3d-shader/dxil.c
```diff
@@ -460,6 +468,14 @@ static dxil_spv_bool dxil_uav_remap(void *userdata, const dxil_spv_uav_d3d_bindi
     resource_flags = dxil_resource_flags_from_kind(d3d_binding->d3d_binding.kind, false);
     use_ssbo = resource_flags != resource_flags_ssbo;
 
+    /* For heap, we have the option to dynamically fall back to texel buffers. */
+    if ((shader_interface_info->flags & VKD3D_SHADER_INTERFACE_HEAP_LOWERING) &&
+        d3d_binding->d3d_binding.kind == DXIL_SPV_RESOURCE_KIND_STRUCTURED_BUFFER &&
+        d3d_binding->d3d_binding.alignment < shader_interface_info->min_ssbo_alignment)
```
doitsujin, line 474: does this work if there is a UAV counter since we sort-of "share" the descriptor?

HansKristian-Work: That should work yes. There's a later commit that will disable heap lowering for BDA based UAV counters. We only enable the sibling texel + SSBO path if we have universal alignment for SSBOs, so there shouldn't be a problem.

doitsujin (approved):

PR #3027, opened 2026-05-15
vkd3d: Fix formatting warnings.

HansKristian-Work:
Extremely annoying and there's a reason why we'd had -Wno-format, but there were tons of (mostly benign) formatting bugs all over the place.

Native and MinGW builds use different types for things mostly due to long. There seems to be no good way to have MinGW assume that %zu works since it still targets msvcrt.dll from prehistoric era (tried various things, but gave up), so needed a hack to get around that too.

### DIFF
--- a/include/private/vkd3d_common.h
+++ b/include/private/vkd3d_common.h
@@ -59,7 +59,9 @@ static inline size_t align(size_t addr, size_t alignment)
 }
 
 #ifdef __GNUC__
-# define VKD3D_PRINTF_FUNC(fmt, args) __attribute__((format(printf, fmt, args)))
+/* Normal printf needlessly warns on %zu since it targets ancient msvcrt.dll.
+ * This is not relevant anymore. */
+# define VKD3D_PRINTF_FUNC(fmt, args) __attribute__((format(gnu_printf, fmt, args)))
 # define VKD3D_UNUSED __attribute__((unused))
 #else
 # define VKD3D_PRINTF_FUNC(fmt, args)
--- a/include/private/vkd3d_native_sync_handle.h
+++ b/include/private/vkd3d_native_sync_handle.h
@@ -72,7 +72,7 @@ static inline int vkd3d_native_sync_handle_release(vkd3d_native_sync_handle hand
             /* Failing to release semaphore is expected if the counter exceeds the maximum limit.
              * If the application does not wait for the semaphore once per present, this
              * will eventually happen. */
-            TRACE("Failed to release semaphore (#%x).\n", GetLastError());
+            TRACE("Failed to release semaphore (#%x).\n", (int)GetLastError());
             prev = -1;
         }
         return prev;
@@ -81,7 +81,7 @@ static inline int vkd3d_native_sync_handle_release(vkd3d_native_sync_handle hand
     {
         if (!SetEvent(handle.handle))
         {
-            ERR("Failed to set event %p (#%x).\n", handle.handle, GetLastError());
+            ERR("Failed to set event %p (#%x).\n", handle.handle, (int)GetLastError());
             return -1;
         }
         return 0;
--- a/libs/d3d12/main.c
+++ b/libs/d3d12/main.c
@@ -47,6 +47,8 @@
 #define DLLEXPORT
 #endif
 
+#include <inttypes.h>
+
 static pthread_once_t library_once = PTHREAD_ONCE_INIT;
 
 static vkd3d_module_t d3d12core_module = NULL;
@@ -152,7 +154,7 @@ HRESULT WINAPI DLLEXPORT D3D12CreateDevice(IUnknown *adapter, D3D_FEATURE_LEVEL 
 HRESULT WINAPI DLLEXPORT D3D12CreateRootSignatureDeserializer(const void *data, SIZE_T data_size,
         REFIID iid, void **deserializer)
 {
-    TRACE("data %p, data_size %lu, iid %s, deserializer %p.\n",
+    TRACE("data %p, data_size %zu, iid %s, deserializer %p.\n",
             data, data_size, debugstr_guid(iid), deserializer);
 
     if (!load_d3d12core())
@@ -174,7 +176,7 @@ HRESULT WINAPI DLLEXPORT D3D12SerializeRootSignature(const D3D12_ROOT_SIGNATURE_
 HRESULT WINAPI DLLEXPORT D3D12CreateVersionedRootSignatureDeserializer(const void *data, SIZE_T data_size,
         REFIID iid, void **deserializer)
 {
-    TRACE("data %p, data_size %lu, iid %s, deserializer %p.\n",
+    TRACE("data %p, data_size %zu, iid %s, deserializer %p.\n",
             data, data_size, debugstr_guid(iid), deserializer);
 
     if (!load_d3d12core())
--- a/libs/d3d12core/debug.c
+++ b/libs/d3d12core/debug.c
@@ -54,7 +54,7 @@ static HRESULT STDMETHODCALLTYPE d3d12_dred_settings_QueryInterface(
 static ULONG STDMETHODCALLTYPE d3d12_dred_settings_AddRef(d3d12_dred_settings_iface *iface)
 {
     struct d3d12_dred_settings *dred_settings = impl_from_ID3D12DeviceRemovedExtendedDataSettings(iface);
-    ULONG refcount = InterlockedIncrement(&dred_settings->refcount);
+    unsigned int refcount = InterlockedIncrement(&dred_settings->refcount);
 
     TRACE("%p increasing refcount to %u.\n", dred_settings, refcount);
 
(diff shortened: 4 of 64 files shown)


mati865:
What versions of mingw-w64 and mingw-w64-gcc do you see the warning with? Or if it will be easier for you, where do you obtain the mingw-w64 toolchain from?

https://github.com/misyltoad/arch-mingw-github-action that seems to be used on the CI is targeting UCRT rather than MSVCRT (it became the default with mingw-w64 12.0 released 2 years ago) and shouldn't have problem with `%zu`. I have also just tested Ubuntu 24.04 container and toolchain there seems to accept `printf("%zu", ...)` just fine.

In a local build on Arch Linux with format warnings enabled I can see:
```
../libs/vkd3d/command.c:19502:22: warning: unknown conversion type character ‘z’ in format [-Wformat=]
19502 |                 WARN("Render target count %u > %zu, ignoring extra descriptors.\n", rt_index, ARRAY_SIZE(list->rtvs));
```
I can look later into why this happens with your macro but doesn't happen with regular `printf`.

HansKristian-Work:
I saw this with Arch Linux MinGW. 16.1. :shrug:

mati865:
Okay, `%zu` is much more fun/cursed (cross one option out) kind of issue that I'm used to deal with.

It can be reproduced with:
```c
#include <stdio.h>

#define VKD3D_PRINTF_FUNC(fmt, args) __attribute__((format(printf, fmt, args)))
void custom_log(const char *fmt, ...) VKD3D_PRINTF_FUNC(1, 2);
void custom_log(const char *fmt, ...) {}

int main(void) {
    size_t foo = 42;
    printf("%zu\n", foo);
    custom_log("%zu\n", foo);
    return 0;
}
```
It results with a warning only with GCC+mingw-w64 combination:
```
❯ x86_64-w64-mingw32-gcc -Wall print.c
print.c: In function ‘main’:
print.c:10:18: warning: unknown conversion type character ‘z’ in format [-Wformat=]
   10 |     custom_log("%zu\n", foo);
      |                  ^
print.c:10:16: warning: too many arguments for format [-Wformat-extra-args]
   10 |     custom_log("%zu\n", foo);
      |                ^~~~~~~
```
Even though it's linked to UCRT:
```
❯ llvm-readobj --coff-imports a.exe | less
...
Import {
  Name: api-ms-win-crt-stdio-l1-1-0.dll
  ImportLookupTableRVA: 0x9228
  ImportAddressTableRVA: 0x93D8
  Symbol: __acrt_iob_func (0)
  Symbol: __p__commode (1)
  Symbol: __p__fmode (2)
  Symbol: __stdio_common_vfprintf (3)
  Symbol: fflush (107)
  Symbol: setvbuf (138)
}
```
Note: when you inspect linker invocation you'll see `-lmsvcrt` but that is just a copy/symlink to the default CRT:
```
8fbf3ca4eb829deaee73f6621fc97b23e2d45893453469e637296ce1ee166c06  /usr/x86_64-w64-mingw32/lib/libmsvcrt.a
22b1d38f32a43399ccf6bbe5172aa723602da1c34397477db93a3fa8472ee024  /usr/x86_64-w64-mingw32/lib/libmsvcrt-os.a
8fbf3ca4eb829deaee73f6621fc97b23e2d45893453469e637296ce1ee166c06  /usr/x86_64-w64-mingw32/lib/libucrt.a
```
This is a long story, but the original MSVCRT is `libmsvcrt-os.a` here.

Back to the issue. I don't know how but that format attribute seems to make GCC think the code is using `ms_printf` rather than `gnu_printf`, so it shows the warning. For the regular function calls the right function is chosen at https://github.com/mingw-w64/mingw-w64/blob/e705546e5ca0314ad58814c6f8d35a71be439aef/mingw-w64-headers/crt/stdio.h#L276-L285
I'm not going to try guessing why it doesn't work with the attribute.

Naturally, using `gnu_printf` inside the attribute avoids the warning.
Interestingly Clang with `-target x86_64-pc-windows-gcc` doesn't show that warning at all, which explains why I haven't seen this thing occur in years - I'm mostly using Clang+mingw-w64 when dealing with Windows.

That would be it about `%zu`. At a quick glance other warnings look like they should be platform independent, so I didn't dig into them. Let me know if that is not a case.

HansKristian-Work:
If using gnu_printf in the macro fixes it, it is a good idea, ye. Didn't think to check that. I'll have to double check what Wine's msvcrt ends up doing here for %zu though ...

mati865:
I'm not familiar with Wine source code and I don't have toolchain that uses MSVCRT by default at hand, but with Clang it's easy to change linked CRT. So I have done that and my previous test case seems to work fine with `gnu_printf`. Still we probably need to test it with toolchain that Proton uses in builds.

Here is how I tested that:
<details>
<summary>Details</summary>

```
/tmp via C v16.1.1-gcc
❯ clang -target x86_64-pc-windows-gnu print.c -Wall

/tmp via C v16.1.1-gcc
❯ wine a.exe
libEGL warning: pci id for fd 33: 10de:2704, driver (null)

pci id for fd 34: 10de:2704, driver (null)
42

/tmp via C v16.1.1-gcc
❯ pci id for fd 63: 10de:2704, driver (null)
pci id for fd 61: 10de:2704, driver (null)


/tmp via C v16.1.1-gcc
❯ llvm-readobj --coff-imports a.exe

File: a.exe
Format: COFF-x86-64
Arch: x86_64
AddressSize: 64bit
Import {
  Name: KERNEL32.dll
  ImportLookupTableRVA: 0xF040
  ImportAddressTableRVA: 0xF1F0
  Symbol: DeleteCriticalSection (295)
  Symbol: EnterCriticalSection (333)
  Symbol: FreeLibrary (458)
  Symbol: GetCPInfo (481)
  Symbol: GetLastError (648)
  Symbol: GetModuleHandleA (670)
  Symbol: GetProcAddress (730)
  Symbol: InitializeCriticalSection (918)
  Symbol: LeaveCriticalSection (1012)
  Symbol: LoadLibraryA (1016)
  Symbol: MultiByteToWideChar (1064)
  Symbol: SetUnhandledExceptionFilter (1433)
  Symbol: Sleep (1449)
  Symbol: TlsGetValue (1485)
  Symbol: VirtualProtect (1533)
  Symbol: VirtualQuery (1535)
  Symbol: WideCharToMultiByte (1588)
}
Import {
  Name: msvcrt.dll
  ImportLookupTableRVA: 0xF0D0
  ImportAddressTableRVA: 0xF280
  Symbol: __C_specific_handler (117)
  Symbol: ___lc_codepage_func (137)
  Symbol: ___mb_cur_max_func (140)
  Symbol: __getmainargs (155)
  Symbol: __initenv (156)
  Symbol: __iob_func (157)
  Symbol: __set_app_type (163)
  Symbol: __setusermatherr (165)
  Symbol: _amsg_exit (194)
  Symbol: _cexit (212)
  Symbol: _commode (226)
  Symbol: _errno (268)
  Symbol: _fmode (302)
  Symbol: _fpreset (305)
  Symbol: _initterm (377)
  Symbol: _lock (479)
  Symbol: _unlock (814)
  Symbol: abort (1010)
  Symbol: atexit (1021)
  Symbol: calloc (1027)
  Symbol: exit (1040)
  Symbol: fflush (1047)
  Symbol: fprintf (1059)
  Symbol: fputc (1061)
  Symbol: free (1066)
  Symbol: localeconv (1113)
  Symbol: malloc (1120)
  Symbol: memcpy (1129)
  Symbol: setvbuf (1158)
  Symbol: signal (1159)
  Symbol: strerror (1178)
  Symbol: strlen (1180)
  Symbol: strncmp (1183)
  Symbol: vfprintf (1213)
  Symbol: wcslen (1233)
}
```

Notes:
- `-D__MSVCRT_VERSION__=0x600` shouldn't be necessary, I just wanted to be sure my UCRT mingw-w64 build is closer to MSVCRT one
- with Clang `-lmsvcrt-os` disables linking of default CRT (UCRT in my case) and links MSVCRT explictly, this won't work with GCC

</details>

HansKristian-Work:
gnu_printf did the trick. Thanks for the tip.

mati865:
Happy to help with other mingw-w64 related issues if you encounter one.

Small nit that the descriptions in PR and the commit are slightly inaccurate. Dunno how much it matters for you, but perhaps some background would help:
MinGW is the original 32-bit project that is no longer active. Mingw-w64 is the updated fork that is used everywhere. And since version 12.0.0, mingw-w64 defaults to UCRT. Note that mingw-w64 versions are unrelated to GCC versions.
You can consider mingw-w64 as CRT like glibc or musl, but with a twist that the real runtime is already provided by the system and mingw-w64 just provides means for the compiler and linker.

AFAICT the assumption that `printf` means msvcrt.dll is made by GCC.

doitsujin (approved):
Haven't looked at every single line, but if this compiles without format warnings it should be good.

HansKristian-Work:
Github is being broken as hell again, just pushed it to master.

PR #3028, opened 2026-05-15
Descriptor heap - Phase 10

HansKristian-Work:
Completes dxil.c integration for non-RT/WG setups.

Sets up proper heap mappings that can be opted into by dxil.c if heap lowering is okay.

Then, if we have raw VA counters, we have to force legacy path since dxil-spirv doesn't support mixing and matching those. (Very rare, so w/e).

Finally, the redzone is set up and pushed appropriately if active.


doitsujin (approved):
was confused about texel buffers not being set up anywhere, but *of course* they are image resources in spirv...

Issue #2991, opened 2026-05-06
Yellow Intro on STALKER Shadow of Chernobyl EE


MiscStuff839:
When I start the game with the below setup via lutris the intro videos as well as opening cutscene are yellow and distorted. The menu and the game itself runs correctly with no artifacts. 

### Software information
Game: STALKER Shadow of Chernobyl EE
Default settings

### System information
- GPU: rtx 3050 6GB
- Driver: 595.58.03
- Proton version: GE-Proton10-34
- VKD3D-Proton version:  The latest version as of March 23 (from proton ge 10-34 patch notes). Unfortunately,I don't know how to get the exact version

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/f47fd613-f5de-42ef-8a33-fae197d2a015" />

[steam-default.log](https://github.com/user-attachments/files/27449968/steam-default.log)

MiscStuff839:
I also tried SteamDeck=1 but it did not work

Blisto91:
The game depends on some d3d12 video features that vkd3d-proton currently doesn't implement https://github.com/HansKristian-Work/vkd3d-proton/issues/2479

MiscStuff839:
Is there any way to workaround it other than SteamDeck=1 ?

Blisto91:
Not that i know of. How did you attempt to use SteamDeck=1 in Lutris?

MiscStuff839:
<img width="915" height="681" alt="Image" src="https://github.com/user-attachments/assets/eee6ade8-13d4-4172-a0d9-f55d05e8849d" />
Through this section

Blisto91:
Alright. Looks fine.
Maybe the devs didn't include the workaround in non Steam releases

Ablixeron:
So, is there any resolution to this issue?  I'm getting the exact same yellow nonsense on the videos for Stalker: Shadow of Chernobyl Enhanced Edition.

Running CachyOS (wayland with an Nvidia GPU), using Lutris to install my GOG version of the game.  The game itself plays fine.  Have tried multiple argument fixes like SteamDeck=1, PROTON_NATIVE_WAYLAND=1, and others, nothing works.

If this is a DirectX12 feature emulation issue for Vulkan, is there a fix in sight?  Or did I misunderstand the root cause.

Blisto91:
Since the game devs only seem to have shipped a workaround in the Steam version of the game i am not aware of any other workarounds no.
[...]

PR #3031, opened 2026-05-16
Add ID3D12VideoDevice stub

laurirasanen:
Adds stubbed functions for the ID3DVideoDevice interface.

Not sure anyone uses this besides the STALKER devs (https://github.com/HansKristian-Work/vkd3d-proton/issues/2991)? But I'd be interested in poking at this if the project wants to support the video API.

I've tested this against https://github.com/laurirasanen/d3d12-video-decode, and it seems to return what I'd expect.

Don't think I own any Steam games that might care about this interface to check.


include/vkd3d_d3d12video.idl
```diff
@@ -0,0 +1,740 @@
...
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public
+ * License along with this library; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
+ */
+
+import "vkd3d_d3d12.idl";
+
+typedef enum D3D12_VIDEO_FIELD_TYPE
```
HansKristian-Work, line 21: Is this IDL based on official headers?

HansKristian-Work:
I don't know where we want to go with this. It's not the kind of project I would readily assign to an external contributor, and I'm very wary about encouraging work in this area. The scope is massive for very low gains. It would require writing a ton of testing code exploring mostly undocumented APIs.

The main reason no one has looked yet is that the ROI on supporting it is virtually zero right now.

HansKristian-Work:
Getting the IDL-only in (not exposing a stub interface which could break apps in theory) should be fine though. Would at least save some time.

include/vkd3d_d3d12video.idl
laurirasanen: Yes.

laurirasanen:
That makes sense. Are there any smaller issues you would want external contributions for?

HansKristian-Work:
The way to start would be to look at issues, try to reproduce them and maybe find fixes or identify which workarounds are needed if it's a game bug. Large scale feature work is not the place to start I think. It's not a very beginner friendly project though. Debugging big D3D12 games is quite hard on average and it's not something I'd recommend people do unless there is very special interest for it.

include/vkd3d_d3d12video.idl
laurirasanen: Does it need to be a clean-room implementation, or did you have some other concern in mind?

HansKristian-Work: Not for public headers that's just IDL (for everything else, yes). The question is mostly, how would I be able to verify that the header is actually correct? If it's just the upstream IDL, that's fine.

laurirasanen: It's based on the upstream IDL of just the base interface, i.e. no extended ID3D12VideoDevice1, 2, 3, and the respective decode and motion estimator stuff that comes with those.

I guess the only way to *really* verify it would be to redo it internally and close this PR. I'm not sure how much time this would really save if someone later implements video stuff? Also I wouldn't want to create extra churn for you if reviewing this is a pain.

HansKristian-Work: Probably yes, verifying this PR would probably take more time than just doing the IDL ourselves.

HansKristian-Work:
Closing based on discussion. This is unfortunately something we should probably do internally when the need arises.

PR #3038, opened 2026-05-18
VK_KHR_opacity_micromap - Phase 2

chrisd-work:
This phase does some basic code cleanup and plumbs in adding the enabling the VK_KHR_opacity_micromap extension. Since there are many commits to actually using the feature, it will be disabled for now.

### DIFF
--- a/libs/vkd3d/device.c
+++ b/libs/vkd3d/device.c
@@ -2934,6 +2934,11 @@ static void vkd3d_trace_physical_device_features(const struct vkd3d_physical_dev
     TRACE("  VkPhysicalDeviceLineRasterizationFeaturesEXT:\n");
     TRACE("    rectangularLines: %u\n", info->line_rasterization_features.rectangularLines);
     TRACE("    smoothLines: %u\n", info->line_rasterization_features.smoothLines);
+
+    TRACE("  VkPhysicalDeviceOpacityMicromapFeaturesEXT:\n");
+    TRACE("    micromap: %#x\n", info->opacity_micromap_features.micromap);
+    TRACE("    micromapCaptureReplay: %#x\n", info->opacity_micromap_features.micromapCaptureReplay);
+    TRACE("    micromapHostCommands: %#x\n", info->opacity_micromap_features.micromapHostCommands);
 }
 
 static HRESULT vkd3d_init_device_extensions(struct d3d12_device *device,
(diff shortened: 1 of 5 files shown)


libs/vkd3d/device.c
```diff
@@ -134,6 +134,7 @@ static const struct vkd3d_optional_extension_info optional_device_extensions[] =
     VK_EXTENSION(EXT_DEPTH_BIAS_CONTROL, EXT_depth_bias_control),
     VK_EXTENSION(EXT_ZERO_INITIALIZE_DEVICE_MEMORY, EXT_zero_initialize_device_memory),
     VK_EXTENSION_COND(EXT_OPACITY_MICROMAP, EXT_opacity_micromap, VKD3D_CONFIG_FLAG_STATIC(DXR_1_2)),
+    VK_EXTENSION_COND(KHR_OPACITY_MICROMAP, KHR_opacity_micromap, VKD3D_CONFIG_FLAG_STATIC(DXR_1_2)),
```
HansKristian-Work: KHR extensions go in their own place in the array above.

libs/vkd3d/device.c
```diff
@@ -2522,6 +2529,18 @@ static void vkd3d_physical_device_info_init(struct vkd3d_physical_device_info *i
     VK_CALL(vkGetPhysicalDeviceFeatures2(device->vk_physical_device, &info->features2));
     VK_CALL(vkGetPhysicalDeviceProperties2(device->vk_physical_device, &info->properties2));
 
+    /* TODO: Until KHR is fully plumbed in, we will only support EXT */
+    info->opacity_micromap_features_khr.micromap = VK_FALSE;
+
+    /* Prefer KHR_opacity_micromap over EXT_opacity_micromap, and only load one. */
+    info->using_khr_opacity_micromap = info->opacity_micromap_features_khr.micromap
+        && info->device_address_commands_features.deviceAddressCommands;
```
HansKristian-Work: Isn't this implied? DAC is a dependency of KHR_omm and it's invalid to enable KHR_omm without DAC.

libs/vkd3d/device.c
HansKristian-Work: Too minor to hold up the review over though ...

HansKristian-Work (approved):

libs/vkd3d/device.c
HansKristian-Work: Pushed in a minor nit fix.

libs/vkd3d/device.c
HansKristian-Work: Pushed a minor nit fix.

chrisd-work:
> Isn't this implied? DAC is a dependency of KHR_omm and it's invalid to enable KHR_omm without DAC.

True, but I ran into it myself - I only plumbed `KHR_omm` through `winevulkan`, not `KHR_dac`. Would a real user ever encounter this, probably not. I tend to be conservative in non-hot code paths. When we turn KHR on for real, I'm happy to remove the `KHR_DAC` check and probe only `KHR_omm`. Since `KHR_omm` is hardcoded off today, we can address it before flipping it on.

I will move the `VK_EXTENSION_COND` for the KHR extension in a follow up PR.

chrisd-work:
I see you actually already fixed these I appreciate that.

PR #3039, opened 2026-05-18
Descriptor heap - Phase 11

HansKristian-Work:
Starts to ponder meta descriptors and heap, everyone's favorite topic.

It's not a very clean setup since we have to support legacy and heap paths no matter what ...

First, we add helpers that place views on the user heap inside the "meta" region.

Second, add heap paths for meta shaders. For any meta shaders that take descriptors and not just pure push constants, we need to setup two sets of meta pipelines. The code fortunately wasn't *that* horrible ...

The final part adds a heap path to ClearUAV. Other meta paths to follow in Phase 12.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -5394,6 +5394,25 @@ void d3d12_command_list_meta_push_data(struct d3d12_command_list *list,
     }
 }
 
+void d3d12_command_list_meta_push_descriptor_index(struct d3d12_command_list *list,
+        VkCommandBuffer vk_command_buffer, uint32_t binding, uint32_t heap_index)
+{
+    const struct vkd3d_vk_device_procs *vk_procs = &list->device->vk_procs;
+    VkPushDataInfoEXT push;
+    memset(&push, 0, sizeof(push));
+    push.sType = VK_STRUCTURE_TYPE_PUSH_DATA_INFO_EXT;
+    push.data.address = &heap_index;
+    push.data.size = sizeof(heap_index);
+    push.offset = VKD3D_DESCRIPTOR_HEAP_META_PUSH_DATA_OFFSET + binding * sizeof(uint32_t);
+
+    /* We don't go through meta heap path if the currently bound heap is invalid. */
+    assert(!list->descriptor_heap.buffers.global_heap_dirty);
+
+    /* Don't bother invalidating root parameters, that's done in push_data
+     * and we never do push_descriptor_index in complete isolation. */
+    VK_CALL(vkCmdPushDataEXT(vk_command_buffer, &push));
+}
+
 static bool d3d12_command_list_gather_pending_queries(struct d3d12_command_list *list)
 {
     /* TODO allocate arrays from command allocator in case
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -10118,7 +10118,7 @@ static void d3d12_command_list_copy_image(struct d3d12_command_list *list,
         pipeline_key.sample_count = vk_samples_from_dxgi_sample_desc(&dst_resource->desc.SampleDesc);
         pipeline_key.dst_aspect_mask = region->dstSubresource.aspectMask;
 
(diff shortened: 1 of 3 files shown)


libs/vkd3d/command.c
```diff
@@ -3589,6 +3596,119 @@ static struct d3d12_command_allocator *d3d12_command_allocator_from_iface(ID3D12
...
+
+    memset(&texel_buffer_info, 0, sizeof(texel_buffer_info));
+    texel_buffer_info.sType = VK_STRUCTURE_TYPE_TEXEL_BUFFER_DESCRIPTOR_INFO_EXT;
+    texel_buffer_info.addressRange.address = va;
+    texel_buffer_info.addressRange.size = range;
+    texel_buffer_info.format = vk_format;
+
+    memset(&desc_info, 0, sizeof(desc_info));
+    desc_info.sType = VK_STRUCTURE_TYPE_RESOURCE_DESCRIPTOR_INFO_EXT;
+    desc_info.type = VK_DESCRIPTOR_TYPE_STORAGE_TEXEL_BUFFER;
```
doitsujin: do we not need uniform texel buffers at all? one of the sampler feedback shaders has a `utextureBuffer`, but not sure where that descriptor comes from.

libs/vkd3d/meta.c
```diff
@@ -110,6 +124,52 @@ static void vkd3d_meta_make_shader_stage(VkPipelineShaderStageCreateInfo *info,
...
+        info->mapping[i].resourceMask = VK_SPIRV_RESOURCE_TYPE_ALL_EXT;
+        info->mapping[i].sourceData.pushIndex.heapOffset = 0; /* We account for any red-zone ourselves. */
+        info->mapping[i].sourceData.pushIndex.pushOffset = VKD3D_DESCRIPTOR_HEAP_META_PUSH_DATA_OFFSET + i * sizeof(uint32_t);
+        info->mapping[i].sourceData.pushIndex.heapArrayStride = device->bindless_state.cbv_srv_uav_size;
+        info->mapping[i].sourceData.pushIndex.heapIndexStride = device->bindless_state.cbv_srv_uav_size;
+    }
+
+    /* TODO: Evaluate later if we can get rid of this and rely on fetches instead.
+     * It's currently used by sampler feedback decode with some gather4s,
+     * which is not impossible to replace with plain fetches if we must. */
```
doitsujin, line 160: right, I guess getting rid of this would allow us to expose larger sampler heaps.

doitsujin (commented):
the clear uav mess is actually painful, but I guess we have enough tests to cover that.

libs/vkd3d/meta.c
HansKristian-Work: Think so, yes. We can ponder it if it becomes necessary, but I'd rather not touch sampler feedback again if I don't have to.

libs/vkd3d/command.c
HansKristian-Work: Good catch. I missed that case.

libs/vkd3d/command.c
```diff
@@ -3589,6 +3596,120 @@ static struct d3d12_command_allocator *d3d12_command_allocator_from_iface(ID3D12
...
+    texel_buffer_info.format = vk_format;
+
+    memset(&desc_info, 0, sizeof(desc_info));
+    desc_info.sType = VK_STRUCTURE_TYPE_RESOURCE_DESCRIPTOR_INFO_EXT;
+    desc_info.type = vk_descriptor_type;
+    desc_info.data.pTexelBuffer = &texel_buffer_info;
+
+    desc_range.address = heap->descriptor_buffer.host_allocation +
+                         heap_index * list->device->bindless_state.cbv_srv_uav_size;
+    desc_range.size = list->device->bindless_state.heap.storage_texel_buffer_size;
```
doitsujin: descriptor size needs updating too, theoretically we can also just use cbv_srv_uav_size.

HansKristian-Work: Right ...

HansKristian-Work: Done.

doitsujin (approved):

PR #3042, opened 2026-05-19
VK_KHR_opacity_micromap - Phase 3

chrisd-work:
The next phase. This phase may be the most contentious, as it establishes how I plan to rotate from EXT -> KHR OMM. There were two major paths that could have been taken, the first is to remove the EXT code and then rebuild the the KHR from the ground up. The second is to temporarly create the ability for both the KHR and the EXT code to live at the same time. The later has the advantage that the changes can be compared in an inline manner, so I preferred it.

The KHR path is plumbed but dormant — using_khr_opacity_micromap is force-cleared during physical-device init, so EXT remains the only active backend. A later PR flips it.

### DIFF
--- a/libs/vkd3d/acceleration_structure.c
+++ b/libs/vkd3d/acceleration_structure.c
@@ -308,7 +308,7 @@ bool vkd3d_acceleration_structure_convert_inputs(struct d3d12_device *device,
                     {
                         omm->micromap = vkd3d_va_map_place_opacity_micromap(
                                 &device->memory_allocator.va_map, device,
-                                geom_desc->OmmTriangles.pOmmLinkage->OpacityMicromapArray);
+                                geom_desc->OmmTriangles.pOmmLinkage->OpacityMicromapArray).ext;
 
                         if (omm->micromap == VK_NULL_HANDLE)
                             ERR("Failed to place OMM at VA 0x%"PRIx64".\n", geom_desc->OmmTriangles.pOmmLinkage->OpacityMicromapArray);
@@ -474,7 +474,7 @@ void vkd3d_acceleration_structure_emit_postbuild_info(
 {
     const struct vkd3d_vk_device_procs *vk_procs = &list->device->vk_procs;
     VkAccelerationStructureKHR vk_acceleration_structure;
-    VkMicromapEXT vk_opacity_micromap;
+    union vkd3d_opacity_micromap vk_opacity_micromap;
     VkDependencyInfo dep_info;
     VkMemoryBarrier2 barrier;
     VkDeviceSize stride;
@@ -509,7 +509,7 @@ void vkd3d_acceleration_structure_emit_postbuild_info(
                 vkd3d_acceleration_structure_write_postbuild_info(list, desc, i * stride, vk_acceleration_structure);
                 continue;
             }
-            else if (vk_opacity_micromap != VK_NULL_HANDLE)
+            else if (vk_opacity_micromap.any_handle != (uint64_t)VK_NULL_HANDLE)
             {
                 vkd3d_opacity_micromap_write_postbuild_info(list, desc, i * stride, vk_opacity_micromap);
                 continue;
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -20223,8 +20223,16 @@ static void d3d12_command_list_free_rtas_batch(struct d3d12_command_list *list)
     vkd3d_free(rtas_batch->omm_infos);
     vkd3d_free(rtas_batch->range_infos);
     vkd3d_free(rtas_batch->range_ptrs);
-    vkd3d_free(rtas_batch->omm_build_infos);
-    vkd3d_free(rtas_batch->omm_usage_infos);
+    if (list->device->device_info.using_khr_opacity_micromap)
+    {
+        vkd3d_free(rtas_batch->omm_build_infos.khr);
+        vkd3d_free(rtas_batch->omm_usage_infos.khr);
+    }
+    else
+    {
+        vkd3d_free(rtas_batch->omm_build_infos.ext);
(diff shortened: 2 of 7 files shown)


libs/vkd3d/acceleration_structure.c
```diff
@@ -509,7 +509,7 @@ void vkd3d_acceleration_structure_emit_postbuild_info(
                 vkd3d_acceleration_structure_write_postbuild_info(list, desc, i * stride, vk_acceleration_structure);
                 continue;
             }
-            else if (vk_opacity_micromap != VK_NULL_HANDLE)
+            else if (vk_opacity_micromap.khr != VK_NULL_HANDLE)
```
HansKristian-Work: Looks broken. It's checking for KHR here and not EXT.

libs/vkd3d/command.c
```diff
@@ -20365,7 +20367,7 @@ static void d3d12_command_list_build_raytracing_opacity_micromap_array(struct d3
 
         vkd3d_opacity_micromap_emit_immediate_postbuild_info(list,
                 num_postbuild_info_descs, postbuild_info_descs,
-                build_info->dstMicromap);
+                micromap);
```
HansKristian-Work, line 20719: micromap.ext is never assigned to anything but NULL, so this seems broken?

libs/vkd3d/command.c
```diff
@@ -20649,7 +20651,7 @@ static void STDMETHODCALLTYPE d3d12_command_list_CopyRaytracingAccelerationStruc
             VKD3D_BREADCRUMB_COMMAND(COPY_RTAS);
             return;
         }
-        else if (src_omm != VK_NULL_HANDLE)
+        else if (src_omm.khr != VK_NULL_HANDLE)
```
HansKristian-Work: Why check for khr?

libs/vkd3d/vkd3d_private.h
```diff
@@ -3090,6 +3090,16 @@ struct d3d12_wbi_batch_state
     size_t batch_len;
 };
 
+union vkd3d_omm_build_info {
```
HansKristian-Work: Brace on new line.

libs/vkd3d/vkd3d_private.h
```diff
@@ -3090,6 +3090,16 @@ struct d3d12_wbi_batch_state
...
+    VkMicromapBuildInfoEXT *ext;
+    VkAccelerationStructureGeometryMicromapDataKHR *khr;
+};
+
+union vkd3d_omm_usage_info {
```
HansKristian-Work: Same.

libs/vkd3d/command.c
HansKristian-Work: Is the idea is to have a generic check for "is union non-null" might make more sense to add a uint64_t object member of the union and check that to make it clear what the intent is. The union is going away eventually anyway.

HansKristian-Work (changes requested):
Some questions about unions and minor nits, but nothing too scary I guess.

libs/vkd3d/acceleration_structure.c
chrisd-work: This is on purpose, NULL check doesn't change for KHR or EXT.

libs/vkd3d/command.c
chrisd-work: Ok I will make this change it addresses the comment above as well. Having a "null" field makes sense.

chrisd-work: Due to the way handles are defined in Vulkan, comparing a `uint64_t` to `VK_NULL_HANDLE` will lead to a warning on some compilers when `VK_USE_64_BIT_PTR_DEFINES==1`. This warning can be squelched by casting `VK_NULL_HANDLE`:

```
if (vk_opacity_micromap.object != (uint64_t)VK_NULL_HANDLE)
```

Another option would be to define a proxy type via `VK_DEFINE_NON_DISPATCHABLE_HANDLE(VkD3DGeneric)` and use that in the union. The proxy would make the comparison against VK_NULL_HANDLE type-correct on both 64-bit (pointer vs pointer) and 32-bit (uint64_t vs uint64_t), eliminating the cast at each call site.

Care has to be taken because, while Vulkan guarantees a non-dispatchable handle is 64 bits regardless of platform, it does not guarantee it is an integer type:

```
#ifndef VK_DEFINE_NON_DISPATCHABLE_HANDLE
    #if (VK_USE_64_BIT_PTR_DEFINES==1)
        #define VK_DEFINE_NON_DISPATCHABLE_HANDLE(object) typedef struct object##_T *(object);
    #else
        #define VK_DEFINE_NON_DISPATCHABLE_HANDLE(object) typedef uint64_t (object);
    #endif
#endif
```

For now I'm addressing this with the `(uint64_t)VK_NULL_HANDLE` cast - happy to revisit with the proxy-type approach if you'd prefer.

libs/vkd3d/acceleration_structure.c
chrisd-work: Fixed.

libs/vkd3d/command.c
chrisd-work: Fixed.

libs/vkd3d/vkd3d_private.h
chrisd-work: Fixed.

libs/vkd3d/vkd3d_private.h
chrisd-work: Fixed.

chrisd-work:
All comments addressed and new commits pushed.

libs/vkd3d/command.c
HansKristian-Work: The code will go away eventually, so it's okay I guess. I just need the clarity of intent when reviewing.

HansKristian-Work (approved):

PR #3044, opened 2026-05-19
Descriptor heap - Phase 13

HansKristian-Work:
Ray-tracing support.

The root signature setup is fairly straight forward and mirrors the global root sig setup with some minor differences that are commented.

Since we no longer have to faff around with local static sampler VkPipelineLayout compatibility shenanigans we fix a long-standing bug where certain valid D3D12 patterns now become implementable.

### DIFF
--- a/libs/vkd3d-shader/dxil.c
+++ b/libs/vkd3d-shader/dxil.c
@@ -316,6 +316,8 @@ static dxil_spv_bool dxil_remap(const struct vkd3d_dxil_remap_userdata *remap,
             /* Not relevant. */
             remap_info.descriptor_table_offset_words = 0;
             remap_info.num_root_descriptors = 0;
+            /* TODO: Remove this. */
+            remap_info.skip_heap_lowering = true;
             return dxil_remap_inner(&remap_info, descriptor_type, d3d_binding, vk_binding, resource_flags);
         }
         else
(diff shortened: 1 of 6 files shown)


libs/vkd3d-shader/dxil.c
```diff
@@ -1784,23 +1833,51 @@ int vkd3d_shader_compile_dxil_export(const struct vkd3d_shader_code *dxil,
...
 
             case D3D12_ROOT_PARAMETER_TYPE_CBV:
-                dxil_spv_converter_add_local_root_descriptor(converter,
-                        DXIL_SPV_RESOURCE_CLASS_CBV, root_parameter->descriptor.binding->register_space,
-                        root_parameter->descriptor.binding->register_index);
+                if (root_desc_heap_lowering)
+                {
+                    /* Dummy parameters to make dxil-spirv happy. These are mapped with proper API elsewhere. */
+                    dxil_spv_converter_add_local_root_descriptor(converter,
+                            DXIL_SPV_RESOURCE_CLASS_SAMPLER, UINT32_MAX, UINT32_MAX);
```
doitsujin, line 1843: why `SAMPLER` on the CBV/SRV/UAV branches?

HansKristian-Work: It's just a dummy parameter. It's only there so dxil-spirv knows the binary layout of the record buffer. (It's linear and computed implicitly by the argument order in local root signature).

doitsujin: right.

doitsujin (approved):

PR #3045, opened 2026-05-19
Descriptor heap - Phase 14

HansKristian-Work:
This is mostly a bunch of very minor cleanup commits that are not very hard to reason about.

The major thing here is that we can dynamically disable the heap redzone if nothing can possibly require extra meta descriptors. On RADV this is the common case unless descriptor heap robustness is enabled.


tests/d3d12_tests.h
HansKristian-Work, line 314: For review purposes: the dupe is here.

doitsujin (approved):

PR #3046, opened 2026-05-19
Descriptor heap - Phase 15/16

HansKristian-Work:
Getting there. Removes legacy path from workgraphs and adds heap path. Workgraphs never worked well with legacy path anyway, and no games rely on it, so only support it when we can do it somewhat correctly ...

Basic idea for workgraph is to take advantage of INDIRECT mappings in Vulkan heap to combine the system push data along with user root parameters. We can in theory be smarter and more adaptive by fusing them together if the combined size is small enough, but that's an optimization for a future time when content actually ships WGs ...


doitsujin (approved):

doitsujin:
kinda annoying to digest the binding model here but I guess it makes sense if everything is backed by memory.

PR #3052, opened 2026-05-20
VK_KHR_opacity_micromap - Phase 4

chrisd-work:
This is the last set before we can enable `VK_KHR_opacity_micromap`. However, `VK_KHR_opacity_micromap` requires `SPIR-V` changes. There is a dependent commit we may be blocked on (https://github.com/KhronosGroup/SPIRV-Tools/pull/6670).

I will test tomorrow if we can live without it. Without the `SPIR-V` changes, `traceRay` calls will simply pass through OMM. So it is less than ideal. We could also continue to commit the removal commits, but leave `KHR_omm` disabled.

### DIFF
--- a/libs/vkd3d/resource.c
+++ b/libs/vkd3d/resource.c
@@ -4863,7 +4863,12 @@ static void vkd3d_view_destroy(struct vkd3d_view *view, struct d3d12_device *dev
             break;
         case VKD3D_VIEW_TYPE_ACCELERATION_STRUCTURE_OR_OPACITY_MICROMAP:
             if (view->info.buffer.rtas_is_micromap)
-                VK_CALL(vkDestroyMicromapEXT(device->vk_device, view->vk_micromap.ext, NULL));
+            {
+                if (device->device_info.using_khr_opacity_micromap)
+                    VK_CALL(vkDestroyAccelerationStructureKHR(device->vk_device, view->vk_micromap.khr, NULL));
+                else
+                    VK_CALL(vkDestroyMicromapEXT(device->vk_device, view->vk_micromap.ext, NULL));
+            }
             else
                 VK_CALL(vkDestroyAccelerationStructureKHR(device->vk_device, view->vk_acceleration_structure, NULL));
             break;
@@ -5258,27 +5263,45 @@ bool vkd3d_create_opacity_micromap_view(struct d3d12_device *device, const struc
         struct vkd3d_view **view)
 {
     const struct vkd3d_vk_device_procs *vk_procs = &device->vk_procs;
+    VkAccelerationStructureCreateInfo2KHR create_info_khr;
     union vkd3d_opacity_micromap vk_micromap;
-    VkMicromapCreateInfoEXT create_info;
+    VkMicromapCreateInfoEXT create_info_ext;
     struct vkd3d_view *object;
     VkResult vr;
 
-    create_info.sType = VK_STRUCTURE_TYPE_MICROMAP_CREATE_INFO_EXT;
-    create_info.pNext = NULL;
-    create_info.type = VK_MICROMAP_TYPE_OPACITY_MICROMAP_EXT;
-    create_info.createFlags = 0;
-    create_info.deviceAddress = 0;
-    create_info.buffer = desc->buffer;
-    create_info.offset = desc->offset;
-    create_info.size = desc->size;
+    if (device->device_info.using_khr_opacity_micromap)
+    {
+        memset(&create_info_khr, 0, sizeof(create_info_khr));
 
-    vr = VK_CALL(vkCreateMicromapEXT(device->vk_device, &create_info, NULL, &vk_micromap.ext));
+        create_info_khr.sType = VK_STRUCTURE_TYPE_ACCELERATION_STRUCTURE_CREATE_INFO_2_KHR;
+        create_info_khr.type = VK_ACCELERATION_STRUCTURE_TYPE_GENERIC_KHR;
+        create_info_khr.addressRange.address = vkd3d_get_buffer_device_address(device, desc->buffer) + desc->offset;
+        create_info_khr.addressRange.size = desc->size;
+
+        vr = VK_CALL(vkCreateAccelerationStructure2KHR(device->vk_device, &create_info_khr, NULL, &vk_micromap.khr));
+    }
+    else
+    {
+        memset(&create_info_ext, 0, sizeof(create_info_ext));
+
+        create_info_ext.sType = VK_STRUCTURE_TYPE_MICROMAP_CREATE_INFO_EXT;
+        create_info_ext.buffer = desc->buffer;
+        create_info_ext.offset = desc->offset;
+        create_info_ext.size = desc->size;
+        create_info_ext.type = VK_MICROMAP_TYPE_OPACITY_MICROMAP_EXT;
+
+        vr = VK_CALL(vkCreateMicromapEXT(device->vk_device, &create_info_ext, NULL, &vk_micromap.ext));
+    }
+
     if (vr != VK_SUCCESS)
         return false;
 
     if (!(object = vkd3d_view_create(VKD3D_VIEW_TYPE_ACCELERATION_STRUCTURE_OR_OPACITY_MICROMAP)))
     {
-        VK_CALL(vkDestroyMicromapEXT(device->vk_device, vk_micromap.ext, NULL));
+        if (device->device_info.using_khr_opacity_micromap)
+            VK_CALL(vkDestroyAccelerationStructureKHR(device->vk_device, vk_micromap.khr, NULL));
+        else
+            VK_CALL(vkDestroyMicromapEXT(device->vk_device, vk_micromap.ext, NULL));
         return false;
     }
 
--- a/libs/vkd3d/acceleration_structure.c
+++ b/libs/vkd3d/acceleration_structure.c
@@ -21,7 +21,7 @@
 
 #define RT_TRACE TRACE
 
-static VkBuildAccelerationStructureFlagsKHR d3d12_build_flags_to_vk(
+static VkBuildAccelerationStructureFlagsKHR d3d12_build_flags_to_vk_ext(
         D3D12_RAYTRACING_ACCELERATION_STRUCTURE_BUILD_FLAGS flags)
 {
     VkBuildAccelerationStructureFlagsKHR vk_flags = 0;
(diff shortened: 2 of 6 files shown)


libs/vkd3d/resource.c
```diff
@@ -5258,27 +5263,45 @@ bool vkd3d_create_opacity_micromap_view(struct d3d12_device *device, const struc
...
 
-    create_info.sType = VK_STRUCTURE_TYPE_MICROMAP_CREATE_INFO_EXT;
-    create_info.pNext = NULL;
-    create_info.type = VK_MICROMAP_TYPE_OPACITY_MICROMAP_EXT;
-    create_info.createFlags = 0;
-    create_info.deviceAddress = 0;
-    create_info.buffer = desc->buffer;
-    create_info.offset = desc->offset;
-    create_info.size = desc->size;
+    if (device->device_info.using_khr_opacity_micromap)
```
HansKristian-Work, line 5272: Why is create_opacity_micromap_view even called? It should just be a normal RTAS so why not call that instead and leave this function EXT only?

HansKristian-Work: If this is cleaned up later when EXT is removed, that's fine.

HansKristian-Work:
> traceRay calls will simply pass through OMM.

What do you mean here? RTPSO requires a flag to enable OMM. SM 6.9 ray query requires the flag in shader though.

HansKristian-Work (changes requested):
Some questions, but nothing too concerning.

libs/vkd3d/resource.c
chrisd-work: Yes, agreed - `vkd3d_create_opacity_micromap_view` shouldn't exist as a separate entry point under KHR. This is addressed later in the series by "vkd3d: Merge OMM view creation into `vkd3d_create_acceleration_structure_view`" (preceded by "vkd3d: Collapse `vkd3d_view` AS/OMM handle aliases" which unifies the underlying field). In this PR the split persists because EXT is still the active backend and the EXT-side micromap-create path needs the dedicated function; once EXT is removed the OMM-create call site collapses into the AS-create call site directly.

chrisd-work:
> > traceRay calls will simply pass through OMM.
>
> What do you mean here? RTPSO requires a flag to enable OMM. SM 6.9 ray query requires the flag in shader though.

To clarify "pass through OMM": I mean OMM evaluation is bypassed at trace time. The OMM data is built into the BLAS and bound, but the ray-triangle intersection ignores it; rays hit the underlying triangles as if OMM were never attached. The data is correctly present, the trace just doesn't consult it.

On the spec position, both ray-tracing entry points are symmetric - from the DXR functional spec (https://microsoft.github.io/DirectX-Specs/d3d/Raytracing.html):

> Opacity micromaps and TraceRay: "If the flag is omitted and an OMM is encountered during ray traversal, behavior is undefined."
>
>  Opacity micromaps and TraceRayInline: "If the flag is omitted and an OMM is encountered during ray traversal, behavior is undefined."

The relevant flags are `D3D12_RAYTRACING_PIPELINE_FLAG_ALLOW_OPACITY_MICROMAPS` on the RTPSO and `RAYQUERY_FLAG_ALLOW_OPACITY_MICROMAPS` on the RayQuery template. Notably, there is no inverse - D3D12 provides no mechanism to explicitly request OMM be skipped. Apps either set the flag (defined: evaluate OMM) or leave it unset (UB if an OMM is present). There is no D3D12 program that legitimately intends "skip OMM" semantics, because the API doesn't expose that intent.

Two practical paths reach OMM evaluation in real apps:

1. NVAPI + TraceRay (pre-SM-6.9). NVAPI exposes OMM directly via vendor extension. Apps using this path don't touch the DXR flag and don't need SM 6.9; OMM is consulted on every TraceRay. This is the legacy path most existing OMM-using games take and it works on any shader model NVAPI supports.
2. SM 6.9 RayQuery. Goes through the spec'd path. If the RayQuery template doesn't carry `RAYQUERY_FLAG_ALLOW_OPACITY_MICROMAPS`, the spec says any OMM encountered is UB per the quote above.

On NVIDIA hardware specifically, I tested what that UB resolves to in practice: the OMM gets evaluated anyway. Apps that target NVIDIA without setting the flag still hit OMM at trace time. Games will almost certainly end up relying on that behavior since that's what shipping content does. So at least for the first cut we should implement the UB consistent with NVIDIA's resolution - evaluate OMM by default on the Vulkan side, matching what NVIDIA does on D3D12. If we later find hardware that resolves the UB differently, we'll need to gate this behind a flag.

The Vulkan side of the asymmetry is what makes the SPIR-V dependency load-bearing for this MR: in `VK_KHR_opacity_micromap`, the default behavior of the equivalent trace path is to skip OMM, not evaluate it. OMM evaluation is opt-in via pipeline-create flag + SPIR-V capability declaration in the shader. Even with the pipeline-side flag wired up, an SM 6.9 RayQuery shader that doesn't explicitly opt in gets "evaluate OMM" semantics on D3D12+NVIDIA but "skip OMM" semantics on us.

Concrete example of the impact: I tested this in Black Myth: Wukong. Without the SPIR-V opcodes that ensure traceRay calls evaluate OMM, the leaves render incorrectly in shadow - the shadow rays hit the underlying foliage triangles instead of consulting the OMM, so shadows on the leaves look wrong. That's the kind of user-visible regression we'd ship if we flipped KHR OMM on without the SPIR-V side in place.

HansKristian-Work:
> in VK_KHR_opacity_micromap, the default behavior of the equivalent trace path is to skip OMM, not evaluate it

Is this true even for ray pipelines with RTPSO flag? That's a breaking change from EXT. The way I understood it that RayQuery needs the per-shader mode, but not RTPSO since it's controlled with the PSO flag.

Can you refer to spec language to back this up?

The discussion in https://github.com/KhronosGroup/SPIRV-Tools/pull/6670#discussion_r3255289729 agrees with my interpretation at least.

> So at least for the first cut we should implement the UB consistent with NVIDIA's resolution - evaluate OMM by default on the Vulkan side, matching what NVIDIA does on D3D12

RADV will not like that. We can add app-opt to force-evaluate if applications turn out to need it, but let's not opt-in to emulating UB behavior unless proven to be absolutely necessary. Forcing that path for NvAPI OMM is fine though, since I don't think RADV is going to care about those paths anyway.

chrisd-work:
> Is this true even for ray pipelines with RTPSO flag? That's a breaking change from EXT. The way I understood it that RayQuery needs the per-shader mode, but not RTPSO since it's controlled with the PSO flag.
> 
> Can you refer to spec language to back this up?
> 

You're right, my earlier wording was too broad. The pipeline flag does suffice for the `TraceRay` path. The breaking change from EXT is specifically `RayQuery`. From the `VK_KHR_opacity_micromap` proposal (https://github.khronos.org/Vulkan-Site/features/latest/features/proposals/VK_KHR_opacity_micromap.html), §3.4 and §3.8.3:

> In order to use micromaps with ray pipelines, the following flag must be provided: ... VK_PIPELINE_CREATE_2_RAY_TRACING_OPACITY_MICROMAP_BIT_KHR
>
> Ray query operations could use opacity micromaps in the `VK_EXT_opacity_micromap` extension without needing to supply a flag like ray pipelines. This has changed in this extension, which adds a new execution mode in the `SPV_KHR_opacity_micromap` extension to enable opacity micromaps with ray query in shaders: `OpacityMicromapKHR` - this execution mode takes a specialization constant boolean value to determine if ray queries in that shader can use opacity micromaps.

And the load-bearing carve-out:

> If the `VK_EXT_opacity_micromap` extension is not enabled, this execution mode must be supplied to use opacity micromaps with ray queries, even if executed inside a ray pipeline shader with the `VK_PIPELINE_CREATE_RAY_TRACING_OPACITY_MICROMAP_BIT_KHR` flag specified.

To your second point:

> RADV will not like that. We can add app-opt to force-evaluate if applications turn out to need it, but let's not opt-in to emulating UB behavior unless proven to be absolutely necessary. Forcing that path for NvAPI OMM is fine though, since I don't think RADV is going to care about those paths anyway.

Selecting between UBs is always a bit of a coin flip. If I had a RADV GPU to test, I would. I am happy to defer to you on which we should default to.

I consider this a gap in D3D12: there should be a way to force opt out, or opt in. OMM being UB in any situation is kinda wild to me. Vulkan addresses this gap by defining both states; opt-in evaluates OMM and the absence of opt-in skips it, with no UB on either side.

Worth verifying what RADV actually does in the UB case. Do you happen to know what RADV picks, if not we should test. RADV could rationally have landed on the same UB resolution as NVIDIA. The OMM perf concern only kicks in when an OMM is attached to the AS in the first place, and an app that built an OMM-containing AS and traced against it presumably intended to use the OMM, so trusting that signal is a reasonable default regardless of vendor.

HansKristian-Work:
> Worth verifying what RADV actually does in the UB case.

RADV currently does not implement OMM, but it has added shader cost if OMM has to be considered, which is why the flag is important, otherwise we pessimize every case.

> The OMM perf concern only kicks in when an OMM is attached to the AS in the first place

I'm not sure that's true. I'll let some RADV people comment.

> If the VK_EXT_opacity_micromap extension is not enabled, this execution mode must be supplied to use opacity micromaps with ray queries, even if executed inside a ray pipeline shader with the VK_PIPELINE_CREATE_RAY_TRACING_OPACITY_MICROMAP_BIT_KHR flag specified.

Doesn't D3D12 have the same carveout, or is the incompatibility that if RTPSO flag is set, then ray queries implicitly allow OMM too, even without the SM 6.9 flag? Based on my reading of DXR spec, you still need the flag. There's no carveout from what I can see, so I still don't understand what the problem is.

> I am happy to defer to you on which we should default to.

Both specs say UB, so we'll follow that. We can app-opt if needed.

> Concrete example of the impact: I tested this in Black Myth: Wukong. Without the SPIR-V opcodes that ensure traceRay calls evaluate OMM

That sounds like a driver bug, no? TraceRay OMM evaluation is controlled by RTPSO flag, not SPIR-V flags.
If you meant to say Wukong is using _InlineRayTracing_ for shadows, then yes. However, that requires SM 6.9, and we do not expose SM 6.9 right now. Once SM 6.9 is wired up in dxil-spirv, it should be signaling those execution modes as per-spec. If we need sideband signaling for NvAPI, that could be done sure, but that doesn't block this PR or landing KHR_omm. We don't expose NvAPI OMM right now so there is no regression.

KonstantinSeurer:
> The OMM perf concern only kicks in when an OMM is attached to the AS in the first place

There will be additional overhead if OMM is enabled via execution mode or pipeline flag. We don't have hardware support for OMM so we need to emulate it. Part of the emulation is loading additional data and checking if micromaps are present when a triangle is hit.

chrisd-work:
Quick clarification on the SM 6.9 point: RayQuery itself was added in SM 6.5 (DXR 1.1), not SM 6.9. What's new in SM 6.9 is the second template parameter of RayQuery<>. From the DXR functional spec:

> For inline raytracing via RayQuery objects, the RayQuery object takes an additional template parameter (as of Shader Model 6.9) allowing each instance of the object to indicate that the opacity micromap feature is going to be used.

The single-argument RayQuery<RAY_FLAGS> form has been around since SM 6.5. Pre-SM-6.9 shaders using that form cannot specify RAYQUERY_FLAG_ALLOW_OPACITY_MICROMAPS because the template slot for it does not exist. From the spec on the UB case:

> If the flag is omitted and an OMM is encountered during ray traversal, behavior is undefined.

An argument can be made that this clause is specifically limited to the two-parameter SM 6.9 form, where the flag exists and is omitted. However, then the spec does not directly address what happens with the single-parameter SM 6.5 RayQuery<> form when an OMM is encountered, but that silence also leaves it UB.

On the spec carveout question: you're right, both specs are symmetric on defining a per-shader opt-in flag for OMM evaluation in `RayQuery`. They differ on the reverse: D3D12 has no opt-out (absence of the flag is the UB case quoted above), while Vulkan `KHR_OMM` defines absence of the execution mode to skip and also offers an explicit opt-out via the same execution mode set to FALSE. The flag mechanism differs per platform: D3D12's is the SM 6.9 HLSL template parameter `RAYQUERY_FLAG_ALLOW_OPACITY_MICROMAPS`, UB if omitted; Vulkan's is the OpacityMicromapKHR SPIR-V execution mode, which takes a boolean and is SM-agnostic (dxil-spirv emits the execution mode for any RayQuery shader when opacity_micromap_enabled is set, regardless of source HLSL SM). The pipeline flag covers `TraceRay` only in both. My earlier "carveout" wording was loose; I was pointing at the UB clause. Furthermore, Vulkan has no equivalent "driver picks what it thinks best" clause; `KHR_OMM` defines both states explicitly. Maybe that is the resolution: when D3D12 UB is encountered, we define it on a per-device basis to match what each vendor's hardware does in the wild, giving maximum compatibility with the user's Windows experience.

On Wukong: NVAPI-not-exposed is fair, no current regression for BMW specifically since it only uses NVAPI to create OMMs. My earlier "traceRay calls" wording was imprecise; the shadow path is `RayQuery`, not RT-pipeline `TraceRay`. That said, my initial report that the EXT->KHR cut-over introduces a regression remains true for the general case: any game using SM 6.5 `RayQuery` against DXR 1.2-built OMM ASes will hit it after this MR lands. EXT had implicit OMM in RayQuery; KHR requires per-shader execution mode. Wukong isn't currently in that bucket, but the next PR is what enables the bucket to exist.

Thanks to Konstantin for the RADV clarification. Default-off on RADV is the right call given the per-hit emulation cost.

Based on Konstantin's note (no hardware OMM, software emulation with per-hit presence check), my inference is that AMD's Windows D3D12 driver also skips OMM in the UB case. Without hardware acceleration to make "always-evaluate" cheap, skipping the emulation overhead when the flag is absent is the natural implementation choice. Please let me know if AMD-Windows actually evaluates here; otherwise I'm assuming the inference holds.

If it holds, we have a confirmed vendor disagreement on UB resolution: NVIDIA-with-hardware evaluates, AMD-without-hardware skips. A UB implementations could produce visually correct output, but on foliage-heavy content like BMW (Treat OMM as triangles, AnyHit always ran) and it would run ~16% slower than the UB-on path (evaluate OMM), and the user would have no way of knowing why. Vendor-detect-and-default then becomes unambiguously the right policy: on NVIDIA, match the evaluate behavior; on AMD, match the skip behavior. Each user gets the Windows-native experience on their hardware (vkd3d already gates many behaviors on VK_VENDOR_ID_* in device.c). App-opt is still useful as a per-game override regardless of the global default. App+device combinations work for me too if we want finer granularity (e.g., a game opted in on NVIDIA but not on AMD).

chrisd-work:
@HansKristian-Work I moved this PR to a draft on purpose; a bug got introduced into the commit chain during rebase. Fortunately, it occurs in this chain of commits. So, I have added a commit that should be squashed. However, before I squash it, I want it to be very clear what the change was, so you can focus your attention on it. Once you give the all clear, I will squash the fix and move the PR out of draft.

chrisd-work:
I discussed this with @HansKristian-Work.  The real sticking point is RayQuery, especially once SM 6.9 lands.

The plan is as follows:

TraceRay -> Nothing to do, will work as is. No spec ambiguity.

RayQuery -> If we restrict ourselves to OMM-off in the UB case, in SM 6.9 when two RayQueries exist (one UB and one opted into OMM), it is not possible to map this onto Vulkan. Notably, Vulkan only lets you set the execution mode OpacityMicromapIdKHR once per shader as a constant.

This means that in Vulkan's current shape, it isn't possible to go back and forth.

So there are three solutions:

* UB always OMM-off: Violates SM 6.9 opt-ins.
* UB always OMM-on: Some vendors will have negative perf impacts.
* UB always OMM-off, unless there is at least one RayQuery that probes the OMM; then all RayQueries will.

The last choice is the selected one. Balancing perf and correctness.

There are two caveats:

1. On NVIDIA GPUs, VKD3D will opt into always on. This matches NVIDIA's behavior on Windows, and there is no perf concern on NVIDIA GPUs.
2. NVAPI will also default to always on regardless of what device it runs on. This is motivated by games like BMW that will not operate properly otherwise.

Both of these caveats need to be documented in code. In the future, this may change as games use DXR 1.2, via app overrides, device overrides, and/or picking a different global default.

HansKristian-Work (approved):

PR #3059, opened 2026-05-27
Descriptor heap - Final Phase

HansKristian-Work:
Implement NvAPI support (for DLSS).
Flips over the name of the config to "descriptor_heap". The goal is to eventually change the flag to no_descriptor_heap when the failure rates on NV is small enough, but for now that doesn't seem to be safe.

This completes the absolute nightmare it has been to implement (and review) this monster.


moi952:
This completes the absolute nightmare it has been to implement and review 😄
Huge congratulations for all this work!

Proton-CachyOS already enables descriptor heaps through launch options/environment variables. Does that mean it already benefits from most of this implementation, or are important parts from this final phase still missing?

I'm asking because enabling descriptor heaps still causes issues in some games on NVIDIA (visual glitches, instability, etc.). I'm wondering whether those remaining problems are mostly on the VKD3D-Proton side, or if NVIDIA drivers still need additional work/improvements for descriptor heaps to become fully reliable.

Thanks again for the amazing work!

HansKristian-Work:
There is no obvious work on vkd3d-proton side left. NVIDIA drivers needs to be fixed.

moi952:
> There is no obvious work on vkd3d-proton side left. NVIDIA drivers needs to be fixed.

Thanks for your quick reply! That's clear then 😊

Best of luck with your hard work!

libs/vkd3d/device_vkd3d_ext.c
```diff
@@ -449,30 +547,58 @@ static HRESULT STDMETHODCALLTYPE d3d12_device_vkd3d_ext_GetCudaIndependentDescri
         params->pNext = NULL;
     }
 
-    desc = d3d12_desc_decode_va(params->desc);
+    if (d3d12_device_use_descriptor_heap(device))
+    {
+        if (params->type != D3D12_GET_CUDA_INDEPENDENT_DESCRIPTOR_OBJECT_SURFACE &&
+            params->type != D3D12_GET_CUDA_INDEPENDENT_DESCRIPTOR_OBJECT_TEXTURE)
```
doitsujin, line 553: I guess `SAMPLER` just isn't needed? Should be implementable now at least, but if it's not used by anything....

include/private/config_flag_decl.h
```diff
@@ -62,4 +62,4 @@ VKD3D_DECL_CONFIG("damage_not_zeroed_allocations", DAMAGE_NOT_ZEROED_ALLOCATIONS
 VKD3D_DECL_CONFIG("defer_resource_destruction", DEFER_RESOURCE_DESTRUCTION)
 VKD3D_DECL_CONFIG("prefer_thin_uav_tiling", PREFER_THIN_UAV_TILING)
 VKD3D_DECL_CONFIG_PLAIN(EXTENDED_DEBUG_UTILS)
-VKD3D_DECL_CONFIG("i_want_broken_descriptor_heap", DESCRIPTOR_HEAP)
+VKD3D_DECL_CONFIG("descriptor_heap", DESCRIPTOR_HEAP)
```
doitsujin, line 65: yay

doitsujin (approved):
annoing amounts of NVAPI churn, but it makes sense. I guess it's strictly required for the app to bind the correct heap by hand before doing DLSS stuff?

libs/vkd3d/device_vkd3d_ext.c
HansKristian-Work: Sampler could be omitted from the VA map ye, but figured why not. It's not like application have a ton of shader visible sampler heaps either.

HansKristian-Work:
Indeed, the application must bind the heap, but that's the case for NVAPI D3D12 anyway, and it has survived all known testing I think. It would be a disaster to have to switch back to legacy heap. It would break the heap implementation in half.

PR #3069, opened 2026-05-28
VK_KHR_opacity_micromap - Phase 5

chrisd-work:
This is a big change in the phasing. We actually now switch from using EXT_omm to using KHR_omm. In addition when OMM several tests are moved out of todo. In addition, a small bug is fixed in the tests as well.

Sample of tests runs for `VKD3D_CONFIG=dxr12 VKD3D_TEST_FILTER=test_raytracing_opacity_micro_map` when running each commit in this set, and TOT as of the last rebase.

```
HEAD is now at 492fda465 dxil-spirv: Update submodule.
d3d12: 45781 tests executed (9 failures, 12 successful todo, 0 skipped, 9 todo, 0 bugs).

HEAD is now at 1213964d9 vkd3d: Prefer VK_KHR_opacity_micromap when available.
d3d12: 45781 tests executed (0 failures, 21 successful todo, 0 skipped, 0 todo, 0 bugs).

HEAD is now at a6e487a3d vkd3d: Stop enabling VK_EXT_opacity_micromap.
d3d12: 45781 tests executed (0 failures, 21 successful todo, 0 skipped, 0 todo, 0 bugs).

HEAD is now at c0e16a3f3 tests: Fix opacity micromap postbuild CURRENT_SIZE queries.
d3d12: 45781 tests executed (0 failures, 0 successful todo, 0 skipped, 0 todo, 0 bugs).
```

The next phase will remove EXT in 11 commits.

### DIFF
--- a/libs/vkd3d/device.c
+++ b/libs/vkd3d/device.c
@@ -2549,9 +2549,6 @@ static void vkd3d_physical_device_info_init(struct vkd3d_physical_device_info *i
     VK_CALL(vkGetPhysicalDeviceFeatures2(device->vk_physical_device, &info->features2));
     VK_CALL(vkGetPhysicalDeviceProperties2(device->vk_physical_device, &info->properties2));
 
-    /* TODO: Until KHR is fully plumbed in, we will only support EXT */
-    info->opacity_micromap_features_khr.micromap = VK_FALSE;
-
     /* Prefer KHR_opacity_micromap over EXT_opacity_micromap, and only load one. */
     info->using_khr_opacity_micromap = info->opacity_micromap_features_khr.micromap;
 
--- a/libs/vkd3d/device.c
+++ b/libs/vkd3d/device.c
@@ -134,7 +134,6 @@ static const struct vkd3d_optional_extension_info optional_device_extensions[] =
     VK_EXTENSION_COND(EXT_DEVICE_ADDRESS_BINDING_REPORT, EXT_device_address_binding_report, VKD3D_CONFIG_FLAG_STATIC(FAULT)),
     VK_EXTENSION(EXT_DEPTH_BIAS_CONTROL, EXT_depth_bias_control),
     VK_EXTENSION(EXT_ZERO_INITIALIZE_DEVICE_MEMORY, EXT_zero_initialize_device_memory),
-    VK_EXTENSION_COND(EXT_OPACITY_MICROMAP, EXT_opacity_micromap, VKD3D_CONFIG_FLAG_STATIC(DXR_1_2)),
     VK_EXTENSION(EXT_SHADER_FLOAT8, EXT_shader_float8),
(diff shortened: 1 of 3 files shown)


libs/vkd3d/device.c
```diff
@@ -2552,10 +2545,7 @@ static void vkd3d_physical_device_info_init(struct vkd3d_physical_device_info *i
     /* Prefer KHR_opacity_micromap over EXT_opacity_micromap, and only load one. */
     info->using_khr_opacity_micromap = info->opacity_micromap_features_khr.micromap;
 
-    if (info->using_khr_opacity_micromap)
-        info->opacity_micromap_features_ext.micromap = VK_FALSE;
-
-    info->supports_opacity_micromap = info->using_khr_opacity_micromap || info->opacity_micromap_features_ext.micromap;
+    info->supports_opacity_micromap = info->using_khr_opacity_micromap;
```
HansKristian-Work, line 2548: I assume that eventually, the code will just check the opacity feature struct directly rather than going through a dummy bool.

libs/vkd3d/device.c
```diff
@@ -2473,12 +2472,6 @@ static void vkd3d_physical_device_info_init(struct vkd3d_physical_device_info *i
         vk_prepend_struct(&info->features2, &info->opacity_micromap_features_khr);
     }
 
-    if (vulkan_info->EXT_opacity_micromap)
```
HansKristian-Work, line 2476: I assume the opacity_micromap_features_ext struct will be removed from the feature struct as well.

HansKristian-Work (approved):

libs/vkd3d/device.c
chrisd-work: This is part of the next commit set.

libs/vkd3d/device.c
chrisd-work: It stops being dummy once the DXR-1.2/NVAPI split lands later in this branch. The DXR-1.2-path thunks pass `tier_1_2`; the NVAPI thunks pass `supports_opacity_micromap` directly. They're equal today only because `tier_1_2` is currently defined as "OMM supported" alone, while Shader Execution Reordering isn't wired yet. Once SER lands, `tier_1_2` tightens to OMM-AND-SER, and the gates necessarily diverge:

- DXR-1.2 callers see the full DXR-1.2 contract.
- NVAPI callers see the OMM extension whenever the Vulkan extension is exposed, which is NVAPI's intended semantic (it predates DXR 1.2 and treats OMM as an independent feature, not part of a tier).

Reading `opacity_micromap_features.micromap` at every NVAPI callsite would inline what's effectively a named semantic boundary, and would re-conflate the two gates the moment SER lands.

When we get to that commit in review, happy to take it in one of three directions, whichever you prefer:

1. Keep `supports_opacity_micromap` as-is.
2. Rename it to something more NVAPI-OMM specific (e.g. `nvapi_omm`).
3. Drop the field entirely and add a `d3d12_device_supports_opacity_micromap(device)` helper mirroring how `d3d12_device_supports_ray_tracing_tier_1_2` is exposed, with `tier_1_2` chaining into it.

PR #3074, opened 2026-05-29
Rewrite how resources get retained for workaround purposes

HansKristian-Work:
Basic gist here is to do the "proper" workaround and wait until GPU is actually done. Retaining resources based on direct API use in the command buffer does not seem to actually work.

Removes the old clunky paths, so this is probably a good cleanup overall if anything.

This should resolve lots of lingering use-after-free issues in FSR3. There is a ping-ponged history Luma1/2 resource which does not participate in any barrier every other frame since it's only used as SRV, so we cannot rely on the existing resource retain mechanism at all.


libs/vkd3d/resource.c
```diff
@@ -2274,11 +2262,75 @@ static void d3d12_device_add_pending_resource_decref(struct d3d12_device *device
...
+                pthread_mutex_lock(&queue->queue_lock);
+                if (idle_queue)
+                    idle_queue = queue->last_submission_timeline_value <= last_observed;
+
+                if (!idle_queue)
+                {
+                    d3d12_resource_incref(resource);
+                    sub.type = VKD3D_SUBMISSION_RESOURCE_RETAIN;
+                    sub.resource = resource;
+                    d3d12_command_queue_add_submission_locked(queue, &sub);
```
doitsujin, line 2318: I *think* there's a theoretical race condition if a command queue gets destroyed:

`d3d12_command_queue_Release` sends the `STOP` command first, waits for the worker to complete, *then* unmaps the queue, so if we send a `RETAIN` after the `STOP` it'll never get processed and we leak the resource.

Think changing `Release` to unmap first should completely fix the problem. This has not been an issue before because adding work to the queue was only ever possible via the queue itself.

HansKristian-Work: Excellent catch. Fixed.

doitsujin (approved):

PR #3075, opened 2026-05-29
 VK_KHR_opacity_micromap - Phase 7

chrisd-work:
This change cleans up most of the dual-use issues that `VK_EXT_omm` required. One bookkeeping flag remains:

`rtas_is_micromap` exists because my reading of the D3D12 spec is that SERIALIZATION postbuild is permitted on OMM-Arrays, but `VK_KHR_opacity_micromap` disallows the BLP query on non-TLAS (VUID-vkCmdWriteAccelerationStructuresPropertiesKHR-pAccelerationStructures-12425). The kind tag lets us fillbuffer(0) the NumBottomLevelPointers field for OMM-Arrays instead of issuing the illegal Vulkan query.

### DIFF
--- a/libs/vkd3d/acceleration_structure.c
+++ b/libs/vkd3d/acceleration_structure.c
@@ -430,7 +430,7 @@ void vkd3d_acceleration_structure_emit_postbuild_info(
 {
     const struct vkd3d_vk_device_procs *vk_procs = &list->device->vk_procs;
     VkAccelerationStructureKHR vk_acceleration_structure;
-    VkAccelerationStructureKHR vk_opacity_micromap;
+    bool is_micromap;
     VkDependencyInfo dep_info;
     VkMemoryBarrier2 barrier;
     VkDeviceSize stride;
@@ -458,16 +458,14 @@ void vkd3d_acceleration_structure_emit_postbuild_info(
         if (d3d12_device_supports_ray_tracing_tier_1_2(list->device))
         {
             vkd3d_va_map_try_read_rtas(&list->device->memory_allocator.va_map,
-                    list->device, addresses[i], &vk_acceleration_structure, &vk_opacity_micromap);
+                    list->device, addresses[i], &vk_acceleration_structure, &is_micromap);
 
             if (vk_acceleration_structure != VK_NULL_HANDLE)
             {
-                vkd3d_acceleration_structure_write_postbuild_info(list, desc, i * stride, vk_acceleration_structure);
-                continue;
-            }
-            else if (vk_opacity_micromap != VK_NULL_HANDLE)
-            {
-                vkd3d_opacity_micromap_write_postbuild_info(list, desc, i * stride, vk_opacity_micromap);
+                if (is_micromap)
+                    vkd3d_opacity_micromap_write_postbuild_info(list, desc, i * stride, vk_acceleration_structure);
+                else
+                    vkd3d_acceleration_structure_write_postbuild_info(list, desc, i * stride, vk_acceleration_structure);
                 continue;
             }
             else
@@ -475,7 +473,7 @@ void vkd3d_acceleration_structure_emit_postbuild_info(
         }
 
         vk_acceleration_structure = vkd3d_va_map_place_acceleration_structure(
-                &list->device->memory_allocator.va_map, list->device, addresses[i]);
+                &list->device->memory_allocator.va_map, list->device, addresses[i], false);
         if (vk_acceleration_structure != VK_NULL_HANDLE)
             vkd3d_acceleration_structure_write_postbuild_info(list, desc, i * stride, vk_acceleration_structure);
         else
@@ -550,7 +548,7 @@ void vkd3d_acceleration_structure_copy(
     VkCopyAccelerationStructureInfoKHR info;
     VkAccelerationStructureKHR dst_as;
 
-    dst_as = vkd3d_va_map_place_acceleration_structure(&list->device->memory_allocator.va_map, list->device, dst);
+    dst_as = vkd3d_va_map_place_acceleration_structure(&list->device->memory_allocator.va_map, list->device, dst, false);
     if (dst_as == VK_NULL_HANDLE)
     {
         ERR("Invalid dst address #%"PRIx64" for RTAS copy.\n", dst);
--- a/libs/vkd3d/acceleration_structure.c
+++ b/libs/vkd3d/acceleration_structure.c
@@ -542,25 +542,31 @@ static bool convert_copy_mode(
 void vkd3d_acceleration_structure_copy(
         struct d3d12_command_list *list,
         D3D12_GPU_VIRTUAL_ADDRESS dst, VkAccelerationStructureKHR src_as,
-        D3D12_RAYTRACING_ACCELERATION_STRUCTURE_COPY_MODE mode)
+        D3D12_RAYTRACING_ACCELERATION_STRUCTURE_COPY_MODE mode,
+        bool rtas_is_omm)
 {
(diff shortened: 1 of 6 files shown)


HansKristian-Work (commented):
I do not understand why "rtas_is_omm" is kept around. It's broken since it's not possible to know if a placed RTAS is RTAS or OMM at any time. It can dynamically change over time. The code should be completely agnostic w.r.t. placement, copies, queries, etc. For logging purposes, just treat everything as RTAS, since we cannot know any better.

chrisd-work:
I explained it in the PR description the problem is one function. That does come in some latter commits it is easier to see there. The commit that makes it easiest to see https://github.com/HansKristian-Work/vkd3d-proton/pull/3005/changes/83d964dc4e095081de124a73476d6ae4fb4e65a9. 

Thinking about this again, maybe this is a post build only issue, so today I am going to check. It maybe possible remove it by the time we get to the final commit. Since it is already late for you, I will just say, I will get back to you by the EOD, so you can digest it tomorrow.

HansKristian-Work:
Right. That's an unfortunate mismatch in API behavior ...

Serialization isn't really something that games actually use though. The only tests of this we have are basic query tests to check that something sensible is returned, but that's about it.

Rather than trying to maintain a bool property that we cannot possibly maintain correctly, I think the right approach is to just let the serialization calls through as-is and just add a FIXME call if OMM is enabled on the device that we may be out of spec (we cannot know at record time if we are). That way, if stuff breaks in the wild, we can figure something out and it will also let us know which games (if any) use serialization APIs.

HansKristian-Work (approved):
I guess the is_omm thing can be removed later, just moving the phases along for now.

chrisd-work:
Ok sure I will give you my complete thoughts we can fix things robustly later. And I had to back through the code and had to remind myself clearly why this flag is still required, and no removing it as I suggested is probably not the right path forward.

The changes in all of the commits for VK_KHR_omm fix most of the dual-use issues that `VK_EXT_opacity_micromap` required. One residual question for the BLP slot of `SERIALIZATION` postbuild:

## The gap

D3D12 lets the app ask for `NumBottomLevelAccelerationStructurePointers` on any AS type:

|             | `NumBottomLevelAccelerationStructurePointers` |
|-------------|-----------------------------------------------|
| TLAS        | real count                                    |
| BLAS        | 0 (spec: "For a bottom-level acceleration structure this will be 0") |
| OMM-Array   | implicit 0 (not TLAS)                         |

Vulkan's `VK_QUERY_TYPE_ACCELERATION_STRUCTURE_SERIALIZATION_BOTTOM_LEVEL_POINTERS_KHR` is restricted to TLAS only by `VUID-vkCmdWriteAccelerationStructuresPropertiesKHR-pAccelerationStructures-12425`. The VUID is anchored to the build-time type (`VkAccelerationStructureBuildGeometryInfoKHR::type`), not the create-time type - vkd3d-proton creates every AS as `VK_ACCELERATION_STRUCTURE_TYPE_GENERIC_KHR` and builds it with the concrete type (`TOP_LEVEL` / `BOTTOM_LEVEL` / `OPACITY_MICROMAP`), which is what the VUID checks. So a TLAS is fine; a BLAS or OMM-Array is not. Calling the BLP query on either is a VUID violation. A faithful translation has to substitute `fillbuffer(0)` for non-TLAS rather than forward the query.

(Generalized SERIALIZE/DESERIALIZE has even bigger problems - incompatible blob formats between D3D12 and Vulkan - which is why `D3D12_RAYTRACING_ACCELERATION_STRUCTURE_COPY_MODE_SERIALIZE` /  `_DESERIALIZE` are already FIXME-unsupported in `vkd3d_acceleration_structure_copy`. `SERIALIZATION` postbuild's broader usefulness is bounded by that.)

## VA reuse

D3D12 lets the app retire whatever's at VA `X` and build a different-kind structure on the same VA. This was a problem in the `VK_EXT_opacity_micromap` implementation. The cache here handles it: `vkd3d_va_map_place_acceleration_structure` atomically mutates the cached kind on every call (`uint32_t` with relaxed ordering). D3D12's per-VA serialization contract guarantees the mutation is visible by the time the corresponding postbuild dispatches.

## The flag should be `rtas_is_tlas`, not `rtas_is_omm`

The Vulkan VUID gives us a binary split (TLAS vs everything else), but the current flag distinguishes OMM-vs-not-OMM. Under that flag, BLAS inputs still get routed through the BLP query - a latent BLP-on-non-TLAS issue that the AS-side code has carried since BLP queries were first dispatched. The call has never been gated on AS type, and a "bottom-level pointer count" is a structural TLAS concept - asking it of a BLAS or OMM-Array has always been a category error, regardless of `VUID-12425`. To fix it cleanly the discriminator should be `rtas_is_tlas` (tracking the build-time type): `true` for TLAS only, `false` for BLAS or OMM-Array. Same plumbing, inverted polarity, correct for all cases.

## Four options

1. **Keep the flag, rename to `rtas_is_tlas`.** Route TLAS through the real Vulkan BLP query, `fillbuffer(0)` for BLAS or OMM-Array. Spec-correct on every case, including the latent BLAS one the AS-side code has carried throughout.
2. **Drop `SERIALIZATION` postbuild entirely.** FIXME both fields to 0, matching the FIXME state of `COPY_MODE_SERIALIZE`. The flag, the per-view atomic field, the `va_map` kind cache, and the parameter threading all disappear. The latent BLP-on-BLAS issue also disappears (we no longer call the BLP query at all).
3. **Zero-fill BLP universally, keep size query.** No flag needed, no VUID violation, but TLAS apps get `0` instead of the real count.
4. **Drop the flag entirely, call BLP on all types.** This violates `VUID-12425` on non-TLAS inputs. The BLAS-side code does this today (the latent violation); option 4 extends the same behavior to OMM-Array. The spec says behavior is undefined when the VUID is violated - drivers are free to return garbage or crash, and behavior may change between driver versions.

If we want TLAS `NumBottomLevelAccelerationStructurePointers` working correctly **and** `VUID-12425` respected on non-TLAS, option 1 (renamed) is the only path. The reality is I don't know if games use `NumBottomLevelAccelerationStructurePointers` in a way where option 3's universal `0` would be a problem. In some ways that feels like the safest route.

Happy to take this either way.

## References

D3D12 / DXR:
  - `D3D12_RAYTRACING_ACCELERATION_STRUCTURE_POSTBUILD_INFO_SERIALIZATION_DESC` (field semantics, including "For a bottom-level acceleration structure this will be 0"):
  https://learn.microsoft.com/en-us/windows/win32/api/d3d12/ns-d3d12-d3d12_raytracing_acceleration_structure_postbuild_info_serialization_desc
 - DXR functional spec (OMM-Array serialize/deserialize, copy modes): https://microsoft.github.io/DirectX-Specs/d3d/Raytracing.html

Vulkan:
- `vkCmdWriteAccelerationStructuresPropertiesKHR` (source of VUID-12425 + the "built with" language): https://docs.vulkan.org/spec/latest/chapters/accelstructures.html#vkCmdWriteAccelerationStructuresPropertiesKHR
- `VkQueryType` enumerants (BLP / SERIALIZATION_SIZE / COMPACTED_SIZE descriptions): https://docs.vulkan.org/spec/latest/chapters/queries.html#VkQueryType
- `VkAccelerationStructureTypeKHR` (TLAS / BLAS / OPACITY_MICROMAP / GENERIC, including the create-vs-build constraint): https://docs.vulkan.org/spec/latest/chapters/accelstructures.html#VkAccelerationStructureTypeKHR
- `VK_KHR_ray_tracing_maintenance1` (defines the BLP query type): https://docs.vulkan.org/spec/latest/appendices/VK_KHR_ray_tracing_maintenance1.html

PR #3083, opened 2026-06-01
vkd3d: Massively simplify anti-lag implementation.

HansKristian-Work:
The current implementation where we try to keep track of frame counters is fundamentally broken and there is no good fix. Just make use of the simplified API since it seems to map better to DX12 AntiLag API.

### DIFF
--- a/libs/vkd3d/swapchain.c
+++ b/libs/vkd3d/swapchain.c
@@ -1194,13 +1194,10 @@ static HRESULT STDMETHODCALLTYPE dxgi_vk_swap_chain_Present(IDXGIVkSwapChain2 *i
     }
     else if (chain->queue->device->device_info.anti_lag_amd.antiLag)
     {
-        request->low_latency_frame_id = vkd3d_atomic_uint64_load_explicit(
-                &chain->queue->device->frame_markers.present, vkd3d_memory_order_acquire);
-
         spinlock_acquire(&chain->queue->device->low_latency_swapchain_spinlock);
-        request->requested_anti_lag_state.mode = chain->queue->device->swapchain_info.mode;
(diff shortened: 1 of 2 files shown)


libs/vkd3d/swapchain.c
```diff
@@ -1192,17 +1192,6 @@ static HRESULT STDMETHODCALLTYPE dxgi_vk_swap_chain_Present(IDXGIVkSwapChain2 *i
...
     }
-    else if (chain->queue->device->device_info.anti_lag_amd.antiLag)
-    {
-        request->low_latency_frame_id = vkd3d_atomic_uint64_load_explicit(
-                &chain->queue->device->frame_markers.present, vkd3d_memory_order_acquire);
-
-        spinlock_acquire(&chain->queue->device->low_latency_swapchain_spinlock);
-        request->requested_anti_lag_state.mode = chain->queue->device->swapchain_info.mode;
-        request->requested_anti_lag_state.max_fps = chain->queue->device->swapchain_info.max_fps;
-        low_latency_enable = request->requested_anti_lag_state.mode;
```
doitsujin, line 1203: do we not need to track enablement anymore?

HansKristian-Work: The only place the API is called is in this entry point, so I don't think so, no.

doitsujin (approved):

libs/vkd3d/swapchain.c
HansKristian-Work: Or hm ... It is used by dxgi_vk_swap_chain_wait_internal_handle to figure out how to deal with sleeping pattern, so it might make sense to plumb that through after all.

HansKristian-Work: Updated.

doitsujin (approved):

PR #3084, opened 2026-06-01
 VK_KHR_opacity_micromap - Phase 8

chrisd-work:
This is a pretty big inflection point for the VK_KHR_omm feature.

This finishes up cleanup related to merging OMM and non-OMM AS code paths. In addition, this adds UINT8 support for OMMs.

After this is merged, all changes for VK_KHR_OMM are done except the following caveats:

- The active discussion we are having in https://github.com/HansKristian-Work/vkd3d-proton/pull/3075 related to `rtas_is_omm`. No matter what is decided, I think there will be 1 follow-up commit at a minimum for this.
- RayQueries need to be resolved. Two outstanding items: (1) Still waiting on the SPIRV-Tools PR to merge. (2) Once it does, I will prototype an NVIDIA-only opt-in for the UB.
- NVAPI. I have updated all the code on the vkd3d-proton side. However, I have to clean up the code on the dxvk-nvapi side. I hope to get that into review by EOW.


libs/vkd3d/opacity_micromap.c
```diff
@@ -309,8 +309,7 @@ bool vkd3d_acceleration_structure_convert_opacity_micromap(struct d3d12_device *
 
     if (geom_desc->OmmTriangles.pOmmLinkage->OpacityMicromapArray)
     {
-        omm_triangles_info->micromap = vkd3d_va_map_place_acceleration_structure(
-                &device->memory_allocator.va_map, device,
+        omm_triangles_info->micromap = vkd3d_va_map_read_rtas(&device->memory_allocator.va_map, device,
```
HansKristian-Work: I don't think this okay.

Command buffer A could record this command before the OMM ever existed. Then Command Buffer B records the OMM creation.

Then, application could submit B before A, which would be valid. Placing the RTAS would work fine.

libs/vkd3d/acceleration_structure.c
```diff
@@ -455,29 +455,21 @@ void vkd3d_acceleration_structure_emit_postbuild_info(
 
     for (i = 0; i < count; i++)
     {
-        if (d3d12_device_supports_ray_tracing_tier_1_2(list->device))
+        if (!vkd3d_va_map_try_read_rtas(&list->device->memory_allocator.va_map,
```
HansKristian-Work: This looks like a bug to me. It's valid to record the EmitPostBuildInfo() before the RTAS is ever built as long as application submits in proper order.

Seems like the old code was broken here too though ...

libs/vkd3d/va_map.c
```diff
@@ -277,11 +274,35 @@ void vkd3d_va_map_try_read_rtas(struct vkd3d_va_map *va_map,
...
+    {
+        ERR("Failed to query %s for VA 0x%"PRIx64".\n", expected_rtas_is_omm ? "opacity micromap" : "RTAS", va);
+        return VK_NULL_HANDLE;
+    }
+    if (rtas_is_omm != expected_rtas_is_omm)
+    {
+        ERR("Expected a %s at VA 0x%"PRIx64" got %s, missing new build?.\n",
+                expected_rtas_is_omm ? "opacity micromap" : "RTAS", va,
+                rtas_is_omm ? "opacity micromap" : "RTAS");
+        return VK_NULL_HANDLE;
```
HansKristian-Work: Returning a NULL_HANDLE here is not okay. Depending on API usage valid D3D12 code could end up here by mistake. At worst, a FIXME or something, but this whole "expected_rtas_is_omm" thing needs to go away eventually.

HansKristian-Work (changes requested):
Comments.

libs/vkd3d/opacity_micromap.c
chrisd-work: This is a really good point. I forgot to consider the case that command buffer generation doesn't occur in the order as command buffer submission to the device. So I will remove this change from the queue. If I recall correctly I added it because there are some VUIDs I was running into, and I don't know if there is a way now to avoid those VUIDs. Well there is one way, but it would be very expensive: split up into n command buffers and resolve the issue at submission time by amending the tail of the command buffers.

This also has implications for the other statement I made about rtas_is_omm on the phase 7 thread. Notably it has the same problem. As I noted, that value really needs to be called rtas_is_tlas because BLP queries on both BLAS and OMM-Arrays are VUID violations. So we may have to discuss what to do about them. It is a narrower window than what was present in EXT_OMM. Notably, only BLP queries are impacted, but we may not be able to avoid the VUID violations. I can find out what UB NVIDIA will invoke in this case. However, obviously that can change and is specific only to NVIDIA.

chrisd-work: I am crossing my fingers that BLP queries will be OK, because there has been a latent bug here for a long time.

chrisd-work: Ok I have thought about it, for the TLAS BLP. This is what I am thinking, and will push as a change. `rtas_ever_tlas`. If this gets set to true, we will always do the BLP query, VUID violation. If it is set to true and the value is different later on, before it is removed FIXME. This seems to be the most defensive thing we can do.

chrisd-work: I verified on NVIDIA drivers, this VUID violation will end-up matching D3D12 contractual behavior on TOT. I am going to also bring up adding an extension to close this gap on Khronos side. It doesn't hurt to ask even if it is unlikely. That is the only real robust fix.

chrisd-work: > Ok I have thought about it, for the TLAS BLP. This is what I am thinking, and will push as a change. rtas_ever_tlas. If this gets set to true, we will always do the BLP query, VUID violation. If it is set to true and the value is different later on, before it is removed FIXME. This seems to be the most defensive thing we can do.

Actually this won't work either :( sigh.

chrisd-work: I think we can either do universal zero-fill, or universal BLP. The problem is that in the extreme case the Builds and the Emits could be inverted, as you point out: Emit-before-Build at command buffer record time, Build-before-Emit at command buffer submission time. This means there is no robust CPU-record-time solution to the problem; the only correct answer is to resolve it at submission time, which is expensive. Perhaps a device flag for the choice?

HansKristian-Work: Just let the theoretical UB through and flag a FIXME that we cannot be 100% compliant in all cases. If we run into cases where the UB causes real-world pain, we will know and we will have a clear lead and we can re-evaluate what to do then. We should raise a potential candidate for RT_maintenance2 or something like that so we can plug the hole. If we had an easy workaround we could, but as I've demonstrated with this out-of-order submit issue, it's not feasible.

Of course, for the post-build that happens immediately at BuildRTAS time, we can do the right thing. This concern is really just for the standalone EmitPostBuildInfo() call.

chrisd-work: The only concern I have is that if the Emit is done before the Build of a TLAS, I could assume BLP is valid to query. But then if it is Emit before BLAS Build, I would be wrong. And vice versa, if I assumed at the Emit BLP was not valid. SIGH. So AFAICT in the Emit case, you have to assume either BLP always or BLP never. Even if the TLAS state of VA is known at Emit time, it could change by the time the command list is submitted to the device. In other words, you cannot rely on the state of the VA being correct for Emit. I am going to assume BLP always because that is what the code does now, even before the OMM changes. I will also add a CONFIG to force it to zero, for testing. I am not even sure in the Emit -> Build case how you would add a FIXME robustly. At Emit we don't know what it should be. The only FIXME we could add is if a VA is reused across TLAS and non-TLAS; though it would be overly conservative.

Saancreed: > I am not even sure in the Emit -> Build case how you would add a FIXME robustly.

Have two flags, set the first one when the first AS of micromap type is built, set the second one when standalone Emit is called with `SERIALIZATION` query, `FIXME_ONCE` when both are set? Unless we want to strictly avoid false positives, which this would report.

That said, aren't we already running into invalid usage with BLP query when we issue them on AS for which a build command is recorded later as a bottom level AS?

HansKristian-Work: > how you would add a FIXME robustly

I don't understand the problem here. It's fine to emit a FIXME even if we happen to hit a legal code pattern. The whole point of the FIXME is to give us a hint for debugging later should it be needed. A false positive FIXME is not the end of the world.

chrisd-work:
This took some thinking. The goal was to surface maximum information when the ambiguous case actually happens, without firing `FIXME`s in scenarios that resolve correctly.

Example: a `VA` may be `UNKNOWN` (an `Emit` recorded before any Build, with no `SERIALIZATION` info requested), then upgraded to a known type by a later Build, then `SERIALIZATION` postbuild against the now-known type. That sequence is fine and shouldn't `FIXME`. The state machine handles it: `UNKNOWN` can upgrade to `TLAS` or `NON_TLAS`, and the `SERIALIZATION` writer only consults the kind at the moment it's invoked. The `FIXME` fires only when `SERIALIZATION` lands at `UNKNOWN` or `MUTATED`, which is the genuinely ambiguous case.

This works around a gap between `D3D12` and Vulkan: D3D12 permits `EmitPostBuildInfo` to be recorded before the corresponding build, with submission order resolving the temporal ambiguity. Vulkan's VUID-12425 restricts BLP queries to AS built as TLAS, with no record-time escape hatch. There's no way to bridge that fully at CPU record time; the `zero_fill_blp` config flag lets you trade the D3D12-default behavior for spec conformance per workload.

Worth noting: the BLAS side of this VUID violation predates the OMM work. Before any of the KHR_OMM commits, the `SERIALIZATION` branch in `vkd3d_acceleration_structure_write_postbuild_info` emitted BLP unconditionally - gated only on `rayTracingMaintenance1`, with no per-AS kind check at all. Every BLAS postbuild already tripped VUID-12425. It was never caught because NV tolerates the violation benignly and VVL didn't flag it at the time. The kind enum exposes the gap instead of papering over it.

libs/vkd3d/command.c
```diff
@@ -20841,37 +20846,25 @@ static void STDMETHODCALLTYPE d3d12_command_list_CopyRaytracingAccelerationStruc
...
-
     if (src_as != VK_NULL_HANDLE)
     {
-        vkd3d_acceleration_structure_copy(list, dst_data, src_as, mode, false);
+        vkd3d_acceleration_structure_copy(list, dst_data, src_as, mode, rtas_kind);
         VKD3D_BREADCRUMB_AUX64(dst_data);
         VKD3D_BREADCRUMB_AUX64(src_data);
         VKD3D_BREADCRUMB_AUX32(mode);
         VKD3D_BREADCRUMB_COMMAND(COPY_RTAS);
+        return;
```
HansKristian-Work, line 20867: Weird return, but I'll let it slide if there's nothing else.

libs/vkd3d/acceleration_structure.c
```diff
@@ -395,7 +396,32 @@ void vkd3d_acceleration_structure_write_postbuild_info(
 
     if (desc->InfoType == D3D12_RAYTRACING_ACCELERATION_STRUCTURE_POSTBUILD_INFO_SERIALIZATION)
     {
-        if (is_micromap)
+        /* FIXME: The current Vulkan spec makes it a VUID to use BLP on BLAS / OMM-Array. We try to avoid this where
+         * possible but it isn't always possible. Some drivers such as NVIDIA will work correctly even under the VUID
+         * violation */
+        bool is_tlas;
+
+        if (rtas_kind == VKD3D_RTAS_KIND_TLAS)
```
HansKristian-Work, line 404: There is a technical weakness here since while at record time we might have observed only BOTTOM being used, which would lead us to fill the buffer to 0, but by the time we submit, there might have been a record of a TOP which took precedence, meaning any check here cannot outright block the query.

Similarly, while we may only have observed a TLAS at record time, a BLAS may have taken precedence by the time we actually execute, triggering invalid behavior.

Even with all this instrumentation in place we cannot be 100% sure of anything, but the chance this actually goes wrong in the wild is microscopic I would assume.

I suppose I can just accept this code as-is since it's very likely going to be enough, but if this creates issues in the wild, I will likely just remove all of this code and allow the serialization to go through (or force to 0 given the config flag).

HansKristian-Work:
I'm hitting a VVL error in test_raytracing_opacity_micro_map:

```
19cd19:err:vkd3d_debug_messenger_callback: VUID-vkGetAccelerationStructureBuildSizesKHR-pBuildInfo-03619: vkGetAccelerationStructureBuildSizesKHR(): pBuildInfo->geometryCount is 1, but pMaxPrimitiveCounts is NULL.
19cd19:err:vkd3d_debug_messenger_callback: VUID-vkCmdBuildAccelerationStructuresKHR-ppBuildRangeInfos-03676: vkCmdBuildAccelerationStructuresKHR(): ppBuildRangeInfos[0] is NULL.
```

... but it's not a regression in this PR. Maybe one of the earlier phases?

HansKristian-Work:
Might be older VVL:

```
VUID-vkGetAccelerationStructureBuildSizesKHR-pMaxPrimitiveCounts-11612
If pBuildInfo->geometryCount is not 0 and pBuildInfo->type is not [VK_ACCELERATION_STRUCTURE_TYPE_OPACITY_MICROMAP_KHR](https://registry.khronos.org/vulkan/specs/latest/html/vkspec.html#VkAccelerationStructureTypeNV) , pMaxPrimitiveCounts must be a valid pointer to an array of pBuildInfo->geometryCount uint32_t values
```

03619 doesn't exist anymore.

HansKristian-Work (approved):

chrisd-work:
I will double check the VVL on TOT of Vulkan Validation repo.

Re the technical weakness I agree. It's yet another edge case. I probably will follow up with a small commit upgrading the WARN in va_map to a FIXME thinking about it. The code now attempts to do the best thing it can based on the information it has when it records, but by submission its assumptions may be invalidated.

Sadly there are only two robust fixes.

1) Extend BLP on the Vulkan side to work like D3D12. Basically do what NV drivers do. Accept BLP queries on all AS, return 0 when the AS is not a TLAS.
2) You could "cut" the command buffer at these points when serialization is used. And resolve it at submission time. However, this is even more complex than the coded I put in, and slower.

So I agree, if this all becomes a problem feel free to burn it with fire, and leave the config bit that either fills 0 / always BLP. I am going to discuss this with my contacts on the Khronos side. This is something that undoubtedly a game will hit at some point.

chrisd-work:
Ok so I think the VVL you are running is old. The new langue reads:

> VUID-vkGetAccelerationStructureBuildSizesKHR-pMaxPrimitiveCounts-11612
If pBuildInfo->geometryCount is not 0 and pBuildInfo->type is not [VK_ACCELERATION_STRUCTURE_TYPE_OPACITY_MICROMAP_KHR](https://docs.vulkan.org/refpages/latest/refpages/source/VkAccelerationStructureTypeKHR.html) , pMaxPrimitiveCounts must be a valid pointer to an array of pBuildInfo->geometryCount uint32_t values

Older VVL would create a VUID here when the code is conforming to VK_ACCELERATION_STRUCTURE_TYPE_OPACITY_MICROMAP_KHR carve out.

PR #3090, opened 2026-06-03
tests: fix few issues found with ASAN

hakzsam:
Found these when running the testsuite against RADV.


HansKristian-Work:
Curious how you even hit these, but I guess the fixes are fine.

HansKristian-Work (approved):

PR #3101, opened 2026-06-09
Improve expect assume

HansKristian-Work:
Reports fault codes instead of just derping out instruction number which has to be fished out from spirv-cross.


libs/vkd3d/descriptor_debug.c
```diff
@@ -263,6 +263,30 @@ uint32_t vkd3d_descriptor_debug_clear_bloom_filter(
...
+        TAG(GROUP_SHARED_GEP_OOB);
+        TAG(CONSTANT_GEP_OOB);
+        TAG(ALLOCA_GEP_OOB);
+        TAG(QUESTIONABLE_BAB_OVERFLOW);
+        TAG(QUESTIONABLE_STRUCTURED_OVERFLOW);
+        default: break;
+    }
+#undef TAG
+
+    return NULL;
```
doitsujin, line 287: is null safe here or should we return something like "unknown"?

HansKristian-Work: It's null checked later, and replaced with QA - non-normal.

doitsujin (approved):

PR #3117, opened 2026-06-22
Bump supported SDK to 619

HansKristian-Work:
- OpenHeapFromAddress1 adds an explicit size argument, which makes it implementable on Linux.
- CPU resolve query heap, which fortunately don't have any sync requirements, so it's fairly straight forward.
- Byte offset buffer descriptors which is more Vulkan-like. As always, the native impls are broken and runtime is super broken ... :frog:
- Add "stub" implementations of Try*CreateBlah(). Can be fixed later, but this PR is getting quite chungus already ...

Also fixes a goofy bug where we used wrong enum type when allocating the query heap fallback buffer :V


libs/vkd3d/device.c
```diff
@@ -6183,6 +6244,92 @@ static HRESULT STDMETHODCALLTYPE d3d12_device_CheckFeatureSupport(d3d12_device_i
...
+            D3D12_FEATURE_DATA_BYTECODE_BYPASS_HASH_SUPPORTED *data = feature_data;
+
+            if (feature_data_size != sizeof(*data))
+            {
+                WARN("Invalid size %u.\n", feature_data_size);
+                return E_INVALIDARG;
+            }
+
+            /* This seems to be related to MLIR, which hasn't shipped. */
+            data->Supported = FALSE;
```
Saancreed: Actually, this is just https://github.com/microsoft/hlsl-specs/blob/main/proposals/infra/INF-0004-validator-hashing.md#reserved-static-hash-values, should be trivially supported, if it isn't already by virtue of ignoring shader hashes altogether.

HansKristian-Work: Right. Can't exactly be blamed for assuming that D3D12 specs don't live in the HLSL specs repo as a proposal :frog: ...

libs/vkd3d/resource.c
```diff
@@ -10506,6 +10559,12 @@ HRESULT d3d12_query_heap_create(struct d3d12_device *device, const D3D12_QUERY_H
         pool_info.flags = 0;
         pool_info.queryCount = desc->Count;
 
+        /* We won't have the time to reset the query pool, but don't gate
```
doitsujin: Bit confused what this means, we're creating the pool in reset state and host_query_reset also exists.

libs/vkd3d/resource.c
```diff
@@ -10565,14 +10627,50 @@ HRESULT d3d12_query_heap_create(struct d3d12_device *device, const D3D12_QUERY_H
...
+        if (device->device_info.zero_initialize_device_memory_features.zeroInitializeDeviceMemory)
+            alloc_flags |= VK_MEMORY_ALLOCATE_ZERO_INITIALIZE_BIT_EXT;
+
+        hr = vkd3d_allocate_internal_buffer_memory(device, object->vk_buffer,
+                memory_properties, alloc_flags, &object->device_allocation);
+
+        if (FAILED(hr) && (flags & D3D12_QUERY_HEAP_FLAG_CPU_RESOLVE))
+        {
+            /* Technically, cached | coherent isn't guaranteed (but in reality any relevant impl supports it).
+             * Prefer coherent over cached since it's plausible that apps might spin. */
```
doitsujin: `vkd3d_try_allocate_device_memory` already has a fallback that does the same thing, so this code is probably dead in practice.

HansKristian-Work: Right ... Fixed.

libs/vkd3d/resource.c
HansKristian-Work: Ye, it's misleading. Anyway it's UB it turns out to query the heap without ever having completed a write it seems according to Jesse, so will just leave it as-is.

doitsujin (approved):

PR #3118, opened 2026-06-23
VK_KHR_opacity_micromap - Phase 9 (NVAPI)

chrisd-work:
This PR will commit all the changes for NVAPI support. This change interoperates with https://github.com/jp7677/dxvk-nvapi/pull/371. Since the PR to dxvk-nvapi appears to be down to a few minor issues, I am putting this change into review as I do not expect API changes at this point. These changes are expected to be able to land in either order. For example, vkd3d-proton will not crash if the dxvk-nvapi changes have not been merged, and vice versa.

After this PR only one minor change is left: bumping dxil-spirv, and enabling `ray_query_force_omm` on NVIDIA hardware. I will post this as a separate phase to keep things concise.

### DIFF
--- a/libs/vkd3d/vkd3d_private.h
+++ b/libs/vkd3d/vkd3d_private.h
@@ -2985,7 +2985,7 @@ struct vkd3d_rendering_info
 };
 
 /* ID3D12CommandListExt */
-typedef ID3D12GraphicsCommandListExt1 d3d12_command_list_vkd3d_ext_iface;
+typedef ID3D12GraphicsCommandListExt2 d3d12_command_list_vkd3d_ext_iface;
 
 struct d3d12_rt_state_object;
 
--- a/libs/vkd3d/vkd3d_private.h
+++ b/libs/vkd3d/vkd3d_private.h
@@ -5837,6 +5837,11 @@ struct d3d12_device
         HMODULE amdxc64;
 #endif
         struct vkd3d_nv_shader nv_shader;
+
+        /* Set by SetCreatePipelineStateOptions (NVAPI). NVAPI has no per-pipeline OMM opt-in
+         * (unlike DXR 1.2's ALLOW_OPACITY_MICROMAPS flag), so this global is OR'd into every
+         * RT pipeline creation to implement the device-wide enable/disable semantic. */
+        VkPipelineCreateFlags global_ray_tracing_pipeline_create_flags;
     } vendor_hacks;
 
     bool independent_device;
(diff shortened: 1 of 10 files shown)


libs/vkd3d/vkd3d_private.h
```diff
@@ -5761,6 +5761,11 @@ struct d3d12_device
     LUID adapter_luid;
     D3DKMT_HANDLE kmt_local;
 
+    /* Set by SetCreatePipelineStateOptions (NVAPI). NVAPI has no per-pipeline OMM opt-in
+     * (unlike DXR 1.2's ALLOW_OPACITY_MICROMAPS flag), so this global is OR'd into every
+     * RT pipeline creation to implement the device-wide enable/disable semantic. */
+    VkPipelineCreateFlags global_ray_tracing_pipeline_create_flags;
```
HansKristian-Work: This should go in vendor_hacks struct of struct d3d12_device.

HansKristian-Work (changes requested):
Trivial nit, otherwise LGTM.

Right now OMM is gated behind VKD3D_CONFIG=dxr12, I assume we want to remove that gate? The reason it was gated was because EXT_omm was simply broken and non-compliant, but seems more appropriate to gate it behind `!no_dxr` now?

libs/vkd3d/vkd3d_private.h
chrisd-work: When looking at the `vendor_hacks` today, I see them used in cases where AMD / NVIDIA have a feature that is only used on that GPU. The NVAPI OMM seems different. Notably, it is implemented in such a way that any vendor that implements `VK_KHR_opacity_micromap` can use the feature. Given the name "hack", I assumed this was for situations attuned to a specific GPU running on the end-user's system rather than something specific to an API. I can see the counter-argument, however: NVAPI is an NVIDIA-specific API, hence it belongs in `vendor_hacks`. If you still feel it should be moved after reading my thought process, just let me know and I will move it.

chrisd-work:
I agree I will amend the final commit to lower the DXR12 requirement -> general DXR

chrisd-work:
Updated the code as you suggested to pull in KHR_omm if DXR is enabled, rather than requiring DXR 1.2. I responded to your second concern.

libs/vkd3d/vkd3d_private.h
HansKristian-Work: It's for any vendor-specific implementation details that have nothing to do with the standard D3D12 API (e.g. amdxc64.dll hackery), and NVAPI OMM flag overrides belong there, along with the NVAPI space/binding stuff, even if multiple vendors can technically implement it.

chrisd-work: Thanks for clarifying the intent. I will move it.

chrisd-work: Moved.

HansKristian-Work (approved):

PR #3119, opened 2026-06-23
VK_KHR_opacity_micromap - Phase 10

chrisd-work:
This change is orthogonal to Phase 9 https://github.com/HansKristian-Work/vkd3d-proton/pull/3118, so I am pushing both at the same time. This change bumps dxil-spirv, enables using opacity micromap when generating shaders, and finally enables querying the OMM when RayQuery is called in the UB case.


libs/vkd3d/device.c
```diff
@@ -10642,8 +10642,23 @@ static void vkd3d_init_shader_extensions(struct d3d12_device *device)
 
     if (device->device_info.supports_opacity_micromap)
     {
+        bool ray_query_omm_when_unspecified;
+
         device->vk_info.shader_extensions[device->vk_info.shader_extension_count++] =
                 VKD3D_SHADER_TARGET_EXTENSION_OPACITY_MICROMAP;
+
+        ray_query_omm_when_unspecified =
```
HansKristian-Work: A comment about rationale here is needed. I.e. basically based on discussion:

- NV HW deals natively with it
- Be defensive against bad UB in the wild
- D3D12 driver already assumes it
- Work with NVAPI OMM

chrisd-work: Pushed with the following comment

```
        /* Default-on for NVIDIA for three reasons:
         *   1. The DXR 1.2 spec leaves the unflagged ray-query case unspecified.
         *      Worse, Vulkan has no equivalent flag that defers the choice to the
         *      driver. The choice has to be baked into the SPIR-V at shader-emit
         *      time.
         *   2. NV's D3D12 driver on Windows resolves the unflagged case by
         *      considering OMM; NV hardware traverses OMM cheaply so the cost is
         *      minimal. Default-on matches the behavior apps observe on
         *      NVIDIA's D3D12 driver.
         *   3. NVAPI has no per-ray-query OMM opt-in. NVAPI OMM titles (e.g.
         *      Black Myth: Wukong) rely on the default-on path because there
         *      was no other surface for them to request OMM in a ray query.
         *
         * Default-off elsewhere; other vendors may take a significant perf hit
         * when OMM is consulted in the unflagged case. */
```

libs/vkd3d-shader/dxil.c
```diff
@@ -1157,18 +1157,27 @@ static int vkd3d_dxil_converter_set_options(dxil_spv_converter converter,
                     return VKD3D_ERROR_NOT_IMPLEMENTED;
                 }
             }
-#if 0
             else if (compiler_args->target_extensions[i] == VKD3D_SHADER_TARGET_EXTENSION_OPACITY_MICROMAP)
             {
-                static const dxil_spv_option_opacity_micromap helper =
-                        { { DXIL_SPV_OPTION_OPACITY_MICROMAP }, DXIL_SPV_TRUE };
+                dxil_spv_option_opacity_micromap helper =
+                        { { DXIL_SPV_OPTION_OPACITY_MICROMAP }, DXIL_SPV_TRUE, DXIL_SPV_FALSE };
```
HansKristian-Work: This isn't quite accurate. For RTPSO (dxil_compile_export path), the first member here is meant to mirror the RTPSO flag state. Otherwise, we will enable OMM capability `RayTracingOpacityMicromapKHR` without RTPSO flag being set, which is not legal I think. I think to make this correct, the first member needs to be enabled with a condition:

`if (!stage_is_rtpso(stage) || rtpso_flag_is_enabled)`

For rayquery, support will be keyed off the second member or SM 6.9, so anything non-rtpso will work fine.

You can add a VKD3D_SHADER_INTERFACE flag that signals this from raytracing_pipeline.c.

HansKristian-Work: NVAPI OMM already overrides the RTPSO flag on pipeline creation and that will be forwarded here just fine.

chrisd-work:
I am fixing this PR as you advised. However, something came up in testing: enabled being tied to the RTPSO as you advised doesn't work in BMW, specifically the test in dxil-spirv options.opacity_micromap_enabled && options.ray_query_force_omm_execution_mode. Notably, tying the RTPSO to the enabled flag disables RQ even if the force flag is on for a shader that doesn't use OMM in the RTPSO. The fix is to have three flags rather than two: a general enable flag stating that the KHR_omm feature is present, and another flag stating that the RTPSO of the shader consults the OMM. The final flag's meaning is unchanged: ray_query_force_omm_execution_mode. This conforms with our prior discussions that the RTPSO-enabled TraceRay is fully orthogonal to RayQuery, both in DXR 1.2/NVAPI and Vulkan.

I will be following up with a small PR for dxil-spirv as well. We should land that first since it will change the dxil-spirv submodule pin in vkd3d-proton.

Placing this PR into draft until the dxil-spirv PR lands

chrisd-work:
Ok I have updated this PR and created a new on for dxil-spirv. I will leave this PR in draft but they work in tandem https://github.com/HansKristian-Work/dxil-spirv/pull/286.

libs/vkd3d-shader/dxil.c
```diff
@@ -1157,18 +1157,31 @@ static int vkd3d_dxil_converter_set_options(dxil_spv_converter converter,
...
             }
-#if 0
             else if (compiler_args->target_extensions[i] == VKD3D_SHADER_TARGET_EXTENSION_OPACITY_MICROMAP)
             {
-                static const dxil_spv_option_opacity_micromap helper =
-                        { { DXIL_SPV_OPTION_OPACITY_MICROMAP }, DXIL_SPV_TRUE };
+                dxil_spv_option_opacity_micromap helper =
+                        { { DXIL_SPV_OPTION_OPACITY_MICROMAP }, DXIL_SPV_TRUE, DXIL_SPV_FALSE, DXIL_SPV_FALSE };
+
+                helper.rtpso_has_omm =
```
HansKristian-Work: The way I implemented dxil-spirv was for the first member to be "enabled && supported". There should be no need to split it like this.

If SM 6.9 RQ OMM is used, it will force OMM on.

HansKristian-Work: What exactly is BMW doing wrong here? Does it use OMM + RayQuery, but doesn't even set the NVAPI PSO flag?

HansKristian-Work: If that is the case (game bug?) I think the reasonable option is to change the OMM_enable && rq_force_omm check in dxil-spirv to just check for rq_force_omm rather than introducing a new member. It was my assumption in dxil-spirv that RQ OMM would only be enabled when app uses the NVAPI flag.

chrisd-work: Forgive my long discussion however I am attempting to ensure all the relevant issues are addressed.

1. There was a bug in my code

There was a small bug in my code that didn't get the RTPSO flag that BMW was setting to the SPIR-V shader. I have fixed this in the latest commit, but it doesn't solve the issues alone.

2. The original logic seems faulty even if RTPSO was enabled for every RQ shader.

The original logic you wrote:

```
(options.opacity_micromap_enabled && options.ray_query_force_omm_execution_mode) ||
        shader_analysis.ray_query.requires_opacity_micromap_tracing;
```

The otter or of `requires_opacity_micromap_tracing` is problematic because the location that sets it to `true`
is not gated by `options.opacity_micromap_enabled`. I think the intention was

```
options.opacity_micromap_enabled &&
    (options.ray_query_force_omm_execution_mode || shader_analysis.ray_query.requires_opacity_micromap_tracing);
```

However, as we will see this gate isn't enough.

3. Shaders can exist w/o RTPSO and with RQ

As we discussed in #3052, RTPSO `TraceRay` and `RayQuery` are orthogonal. More critically, `TraceRay` can only be used in RT shaders like raygen, anyhit, miss etc. However, there is no such limitation on `RayQuery` calls. Here is a breakdown on BMW shader usage (entering the first level):

| # | Path | What it is |
|---|---|---|
| 192 | RTPSO raygen + `RayQuery`, ray flags include `RAY_FLAG_FORCE_OPACITY_MICROMAP_2_STATE` | Lumen `OcclusionRGS` and friends, asking for OMM 2-state |
| 12 | RTPSO raygen + `RayQuery`, ray flags don't include `RAY_FLAG_FORCE_OPACITY_MICROMAP_2_STATE` | Same family, branches not asking for OMM at the ray-cast level |
| 11 | RTPSO raygen + classic `TraceRay()`, no inline `RayQuery` | Lumen `RayTracingReflectionsRGS` (10) + `VoxelShadowRGS` (1) |
| 5 | Non-RT fragment (4) + compute (1) + inline `RayQuery` | UE5 SSGI/AO-style screen-space ray tracing |

The last row is the critical one. There are five shaders that do not have any RTPSO concept at all - they aren't in
an RTPSO. The first four are fragment shaders, and the fifth is a compute shader, each using inline `RayQuery` from
a graphics or compute pipeline. This is not an application bug; it is the by-design configuration both DXR 1.2 and
`VK_KHR_opacity_micromap` support.

The RTPSO opt-in and the `RayQuery` opt-in are independent. Apps can opt into one without the other, and `RayQuery` isn't bound to an RTPSO in the first place. BMW exercises two patterns that exploit this independence:

- RT exports relying on NVIDIA's UB resolution. The 204 Lumen raygens use `RayQuery` inline inside RTPSO exports. BMW supplies the RTPSO-level opt-in via NVAPI for the `TraceRay` path but does not set a per-`RayQuery` opt-in
(`RAYQUERY_FLAG_ALLOW_OPACITY_MICROMAPS`). Per DXR `Raytracing.md` "Opacity Micromaps and TraceRayInline" this is UB-covered. NVIDIA resolves the UB deterministically by consulting OMM, and the NVAPI ecosystem ships against that resolution.
- Non-RT shaders with no RTPSO at all. The 4 fragment + 1 compute shaders use inline `RayQuery` from graphics and compute pipelines. There is no RTPSO subobject to opt into; `ALLOW_OPACITY_MICROMAPS` is structurally inapplicable. On the Vulkan side, `RayQuery` OMM in graphics/compute pipelines is gated by SPIR-V execution mode (`OpExecutionModeId OpacityMicromapIdKHR`), not by any pipeline-state flag - the pipeline-flag scope only exists for RT pipelines (`VK_PIPELINE_CREATE_RAY_TRACING_OPACITY_MICROMAP_BIT_KHR`). `SPV_KHR_opacity_micromap` requires `SPV_KHR_ray_query` OR `SPV_KHR_ray_tracing` (disjunctive), so ray query alone satisfies it: the OMM capability is 
legal outside an RT shader. 

My changes to vkd3d-proton respect both patterns: explicit per-`RayQuery` opt-in via `requires_opacity_micromap_tracing` when present in the DXIL, and the NVIDIA-platform implicit case via `ray_query_force_omm_execution_mode=TRUE` default-on (the PR #3052 policy choice). Any gate that demanded the RTPSO flag for `RayQuery`-OMM emission would break both BMW patterns: the first by failing to match the NVAPI ecosystem behavior, the second by demanding a flag that has no place to live.

In both cases, the state of how `RayQuery` interacts with OMM is independent of how `TraceRay` interacts with OMM. `RayQuery` is fully independent of any RTPSO flags wrt the OMM.

This is why my change separates out `enabled`. I think you intended `enabled` to be `KHR_omm` present and RTPSO flags indicate OMM. However, as this shows, it is legal and expected that an application can use RQ without enabling RTPSO OMM in an RT shader, or use RQ in a non-RT shader where it is not even possible to enable OMM through the PSO flags. So the state `enabled` (`KHR_omm` supported) and `rtpso_has_omm` (the PSO is RT and OMM is enabled in the RTPSO) must be separated as two independent flags.

4. A final gap no app I know of requires a fix for.

Up to this point we have dealt with the unspecified behavior of `RayQuery` because that shows up in games like BMW and CP2077. However, there is another spec gap that may come up. Notably, if an RTPSO is created without
`ALLOW_OPACITY_MICROMAPS`, `TraceRay` is called, and an OMM is encountered, both D3D12 and Vulkan treat this as undefined. In theory the two APIs could pick different UB resolutions on the same vendor, which would show up as a behavior difference between D3D12 native and vkd3d-proton.

In practice no applications I know of exercise this path, and the expectation is that for a given vendor the UB
resolution will be consistent between D3D12 and Vulkan (same hardware, same driver code under the hood). So this problem could come up, but it might not ever come up either. Basically just noting it; I think it is fine to leave
it as is for now.

Fixing this is trivial, ideally with a flag, but may have a large blast radius. Since no applications are impacted by it, I think it is fine to defer.

chrisd-work:
Based on some private conversations I think it is best to clarify some points 

1. RTPSO controls `TraceRay`. `RayQuery` is independent of it.

Both D3D12 and Vulkan partition OMM opt-in into two independent surfaces:

- An RTPSO/RT-pipeline flag that gates `TraceRay`. RT-pipelines only. No equivalent for graphics or compute
  pipelines.
- A per-shader (or per-`RayQuery`) opt-in that gates `RayQuery`. Works in any shader stage, including fragment
  and compute where there is no RTPSO at all. Even in RT shaders (raygen, closesthit, anyhit, miss, intersection)
  where the RTPSO is in scope, RayQuery is fully independent of any RTPSO flags wrt the OMM. Only the per-RayQuery
  opt-in controls OMM for RayQuery.

Apps can opt into one without the other.

## D3D12 / DXR 1.2

`TraceRay`: [RTPSO flag](https://github.com/microsoft/DirectX-Specs/blob/master/d3d/Raytracing.md#opacity-micromaps-and-traceray):

> A new pipeline flag, D3D12_RAYTRACING_PIPELINE_FLAG_ALLOW_OPACITY_MICROMAPS, has been added to let the driver
> know that the opacity micromap feature is going to be used.
> ...
> If the flag is omitted and an OMM is encountered during ray traversal, behavior is undefined.

`RayQuery`: [per-`RayQuery` template parameter](https://github.com/microsoft/DirectX-Specs/blob/master/d3d/Raytracing.md#opacity-micromaps-and-tracerayinline):

> For inline raytracing via `RayQuery` objects, the `RayQuery` object takes an additional template parameter 
> (as of Shader Model 6.9) allowing each instance of the object to indicate that the opacity micromap feature 
> is going to be used.
> ...
> If the flag is omitted and an OMM is encountered during ray traversal, behavior is undefined.

Two parallel sections, no cross-reference.

## Vulkan / VK_KHR_opacity_micromap

`TraceRay`: [pipeline-create flag (VUID-11640)](https://github.com/KhronosGroup/Vulkan-Docs/blob/main/chapters/commonvalidity/trace_rays_common_khr.adoc?plain=1):

> If a `VK_DESCRIPTOR_TYPE_ACCELERATION_STRUCTURE_KHR` descriptor is accessed as a result of this command,
> the access is not through an `OpRayQueryProceedKHR` operation,  and the acceleration structure is built using 
> geometry with a `VkAccelerationStructureTrianglesOpacityMicromapKHR` structure included in its `pNext` chain, 
> the `VkPipeline` bound to the pipeline bind point used by this command must: have been created
> with `flags` that includes `VK_PIPELINE_CREATE_RAY_TRACING_OPACITY_MICROMAP_BIT_KHR`.

VUID-11640 only lives in the trace_rays chapter and explicitly excludes `OpRayQueryProceedKHR` access.

`RayQuery`: [SPIR-V execution mode (VUID-11636)](https://github.com/KhronosGroup/Vulkan-Docs/blob/main/chapters/commonvalidity/draw_dispatch_common.adoc?plain=1):

> If a `VK_DESCRIPTOR_TYPE_ACCELERATION_STRUCTURE_KHR` descriptor is accessed as a result of this command through an
> `OpRayQueryProceedKHR` operation, the `VK_EXT_opacity_micromap` extension is not enabled, and the acceleration 
> structure is built using geometry with a `VkAccelerationStructureTrianglesOpacityMicromapKHR` structure
> included in its `pNext` chain, the shader must: have specified a `OpacityMicromapIdKHR` execution mode 
> with its specialization constant operand equal to `VK_TRUE`.

VUID-11636 lives in draw, dispatch, and trace_rays chapters - it covers RayQuery in fragment, compute, and RT
shaders. Same rule everywhere. No pipeline-create flag in the VUID.

The two VUIDs are mutually exclusive on access type. The pipeline-create flag does not cover `RayQuery`. Even
inside an RT pipeline that has `VK_PIPELINE_CREATE_RAY_TRACING_OPACITY_MICROMAP_BIT_KHR`, any `RayQuery` that
needs OMM must independently specify the `OpacityMicromapIdKHR` execution mode.

## The rules are identical in shape between D3D12 and Vulkan

| Path | D3D12 / DXR 1.2 | Vulkan KHR |
|---|---|---|
| `TraceRay` (RT pipeline only) | `D3D12_RAYTRACING_PIPELINE_FLAG_ALLOW_OPACITY_MICROMAPS` on the RTPSO. Missing with OMM AS -> UB. | `VK_PIPELINE_CREATE_RAY_TRACING_OPACITY_MICROMAP_BIT_KHR` on the RT pipeline. Missing with OMM AS -> VUID-11640. |
| `RayQuery` in RT shader | Per-`RayQuery` `RAYQUERY_FLAG_ALLOW_OPACITY_MICROMAPS` (SM 6.9+). Missing with OMM AS -> UB. | Per-shader `OpacityMicromapIdKHR` execution mode with spec-const `VK_TRUE`. Missing with OMM AS -> VUID-11636. |
| `RayQuery` in fragment/compute (no RTPSO) | Same per-`RayQuery` template parameter. RTPSO flag doesn't exist for these pipelines. | Same `OpacityMicromapIdKHR` execution mode. Pipeline-create flag (VUID-11640) doesn't apply outside trace_rays. |

Pipeline-level opt-in is RT-pipeline-only in both APIs. Per-shader opt-in is univers
APIs. RayQuery is fully independent of any RTPSO flags wrt the OMM in both APIs.

2. `RayQuery` UB resolution: spec defers it, we pick.

Per spec, `RayQuery` without the OMM opt-in is UB in both APIs. D3D12 says so verbatim in `Raytracing.md` "Opacity
Micromaps and TraceRayInline". Vulkan VUID-11636 requires `OpacityMicromapIdKHR` execution mode with spec-const
`VK_TRUE` for any RayQuery on an OMM-bearing AS - absence (or spec-const `VK_FALSE`) is a VUID violation. RTPSO
flags don't enter into this on either API; this is purely a RayQuery opt-in question.

NVIDIA's D3D12 native driver doesn't resolve this the same way across versions:

- **595 driver:** RayQuery without the OMM flag evaluates the OMM.
- **610 driver:** RayQuery without the OMM flag ignores the OMM, unless SM < 6.9 (then it still consults). The
  driver comments say this is NVAPI backwards compat for pre-SM-6.9 titles.

vkd3d-proton is a separate implementation. We pick our own policy. The spec-aligned (VUID violation) default is
`OpacityMicromapIdKHR = VK_FALSE` ("do not evaluate"), but that breaks NVAPI titles like BMW and CP2077 that ship
against NVIDIA's UB-consulting behavior and don't supply the per-RayQuery flag.

My proposal: tie the RayQuery UB default to the NVAPI `ENABLE_OMM_SUPPORT` opt-in. When the app sets the NVAPI
flag (on any RTPSO under this device), default `OpacityMicromapIdKHR = VK_TRUE` for all RayQuery shaders compiled
under that device. When the app doesn't set it, default to `VK_FALSE`.

This is technically a lie. The NVAPI flag is the TraceRay opt-in - it maps to
`VK_PIPELINE_CREATE_RAY_TRACING_OPACITY_MICROMAP_BIT_KHR` - and per the spec the two opt-ins are independent. But
pragmatically it matches what NVAPI titles expect: opt in via NVAPI => OMM consulted everywhere. It generalizes
across vendors. And it doesn't pin vkd3d-proton to a Windows driver version.

This replaces the current vendor-keyed `ray_query_force_omm_execution_mode=TRUE` default-on for NVIDIA with an
app-intent-keyed default. The existing `ray_query_omm_when_unspecified` / `no_ray_query_omm_when_unspecified`
config flags stay as manual overrides.

Minor concern: shader-compile timing. The NVAPI flag is sticky-global in
`vendor_hacks.global_ray_tracing_pipeline_create_flags`. A RayQuery shader compiled before the app calls
`NvAPI_D3D12_SetCreatePipelineStateOptions` would bake `OpacityMicromapIdKHR = VK_FALSE` and not consult OMM at
runtime even after the flag is set later. I'm going to test BMW and CP2077. If both work this is academic and I
move forward. If either is broken I need a different mechanism.

3. A residual gap, probably theoretical.

NVIDIA's implementation agrees with the natural "no flag = no OMM" baseline on both APIs. On DXR 1.2 SM 6.9+ with
the per-RayQuery flag absent, the 610 driver does not evaluate OMM. On Vulkan KHR with `OpacityMicromapIdKHR`
spec-const `VK_FALSE`, the 620 driver (first to officially support VK_KHR_opacity_micromap) does not evaluate
either. It's still UB per Vulkan spec text but vendors may de-facto converge here. The policy above bets on that
convergence. May need to revisit if a divergent vendor surfaces.

The residual spec gap no title appears to exercise: there's no clean mechanism for a single shader to toggle OMM
consultation between different RayQuery invocations. The `OpacityMicromapIdKHR` execution mode is per-shader,
baked at pipeline create. If a game wants OMM-consulting and non-OMM-consulting ray casts in the same shader,
the spec gives no per-call switch.

No game I know of requires this and even if one shows up it will be rare. Hopefully Khronos adds a resolution
before that happens. As a worst-case fallback vkd3d-proton could hijack an unused bit in the instance mask to
route rays to OMM-enabled vs non-OMM instances - in theory this would work but has significant performance and
memory costs and only applies when the game leaves an instance-mask bit unused. Better to push for a proper
Vulkan-side mechanism than to deploy this preemptively.

chrisd-work:
Ok two findings with the proposed change

BMW works
CP2077, I never dug this deeply but AFICT it only uses TraceRay.

chrisd-work:
removed out of draft. Reworked code based on the changes you made @HansKristian-Work . This is the best compromise I think we can achieve for now.

HansKristian-Work (approved):
Technically a bit questionable w.r.t. shader caching interactions, but doubt it will ever matter in practice.

PR #3138, opened 2026-07-01
vkd3d: Unwrap potentially hooked queue interface before assuming that is our object.

gofman:
That fixes HITMAN World of Assassination (1659040)  in VR mode which otherwise crashes. The queue which is passed through VR texture structure for IVRCompositor::Submit() is wrapped by sl.interposer, so using that as vkd3d-proton object in GetVulkanQueueInfo() doesn't work very well. Meanwhile, any sane wrapper would pass any unknown interface query to the wrapped object (like it happens in this case) and so we get our queue.

It could be made more idiomatic by, e. g., moving the queue related methods to a new queue extension interface, but it looks to me using an universal "canary" guid reduces a lot of unneeded boilerplate, the same uuid can be used for other objects once needed.

### DIFF
--- a/include/vkd3d_device_vkd3d_ext.idl
+++ b/include/vkd3d_device_vkd3d_ext.idl
@@ -221,3 +221,16 @@ interface IAmdExtAntiLagApi : IUnknown
 {
     HRESULT UpdateAntiLagState(void *pData);
 }
+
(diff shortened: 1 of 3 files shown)


HansKristian-Work (approved):
Annoying, but ok.

PR #3140, opened 2026-07-02
vkd3d: Disable suballocation when EXT_pageable_device_local_memory is supported

pixelcluster:
Apps having full control over their priorities can significantly improve perf (as well as making good perf not a matter of luck) when VRAM ends up being contended. Suballocation removes the ability to set prios, so don't do suballocation when prios are relevant.

Marked as draft because while I think this could be useful, I want to be more sure it doesn't regress gameplay experience if apps spam allocations.


HansKristian-Work (approved):
LGTM. The major concerns with not suballocating image heaps should have been addressed by newer kernels anyway.

HansKristian-Work (changes requested):
Actually, I think it makes more sense to just skip the ALLOW_IMAGE_SUBALLOCATION flags when pageable is set rather than filtering it out here.

doitsujin (approved):

PR #3143, opened 2026-07-02
Add mechanism to do per-draw feedback loop workarounds

HansKristian-Work:
Should address a bad SOTTR workaround more cleanly than just disabling compression across the board.


doitsujin (approved):
cursed

PR #3144, opened 2026-07-03
vkd3d: Add 6_9 to VKD3D_SHADER_MODEL override

jp7677:
Allows to spoof shader model 6.9 with `VKD3D_SHADER_MODEL`. 
Value taken from: https://github.com/microsoft/DirectX-Headers/blob/5c0f4bace522a37fda02421a97ccaa8a0d6a739a/include/directx/d3d12.idl#L1325

I just wanted to know how far I can get with starting https://github.com/NVIDIA-RTX/RTXPT/releases/tag/v.1.8.1 with just pretending SM 6.9, since the demo otherwise just bails out early with a message box that SM 6.9 is needed.

(Turns out it wants `Long Vector`, `vkd3d_dxil_log_callback: dxil-spirv: Long vector is not supported.`)


jp7677:
looks like CI is having a bad day:

```
ERROR: failed to build: failed to solve: archlinux:base-devel: failed to resolve source metadata for docker.io/library/archlinux:base-devel: unexpected status from HEAD request to https://registry-1.docker.io/v2/library/archlinux/manifests/base-devel: 500 Internal Server Error
```

HansKristian-Work (approved):

oscarbg:
tested also PT SDK 1.8.1 demo with same result: long vector missing..
curious about:

1) VK_EXT_shader_long_vector is enough to implement SM6.9 long vector support?.. for this VK ext there is NV support and AMD Windows support but not yet RADV sadly.. so I asked for it:
https://gitlab.freedesktop.org/mesa/mesa/-/work_items/16071

2) curious to know if @chrisd-work who has brought us other SM6.9/DXR1.2 features like SER and OMM is interested to work on it eventually..

chrisd-work:
>     2. curious to know if @chrisd-work who has brought us other SM6.9/DXR1.2 features like SER and OMM is interested to work on it eventually..

I work with peterkohaut, who has been working a lot on patches for DXIL-spriv to support SM 6.9. Their expertise on these topics is much better than mine. I am still working on Linux gaming and will def return to supporting VKD3D from time to time when it makes sense.

oscarbg:
thanks @chrisd-work for answering.. I see now @peterkohaut has worked recently on SM6.9 SER support (https://github.com/peterkohaut/dxil-spirv/tree/sm69ser).. so hope he is interested on this remaining work..

peterkohaut:
All changes related to SER were merged to dxil-spirv. For vkd3d-proton I have https://github.com/HansKristian-Work/vkd3d-proton/compare/master...peterkohaut:vkd3d-proton:ser that I need to rebase and file a PR.
Long vector is a problem, I'm trying to hack support for VK_EXT_shader_long_vector, but that will not work with MESA as mentioned. I'm also checking breaking long-vector operation to smaller operation on 1-4 components, but that is much more work.

chrisd-work:
I wanted to confirm something that I suspected and was told in passing. I have confirmed that if you compile an otherwise normal SM 6.8 shader using DXC in SM 6.9, it will use long vectors. This means that in order to even have the most baseline support for an SM 6.9 game, long vector support will be required. E.g. a float4 `RawBufferLoad` becomes a `RawBufferVectorLoad`.

Per @peterkohaut, you might think that supporting only 1-4 components gets you in the clear. It doesn't. Consider the following very simple shader:

```
StructuredBuffer<float4x4> In : register(t0);
RWStructuredBuffer<float4x4> Out : register(u0);

[numthreads(1,1,1)]
void main(uint3 id : SV_DispatchThreadID)
{
    float4x4 a = In[0];
    float4x4 b = In[1];
    float4x4 c = a + b;
    Out[0] = c;
}
```

It uses a RawBufferVectorLoad to load a and b. It then uses fadd to add them, and then a RawBufferVectorStore to store the result. Each operation is done as a 16-component long vector.

This means that to support SM 6.9, long vector support likely has to be complete or near complete in order to consider it complete.

However, in the case of a machine that doesn't support `VK_EXT_shader_long_vector`, long vectors could be lowered into traditional operations. It will bloat the generated SPIR-V significantly, but it may be possible.

PR #3157, opened 2026-07-10
Add World of Warcraft workarounds to Classic as well and enable descriptor heap for both Retail and Classic

runar-work:
The same ray tracing GPU hang in retail WoW turns up in Classic as well, so it needs the same workarounds.


HansKristian-Work (approved):
hng, lgtm.

PR #3160, opened 2026-07-12
vkd3d: Add DirectStorage integration API

infinityabundance:
## Summary

Adds 7 new exported functions to d3d12.dll that enable dstoragecore.dll (Wine's DirectStorage implementation, see wine-mr/11371) to perform GPU-accelerated GDeflate decompression directly against Vulkan resources.

## New Exports

| Function | Purpose |
|---|---|
| `vkd3d_dstorage_get_version` | API version check (0x0002000D = v2.13) |
| `vkd3d_dstorage_get_vk_buffer` | Extract VkBuffer + GPU address from ID3D12Resource |
| `vkd3d_dstorage_get_vk_device` | Return VkDevice backing an ID3D12Device |
| `vkd3d_dstorage_get_compute_queue` | Get a compute-capable VkQueue (async > internal > primary) |
| `vkd3d_dstorage_export_dma_buf` | Export VkBuffer memory as Linux dma-buf fd for zero-copy io_uring I/O |
| `vkd3d_dstorage_signal_fence` | Signal a D3D12 fence's underlying Vulkan timeline semaphore |
| `vkd3d_dstorage_submit_compute` | Submit a command buffer with GDeflate decompression dispatches |

## Implementation Notes

- All functions access existing vkd3d-proton internal structs (`d3d12_resource`, `d3d12_device`, `d3d12_fence`) via `impl_from_*` helpers
- No new Vulkan extensions required — reuses existing device/queue/fence infrastructure
- dma-buf export uses `VK_KHR_external_memory_fd` + `VK_EXT_external_memory_dma_buf`, both already supported by vkd3d-proton
- The cs_gdeflate.comp compute shader (already shipped) is dispatched via `vkd3d_dstorage_submit_compute`

## Dependencies

- Wine MR !11371 (dstorage/dstoragecore PE DLLs)

https://gitlab.winehq.org/wine/wine/-/merge_requests/11371

- vkd3d-proton 2.13+ (cs_gdeflate.comp available since 2.13)
https://github.com/HansKristian-Work/vkd3d-proton/blob/master/libs/vkd3d/shaders/cs_gdeflate.comp

### DIFF
--- a/libs/d3d12/d3d12.def
+++ b/libs/d3d12/d3d12.def
@@ -10,3 +10,12 @@ EXPORTS
     D3D12SerializeRootSignature
     D3D12SerializeVersionedRootSignature
     D3D12GetInterface
+
+    ; DirectStorage integration (vkd3d_dstorage)
+    vkd3d_dstorage_get_version
+    vkd3d_dstorage_get_vk_buffer
+    vkd3d_dstorage_get_vk_device
+    vkd3d_dstorage_get_compute_queue
+    vkd3d_dstorage_export_dma_buf
+    vkd3d_dstorage_signal_fence
+    vkd3d_dstorage_submit_compute
--- a/libs/vkd3d/meson.build
+++ b/libs/vkd3d/meson.build
@@ -82,6 +82,7 @@ vkd3d_src = [
   'debug_ring.c',
   'va_map.c',
   'vkd3d_main.c',
+  'vkd3d_dstorage.c',
   'raytracing_pipeline.c',
   'state_object_common.c',
   'acceleration_structure.c',
--- a/libs/vkd3d/vkd3d_dstorage.c
+++ b/libs/vkd3d/vkd3d_dstorage.c
@@ -1,0 +1,534 @@
+/*
+ * vkd3d_dstorage.c — vkd3d-proton DirectStorage integration implementation
+ *
+ * This file implements the functions declared in vkd3d_dstorage.h.
+ * It is designed to be compiled into vkd3d-proton's d3d12.dll.
+ *
+ * Integration with vkd3d-proton internals:
+ *   This file accesses the following vkd3d-proton internal structures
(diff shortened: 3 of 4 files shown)


infinityabundance:
https://gitlab.winehq.org/wine/wine/-/merge_requests/11372

doitsujin:
This entire concept seems backwards.

First of all, I don't think you're going to have a huge amount of luck getting a wine PR merged that hard-requires software developed *outside of wine* to even work. That's just not how things work over there, and before dumping massive feature PRs on people out of nowhere, it would be nice to at least... discuss the general direction of things with the respective maintainers?

Secondly, I don't see *a single good reason* why we need *another* interop API for queue submissions etc here when
- a) all of these things should be happening on the D3D12 side. Native DStorage doesn't need any of this; the metacommand API exists and you can always record a proper D3D12 command buffer and dispatch a custom shader.
- b) the API cements implementation details in such a way that further development in the affected areas essentially becomes impossible (not like we've rewritten the whole fence / queue sync stuff 10 times already....), but fucks even that up because non-shared fences *do not actually have* a timeline semaphore associated with them. Was this even tested with anything at all?
- c) we *already have* a bunch of APIs to query device handles and do external queue submissions in the shape of COM interfaces.

<sub>and d) I have a strong suspicion that an undisclosed amount of LLM usage was at work here.</sub>

Anyway, this whole thing is obviously not going to work as-is.

infinityabundance:
Thank you for the feedback.

mati865:
Just in case this wasn't clear, Wine has a strict no-go policy for LLM code like MRs mentioned in the description.

HansKristian-Work:
Agreed that further interop APIs are absolutely not needed for DirectStorage since normal DirectStorage does not need it and even then we have lots of interop stuff already. This is a no-go, and not going to bother trying to review this into something shippable.

PR #3166, opened 2026-07-24
Add UE configs for new Halo Campaign Evolved

Astyyyyy:
New halo game is running under unreal engine and does not use the standard win64-shipping naming scheme for exe


HansKristian-Work (approved):

HansKristian-Work (changes requested):
Did you try to build this? You're missing a "," at the end of line 733.

HansKristian-Work (approved):
Fixed the issue inline.

PR #3181, opened 2026-07-30
vkd3d: Intel vendor hack DLL fix

damian-tomczak-intel:
.inf file for an Intel device in Wine is named igd_faux.inf


damian-tomczak-intel:
https://github.com/ValveSoftware/wine/commit/4c53233bc9dd083509198dfba8a3afc9bfd80d3b

PR #3184, opened 2026-07-30
android: allow native Android build using NDK

olvaffe:
This is enough to run headless tests. It serves the same purpose as the native linux build and is for development purpose only.


HansKristian-Work (approved):
Guess it's fine, but it's not really a supported config.

PR #3200, opened 2026-08-06
vkd3d: Add a workaround to limit GPU timeouts in Farming Simulator 25.

HansKristian-Work:
On loading some 4x maps, the game ends up spamming well over 100k copy commands in a single command list. On some weaker GPU, this causes a GPU timeout. There seems to be no observable effect from dropping copies beyond a certain limit. It's unclear if there exists a better fix.

Proton: Farming Simulator 25 (2300320)


doitsujin:
Only thing I can think of as a more "proper" fix would be to split not only the command list but also submissions, which is... probably not all that viable currently.

Do we know what the excess copies even do? Wondering if we're breaking small mips or something.

HansKristian-Work:
Superseded by #3203.

Issue #3134, opened 2026-06-29
IL-2 Korea - Textures not loading in-game


r3animate:
Game launches okay (minor pixelated squares in menus, issue with film grain filter maybe) but when loading into a map, textures on the ground don't display properly. Attached log, end of log seems to have a lot of repeat lines with :

6827.358:02c4:02c8:warn:vkd3d-proton:d3d12_command_list_ResourceBarrier: Issuing split barrier(s) on D3D12_RESOURCE_BARRIER_FLAG_END_ONLY.

and 
8370.662:0344:0484:warn:debugstr:OutputDebugStringA "error: no saver for property: FieldName = Flags, SubPropertyName = Block, FieldValue = 0\n"
8370.662:0344:0484:warn:seh:dispatch_exception "error: no saver for property: FieldName = Flags, SubPropertyName = Block, FieldValue = 0\n"

seems present throughout

If this is redundant to the main Proton issues thread(https://github.com/ValveSoftware/Proton/issues/9906) I apologize though I suspect this issue is VK3D specific as the main thread covers other issues (wine related).

### Software information
Name of the game, settings used etc.
Il-2 Korea, any settings but tested on high

### System information
- GPU: 9070XT
- Driver: mesa 3:26.1.3-1
- Wine version: 11
- VKD3D-Proton version: 1782735980 experimental-bleeding-edge-11.0-387225-20260629-p7230a5-w7e906e-d60f032-v074c5b

### Log files
Edit: (corrected log upload)
[steam-12814649275476606976.tar.gz](https://github.com/user-attachments/files/29466992/steam-12814649275476606976.tar.gz)

<img width="2554" height="1440" alt="Image" src="https://github.com/user-attachments/assets/1bb3ce83-9dfc-4401-89e1-f039683213e2" />

silv3rshi3ld:
I reproduced this on an RX 7800 XT with Mesa/RADV 26.1.6. single_queue,
disabling descriptor buffers, and an unmodified current VKD3D-Proton build do
not change the corruption. The traced mission path uses no reserved-resource
or tile-mapping calls, and the placed resources checked so far do not overlap.

The strongest lead is an invalid-copy pattern on the active terrain cache. The
game creates 164 placed 2048x2048 one-mip BC3 textures. A trace recorded 432
narrow uploads to this pool:

- 118 x 128x1
- 112 x 1x128
- 110 x 64x1
- 92 x 1x64

[...]

PR #3202, opened 2026-08-07
vkd3d: Convert buffer-image copies between block formats

silv3rshi3ld:
Addresses #3134.

## What changed

Convert placed-buffer footprint geometry through physical block counts only
when the footprint and destination image formats have the same physical
element size but different block dimensions. This expresses
`bufferRowLength`, `bufferImageHeight`, and `imageExtent` in destination-image
texels before emitting the Vulkan buffer-to-image copy. Formats whose block
geometry already matches remain on the original path.

A focused regression test covers a `64x64 R32G32B32A32_UINT` placed footprint
copied to a `256x256 BC3_UNORM` texture.

## Root cause

IL-2 Korea (Steam AppID 247970) uploads terrain-cache pages with that format
pattern. Both physical elements are 16 bytes, so one source texel represents
one 4x4 BC3 block. The old path retained the source-format geometry and copied
only a 64x64 BC3 region, leaving most of each 256x256 terrain page unpopulated.

VKD3D-Proton already converts the corresponding compressed/uncompressed
image-to-image copies through physical blocks. This patch adds the demonstrated
placed-buffer-to-image case. It contains no executable or AppID check.

## Validation

- Focused test: four deterministic failures on the old helper; 22/22
  assertions pass with this change.
- `VKD3D_TEST_FILTER=copy`: 6,429,713 tests executed, zero failures.
- Existing neighboring compressed-copy tests: 147/147 and 50/50 pass.
- Native and MinGW x64 test builds compile; x64 and x86 packages build.
- A gated diagnostic adjusted 522/522 observed IL-2 terrain page/border copies
  with zero rejects.
- A clean build of the general conversion, without the diagnostic gate,
  restores continuous terrain at 4,813 m, 2,427 m, and 742 m on RX 7800 XT /
  RADV 26.1.6. The narrowed condition selects the same conversion for this
  `1x1`-to-`4x4` block geometry.
- No device loss, GPU hang/reset, or OOM signature was observed.

Repaired terrain at 742 m

The separate main-menu block/shimmering artifact and OpenMP/NUMA startup issue
remain unresolved.

Investigation notes and reproducible evidence:
https://github.com/silv3rshi3ld/il2-korea-proton-investigation

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -9634,8 +9634,8 @@ static void vk_buffer_image_copy_from_d3d12(VkBufferImageCopy2 *copy,
         const struct vkd3d_format *dst_format, const D3D12_BOX *src_box, unsigned int dst_x,
         unsigned int dst_y, unsigned int dst_z)
 {
-    VkExtent3D src_extent;
     bool convert_block_geometry;
+    VkExtent3D src_extent;
 
     /* D3D12 describes the buffer layout in source-format texels, while Vulkan
      * describes it in destination image texels. Convert through physical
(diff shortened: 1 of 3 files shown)


HansKristian-Work:
Nice find. Seems reasonable at a first glance, but I'll do a more detailed review on Monday and merge it after doing some minor nits I'm not going to bother you with.

silv3rshi3ld:
> Nice find. Seems reasonable at a first glance, but I'll do a more detailed review on Monday and merge it after doing some minor nits I'm not going to bother you with.

Thanks, really appreciate that! 

This one took a fair bit of digging, so I’m pretty stoked to see it heading towards a merge. 

I’ve done plenty of PRs professionally, but I don’t get to contribute to open source very often, so I’m especially thrilled about this one. Looking forward to the review on Monday, if you happen to post any notes publicly. 

Have a great weekend!

HansKristian-Work (approved):

PR #3203, opened 2026-08-07
Optimize some completely degenerate copy patterns in Farming Simulator 25

HansKristian-Work:
Native drivers (especially NV) do absolute degenerate shit to make this code fast, so our hand is forced to be just as degenerate to claw back a 80x speed boost on NV for some very stupid copy patterns.

A test case is added that attempts to spam the GPU really hard. On some GPUs, this can hit TDRs if not sufficiently optimized, so the test is more of a "does this hang the GPU" kind of thing. It also functions as a benchmark to validate the implementation uplift. It's ~80x on NV and ~8x on RADV. Both Vulkan drivers are now on-par or better than the native driver.

Second part is to add an image rect tracker to hazard tracking. Before we used to only track on a subresource level, but that's clearly a bit too naive.

The final part is a rewrite of how RESOURCE states are synchronized with COPY_DEST. To combat real world shitty game code, we can defer any RESOURCE <-> COPY_DEST barriers and instead resolve them as needed.
Fortunately, the image layout for RESOURCE and COPY_DEST are the same even without unified layouts, so we only have to consider memory barriers.

For COPY_DEST -> RESOURCE, we resolve it in end_transfer_barrier by marking the end_transfer_barrier as needing shader resource access. The RAW hazard is resolved there.

For RESOURCE -> COPY_DEST, we add a new "begin_transfer" call to every COPY command. Here, we perform a WAR barrier as needed. We have to be a bit careful, since RESOURCE -> COPY -> RESOURCE barrier roundtrips would re-arm that WAR barrier. We only need to do the barrier if we have observed shaders actually being executed that would access RESOURCES. We abuse the end_transfer_barrier() flag to mark that there are potential threads in flight that might be reading from a resource.

Overall, the implementation isn't as complicated as it sounds I think, and should better align with the optimizations that at least NV seem to employ in their drivers.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -42,7 +42,7 @@ static HRESULT d3d12_fence_signal_cpu_timeline_semaphore(struct d3d12_fence *fen
 
 /* This must be at least twice the number of texture region batches, since we must be able to resolve
  * a source + destination memory barrier per copy without incurring a barrier flush.
- * A barrier flush breaks our ability to eliminate redundant transitions for e.g. source copies. */
+ * A barrier flush breaks our ability to eliminate redundant transitions for e.g. source and dest copies. */
 #define MAX_BATCHED_IMAGE_BARRIERS (VKD3D_COPY_TEXTURE_REGION_MAX_BATCH_SIZE * 2)
 struct d3d12_command_list_barrier_batch
 {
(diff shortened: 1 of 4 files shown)


doitsujin (approved):
Makes sense with the whole "transfer pass" approach, just annoying that it's *yet another* thing that requires handling in almost every single entry point.

PR #296, opened 2020-10-16
Support raw SSBO path for root descriptors

doitsujin:
Only on hardware where `minStorageBufferAlignment <= 4`, since we know nothing about root descriptors except their VA.

PR #3207, opened 2026-08-10
vkd3d-shader: Work around IL-2 tiled-light allocator

silv3rshi3ld:
> [!IMPORTANT]
> This is a dependent draft. Its compiler-aware VKD3D-Proton integration depends on https://github.com/HansKristian-Work/dxil-spirv/pull/296 and pins its current public commit `afff4dfb3e51ab81a4d541011bcf7ec2f65e2ffa`. The dependency is not merged, so this PR remains draft and may need matching API adjustments during review.

## Summary

Fix the square lighting blocks and broad flicker reported in #3134 for IL-2: Korea.

The game performs a scalar 32-bit atomic through an `R16_UINT` typed UAV. The revised implementation separates mechanism from application policy:

- dxil-spirv generically lowers eligible `I32` and `U32` typed-UAV atomics.
- It temporarily requests raw-buffer remapping and accepts the lowering only when the remapper returns an SSBO.
- On failure or a non-SSBO result, it restores the typed binding and retries normally.
- It does not extend any public C callback structure.
- It excludes 64-bit atomics, sparse operations, non-atomic resources, and the SM 6.6 heap path.
- VKD3D-Proton enables the compiler quirk only for exact executable `IL2Series.exe`, shader `0x7cefa1bc80bb4c70`, and layouts satisfying `RAW_SSBO && !MUTABLE_TYPE_RAW_SSBO`.

The earlier `9b6e15be` implementation remains useful causal and visual evidence, but it is not the implementation proposed for merging.

## Validation

- Clean dxil-spirv resources suite.
- Full suite with only the baseline-reproduced `control-flow/switch-continue.frag` validator failure.
- Valid candidate, fallback, and baseline outputs for the private captured shader.
- Clean VKD3D-Proton x86-64 and x86 builds and complete package build.
- Remapper harness: capability ON produces `StorageBuffer` plus `OpAtomicIAdd`; capability OFF is byte-identical to the typed baseline.
- One verified-loaded RX 7800 XT/RADV game run with correct menu, short flight, terrain, map, lighting, and shadows, without the blocks or broad flicker.

The native-Windows film-grain effect is excluded. This is not yet cross-hardware validation.

## Before and after

The baseline still catches a relatively faint frame. The square blocks flashed and became more pronounced during normal runtime.

| Before, unmodified VKD3D-Proton | After, D49 compiler-aware candidate |
| --- | --- |
| Faint captured frame of the flashing tiled-light blocks | D49 result without tiled-light blocks |

Exact D49 provenance and evidence: https://github.com/silv3rshi3ld/il2-korea-proton-investigation/blob/main/docs/evidence-d49-compiler-aware-result.md

## Separate work

The terrain-page corruption was fixed independently by #3202. The Wine NUMA startup work is also independent and is not part of this PR.

### DIFF
--- a/include/vkd3d_shader.h
+++ b/include/vkd3d_shader.h
@@ -267,6 +267,7 @@ enum vkd3d_shader_interface_flag
     VKD3D_SHADER_INTERFACE_INLINE_REDZONE_CBV                  = 0x00002000u,
     VKD3D_SHADER_INTERFACE_RAYTRACING_OPACITY_MICROMAP         = 0x00004000u,
     VKD3D_SHADER_INTERFACE_RAY_QUERY_OMM_DEVICE_GLOBAL         = 0x00008000u,
+    VKD3D_SHADER_INTERFACE_TYPED_BUFFER_RAW_SSBO_ALIAS         = 0x00010000u,
 };
 
 struct vkd3d_shader_stage_io_entry
@@ -578,6 +579,9 @@ enum vkd3d_shader_target_extension
  * Used to workaround games where an image is rendered to while being sampled as an SRV,
  * which is highly illegal and only way out of it is to use filthy trickery. */
 #define VKD3D_SHADER_QUIRK_FORCE_FEEDBACK_LOOP (1ull << 38)
+
+/* Converts 32-bit atomics on typed UAVs to raw buffer atomics. */
+#define VKD3D_SHADER_QUIRK_TYPED_UAV_ATOMIC_AS_RAW (1ull << 39)
 
 typedef uint64_t vkd3d_shader_quirks_t;
 
--- a/libs/vkd3d-shader/dxil.c
+++ b/libs/vkd3d-shader/dxil.c
@@ -710,6 +710,15 @@ static bool vkd3d_dxil_converter_set_quirks(dxil_spv_converter converter,
         }
     }
 
+    if ((quirks & VKD3D_SHADER_QUIRK_TYPED_UAV_ATOMIC_AS_RAW) &&
+            (shader_interface_info->flags & VKD3D_SHADER_INTERFACE_TYPED_BUFFER_RAW_SSBO_ALIAS))
+    {
+        const dxil_spv_option_shader_quirk helper =
+                { { DXIL_SPV_OPTION_SHADER_QUIRK }, DXIL_SPV_SHADER_QUIRK_TYPED_UAV_ATOMIC_AS_RAW };
+        if (dxil_spv_converter_add_option(converter, &helper.base) != DXIL_SPV_SUCCESS)
+            return false;
+    }
+
     if (quirks & VKD3D_SHADER_QUIRK_FORCE_LOOP)
     {
         struct dxil_spv_option_branch_control helper = { { DXIL_SPV_OPTION_BRANCH_CONTROL } };
--- a/libs/vkd3d/device.c
+++ b/libs/vkd3d/device.c
@@ -1113,6 +1113,15 @@ static const struct vkd3d_shader_quirk_info empire_of_the_ants_quirks = {
     NULL, 0, VKD3D_SHADER_QUIRK_CLAMP_WAVE_SIZE_TO_THREAD_GROUP32,
 };
 
+static const struct vkd3d_shader_quirk_hash il2_korea_hashes[] = {
+    /* ComputeLightsFirstRef uses a 32-bit atomic through an R16_UINT UAV. */
+    { NULL, 0x7cefa1bc80bb4c70, VKD3D_SHADER_QUIRK_TYPED_UAV_ATOMIC_AS_RAW },
+};
+
+static const struct vkd3d_shader_quirk_info il2_korea_quirks = {
+    il2_korea_hashes, ARRAY_SIZE(il2_korea_hashes), 0,
+};
+
 static const struct vkd3d_shader_quirk_meta application_shader_quirks[] = {
     /* F1 2020 (1080110) */
     { VKD3D_STRING_COMPARE_EXACT, "F1_2020_dx12.exe", &f1_2019_2020_quirks },
@@ -1207,6 +1216,8 @@ static const struct vkd3d_shader_quirk_meta application_shader_quirks[] = {
     /* Empire of the Ants (2287330). With Terrain Tessellation on, some shaders
      * seem to assume Wave32 by mistake leading to GPU timeout. */
     { VKD3D_STRING_COMPARE_EXACT, "Empire-Win64-Shipping.exe", &empire_of_the_ants_quirks },
+    /* IL-2 Sturmovik: Korea (247970). */
+    { VKD3D_STRING_COMPARE_EXACT, "IL2Series.exe", &il2_korea_quirks },
     /* Unreal Engine 4 */
     { VKD3D_STRING_COMPARE_ENDS_WITH, "-Shipping.exe", &ue4_quirks },
     { VKD3D_STRING_COMPARE_NEVER, NULL, NULL },
--- a/libs/vkd3d/state.c
+++ b/libs/vkd3d/state.c
@@ -2290,6 +2290,9 @@ unsigned int d3d12_root_signature_get_shader_interface_flags(const struct d3d12_
         flags |= VKD3D_SHADER_INTERFACE_HEAP_LOWERING;
     if (root_signature->heap.redzone_style == VKD3D_ROOT_SIGNATURE_HEAP_REDZONE_STYLE_INLINE)
         flags |= VKD3D_SHADER_INTERFACE_INLINE_REDZONE_CBV;
+    if (d3d12_device_use_ssbo_raw_buffer(root_signature->device) &&
+            !(root_signature->device->bindless_state.flags & VKD3D_BINDLESS_MUTABLE_TYPE_RAW_SSBO))
+        flags |= VKD3D_SHADER_INTERFACE_TYPED_BUFFER_RAW_SSBO_ALIAS;
 
     if (vkd3d_atomic_uint32_load_explicit(&root_signature->device->vendor_hacks.force_ray_query_omm,
             vkd3d_memory_order_relaxed))


HansKristian-Work:
This is uh ... quite something. I don't see how this PR would work as-is though. dxil-spirv does not know how to interpret a texel buffer as an SSBO, so this should not work, at least not on most hardware.

I'll have to verify this bug myself first.

silv3rshi3ld:
Thanks, that is a fair concern. I only validated this on an RX 7800 XT with RADV, where the visual A/B result was repeatable. I should have made that limited scope clearer in the PR.

I will wait for your verification before changing anything. Even if the implementation needs to change, I hope the identified shader and captured before-and-after behavior are still useful. I can provide the shader dumps, SPIR-V files, or capture details if needed.

HansKristian-Work:
Finding the root cause is most of the battle, so that's appreciated even if the suggested fix won't make it in as-is.

silv3rshi3ld:
I had another thought after looking more closely into your concern.

If your native testing confirms that the R16_UINT mismatch is the trigger, would it make more sense to handle this as an explicit typed-atomic-to-raw legalization in dxil-spirv, similar in spirit to dxbc-spirv's `useRawForTypedAtomic`, while keeping vkd3d-proton limited to enabling it for this exact shader?

That would let the compiler understand the resource conversion instead of only changing the descriptor mapping.

Another possibility might be an R32_UINT texel-buffer alias, if that turns out to match native driver behavior more closely. That would probably need to remain shader-selectable because the same backing buffer is also used normally as R16 elsewhere.

HansKristian-Work:
> If your native testing confirms that the R16_UINT mismatch is the trigger, would it make more sense to handle this as an explicit typed-atomic-to-raw legalization in dxil-spirv, similar in spirit to dxbc-spirv's useRawForTypedAtomic, while keeping vkd3d-proton limited to enabling it for this exact shader?

Most likely that's what we'll have to do, yes.

silv3rshi3ld:
> Most likely that's what we'll have to do, yes.

Thanks for your feedback. I went ahead and reworked the fix along those lines.

The generic part now lives in dxil-spirv. For an eligible scalar 32-bit typed-UAV atomic, it temporarily requests raw-buffer remapping and accepts the lowering only when the remapper returns an SSBO. Otherwise it restores the typed binding and retries the normal path. This uses the existing remapper ABI without adding fields to the public callback structures.

VKD3D-Proton is now limited to selecting the exact IL-2 executable and shader, and enabling the compiler quirk only when the descriptor layout can safely provide the raw SSBO sibling.

I opened the compiler part as a draft here: https://github.com/HansKristian-Work/dxil-spirv/pull/296

The resources suite passes, the captured shader validates in both candidate and fallback modes, and the x86-64 and x86 VKD3D builds pass. I also completed a verified D49 game run covering the menu, a short flight, and the map without the blocks or broad flicker.

The dxil-spirv commit is now reachable through [dxil-spirv PR #296](https://github.com/HansKristian-Work/dxil-spirv/pull/296). After verifying the dependency from a clean recursive clone, I updated this draft PR to the compiler-aware VKD3D-Proton integration at `e8868dfd`, pinning dxil-spirv commit `afff4dfb`.

Does this approach look more in line with what you expect?

HansKristian-Work:
Which game is this exactly? IL-2 Sturmovik: Korea does not seem to exist and IL-2 Sturmovik is a really old game which is not D3D12.

silv3rshi3ld:
> Which game is this exactly? IL-2 Sturmovik: Korea does not seem to exist and IL-2 Sturmovik is a really old game which is not D3D12.

My apologies, [IL2 Korea](https://store.steampowered.com/app/247970/Korea_IL2_Series/). I've typed IL2 Sturmovik so many times that it is a bit of automatism.

HansKristian-Work:
The game crashes with:

Screenshot_20260811_145052

How did you get past that?

silv3rshi3ld:
> The game crashes with:
> Screenshot_20260811_145052
> 
> How did you get past that?

`OMP_NUM_THREADS=16 KMP_AFFINITY=disabled %command%` in the launch option. There is already an upstream fix for it in Wine incoming [[Wine MR !11604](https://gitlab.winehq.org/wine/wine/-/merge_requests/11604)]. Edit: further info [ValveSoftware/Proton issue #9906](https://github.com/ValveSoftware/Proton/issues/9906) for reference.

HansKristian-Work:
Ok, I can reproduce the game bug in RenderDoc.

HansKristian-Work:
It's somewhat confusing that this wouldn't just work already. From what I understand, the texel buffer atomic should ignore the descriptor format here and it's only writing to offset 0. Need to write some test cases to see what happens ...

silv3rshi3ld:
That makes sense, especially since using coordinate 0 removes any format-dependent stride from the address calculation. The reason I pursued the `R16_UINT` mismatch was based on an empirical result rather than an assumption about how the atomic itself should behave.

I built a small headless Vulkan probe around the exact captured SPIR-V. On both RADV devices I tested, and using both the descriptor-set and `VK_EXT_descriptor_buffer` paths, I observed the following:

* `R32_UINT` view: one correct global allocation, 8,160 entries, no overlaps
* `R16_UINT` view: all 50 workgroups restarted from zero, resulting in 7,966 overlapping and 7,966 missing entries
* Raw SSBO variant: one correct global allocation again, with no overlaps

The `R16_UINT` case was deliberately non-portable, and validation did not report an error. I agree that this result does not establish that SSBO legalization is the correct upstream fix. It may instead be exposing some lower-level descriptor or driver behavior.

What stood out to me was specifically the per-workgroup restart pattern. That is the part I could not explain purely from the format mismatch or address calculation.

If useful, the probe source and test setup are available here:

https://github.com/silv3rshi3ld/il2-korea-proton-investigation/tree/main/probes/d21-vulkan-atomic

Does this behavior fit with what you would expect from the texel-buffer atomic path, or am I overlooking something about how the operation is handled or lowered here?

silv3rshi3ld:
I would have rather preferred to solve this within the existing texel-buffer path. But I was unable to identify why that path failed, which resulted in why i pursued the SSBO route. But still when looking back into it, I just couldn't make sense out of it. 

Or to be more specific, I was able to isolate the failure to the R32ui atomic being used with the R16UINT view, but I couldn't explain why that combination produced a per workgroup restart at coordinate 0.

HansKristian-Work:
https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/43672 also fixes it. If this MR is accepted I would rather do this than implementing a super awkward quirk in vkd3d-proton.

NVIDIA also passes the test with descriptor heap, so I think we can harmonize with native D3D12 without adding further hacks.

silv3rshi3ld:
Thanks, this looks like it closes the loop.

I agree with that direction. If [Mesa MR !43672](https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/43672) passes review, matching RADV to native AMD D3D12 and pre-GFX10 behavior is clearly cleaner and more general than either dxil-spirv lowering or a per-game VKD3D-Proton quirk.

Also, apologies for moving to dxil-spirv too quickly after your earlier reply. While rubber-ducking the results earlier today, I noticed a gap in my previous comparison. The successful R32 control used a four-byte counter, while the failing R16 case used the game's full 87,040-byte allocation. I had therefore not cleanly separated the view format from the buffer size and range.

I added two tighter tests. First, I ran the same minimal atomic shader against the same 87,040-byte buffer in an R32, R16, R32 sequence, changing only the buffer-view format. On both RADV devices here, both R32 runs completed correctly, while the R16 run reproduced the per-workgroup restart pattern. Vulkan validation also reports the R32ui/R16_UINT format mismatch and the lack of storage-texel-buffer atomic support for R16_UINT.

Second, I ran the exact captured game shader against a full-size R32_UINT alias. It completed all 8,160 allocations without overlap or missing entries through both the mutable descriptor-set and descriptor-buffer paths on both devices.

That suggested that I did not need to change the atomic lowering at all. The generated shader already uses an R32ui texel buffer with OpImageTexelPointer and OpAtomicIAdd. I therefore built a narrowly scoped VKD3D-Proton-only D52 diagnostic candidate which retains the normal R16_UINT descriptor, creates an R32_UINT texel-buffer sibling for the exact IL-2 UAV shape, and routes only shader 0x7cefa1bc80bb4c70 u1 to that sibling. The dxil-spirv submodule remains unchanged at cc75a0c9.

The runtime log confirms that the alias was created. The captured DXIL is byte-identical to the baseline, and the SPIR-V differs only in that resource's descriptor set and binding, from 1/1 to 2/0. The R32ui image type and atomic instructions are unchanged. The blocks were absent in two game runs, including a repeat without VKD3D logging or shader-dump options. Both runs still used the separate OpenMP startup workaround because this isolated build uses stock Experimental Wine.

I regard that alias as a diagnostic discriminator, not as the desired upstream fix. It confirms that the texel-buffer descriptor and OOB behavior is the important variable, while your Mesa change provides the cleaner driver-level solution and the missing explanation.

Fortunately, the detour caused no lasting harm. Both submissions are still drafts, nothing was merged, and the additional A/B testing helped isolate the descriptor-format boundary and support the Mesa explanation. I am glad we got to the bottom of it, I had fun while doing it.

I will update my evidence accordingly and treat the current dxil-spirv and VKD3D-Proton implementations as superseded while the Mesa MR is reviewed. 

Thanks for reproducing the issue and finding the missing driver-side piece. Was fun! Thanks a lot for the interaction.

silv3rshi3ld:
For completeness, I have now published the source for the later D52 VKD3D-Proton-only diagnostic mentioned above:

https://github.com/silv3rshi3ld/vkd3d-proton/commit/8cd28e8f98751afe3b85c3b08519464907aa5143

This is a single commit on current VKD3D-Proton master with stock dxil-spirv. It retains the normal `R16_UINT` descriptor, creates an `R32_UINT` texel-buffer sibling for the exact IL-2 UAV shape, and routes only shader `0x7cefa1bc80bb4c70` resource `u1` to that sibling when the descriptor layout supports it.

The x86-64 and x86 builds pass. The identical D52 source change removed the blocks in two earlier game runs, but I have not repeated the game test after forwarding it to current master.

I am not proposing that this PR be reopened or that the diagnostic be merged. Mesa MR !43672 remains the preferred general solution. This only makes the final VKD3D-Proton discriminator available for inspection.

PR #3208, opened 2026-08-10
vkd3d: Implement batching of postbuild info

HansKristian-Work:
CP77 hits this in some bad ways that obliterates perf. Batch postbuilds since we have evidence of this being important now. The postbuild writes are considered UAV, so we can be quite loose with synchronization.


doitsujin (approved):
hngh

PR #3214, opened 2026-08-14
Optimize ClearUnorderedAccessView with fixed function commands when possible

HansKristian-Work:


libs/vkd3d/command.c
```diff
@@ -16371,6 +16372,315 @@ static void vkd3d_texture_view_desc_convert_from_metadata(
...
+            return false;
+
+        /* A float format image can only be cleared with a FP32 value.
+         * Ignore this if we're clearing to zero anyway. */
+        if (is_float && format->type != VKD3D_FORMAT_TYPE_OTHER && memcmp(values, zero_pattern, sizeof(zero_pattern)) != 0)
+            return false;
+
+        /* If we're clearing an integer format image, we can only clear it with SINT format,
+         * otherwise we risk losing proper sign extension. */
+        if (!is_float && format->type != VKD3D_FORMAT_TYPE_UINT && memcmp(values, zero_pattern, sizeof(zero_pattern)) != 0)
```
doitsujin, line 16607: is it just me or does the comment contradict the code here? We're noping out if SINT and clear color != 0, which intuitively makes sense given how ClearUAVUint works.

HansKristian-Work: I think the comment intends to say that we can only clear it with proper SINT format, i.e. a raw fixed function clear does not work, and we have to fallback.

HansKristian-Work: Updated the comment.

doitsujin (approved):

libs/vkd3d/command.c
doitsujin: right, yeah, initial wording was very confusing indeed.

PR #3217, opened 2026-08-16
Support hybrid_patchable on the command queue vtable

Sonicadvance1:
Some games (notably GTA5E) hook vtable entries that are returned from D3D12 as part of their anti-tamper measures. When these vtable entries are compiled for ARM64ec/ARM64X then the x86 game ends up modifying ARM64 code with x86 instructions, causing crashes to occur.

The workaround to fix this issue is called "Fast-forward sequences" documented by
[Microsoft](https://learn.microsoft.com/en-us/windows/arm/arm64ec-abi) and
[WINE](https://gitlab.winehq.org/wine/wine/-/wikis/ARM64EC-Toolchain#Patchable_functions)

Instead of returning the direct arm64 code, instead there is a little x86 trampoline sequence that allows applications to still hook those entrypoints. There's a bit of magic here in that it *shouldn't* add any overhead to the emulation by way of the "Call-checker" directly forwarding the emulation past the trampoline unless the code has actually been modified.

This is the only vtable that GTA5E needs, but we may want to extend this to more vtables in the future. As such, with this change, GTA5E is able to get in-game on ARM hosts.

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -22692,7 +22692,7 @@ static inline struct d3d12_command_queue *impl_from_ID3D12CommandQueue(ID3D12Com
     return CONTAINING_RECORD(iface, struct d3d12_command_queue, ID3D12CommandQueue_iface);
 }
 
-HRESULT STDMETHODCALLTYPE d3d12_command_queue_QueryInterface(ID3D12CommandQueue *iface,
+VKD3D_METHODENTRY_NONSTATIC(HRESULT) d3d12_command_queue_QueryInterface(ID3D12CommandQueue *iface,
         REFIID riid, void **object)
 {
     struct d3d12_command_queue *command_queue = impl_from_ID3D12CommandQueue(iface);
@@ -22741,7 +22741,7 @@ HRESULT STDMETHODCALLTYPE d3d12_command_queue_QueryInterface(ID3D12CommandQueue 
     return E_NOINTERFACE;
 }
 
-ULONG STDMETHODCALLTYPE d3d12_command_queue_AddRef(ID3D12CommandQueue *iface)
+VKD3D_METHODENTRY_NONSTATIC(ULONG) d3d12_command_queue_AddRef(ID3D12CommandQueue *iface)
 {
     struct d3d12_command_queue *command_queue = impl_from_ID3D12CommandQueue(iface);
     unsigned int refcount = InterlockedIncrement(&command_queue->refcount);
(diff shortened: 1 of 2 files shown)


libs/vkd3d/command.c
```diff
@@ -22795,7 +22801,7 @@ ULONG STDMETHODCALLTYPE d3d12_command_queue_Release(ID3D12CommandQueue *iface)
     return refcount;
 }
 
-static HRESULT STDMETHODCALLTYPE d3d12_command_queue_GetPrivateData(ID3D12CommandQueue *iface,
+HRESULT VKD3D_HPCALLATTR STDMETHODCALLTYPE d3d12_command_queue_GetPrivateData(ID3D12CommandQueue *iface,
```
HansKristian-Work: Why is static removed?

libs/vkd3d/command.c
```diff
@@ -22692,7 +22698,7 @@ static inline struct d3d12_command_queue *impl_from_ID3D12CommandQueue(ID3D12Com
     return CONTAINING_RECORD(iface, struct d3d12_command_queue, ID3D12CommandQueue_iface);
 }
 
-HRESULT STDMETHODCALLTYPE d3d12_command_queue_QueryInterface(ID3D12CommandQueue *iface,
+HRESULT VKD3D_HPCALLATTR STDMETHODCALLTYPE d3d12_command_queue_QueryInterface(ID3D12CommandQueue *iface,
```
HansKristian-Work: Is it possible to redefine STDMETHODCALLTYPE to be `__attribute__((hybrid_patchable))` instead so we don't have to spam this everywhere?

libs/vkd3d/command.c
Sonicadvance1: I was told that the hybrid_patchable attribute is incompatible with static. Although I guess the attribute is too new to have error checking with this combination.

HansKristian-Work: hnnnnng ... Maybe we need some kind of macro to do:

`VKD3D_HRESULT_ENTRY d3d12_command_queue_Blah()` that expands to

`HRESULT __attribute__(()) STDMETHODCALLTYPE` on arm64ec and

`static HRESULT STDMETHODCALLTYPE d3d12_command_queue_Blah()` everywhere else?

I don't feel comfortable just dropping static everywhere.

Sonicadvance1: I'll double check today if the attribute actually works in testing with static or not. If it does work then it's easy and we can keep static. If not...Then I'll create some monster of a `#define VKD3D_RESULT_ENTRY(return_type)` define.

HansKristian-Work: A few of the entry points aren't static, but could have two variants. Once the macro hell is in place, some simple regexes should be enough to replace all the entry points at least.

Sonicadvance1: Okay, confirmed that static functions still can't have the hybrid_patchable attribute applied and it even becomes a compiler error which is good.

Added two defines that do what we need, `VKD3D_METHODENTRY` for entries that still desire `static` qualifier and `VKD3D_METHODENTRY_NONSTATIC` for the few that don't want static. Depending on if the target is arm64ec or not, it'll choose the correct set of qualifiers and attributes.

Bit of a monster but should be able to be used wherever necessary. Confirmed this latest set still works for GTA5E.

HansKristian-Work (approved):
Guess it's as pretty as it's going to get :')

PR #3223, opened 2026-08-19
SM 6.9 - Phase 4/4

HansKristian-Work:
Completes SM 6.9 exposure. Special shoutout to Peter Kohaut who did most of the heavy SM 6.9 lifting in dxil-spirv.

Passes on NV with some driver workarounds. The bugs have been reported and fixes are on the way, so I'm not too worried.

Tested against the RTXPT 1.8.1 demo which uses standardized OMM, SER and hitobjects in SM 6.9.


oscarbg:
Nice work as always.. can share if the NV SER bugs have been reported to NV?

libs/vkd3d/device.c
```diff
@@ -10901,6 +10920,25 @@ static void d3d12_device_caps_init_shader_model(struct d3d12_device *device)
...
+         * - Shader Execution Reordering
+         * - 16/64-bit is{inf,nan,normal} etc
+         */
+        if (device->d3d12_caps.max_shader_model == D3D_SHADER_MODEL_6_8 &&
+            device->device_info.opacity_micromap_features.micromap &&
+            /* For maint9 bitops which support 16-bit / 64-bit. Needed for vector version of these. */
+            device->device_info.maintenance_9_features.maintenance9 &&
+            device->device_info.long_vector_features.longVector &&
+            device->device_info.long_vector_properties.maxVectorComponents >= 1024 &&
+            device->device_info.invocation_reorder_features.rayTracingInvocationReorder)
```
doitsujin, line 10937: do we strictly require the reorder extension even if `nop` is technically a valid impl?

HansKristian-Work: Yes, because all the hit object stuff is part of it, and hit object cannot be emulated.

doitsujin (approved):

PR #3229, opened 2026-08-24
vkd3d: Add method to simulate pure wave64 device for debug.

HansKristian-Work:
Wave32-only bugs are getting more and more annoying. To make debug easier, allow pretending that RDNA devices are more GCN-like.


doitsujin (approved):

phly95:
While it sounds like this is meant for debugging behavior specific to Wave32 by forcing games into Wave64 mode, perhaps it might prove handy when debugging games that are broken on Wave64 but not Wave32 like this case in Crimson Desert.
https://github.com/HansKristian-Work/vkd3d-proton/issues/3260

Perhaps it might prove useful for keeping GCN cards usable for modern applications.

HansKristian-Work:
Yes, that's the point of the PR, but Crimson Desert is known to rely on patterns that cannot be emulated with simple workarounds.

PR #3231, opened 2026-08-25
Ensure that GPUVA is aligned for placed resources

HansKristian-Work:
Works around quirky implementation details of UE5 + Turnip.

### DIFF
--- a/libs/vkd3d/memory.c
+++ b/libs/vkd3d/memory.c
@@ -1307,6 +1307,14 @@ static bool vkd3d_is_imported_allocation(const struct vkd3d_allocate_memory_info
     return false;
 }
 
+static bool d3d12_device_aligns_bda_64k(struct d3d12_device *device)
+{
+    /* Only seems to affect Turnip for now at least.
+     * TODO: We really should just have an extension for this.
+     */
+    return device->device_info.vulkan_1_2_properties.driverID != VK_DRIVER_ID_MESA_TURNIP;
+}
+
 static HRESULT vkd3d_memory_allocation_init(struct vkd3d_memory_allocation *allocation, struct d3d12_device *device,
         struct vkd3d_memory_allocator *allocator, const struct vkd3d_allocate_memory_info *info)
 {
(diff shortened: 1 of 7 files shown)


libs/vkd3d/vkd3d_private.h
```diff
@@ -874,13 +875,20 @@ struct vkd3d_memory_allocation
...
+     * Once an allocation is sliced later, the padding offset is "consumed". */
+    uint32_t realignment_offset;
+
     struct vkd3d_memory_chunk *chunk;
 };
 
 static inline void vkd3d_memory_allocation_slice(struct vkd3d_memory_allocation *dst,
         const struct vkd3d_memory_allocation *src, VkDeviceSize offset, VkDeviceSize size)
 {
+    offset += src->realignment_offset;
```
doitsujin, line 889: i suppose this part implicitly deals with all the placed resource cases?

libs/vkd3d/memory.c
```diff
@@ -1874,6 +1926,10 @@ static HRESULT vkd3d_memory_allocator_try_add_chunk(struct vkd3d_memory_allocato
         alloc_info.explicit_global_buffer_usage = explicit_global_buffer_usage;
     }
 
+    /* Should normally request aligned GPU address for chunks, but it creates
+     * too much of a mess in implementation of the chunk allocator, so skip it.
+     * It's not needed in any known content. */
```
doitsujin, line 1931: fair to skip for now, but wouldn't this trivially work already when changing the `vkd3d_memory_chunk_insert_range` call inside `vkd3d_memory_chunk_create` to use the requested chunk size rather than the adjusted allocation size?

Everything that's suballocated should be getting the `vkd3d_memory_allocation_slice` treatment anyway.

libs/vkd3d/vkd3d_private.h
HansKristian-Work: Yes. CreatePlacedResource always calls allocation_slice(), which normalizes the offset. The only other case where we had to care about the realignment offset is in UpdateTileMappings as far as I could tell.

libs/vkd3d/memory.c
HansKristian-Work: I failed to prove to myself that the weird offsets would work. Especially on freeing a chunk, it will try to insert an adjusted range which got really weird. It's solveable, but the risk for subtle breakage is too high. We'll fix it more properly with a vendor ext soon-ish.

doitsujin: hm, right.

doitsujin:
Actually one more thing, what's the intended behaviour with committed buffers here? Think there's nothing stopping us from hitting padding path, but we'd need to swap the realignment_offset and offset around for the actual buffer resource as well.

doitsujin:
If we get a vendor ext, do we need a way to ensure sparse resource VAs (i.e. the actual sparse VkBuffer, not the memory bound to it) are also aligned or can we trust drivers there? IIRC some hw supports <64k sparse pages.

HansKristian-Work:
Committed resources are not handled here. UE5 doesn't rely on that. Sparse is the same. The only case we handle here is CreateHeap().

The proposed ext will add VA alignment vkAllocateMemory as well as vkCreateBuffer sparse, indeed.

doitsujin (approved):

PR #3237, opened 2026-08-27
Fix vk_swapchain_count accounting, so that Reflex continues to work after re-creating swapchains.

wcampbell-nv:
vk_swapchain_count is not decremented when destroying a swapchain in this DX-swapchain-destruction path. This can result in Reflex getting shut down by the multi-swapchain disablement logic, even though only one swapchain currently exists.

With this change we decrement vk_swapchain_count alongside every call to vkDestroySwapchainKHR.

I hit this when toggling DLSS-FG on and off in Avatar Frontiers of Pandora. I confirmed that Reflex now continues to work with this fix.

### DIFF
--- a/libs/vkd3d/swapchain.c
+++ b/libs/vkd3d/swapchain.c
@@ -583,6 +583,13 @@ static void dxgi_vk_swap_chain_cleanup_common(struct dxgi_vk_swap_chain *chain)
     for (i = 0; i < ARRAY_SIZE(chain->present.vk_swapchain_fences); i++)
         VK_CALL(vkDestroyFence(chain->queue->device->vk_device, chain->present.vk_swapchain_fences[i], NULL));
 
+    if (chain->queue->device->vk_info.NV_low_latency2 && chain->present.vk_swapchain)
+    {
+        spinlock_acquire(&chain->queue->device->low_latency_swapchain_spinlock);
+        chain->queue->device->swapchain_info.vk_swapchain_count--;
+        spinlock_release(&chain->queue->device->low_latency_swapchain_spinlock);
+    }
+
     VK_CALL(vkDestroySwapchainKHR(chain->queue->device->vk_device, chain->present.vk_swapchain, NULL));
 
     for (i = 0; i < ARRAY_SIZE(chain->user.backbuffers); i++)


libs/vkd3d/swapchain.c
```diff
@@ -583,6 +583,13 @@ static void dxgi_vk_swap_chain_cleanup_common(struct dxgi_vk_swap_chain *chain)
     for (i = 0; i < ARRAY_SIZE(chain->present.vk_swapchain_fences); i++)
         VK_CALL(vkDestroyFence(chain->queue->device->vk_device, chain->present.vk_swapchain_fences[i], NULL));
 
+    if (chain->queue->device->vk_info.NV_low_latency2 && chain->present.vk_swapchain)
+    {
+        spinlock_acquire(&chain->queue->device->low_latency_swapchain_spinlock);
+        chain->queue->device->swapchain_info.vk_swapchain_count--;
```
HansKristian-Work, line 589: This smells really fishy. Why is this vk_swapchain_count tracking not part of d3d12_device_register_swapchain and d3d12_device_remove_swapchain? I would imagine the promotion and demotion logic to happen there.

HansKristian-Work:
I made an alternative PR which I think solves the same problem, but the original LL2 impl was not done by me and it's possible there is some hidden reason behind the existing system that I'm not aware of. This entire tracking system is very annoying ...

https://github.com/HansKristian-Work/vkd3d-proton/pull/3239

Does that also fix the issue?

HansKristian-Work:
Git blaming a bit I found

```
commit 298eaf00036f8f1cb4410c9e21b5208ed9a66c67
Author: Will Campbell <wcampbell@nvidia.com>
Date:   Fri Jan 23 18:17:10 2026 +0000

    Disable Reflex when a second swapchain is first used, rather than created
```

Which seems to explain why the current weird code is there. The original code had no comments about why this was done which I should have rejected in the original PR. I'll need to rethink my PR a bit.

HansKristian-Work:
Reinstated the workaround in a way that is hopefully clearer and avoids needing to do weird refcounts up and down per swapchain.

wcampbell-nv:
Your https://github.com/HansKristian-Work/vkd3d-proton/pull/3239 does also fix the issue, and looks more graceful than what we have now. Feel free to close this in favor of your refactor.

> This entire tracking system is very annoying ...

Agreed. It shouldn't be necessary, as the NVIDIA driver has similar logic to ignore VK_NV_low_latency2 if multiple swapchains are used at once. But I'm hesitant to suggest removing the vkd3d-proton multi-swapchain disablement, since we've always had it and have not tested games without it.

HansKristian-Work:
Ok, thanks for conforming, closing this as superseded.

PR #3239, opened 2026-08-28
Alternative formulation of PR 3237

HansKristian-Work:


libs/vkd3d/device_vkd3d_ext.c
```diff
@@ -1179,6 +1179,70 @@ CONST_VTBL struct ID3D12DXVKInteropDevice3Vtbl d3d12_dxvk_interop_device_vtbl =
     d3d12_dxvk_interop_device_GetVulkanHeapInfo,
 };
 
+void d3d12_device_notify_swapchain_creation(struct d3d12_device *device, struct dxgi_vk_swap_chain *chain)
```
wcampbell-nv: This fires when an creating an underlying Vulkan swapchain on first use of a DXGI swapchain, right? Rather than on DXGI swapchain creation.

If so naming this something like `d3d12_device_notify_vk_swapchain_creation` would be more clear.

libs/vkd3d/device_vkd3d_ext.c
```diff
@@ -1179,6 +1179,70 @@ CONST_VTBL struct ID3D12DXVKInteropDevice3Vtbl d3d12_dxvk_interop_device_vtbl =
...
+void d3d12_device_register_low_latency_swapchain(struct d3d12_device *device, struct dxgi_vk_swap_chain *chain)
+{
+    if (!device->vk_info.NV_low_latency2)
+        return;
+
+    spinlock_acquire(&device->low_latency_swapchain_spinlock);
+
+    device->swapchain_info.vk_swapchain_count++;
+
+    if (device->swapchain_info.vk_swapchain_count == 1 && !device->swapchain_info.low_latency_swapchain)
```
wcampbell-nv: I believe this now counts DXGI swapchains rather than underlying Vulkan swachains. If so, renaming `vk_swapchain_count` would be more clear.

wcampbell-nv (commented):
I tested and confirmed this resolved the issue from my MR #3237. I'm on board with closing my MR in favor of this.

Verified in games 'Dying Light 2: Reloaded Edition', 'Avatar: Frontiers of Pandora' and 'NBA 2K27'. VK_NV_low_latency2 calls now continue to be passed through after toggling DLSS-FG.

I also tested Arc Raiders and The Finals, which are not impacted by the bug. VK_NV_low_latency2 calls make it through with and without this change.

libs/vkd3d/device_vkd3d_ext.c
HansKristian-Work: True. Fixed.

libs/vkd3d/device_vkd3d_ext.c
HansKristian-Work: Fixed.

doitsujin (approved):

PR #3247, opened 2026-08-29
Handle emulated D24 <-> Buffer copies

doitsujin:
Fixes MGS4 lighting in some cutscenes.

Edit: Draft because test isn't amazing currently.


libs/vkd3d/shaders/cs_copy_emulated_d24_to_buffer.comp
```diff
@@ -0,0 +1,36 @@
...
+void main()
+{
+    if (any(greaterThanEqual(gl_GlobalInvocationID.xy, args.extent)))
+        return;
+
+    uint dst_offset = gl_GlobalInvocationID.x +
+      gl_GlobalInvocationID.y * (args.row_pitch / 4u);
+
+    float depth = texelFetch(s_texture, ivec2(gl_GlobalInvocationID.xy) + args.src_offset, 0).x;
+    args.dst_buffer.data[dst_offset] = uint(float(0xffffff) * clamp(depth, 0.0f, 1.0f));
```
HansKristian-Work: Should RTE here.

libs/vkd3d/meta.c
```diff
@@ -2427,6 +2550,10 @@ HRESULT vkd3d_meta_ops_init(struct vkd3d_meta_ops *meta_ops, struct d3d12_device
     vkd3d_resolve_image_ops_cleanup(&meta_ops->resolve_image_heap, device);
 fail_resolve_image_heap_ops:
     vkd3d_copy_image_ops_cleanup(&meta_ops->copy_image_legacy, device);
+fail_copy_buffer_image_legacy_ops:
```
HansKristian-Work: These failures seem a bit fishy. Failure on goto fail_copy_buffer_image_legacy_ops ends up cleaning up itself, which is not consistent with code above.

HansKristian-Work (changes requested):
Seems good, barring minor nits.

libs/vkd3d/meta.c
doitsujin: derp, fixed

doitsujin:
Updated the test so that it also checks offsets, row pitch etc, and ran into another fun quirk: When copying from D24 to buffer, native zeroes the upper 8 bits of the buffer, whereas NV Vulkan preserves them. Gonna ignore that until it becomes a problem though.

libs/vkd3d/command.c
```diff
@@ -10821,19 +10823,27 @@ static void d3d12_command_list_before_copy_texture_region(struct d3d12_command_l
...
 
-            d3d12_command_list_transition_image_layout_with_global_memory_barrier(list, batch,
-                    src_resource,
+            if (info->needs_conversion)
+            {
+                dst_stages = VK_PIPELINE_STAGE_2_COMPUTE_SHADER_BIT;
+                dst_access = VK_ACCESS_2_SHADER_SAMPLED_READ_BIT;
+            }
+
+            d3d12_command_list_transition_image_layout_with_global_memory_barrier(list, batch, src_resource,
```
HansKristian-Work, line 10845: global_transfer_access does not get STORAGE_WRITE_BIT here. I.e. we don't do the transitive COPY -> COMPUTE barrier that we do for buffer -> image path in next commit.

libs/vkd3d/command.c
```diff
@@ -10862,6 +10872,104 @@ static void d3d12_command_list_before_copy_texture_region(struct d3d12_command_l
...
+            VK_SHADER_STAGE_COMPUTE_BIT, sizeof(compute_args), &compute_args);
+
+    VK_CALL(vkCmdDispatch(list->cmd.vk_command_buffer,
+            vkd3d_compute_workgroup_count(info->copy.buffer_image.imageExtent.width, 8),
+            vkd3d_compute_workgroup_count(info->copy.buffer_image.imageExtent.height, 8), 1u));
+
+cleanup_compute:
+    if (src_view)
+        vkd3d_view_decref(src_view, list->device);
+
```
HansKristian-Work, line 11008: stray newline.

libs/vkd3d/command.c
```diff
@@ -10862,6 +10872,104 @@ static void d3d12_command_list_before_copy_texture_region(struct d3d12_command_l
...
+    compute_args.offset.y = info->copy.buffer_image.imageOffset.y;
+    compute_args.extent.width = info->copy.buffer_image.imageExtent.width;
+    compute_args.extent.height = info->copy.buffer_image.imageExtent.height;
+    compute_args.row_pitch = info->dst.PlacedFootprint.Footprint.RowPitch;
+
+    d3d12_command_list_meta_push_data(list, list->cmd.vk_command_buffer,
+            copy_pipeline_info.vk_pipeline_layout,
+            VK_SHADER_STAGE_COMPUTE_BIT, sizeof(compute_args), &compute_args);
+
+    VK_CALL(vkCmdDispatch(list->cmd.vk_command_buffer,
```
HansKristian-Work, line 10987: I think we're missing some kind of barrier after this dispatch, we have to link compute writes into COPY stage.

libs/vkd3d/command.c
```diff
@@ -10878,30 +10986,40 @@ static void d3d12_command_list_copy_texture_region(struct d3d12_command_list *li
...
+                    info->buffer_footprint_size, !!(dst_resource->flags & VKD3D_RESOURCE_RESERVED));
+            VK_CALL(vkCmdCopyImageToBuffer2(list->cmd.vk_command_buffer, &copy_info));
+        }
+
+        if (!unified || info->needs_conversion)
         {
-            d3d12_command_list_transition_image_layout(list, batch,
-                    src_resource,
-                    &info->copy.buffer_image.imageSubresource, VK_PIPELINE_STAGE_2_COPY_BIT, VK_ACCESS_2_NONE,
+            VkPipelineStageFlags2 src_stage = info->needs_conversion
```
HansKristian-Work, line 11119: I suppose conversion path needs an image barrier with global memory barrier here.

HansKristian-Work (changes requested):
Found some sync issues that I missed in original review.

libs/vkd3d/command.c
doitsujin: right, the image barrier doesn't handle the writes. sigh.

libs/vkd3d/command.c
```diff
@@ -10878,31 +10986,44 @@ static void d3d12_command_list_copy_texture_region(struct d3d12_command_list *li
...
-                    info->src_layout, VK_PIPELINE_STAGE_2_COPY_BIT, VK_ACCESS_2_NONE, src_resource->common_layout);
+            VkPipelineStageFlags2 src_stage = info->needs_conversion
+                ? VK_PIPELINE_STAGE_2_COMPUTE_SHADER_BIT : VK_PIPELINE_STAGE_2_COPY_BIT;
+            VkAccessFlags2 src_access = info->needs_conversion
+                ? VK_ACCESS_2_SHADER_STORAGE_WRITE_BIT : VK_ACCESS_2_NONE;
+
+            d3d12_command_list_transition_image_layout_with_global_memory_barrier(list, batch, src_resource,
+                    &info->copy.buffer_image.imageSubresource, src_stage, VK_ACCESS_2_NONE,
+                    info->src_layout, VK_PIPELINE_STAGE_2_COPY_BIT, VK_ACCESS_2_NONE, src_resource->common_layout,
+                    src_access, VK_ACCESS_2_NONE);
```
HansKristian-Work: I think we need VK_ACCESS_2_TRANSFER_WRITE on the buffer since we don't do the COPY_DEST tracking for this path? It's possible that there's a WAW hazard.

doitsujin: right, the tracking exists. Honestly this entire thing is just too complex to keep track of at this point.

how do we even fix this? I guess moving the `d3d12_command_list_mark_copy_buffer_write` out of the if is step 1, but it doesn't know about compute. I don't really want to unconditionally emit barriers here if we can avoid it.

HansKristian-Work: Trying to avoid the extra barriers would be really cumbersome I think. The incoming COPY -> COMPUTE + WRITE barrier is necessary at least once. We'd have to tack on extra tracking for COPY_DEST stage if it's been made visible to compute or not yet, which is non-trivial complexity.

Given we consider this path exceptionally rare in practice, I think it's okay to keep the complexity local. Could add a FIXME or TODO mentioning this problem.

libs/vkd3d/command.c
doitsujin: resolved by adding a barrier before each copy.

libs/vkd3d/command.c
```diff
@@ -11000,12 +11008,71 @@ static void d3d12_command_list_copy_image_to_buffer_compute(struct d3d12_command
...
+    compute_args.src_va = src_resource->res.va + info->src.PlacedFootprint.Offset;
+    compute_args.extent.width = info->copy.buffer_image.imageExtent.width;
+    compute_args.extent.height = info->copy.buffer_image.imageExtent.height;
+    compute_args.row_pitch = info->src.PlacedFootprint.Footprint.RowPitch;
+
+    d3d12_command_list_meta_push_data(list, list->cmd.vk_command_buffer,
+            copy_pipeline_info.vk_pipeline_layout,
+            VK_SHADER_STAGE_COMPUTE_BIT, sizeof(compute_args), &compute_args);
+
+    VK_CALL(vkCmdDispatch(list->cmd.vk_command_buffer,
```
HansKristian-Work, line 11050: Not seeing a barrier that would allow reading COPY_SOURCE as compute here.

HansKristian-Work: nvm

HansKristian-Work (approved):

PR #3252, opened 2026-08-31
Update some test suite todos

HansKristian-Work:


doitsujin (approved):
do we know what the turnip problem could be?

HansKristian-Work:
> do we know what the turnip problem could be?

It seems like Qcom texel buffer descriptor encoding is very cursed and it would explain these issues.

PR #3255, opened 2026-09-01
Attribute queue submits to the right VK_NV_low_latency2 frame ID.

wcampbell-nv:
This change tweaks the frame ID passed in VkLatencySubmissionPresentIdNV to match the frame-attribution rules added in VK_NV_low_latency2 revision 3.

NVIDIA's DirectX driver follows the implicit attribution scheme described in the spec. Vkd3d-proton violates the implicit attribution scheme by deferring its submissions onto background threads. Instead, vkd3d-proton will compute each submission's frame ID during the app's foreground thread calls, and then pass that explicit ID to the corresponding background thread submission.

Workloads are now delimited by PRESENT_END rather than by the RENDERSUBMIT_START marker.

OUT_OF_BAND_RENDER queues are delimited by OUT_OF_BAND_RENDERSUBMIT_START markers.

### Testing:

This has been confirmed to improve latency/stutter with DLSS-FG and NVIDIA Smooth Motion in many games when combined with a driver advertising VK_NV_low_latency2 revision 3. Including:
 - Cyberpunk 2077
 - Dying Light 2: Reloaded Edition
 - The Last of Us Part II
 - Marvel Rivals

### DIFF
--- a/libs/vkd3d/command.c
+++ b/libs/vkd3d/command.c
@@ -23582,6 +23582,52 @@ out:
     return ret;
 }
 
+/* Returns the low-latency frame id to attribute submissions on this queue to. */
+static uint64_t d3d12_command_queue_latency_frame_id(struct d3d12_command_queue *command_queue)
+{
+    struct vkd3d_device_frame_markers *markers;
+    UINT64 best, id;
+    unsigned int i;
+
+    if (command_queue->out_of_band_queue_type == VK_OUT_OF_BAND_QUEUE_TYPE_RENDER_NV)
+    {
+        return vkd3d_atomic_uint64_load_explicit(
+                &command_queue->device->frame_markers.out_of_band_render, vkd3d_memory_order_acquire);
+    }
+    if (command_queue->out_of_band_queue_type == VK_OUT_OF_BAND_QUEUE_TYPE_PRESENT_NV)
+    {
+        return 0;
+    }
+
+    markers = &command_queue->device->frame_markers;
+    spinlock_acquire(&command_queue->device->low_latency_swapchain_spinlock);
+    if (markers->new_frame)
+    {
+        /* This is the first submit in a new frame. The application chooses
+         * the ID of the new frame, and we unfortunately do not know what ID
+         * they will choose next at the PRESENT_END delimiter.
+         *
+         * We'll choose the frame ID of the next SIMULATION_START marker
+         * with an ID greater than the last PRESENT_END marker's ID (which
+         * triggered the new_frame).
+         *
+         * If unknown, we fallback to incrementing the ID by 1, which matches
+         * the behaviour of most applications. */
+        best = 0;
+        for (i = 0; i < VKD3D_RECENT_SIM_STARTS_COUNT; ++i)
+        {
+            id = markers->recent_sim_starts[i];
+            if (id > markers->present_end && (best == 0 || id < best))
+                best = id;
+        }
+        markers->submission = best ? best : markers->present_end + 1;
+        markers->new_frame = false;
+    }
+    id = markers->submission;
+    spinlock_release(&command_queue->device->low_latency_swapchain_spinlock);
+    return id;
+}
+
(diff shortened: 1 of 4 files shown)


libs/vkd3d/device_vkd3d_ext.c
```diff
@@ -1288,30 +1287,35 @@ static HRESULT STDMETHODCALLTYPE d3d12_low_latency_device_SetLatencyMarker(d3d_l
...
-            {
-                WARN("RENDERSUBMIT_START_NV is non-monotonic %"PRIu64" < %"PRIu64".\n",
-                        internal_frame_id, device->frame_markers.render);
-            }
-            device->frame_markers.render = internal_frame_id;
+        case VK_LATENCY_MARKER_SIMULATION_START_NV:
+        {
+            struct vkd3d_device_frame_markers *markers = &device->frame_markers;
+            spinlock_acquire(&device->low_latency_swapchain_spinlock);
+            markers->recent_sim_starts[markers->recent_sim_starts_index] = frameID;
```
HansKristian-Work, line 1296: What is the rationale for this being a ring buffer? I would expect that the last simulation start is the one we should use for subsequent presents. The code seems to just pick the largest value it finds in this list, so I would imagine all of this could just be replaced with a simple running max() or something?

libs/vkd3d/command.c
```diff
@@ -23838,6 +23582,52 @@ static bool vkd3d_submission_has_query_reset_hazard(
...
+         * with an ID greater than the last PRESENT_END marker's ID (which
+         * triggered the new_frame).
+         *
+         * If unknown, we fallback to incrementing the ID by 1, which matches
+         * the behaviour of most applications. */
+        best = 0;
+        for (i = 0; i < VKD3D_RECENT_SIM_STARTS_COUNT; ++i)
+        {
+            id = markers->recent_sim_starts[i];
+            if (id > markers->present_end && (best == 0 || id < best))
```
HansKristian-Work, line 23622: Do we actually want to pick the largest sim start?

Given that renderers are pipelined, I would imagine we want to pick the **smallest** value that is larger than previous present_end. In that model, it makes sense to have a running list of simulation start values.

HansKristian-Work: Oh, I read it backwards. Seems like it does that already.

The comments should clarify this design point and why there needs to be a small ring buffer of sim IDs.

libs/vkd3d/device_vkd3d_ext.c
HansKristian-Work: nvm, see other comment, I misread the other code.

HansKristian-Work (changes requested):
The impl is sound but it really needs comments explaining why the ring buffer is used. It's not immediately obvious.

libs/vkd3d/command.c
```diff
@@ -23838,6 +23582,52 @@ static bool vkd3d_submission_has_query_reset_hazard(
...
+    UINT64 best, id;
+    unsigned int i;
+
+    if (command_queue->out_of_band_queue_type == VK_OUT_OF_BAND_QUEUE_TYPE_RENDER_NV)
+    {
+        return vkd3d_atomic_uint64_load_explicit(
+                &command_queue->device->frame_markers.out_of_band_render, vkd3d_memory_order_acquire);
+    }
+    if (command_queue->out_of_band_queue_type == VK_OUT_OF_BAND_QUEUE_TYPE_PRESENT_NV)
+    {
```
HansKristian-Work: Get rid of braces here, and have a new-line before the new if.

libs/vkd3d/command.c
wcampbell-nv: I tweaked the comment to make things more clear.

libs/vkd3d/command.c
wcampbell-nv: Sure. I got rid of the braces here and for the condition above.

HansKristian-Work (approved):

PR #3269, opened 2026-09-04
vkd3d: Assume strict BAB addressing wrapping by default.

HansKristian-Work:
UE seems to be affected too, and we have no good strategy for workarounding just UE. Be conformant by default.

Proton: UE5 on Turnip


doitsujin (approved):

HansKristian-Work:
Flipping to draft for now. Need to see if there's something smarter we can do w.r.t. UE first.

HansKristian-Work:
Was superseded.

PR #3279, opened 2026-09-08
SM 6.9 - Phase 3/4

HansKristian-Work:
Completes tests.

- Add HitObject test.

The API for this looks very daunting, but the easiest way to think of it is that it represents the state that exists right after TraceRay is called and right before the final closest hit, miss or nothing is dispatched.

Using that hit object, it's possible to introspect and even override the SBT indices are used. The intent is that reordering should happen here before the hit object is "invoked". Of course, there's a million and a half member functions to handle and test similar to RayQuery :(

It's also possible to synthesize miss and nop objects out of thin air.

RayQuery is also supported. So it's sort of possible to use ray query, capture a hit, assign the hit to a RTPSO-like SBT index and dispatch that. I have no idea why that would come up, but it's valid. Might make sense if you want to reorder threads during a trace perhaps? :|

One big issue with HitObject is that it works like a value object in DXIL (even though it's a pointer in IR), and SPIR-V cannot model this exactly due to some unfortunate spec issues (which have been filed). Therefore the dxil-spirv implementation is best effort but should deal with any meaningful usage pattern in the wild, which is going to be some variant of:

- hitobject::trace()
- reorder
- hitobject::invoke()


doitsujin (approved):
Can't say I looked at every bit of test logic here, but the shader code side of things makes sense at least.

PR #3280, opened 2026-09-08
Add more extensive app detect hackery for UE/RE Engine workarounds

HansKristian-Work:
We have some scenarios where we need to perform sweeping workarounds for these engines and doing it per-game with .exe names is not sustainable.

It's a pretty gross PR overall.

First part is a refactor to pull out all the "hack/workaround/app-detect" code to its own file, since it was taking way too much space.

Second is a helper function from Paul that I modified slightly which reads Win32 PE file metadata.
This isn't enough for UE5 though since some games rename the metadata to match their own company, game, etc, so additional means to detect UE5 are also added. I doubt this is fully robust, but I don't know a better solution either. It's extremely frustrating that there is no "application info" equivalent in D3D12 that games actually use.

For RE Engine, the detection keys off CompanyName starting with CAPCOM and some light file system spelunking.

Turnip needs extra workarounds too for UE5, so we can make sure of the until now unused applicationVersion as a sentinel to let Mesa detect engines in a direct way (e.g. pEngineName == vkd3d && applicationVersion == N).

I don't like having to do any of this, but it is what it is ... :(


libs/vkd3d-common/platform.c
```diff
@@ -213,13 +214,25 @@ static bool get_ue_version_from_exe(const WCHAR *path, uint32_t *major, uint32_t
     if (!VerQueryValueA(block, "\\VarFileInfo\\Translation", (void **)&translation, &size) || size != 4)
         goto done;
 
+    sprintf(buf, "\\StringFileInfo\\%08lx\\CompanyName", MAKELONG(HIWORD(*translation), LOWORD(*translation)));
+    if (VerQueryValueA(block, buf, (void **)&s, &size))
+    {
+        TRACE("CompanyName: %s\n", s);
+        if (!strncmp(s, "CAPCOM", 6))
+        {
+            /* It's probably RE-engine, but we only know it's made by CAPCOM. Could be an older game? */
```
doitsujin: Only thing that comes to mind here would be Monster Hunter World, everything else should be RE Engine indeed.

doitsujin: Works fine.

doitsujin:
As mentioned on Discord, RE Engine seems to be a mess where at least Monster Hunter Stories 3 (demo build, not sure about the full game) doesn't have the `CAPCOM` company string.

DD2 and Monster Hunter Wilds hit the detection as expected.

doitsujin (approved):

PR #3284, opened 2026-09-09
Add robust sdiv/srem by 0 handling as well.

HansKristian-Work:
Native d3d12 drivers go out of their way to make this well defined, although it's not entirely well defined. LLVM states this is UB, but native drivers don't care.

- NVIDIA and WARP both return 0xffffffff here, like for udiv/umod.
- AMD returns INT_MAX or INT_MIN depending on the numerator sign.
 
Both are sensible options, but 0xffffffff is the easiest to emulate.


doitsujin (approved):
sigh

PR #3288, opened 2026-09-11
Compute copy fixes

doitsujin:
Yay, more oversights, this time exposed by Company of Heroes 3.

- Compute fallback stuff is just created on demand now, this would previously assert because the logic wasn't aware of the D24 special case.
- My DXVK brain missed the fact that `block_byte_count` and `byte_count` are separate things here for some strange reason, so we were actually under-allocating scratch and that would cause GPU hangs.

And yes, this game is actually hammering buffer->D24 copies during intialization somehow. No idea *what* their plan is though, probably just zeroes or something because nobody has complained about issues so far.


HansKristian-Work (commented):
Missing Proton: tags. Otherwise lgtm.

PR #3297, opened 2026-09-15
build: Add arm64ec cross-file.

HansKristian-Work:


build-arm64ec.txt
```diff
@@ -0,0 +1,20 @@
+[binaries]
+c = 'arm64ec-w64-mingw32-gcc'
+cpp = 'arm64ec-w64-mingw32-g++'
```
mati865, line 3: Is there a point in pretending those are GCC binaries?
They are symlinked to `clang`, so it's just a cosmetics. At least for now.

HansKristian-Work: I reused something from Proton here. I don't really want to deviate unless I have to.

doitsujin (approved):

PR #3300, opened 2026-09-16
vkd3d: Do not engage CPU timeline waits on VkQueue starved platforms.

HansKristian-Work:
We don't want to force CPU waits by default on e.g. Turnip. It only exposes one VkQueue, and we're never going to be able to get low-latency async behavior out of that anyway. Avoids potential GPU performance snafus.


libs/vkd3d/command.c
```diff
@@ -24534,6 +24534,21 @@ static bool d3d12_command_queue_needs_cpu_waits(struct d3d12_command_queue *comm
...
+         * that everyone have to share.
+         * Another scenario where this comes up is RADV where transfer queue are not exposed
+         * and we're oversubscribed for COMPUTE + COPY. Bubbles are not very serious on async queues,
+         * however, the graphics queue is generally the critical path for performance.
+         */
+        if (!VKD3D_CONFIG_FLAG_IS_SET(SINGLE_QUEUE) &&
+            q->desc.Type != command_queue->desc.Type &&
+            (command_queue->vkd3d_queue->vk_queue_flags & VK_QUEUE_GRAPHICS_BIT))
+        {
+            continue;
```
doitsujin, line 24550: We'll potentially hit the path below if we're adding two compute queues to a gfx queue. Probably need to check queue type mismatches separately first.

Also not sure how useful it is to exclude `SINGLE_QUEUE` here.

HansKristian-Work: Not sure it's okay to just exit the loop, but tweaked the logic so that only multiple DIRECT queues will potentially hit this path.

doitsujin (approved):

Issue #3307, opened 2026-09-20
Soulframe Dx12 crashes after latest update, related to tight alignment


WovTTvoW:
```
Game: Soulframe, current Preludes 16 build
Proton: GE-Proton11-7
VKD3D-Proton: 3.1.0, build `af89350cc2eacd9+`
GPU: AMD Radeon RX 9070 XT
Driver: RADV 26.2.3
Kernel: Linux 7.2.6-1-cachyos
```
-DX12 worked before the latest Soulframe update. After the update, DX12 now crashes during startup, while DX11 still works normally.
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - 
Relevant log sequence:
```
Gfx [Info]: D3D12 SDK Version: 619
Gfx [Info]: Device supports tight alignment
warn:vkd3d-proton:d3d12_resource_validate_texture_alignment:
Invalid resource alignment 0x100.
warn:vkd3d-proton:d3d12_device_GetCopyableFootprints1:
Invalid resource desc.
err:seh:dispatch_exception assertion failure exception
```
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
-The assertion occurs inside `Soulframe.x64.exe` immediately after VKD3D rejects the resource description.
-VKD3D otherwise initializes D3D12 successfully and detects the GPU correctly.
- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
Possible cause: Soulframe now appears to use D3D12 SDK 619 tight-alignment behavior and passes a texture with `Alignment = 0x100` into `GetCopyableFootprints1`. VKD3D rejects that alignment before the game asserts.

-Obviously, switching the game to Dx11 '''fixes''' this, but I thought I'd report it.

[steam-15916073279069093888.log](https://github.com/user-attachments/files/32440881/steam-15916073279069093888.log)

doitsujin:
doesn't seem like the game is available yet? how are you testing this?

doitsujin:
FWIW it does look like there is a genuine bug at least, creating a resource with tight alignment always fails currently *except* on the one path that `test_tight_resource_alignment` tests. Amazing.

[...]

PR #3308, opened 2026-09-21
Update Tight Alignment behaviour

doitsujin:
See #3307. The main change here is alignment validation.

619.4 and .5 made some API-breaking changes to this feature, and in good old D3D12 fashion, native behaviour is very undercooked and broken. Passing an alignment greater than the minimum required per spec is legal and intended (think forcing 16b alignment for BAB use cases), but *obviously* this isn't honoured by `GetResourceAllocationInfo` at all, nor is the information written to the resource desc. Native will alwas return the smallest possible alignment for the resource.

Crucially, we cannot allow image tiling modes to change depending on the explicit alignment.


libs/vkd3d/resource.c
```diff
@@ -3084,31 +3087,71 @@ static bool d3d12_resource_supports_small_resource_alignment(const D3D12_RESOURC
...
     if (!desc->Alignment)
         return true;
 
-    if (desc->Alignment != D3D12_DEFAULT_RESOURCE_PLACEMENT_ALIGNMENT
-            && desc->Alignment != D3D12_SMALL_RESOURCE_PLACEMENT_ALIGNMENT
-            && (desc->SampleDesc.Count == 1 || desc->Alignment != D3D12_DEFAULT_MSAA_RESOURCE_PLACEMENT_ALIGNMENT))
+    if (desc->Flags & D3D12_RESOURCE_FLAG_USE_TIGHT_ALIGNMENT)
+    {
+        /* Any power-of-two alignment between 8 and 256 is allowed here */
+        if ((desc->Alignment & (desc->Alignment - 1u)) || desc->Alignment < 8u || desc->Alignment > 256u)
```
HansKristian-Work, line 3098: So passing Alignment = 0 for buffer with tight alignment is actually banned?

doitsujin: 0 is always fine and checked earlier.

HansKristian-Work: Right.

HansKristian-Work (approved):
hng

PR #3314, opened 2026-09-21
vkd3d: Work around First Light Trail shaders on Wave64-only GPUs.

ctrl-kitty:
007 First Light 1.2.1.0 hangs during **Pursue 009** on Wave64-only Vega, waiting for a compute PSO after rejecting a WaveSize(32) shader. This extends #3256 to three Trail shaders, restricted to their known bytecode hashes:

- TrailGen: reuse the existing 32-thread-group workaround.
- TrailCount: allow Wave64 on Wave64-only devices; the algorithm uses the actual lane count.
- TrailSort: replace the wave-dependent sort with a shared-memory radix sort, compiled through the normal DXIL path. Wave32-capable devices keep the original shader.

Tested on RX Vega (`1002:687f`), RADV 26.1.5, NixOS:

- Native and MinGW x64 release builds passed. The included GPU regression passed all 162 sorting cases natively and through Proton Experimental, including CPU comparison and guard checks.
- Automatic workaround selection with local game-shader fixtures: 162 TrailSort cases, 108 TrailCount cases and TrailGen pipeline creation passed. An unrelated executable still rejects the original shader.
- Six synthetic compiler checks confirmed ALLOW_WAVE64 preserves Wave32-capable behavior.
- The generated header reproduces with Microsoft's DXC v1.8.2505.1 release.

The equivalent external workaround let the player pass the freeze location on Proton Experimental. The integrated build has GPU-test coverage; full-game testing, performance and TrailGen geometry validation remain outstanding.

Original shaders were inspected locally but are not included. New code and tests were developed with AI assistance. Related: ValveSoftware/Proton#9820.


HansKristian-Work:
No, no. I'm not adding full on shader replacement like this.

ctrl-kitty:
Understood, thanks for taking a look. I've reported shader issues and findings to IO Interactive support.

Issue #3226, opened 2026-08-20
The Witcher 3 (next-gen): GPU hang (Xid 109 / VK_ERROR_DEVICE_LOST) when enabling ray tracing


Stanni97:
Enabling the master ray tracing toggle in The Witcher 3 next-gen reliably hangs the GPU about 40 seconds later. The game freezes permanently and has to be killed. `dmesg` reports `Xid 109 CTX SWITCH TIMEOUT` for `witcher3.exe`, and vkd3d-proton logs a cascade of `vr -4` (`VK_ERROR_DEVICE_LOST`). Same channel and `Info` value every run.

No individual RT effect is needed to trigger it — the master toggle alone, with GI, reflections, AO and shadows all off, is enough.

Notes that may help narrow it down:

- Cyberpunk 2077 runs ray tracing on this exact stack without issues, so DXR itself works here.
- The same hardware runs Witcher 3 with RT fine under Windows.
- The game is detected correctly (`Detected game witcher3.exe, adding 2 configs` + shader quirks).
- Tested with no effect: `single_queue`, `no_upload_hvv`, `descriptor_heap` (confirmed active in the log), cleared shader cache, stock clocks (no LACT/overclocking).
- DX11 and DX12-without-RT are both stable.

Possibly related to #3085, where downgrading to vkd3d-proton 3.0.0 fixed a similar `DEVICE_LOST` hang. I could not test that — every Proton build available to me ships 3.1.0.

### Software information

The Witcher 3: Wild Hunt — Complete Edition (GOG), next-gen, DX12, 2560x1440. Ray tracing master toggle enabled in the main menu, all individual RT effects off.

### System information

- GPU: NVIDIA GeForce RTX 4070
- Driver: 610.57.04, open kernel modules
- Wine version: proton-cachyos 11.0-20260703 (also reproduced on GE-Proton11-5)
- VKD3D-Proton version: 3.1.0, build `3dfc6f07d0953b1` (GE-Proton11-5: `731c4aae5991b33`)

### Log files

Attached: full log captured via Lutris/umu-launcher with `VKD3D_DEBUG=info` (not a Steam install, so no `steam-xxxxx.log`).

Relevant excerpt:

```
NVRM: Xid (PCI:0000:01:00): 109, pid=60621, name=witcher3.exe, channel 0x0000001d, errorString CTX SWITCH TIMEOUT, Info 0x25c01a
```

```
[...]

PR #3317, opened 2026-09-22
vkd3d: Split RTAS build batch on destination overlap

mhedges1:
Acceleration structure builds are batched into a single
`vkCmdBuildAccelerationStructures` call. Vulkan forbids two builds in one such
call from sharing scratch memory (`VUID-...-scratchData-03704`) or destination
memory (`VUID-...-dstAccelerationStructure-03698` and `-03702`).

`d3d12_command_list_register_rtas_scratch_range()` already does an interval
overlap test on scratch and splits the batch on a collision. There was no
equivalent check for the destination, and the batch is otherwise only split when
the build type changes or `PERFORM_UPDATE` flips — neither of which considers an
address. An application that builds the same destination twice in a row using
different scratch therefore produced a call violating both destination VUIDs.

This adds the matching destination range check, mirroring the scratch one
including the batch-restart retry.

### How I found it

By inspection, while investigating something unrelated, then confirmed with the
validation layers. I don't have a real application that exercises this — the
included tests are the reproducer. The shared-scratch case is a control that the
existing scratch check already handles; only the distinct-scratch case trips the
destination VUIDs.

Before, on the distinct-scratch test:

```
VUID-vkCmdBuildAccelerationStructuresKHR-dstAccelerationStructure-03698:
  pInfos[0].dstAccelerationStructure and pInfos[1].dstAccelerationStructure
  are both VkAccelerationStructureKHR 0x910000000091.

VUID-vkCmdBuildAccelerationStructuresKHR-dstAccelerationStructure-03702:
  pInfos[1].dstAccelerationStructure overlaps with dstAccelerationStructure
  of pInfos[0], on device address range [0xc320000, 0xd300000).
```

After: clean.

### Testing

`VKD3D_TEST_FILTER=raytracing` under `VK_LAYER_KHRONOS_validation`, NVIDIA
driver 615.71.09: 94306 tests, 0 failures. The remaining validation errors in
that run come from `test_raytracing_opacity_micro_map{,_ray_query}` and
`test_raytracing_multi_global_rs` and are present on unmodified master too
(verified by building without this change and getting identical counts), so
they are unrelated.


HansKristian-Work:
What are you even fixing here. This VVL error is known and is benign. Also, the OMM tests pass fine here on NVIDIA.

Overlapping RTAS build is UB in d3d12. Did you identify a real game doing that?

haroldiedema:
@HansKristian-Work this might be relevant here: https://github.com/HansKristian-Work/vkd3d-proton/issues/3226

Blisto91:
Doesn't their last long comment there boil down to they didn't get rid of the crashes and don't know if the issue is vkd3d-proton or driver? They (or a LLM idk) wrote it after this PR was put up.

HansKristian-Work:
So what is the point then? It's a fix that doesn't fix anything?

PR #3319, opened 2026-09-22
vkd3d: Add a config flag to enable extended a7xx compat.

### LINKED ISSUE #3310: Have all workarounds settable via envvars
In Turnip we are testing d3d12 traces in CI, and there is no way to enable certain workarounds like `VKD3D_APPLICATION_FEATURE_ASSUMES_STRICT_BYTE_ADDRESS_WRAP` or `VKD3D_APPLICATION_FEATURE_REQUIRES_MIN16_DENORMS` for the trace, same for the most other workaround I guess. 

Would be useful both for CI and to quickly check if workaround helps a game.

HansKristian-Work:
Useful for debugging games to see if they need some flags to work properly.

Fix #3310.


werman:
Works for me I guess.

Also, should `VKD3D_APPLICATION_FEATURE_BROKEN_WAVE128_REQUESTS` be also set with it? Doesn't matter for CI, but would matter for case of checking if known compat flags are necessary for an app.

NIT: I think `adreno_7xx_extra_compat` would be slightly better, since Intel also has A7XX named GPUs =)

doitsujin (approved):

